PR #1387, opened 2015-01-02
Remove usage of `Date.today` inside the constant

alexmuller:
This usage of `Date.today` caused a bug in production earlier. When the application was started the year was 2014, and the end date that was stored in the constant (and was never updated) was "2014-12-31".

As I've put in the commit message, I don't necessarily expect this to be merged. But I'm opening this pull request because the change is probably a net improvement and is a way to make sure the problem isn't forgotten.


dsingleton:
I don't think this should have been merged without more discussion, or before the postmortem.

It's not actively broken (since the app was restated) or urgent. We should take some time and work out a proper fix, rather than just putting a bigger number on it.

PR #1398, opened 2015-01-09
Use component alpha label rather than slimmer

edds:
Use the new style component for the alpha label rather than the old
slimmer version. This will make it more obvious as to where the label is
coming from.

This needs this pull request to be merged and deployed first:
https://github.com/alphagov/static/pull/527


edds:
Static has now been merged and deployed to production so this should be good to merge.

dsingleton:
:+1: This is a lovely change.

PR #1404, opened 2015-01-13
Update for grid layouts

edds:
So that the pages still align against the correct left hand reading edge
with a narrower central column remove the extra spacing.

This is the matching pull request to go with this one on static: https://github.com/alphagov/static/pull/530


dsingleton:
LGTM.

There's a lot of debt in the SA styles here, but once the changes with this+static and rolled out it'll be a lot easier to cleanup (and is on my list)

edds:
Got the thumbs up, merging so it can be deployed.

PR #1418, opened 2015-01-23
Yes/No radio buttons side by side

robmckinnon:
Adds the `inline` style to Yes and No radio buttons. The inline style results in Yes
and No buttons being placed side by side. This is consistent with the GOV.UK
elements guide, see: http://govuk-elements.herokuapp.com/#form-radio-buttons

Here is a hackpad page where this issue is being discussed:
https://designpatterns.hackpad.com/Smart-answers-9q7BEVu6tOZ


dsingleton:
This is a great improvement, with thumbs up from @antimega and @rooreynolds.

Code wise, ideally the display logic would be in a presenter rather than the template, but with Smart Answer presenters being duplicated, i'm not sure it's worth it at the moment. :+1:

PR #1419, opened 2015-01-23
Highlight radio and checkbox selection buttons

robmckinnon:
Use the `GOVUK.selectionButtons()` function to add visual highlighting to
radio and checkboxes on forms. Logic is added to `updateContent()` function
to ensure highlighting works after content updates.

This change makes smart-answer selection buttons behave according to
recommendations in the GOV.UK elements guide. For example:
http://govuk-elements.herokuapp.com/#form-checkboxes

Here is a hackpad page where this issue is being discussed:
https://designpatterns.hackpad.com/Smart-answers-9q7BEVu6tOZ

### DIFF
--- a/app/assets/javascripts/smart-answers.js
+++ b/app/assets/javascripts/smart-answers.js
@@ -115,6 +115,8 @@ $(document).ready(function() {
     if ($(".outcome").length !== 0) {
       $.event.trigger('smartanswerOutcome');
     }
+    var $blockLabels = $("input[type='radio'], input[type='checkbox']");
+    GOVUK.selectionButtons($blockLabels);
   }
 
   function initializeHistory(data) {


dsingleton:
Good spot! I bugged @tombye about this and he [refactored it to add live events](https://github.com/alphagov/govuk_frontend_toolkit/pull/139), but we're not using them yet.

At the moment `SelectionButton`s are applied on https://github.com/alphagov/static, rather than per application, so it'd be better to update it there (otherwise the call here will be bound  to a particular version of `SelectionButton`).

I'll do a quick pull request for the static change and CC you on that.

dsingleton:
I'm going to this PR, as it's been [updated in static](https://github.com/alphagov/static/pull/535). If you're using a local copy of static you'll need to update it, if you're using our host then it needs a deploy, which should happen this afternoon. :+1:

PR #1427, opened 2015-02-09
Refactor lost or stolen passport

svenagnew:
My first effort at updating smart_answer_flows is to refactor the smart answer for reporting a lost or stolen passport by removing the logic, content and node associated with determining adult or child status.

Please review and merge if suitable or comment here on what I need to modify.


jackscotti:
:stars:

PR #1441, opened 2015-02-25
Remove unnecessary analytics event tracking

benilovj:
Prior to moving to Google Universal Analytics, an audit was carried
out on which analytics events are still relevant, and the events removed
in this commit aren't being used anywhere.

### DIFF
--- a/app/assets/javascripts/smart-answers.js
+++ b/app/assets/javascripts/smart-answers.js
@@ -18,14 +18,8 @@ $(document).ready(function() {
       return false;
     });
 
-    $('.start-right').live('click', function() {
+    $('.start-right, .link-right a').live('click', function() {
       reloadQuestions($(this).attr('href'));
-      return false;
-    });
-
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/smart-answers.js
```diff
@@ -27,7 +26,6 @@ $(document).ready(function() {
 
     // Track when a user clicks on 'Start again' link
     $('.start-right').live('click', function() {
-      window._gaq && window._gaq.push(['_trackEvent', 'MS_smart_answer', getCurrentPosition(), 'Start again']);
```
fofr: Is the `getCurrentPosition` function used by anything else? If not that can also be removed.

benilovj:
On closer analysis, it seems like this tracking was added [very recently](https://github.com/alphagov/smart-answers/commit/aaa81e6a1b789caac325f731770745c81f384753) - closing this PR until we can conclusively determine whether we need to track this information or not.

PR #1444, opened 2015-02-26
Send analytics event through the GOV.UK Analytics API

benilovj:
https://www.pivotaltracker.com/n/projects/1261204/stories/88175032

This change wraps the GA event in the [newly introduced](alphagov/static#549)
analytics API. This will aid in the migration from GA Classic to Universal Analytics, while
remaining functionally equivalent during the migration.

This also removes a commented out line.

/cc @fofr

### DIFF
--- a/app/assets/javascripts/smart-answers.js
+++ b/app/assets/javascripts/smart-answers.js
@@ -3,7 +3,6 @@ function browserSupportsHtml5HistoryApi() {
 }
 
 $(document).ready(function() {
-  //_gaq.push(['_trackEvent', 'Citizen-Format-Smartanswer', 'Load']);
   if(browserSupportsHtml5HistoryApi()) {
     var formSelector = ".current form";
 
@@ -27,7 +26,7 @@ $(document).ready(function() {
 
     // Track when a user clicks on 'Start again' link
     $('.start-right').live('click', function() {
-      window._gaq && window._gaq.push(['_trackEvent', 'MS_smart_answer', getCurrentPosition(), 'Start again']);
+      GOVUK && GOVUK.analytics && GOVUK.analytics.trackEvent && GOVUK.analytics.trackEvent('MS_smart_answer', getCurrentPosition(), {label: 'Start again'});
       reloadQuestions($(this).attr('href'));
       return false;
     });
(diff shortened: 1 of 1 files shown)


fofr:
@edds comment on the social tracking is relevant here too:
https://github.com/alphagov/whitehall/pull/1998#issuecomment-76207953

benilovj:
@fofr, I can see two options here:

1. add the code to check whether the function is present
2. move the analytics code to after the `reloadQuestions` call (so there's no user impact if the analytics call fails) and leave the analytics call as is

I have an ever so slight preference for 2. Thoughts?

fofr:
If you go with option 2 you'll still need to make the `getCurrentPosition` call beforehand, before the reloadQuestions call updates the URL.

benilovj:
> If you go with option 2 you'll still need to make the getCurrentPosition call beforehand, before the reloadQuestions call updates the URL.

That's a very good point. I've updated the code (873cfe0) and the commit comment accordingly.

benilovj:
Reverting to option 1 to make the tests pass.

edds:
The tests are failing because the `GOVUK` object doesn't exist.

    TypeError: 'undefined' is not an object (evaluating 'GOVUK.analytics.trackEvent')

Taking the approach @fofr took with the pull request against whitehall to make the JavaScript error so that we know there was an error then you would probably need to make the methods look more like:

```javascript
$('.link-right a').live('click', function(event) {
  var href = $(this).attr('href');
  event.preventDefault();
  reloadQuestions(href);
  GOVUK.analytics.trackEvent('MS_smart_answer', href, {label: 'Change Answer'});
});
```

Just moving the `trackEvent` call after the `reloadQuestions` call wouldn't be ideal due to the `return false` being use to stop the event from doing the default action when the link is clicked which is something we would want to happen even if the analytics call was to fail.

edds:
Though thinking about it doing that would still cause the tests to fail unless you define `GOVUK.analytics.trackEvent` in the tests somehow. So adding a check for that object to exist might be the easiest option.

benilovj:
@fofr @edds this should be ready to be merged. Are you happy with it as is?

PR #1448, opened 2015-03-02
Bump ruby to 2.1.5

benilovj:
Ruby 1.9.3 is [no longer supported](https://www.ruby-lang.org/en/news/2014/01/10/ruby-1-9-3-will-end-on-2015/).


jackscotti:
LGTM :+1:

PR #1450, opened 2015-03-03
Merge v2 maternity paternity calculator

svenagnew:
As per request from @lutgendorff we have merged the most recent copy and logic changes for maternity-paternity calculations into the V1 multi-file format.


jackscotti:
Good job!

PR #1452, opened 2015-03-03
Send virtual pageview through the GOV.UK Analytics API

benilovj:
This change wraps the GA event in the newly introduced
analytics API. This will aid in the migration from GA Classic to
Universal Analytics, while remaining functionally equivalent
during the migration.

https://www.pivotaltracker.com/n/projects/1261204/stories/87738926


fofr:
The other calls in Smart Answers use `GOVUK && GOVUK.analytics` etc. protection, should this do the same?

benilovj:
@fofr updated the code to be consistent with the other calls in Smart Answers

PR #1458, opened 2015-03-12
Student finance forms yearly document update

BenJanecke:
https://govuk.zendesk.com/agent/tickets/972939

2015 to 2016 forms for continuing EU full-time students will be available from Monday 16 March.

This path:
https://www.gov.uk/student-finance-forms/y/eu-full-time/year-1516/continuing-student

needs to lead to this outcome:
https://github.com/alphagov/smart-answers/blob/master/lib/smartdown_flows/student-finance-forms/outcomes/outcome_eu_ft_1516_continuing.txt

and the file size for the first file needs to change from 109 to 107KB


jackscotti:
:+1:

PR #1461, opened 2015-03-16
Remove Coming soon row from 2014-2015 outcome

BenJanecke:
This was implemented as a quick fix and is no longer necessary as the new 2015 outcome satisfies this need.


jackscotti:
:+1:  merge it when the tests are passed

PR #1464, opened 2015-03-16
Allow to add ?debug=1 in dev env to see the current state of the answer

tadast:
Opening this for a discussion to see if anyone else will find this useful or suggest any alternatives.

As a new person on the project who is changing things in existing smart
answers I found myself spending a decent amount of time inspecting the
current state of the answer given a certain URL. This makes it very easy
to do that by appending a query parameter `debug=1` to the URL.

/cc @BenJanecke @jackscotti @svenagnew

### DIFF
--- a/app/views/smart_answers/_debug.html.erb
+++ b/app/views/smart_answers/_debug.html.erb
@@ -1,0 +1,10 @@
+<% if debug? %>
+  <article>
+    <pre class="application-notice info-notice debug">
+      Debug info:
+      <% @presenter.current_state.to_hash.each do |k, v| %>
+        <%= "#{k}: #{v.to_s}"%>
+      <% end %>
+    </pre>
+  </article>
+<% end %>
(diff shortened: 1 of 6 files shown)


jackscotti:
:+1:  from me. @dsingleton ?

app/views/smart_answers/_debug.html.erb
```diff
@@ -0,0 +1,10 @@
+<% if Rails.env.development? && params[:debug] %>
```
bradleywright: Could we do this in a presenter/controller instead? I'd rather not be poking into Rails' internal state in a view.

tadast:
Added a test and a README entry. Anything else?

bradleywright:
:+1:. Will merge when green.

app/views/smart_answers/_debug.html.erb
```diff
@@ -0,0 +1,10 @@
+<% if @presenter.debug? %>
```
bradleywright: This is currently failing because `debug?` isn't a method on `SmartDownPresenter`.

jackscotti:
fails:

```
ExamplePluginsTest
  test_shows_example_of_plugins_in_use                           ERROR (0.42s)
ActionView::Template::Error:         undefined method `debug?' for #<SmartdownAdapter::Presenter:0x007fb22ce02c80>
```

tadast:
Ok so there are multiple types of presenters. I've moved the check to the controller, should be presenter-neutral now :)

PR #1475, opened 2015-03-20
Remove unnecessary v2 files

jackscotti:


tadast:
LGTM, I like the smell of deleted code

PR #1477, opened 2015-03-20
Correctly display bulleted list in Childcare Costs for Tax Credits SA

issyl0:
- This text with bullets wasn't displaying as with bullets, but in one
  large block, because it was marked as of the `hint` type, not of the
  `body` type. The `body` type seems to consistently support bullet
  points delimited by `-`, instead of the `hint` type.

Before:

screen shot 2015-03-20 at 12 13 09

After:

screen shot 2015-03-20 at 12 25 36


jackscotti:
:+1:

PR #1478, opened 2015-03-20
Calculate married couples allowance: fixed markdown

lutgendorff:


lib/smart_answer_flows/locales/en/calculate-married-couples-allowance.yml
```diff
@@ -80,10 +80,10 @@ en-GB:
           Contact HM Revenue & Customs to claim.
 
 
-          $C
-          **HM Revenue & Customs**
-          **Telephone** 0845 300 0627
-          **Textphone** 0845 302 1408
+           $C
```
tadast: This `$C` is indented with an extra space. This can break things. There is also some trailing whitespace in the new lines, that preferably should not be there :)

PR #1481, opened 2015-03-23
Publish overseas passports changes

tadast:
We've got :+1: from the V2 fact check so this is going live now. Summary of changes is written up in the commit message.

Would be nice if someone could make sure I've dealt with all V2 files that I needed to in a right way as it's the first time I'm doing this. I'll document the V2 lifecycle when I get this merged.


lib/smart_answer_flows/locales/en/overseas-passports.yml
```diff
@@ -1511,6 +1583,7 @@ en-GB:
           3. Check the supporting documents you must send with your application. Non-English supporting documents must be translated into English by a professional translator.
           %{how_to_apply}
 
+          %{incomplete_application_deadline}
 
             %{fco_forms}
```
jackscotti: Extra tab at this line, I know this does not come with your changes but........

BenJanecke:
Looks good to me :+1:

dsingleton:
:+1: to documenting the V2 process.

tadast:
thanks @dsingleton!

V2 docs https://github.com/alphagov/smart-answers/pull/1482/files

PR #1482, opened 2015-03-23
Expand on V2 lifecycle in the readme

tadast:
Is this accurate? Can we make it clearer?


jackscotti:
Pretty clear, thanks for doing this :+1:

dsingleton:
Huge improvement, nice one.

Accurately doc'ing what is done now (and where problems are) is the first step to improving the process. :+1:

PR #1486, opened 2015-03-24
Add capability to assert on smartdown markers

BenJanecke:
* add array comparison method to smartdown_scenarios_test 
* update animal-example-simple to reflect markers

### DIFF
--- a/test/unit/smartdown_content/smartdown_scenarios_test.rb
+++ b/test/unit/smartdown_content/smartdown_scenarios_test.rb
@@ -28,6 +28,7 @@ class SmartdownScenariosTest < ActiveSupport::TestCase
           end
           answers = scenario.question_groups.flatten.map(&:answer)
           outcome_is_reached(smartdown_flow, answers, scenario.outcome)
+          markers_are_included(smartdown_flow, answers, scenario.markers)
         end
       end
     end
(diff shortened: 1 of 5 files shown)


test/unit/smartdown_content/smartdown_scenarios_test.rb
```diff
@@ -56,4 +57,15 @@ def outcome_is_reached(flow, answers, outcome)
                    "In flow #{flow.name} with answers #{answers}, outcome #{state.current_node.name} was reached and not #{outcome}"
     end
   end
+
+  def markers_are_included(flow, answers, markers)
```
tadast: the method name denotes that it's checking for inclusion, but it's actually checking if the elements are equal. I'd either change the name of the method to `markers_match`, `assert_all_markers`, `markers_equal` etc. or modify the method to check for inclusion. Depends on what you want I guess :)

dsingleton: I think `markers_are_included` is reasonable, but it'd be worth renaming `same_elements? (or merging into the method), and using more explicit var names, like `expected_markers` and `actual_markers`.

The behaviour is "does `actual_markers` contain all of `expected_marker`, right? Which the current implementation of `same_elements?` does, eg `(['b','a'] - ['a', 'b','c', 'd']) => []` and `(['b','a', 'e'] - ['a', 'b','c', 'd']) => ['e']` (where the left array is expected, the right actual).

The logic is good, but the naming could be a bit clearer.

How about something like:

```ruby
def markers_are_included(flow, answers, expected_markers)
  state  = flow.state(true, answers)
  actual_markers = state.current_node.markers.map(&:marker_name)
  assert_empty
    (expected_markers - actual_markers),
    "In flow #{flow.name} with answers #{answers} markers #{markers} were expected but #{node_markers} was received"
end
```

(not sure if there is a better built in `assert` we could use there?

PR #1492, opened 2015-03-25
Fix the UK visa checker country list caching issue

tadast:
\<facepalm>
    It was using the same class instance variable to cache two lists, so depending on which one (V1 or V2) was hit first (per app instance) it would cache and serve the same list for both methods 
\</facepalm>


BenJanecke:
:+1:

PR #1494, opened 2015-03-26
Document how to archive a Smart Answer

dsingleton:
First pass at documenting how a Smart Answer needs to be archived,
as it varies from the usual process.

Very much a first pass. 

I'd explicitly appreciate feedback from:
- developers who've worked on SAs: @jackscotti @tadast, @BenJanecke 
- people who work on second line (as they'll be using it)

(I'd also like to move towards the longer-form SA documentation moving to `docs/` and [out of the README](https://github.com/alphagov/styleguides/blob/master/use-of-READMEs.md))

### DIFF
--- a/doc/archiving.md
+++ b/doc/archiving.md
@@ -1,0 +1,46 @@
+# How to archive a Smart Answer
+
+To prevent a Smart Answer being [registered with Panopticon](../blob/master/lib/tasks/panopticon.rake) you will need to change the status of the Smart Answer flow, or remove the Smart Answer flow entirely.
+
+As archiving tends not to be temporary, it's preferable to remove the flow entirely - and it can always be recovered from `git` history.
+
+If a flow needs to be removed urgently then it be can archived quickly by changing the flow's status to 'archived'. You should make sure work is scheduled to properly clean up the, now, unused flows.
+
+Once a flow has been removed the artefact will no longer register itself when the Smart Answers application is released.
+
+You will then need to:
+1. Archive the artefact in Panopticon (same process as other types of content)
+2. Add a redirect `router-data` for the artefact slub. This should be a `prefix` route, so all parts of the Smart Answer URL are matched and preserved. ([This is a good example](https://github.gds/gds/router-data/blob/master/data/slug-changes.csv#L623))
+
+## Flow file naming convention
+
+Files associated to a flow are generally named to match the slug of the flow, eg, `pay-leave-for-parents` may have files named `pay-leave-for-parents.yml` or `pay_leave_for_parents.rb`, depending on the naming convention of the file type.
+
+It's advisable to search the repo for both underscore and snake-case variations of the name.
+
+## Removing a Ruby/YML Smart Answer
+
+Remove these files:
+
+- `lib/smart_answer_flows` has a `.rb` file matching the name of the flow
+- `lib/smart_answer_flows/locales/en` has a `.yml` file matching the name of the flow
+- `test/integration/smart_answer_flows/` has a `_test.rb`, matching the name of the flow
+
+There may also be an associated `Calculator` or data files, these will be referenced from the `.rb` flow. For example: `Calculators::PlanPaternityLeave.new(due_date: due_date)`
+
+- Calculators can be found in `https://github.com/alphagov/smart-answers/tree/master/lib/smart_answer/calculators` - ensure they're not used by any other flows.
+- There may be some flow specific data in `lib/data` - again, ensure thats not used by any other flows.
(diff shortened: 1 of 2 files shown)


jackscotti:
it seems to cover everything. LGTM :+1:

doc/archiving.md
```diff
@@ -0,0 +1,40 @@
+# How to archive a Smart Answer
+
+To prevent a Smart Answer being [registered with Panopticon](../blob/master/lib/tasks/panopticon.rake) you will need to change the status of the Smart Answer flow, or remove the Smart Answer flow entirely.
+
+As archiving tends not to be a temporary, it's preferable to remove the flow entirely - and it can always be recovered from `git` history.
```
jennyd: ' a temporary change', or just 'temporary'?

BenJanecke:
LGTM :+1:  Perhaps mention that a quick scan of the .rb files should be made to see if any helper classes are being used.

mattbostock:
Should this mention router data?

heathd:
really great to have some documentation covering this, thanks David.

Would be helpful to have guidance about how to remove the artefacts from panopticon as well? Could this be an additional rake task within smart answers itself?

doc/archiving.md
```diff
@@ -0,0 +1,40 @@
...
+
+If a flow needs to be removed urgently then it be archived quickly by changing the flows status to 'archived'. You should make sure work is scheduled to properly clean up the, now, unused flows.
```
tadast: minor typos: "then it **can** be archived".. "changing the flow's"

dsingleton:
I've fixed the silly typos.

I'd rather not add too much detail on `panopticon` archiving, or redirects in `router-data`, as thats not SA specific. I'm expecting this documentation to be linked from the Ops Manual, in the context of general archiving. 

@BenJanecke I thought that might be covered this bit under naming convention:
> It's advisable to search the repo for both underscore and snake-case variations of the name.

Do you think it needs to be more explicit?

BenJanecke:
I was thinking more along the lines of something like uk visas using the external UkbaCountry class or marrige abroad using the WorldLocation class. 

So what I'm trying to say is that we should keep in mind that smart answers can have external dependencies.

mattbostock:
@dsingleton: That makes sense. Would you mind taking a look at [this PR](https://github.gds/gds/router-data/pull/365), as I have a relevant question in that PR relating to archiving a Smart Answer.

dsingleton:
I've added a note about `router-data` redirects that is specific to SA (based on @mattbostock's question).

@BenJanecke: I've added an expanded warning:
> Generally, the larger and more complex the Smart Answer, the more likely there may be additional supporting files, and you should check the flow carefully and get review from someone more familiar with Smart Answers.

I think thats strong enough.

BenJanecke:
:+1:  LGTM

doc/archiving.md
```diff
@@ -0,0 +1,42 @@
+# How to archive a Smart Answer
+
+To prevent a Smart Answer being [registered with Panopticon](../blob/master/lib/tasks/panopticon.rake) you will need to change the status of the Smart Answer flow, or remove the Smart Answer flow entirely.
+
+As archiving tends not to be temporary, it's preferable to remove the flow entirely - and it can always be recovered from `git` history.
+
+If a flow needs to be removed urgently then it be can archived quickly by changing the flow's status to 'archived'. You should make sure work is scheduled to properly clean up the, now, unused flows.
+
+Once a flow has been removed the artefact will no longer register itself when the Smart Answers application is released. It will then to be archived in Panopitcon (as for other types of content), and a redirect added to `router-data` as a `prefix` route, so all parts of the Smart Answer URL are matched, not just the `slug` ([example](https://github.gds/gds/router-data/blob/master/data/slug-changes.csv#L623))
```
mattbostock: s/Panopitcon/Panopticon/

mattbostock: > and a redirect added to `router-data` as a `prefix` route

This reads like this is done programatically - I think it would be good to make it clear that adding the router-data redirect is a manual step.

dsingleton: Good spot, i've fixed the typo and rephrased it to be more explicit about requiring action, by YOU, the READER.

> You will then need to:
> 1. Archive the artefact in Panopticon (same process as other types of content)
> 2. Add a redirect `router-data` for the artefact slub. This should bea `prefix` route, so all parts of the Smart Answer URL are matched, not just the slug part of the ([This is a good example](https://github.gds/gds/router-data/blob/master/data/slug-changes.csv#L623))

PR #1514, opened 2015-04-01
Married couples allowance rates 2015-16

tadast:
- whitespace fixes (wanted as a separate commit to reduce noise in the refactoring commit)
- refactors the way rates data is retrieved (see message https://github.com/alphagov/smart-answers/commit/59cd29fb51435bb53db2efa9a8a289f44185b380)
- updates values for the next fiscal year from [this table](https://www.gov.uk/government/publications/tax-and-tax-credit-rates-and-thresholds-for-2015-16/tax-and-tax-credit-rates-and-thresholds-for-2015-16#income-tax-allowances)

Since the refactoring is not very straightforward I'd like a careful review to make sure I didn't change anything I did not want to change /cc @jackscotti @BenJanecke @dsingleton

### DIFF
--- a/lib/data/married_couples_allowance_rates.yml
+++ b/lib/data/married_couples_allowance_rates.yml
@@ -1,0 +1,21 @@
+---
+# the year denotes a starting year of the fiscal year
+# e.g. 2014 for 6th Apr 2014 - 5th April 2015
+personal_allowance:
+  2013: 9440
+  2014: 10000
+over_65_allowance:
+  2013: 10500
+  2014: 10500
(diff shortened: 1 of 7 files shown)


jackscotti:
all as a V1? GREAT refactoring anyway! :+1:

tadast:
It doesn't make much sense to make a V2 as the changes are visible only after 2016-04-06. Also there's little time left before this needs to go live so the FC coming back on time would be unlikely.

jackscotti:
fair enough :+1:

PR #1515, opened 2015-04-01
Split out DSA full and DSA slim to make it clearer to students which they should use

turnbullja:
Split out DSA full and DSA slim to make it clearer to students which they should use.


tadast:
@turnbullja could you please provide the URLs where we could validate the changes when deploying? Otherwise we have to track down the logic to see what questions lead to those outcomes :)
Thanks

PR #1516, opened 2015-04-01
Register a death fixes

tadast:
- Algeria-specific payment option applies only if you are currently there.
- Fixes a typo (V1 and V2) where something birth-related was copied to the deaths registration document
- Resolve the contradictionary 'You will need to' bulletpoint

All V2 + a typo fix for V1.


BenJanecke:
:+1: LGTM

PR #1524, opened 2015-04-07
Births abroad V2

tadast:
Best reviewed commit by commit as each commit has a descriptive message.

I wanted to submit a PR per change but the V2 is shared between this and death registration logic so some of the changes where already made via the other branch causing some of the tests to fail.

V2 changes only


jackscotti:
all good to me apart from the phase5 variable and phraselist naming it

tadast:
@jackscotti renamed "phase 5" to "higher risk" and checked with the content editor regarding the wording, it's correct. The difference is subtle, but important: in some countries it's illegal to register a birth that happened outside of marriage; in some it's illegal to have a child outside of marriage.

jackscotti:
@tadast thank you for looking into this!

PR #1530, opened 2015-04-09
Fix state pension calculator rounding error

tadast:
Floats where used when rounding, therefore in some edge cases the
returned pension weekly rate values were off by 1p.

e.g. /calculate-state-pension/y/amount/female/1977-06-29/0/6/no/no
was £34.78 instead of £34.79
```
>> 35.765.round(2)
=> 35.77
>> 35.775.round(2)
=> 35.78
>> 35.785.round(2)
=> 35.78
```
Using BigDecimals for rounding is more accurate

### DIFF
--- a/lib/smart_answer/calculators/state_pension_amount_calculator.rb
+++ b/lib/smart_answer/calculators/state_pension_amount_calculator.rb
@@ -59,7 +59,7 @@ module SmartAnswer::Calculators
     end
 
     def what_you_get
-      what_you_get_raw.round(2)
+      BigDecimal(what_you_get_raw.to_s).round(2).to_f
     end
 
     def what_you_get_raw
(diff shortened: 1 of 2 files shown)


bradleywright:
:+1: to using `BigDecimal`.

tadast:
It used to be BigDecimals everywhere, but it was removed due to "noise" https://github.com/alphagov/smart-answers/commit/25792143ffc57d3a56b435f888957b0937aac88c

I did not revert this because I don't know the full reasoning behind this commit, so I guess this is good enough for now.

bradleywright:
Native floating points aren't sufficient precision for calculating/rounding money, so that commit is wrong.

tadast:
Should I revert it? I could `require 'bigdecimal/util'` that allows calling `to_d` on floats instead of using `BigDecimal()` if this is the noise the commit message is referring to

bradleywright:
Let's leave it for the moment.

PR #1532, opened 2015-04-09
Fix Democratic Republic of Congo births registration outcome

tadast:
It was misspelled as congo-(democratic-republic) instead of democratic-republic-of-congo, which is how it's represented in the countries dropdown.


test/integration/smart_answer_flows/register_a_birth_v2_test.rb
```diff
@@ -582,6 +582,18 @@ class RegisterABirthV2Test < ActiveSupport::TestCase
     end
   end
 
+  context "Democtaric Republic of Congo" do
```
jackscotti: typo here

jackscotti:
all good apart from the little typo :+1:

tadast:
Thanks Jack

PR #1542, opened 2015-04-10
Remove maternity allowance from adoption flow

BenJanecke:
Removed all occurrences of maternity allowance for the single parent adoption flow.
Removed all occurrences of maternity allowance for the two pre April 5 2015 parent adoption flow.
Removed all occurrences of maternity allowance for the two post April 5 2015 parent adoption flow.


tadast:
LGTM.

For the next time: I would be a lot more confident if I knew why those changes are being made. A sentence or two about the original requirement in a commit message would go a long way imho :)

jackscotti:
:+1:  to the PR and :+1:  to @tadast comment

PR #1547, opened 2015-04-14
Remove duplicated textfrom death registration in Kenya and Nigeria

tadast:
As per FC feedback:
This text appears twice: "You’ll also be sent copies of the registration certificate if you’ve paid for them.
The embassy will contact you when your documents and certificates are ready for collection."
Also, it should be 'high commission' not 'embassy'.

`/register-a-death-v2/y/overseas/kenya/same_country`
`/register-a-death-v2/y/overseas/nigeria/same_country`


tadast:
I'm going to do some refactoring in another branch. Closing this one

PR #1549, opened 2015-04-14
Births and deaths abroad amends

tadast:
Addresses feedback received from Fact Check. All V2 changes. Summary in the commit messages.

/cc @jackscotti


BenJanecke:
LGTM :+1:

PR #1555, opened 2015-04-16
Extract calculator rates for easier uprating

tadast:
When doing uprating for this year I noticed a few problems:
- multiple ways to store rates for calculators
- some rates were "hardcoded" in content, thus requiring a synchronised deploy on the first day of a new fiscal year
- data(rates) and logic were not separate, making it difficult/impossible for content designers to do uprating
- duplicated logic to fetch the rates

Moreover, the rates sometimes come in late'ish and in bulk making it difficult to meet the deadline.

This is an attempt to solve the above problems. I would appreciate some :eyes: and suggestions for further improvements.

/cc @chrisroos @floehopper @dsingleton @BenJanecke 

:warning: please do not merge until 22nd Apr as I'll be away


dsingleton:
:+1: to the overall idea and approach. 

The mixing of data/logic has slowly grown, and this is a good way to cleanly separate them. De-duplicating and solidly testing the fetch logic is also ace.

We've had a few SA bugs in the past caused by [maintaining state between requests](https://github.com/alphagov/smart-answers/commit/2bca3b01ebb30c68734dc38a3a4fe376ddabe5fd), so thats worth double checking with changes like these.

I've not gone through the diffs in full detail yet, so I won't merge yet (and there it'd be opinions of other folks first).

BenJanecke:
LGTM :+1:

tadast:
I took another look after a long weekend and am wondering if it's worth having two data file formats. The [simple format](https://github.com/alphagov/smart-answers/blob/13f33ae5b077b274398558dfd95a575391a898a7/lib/data/rates/state_pension.yml) is simple to understand, but the [more complex](https://github.com/alphagov/smart-answers/blob/13f33ae5b077b274398558dfd95a575391a898a7/lib/data/rates/maternity_paternity_adoption.yml) one is necessary anyway and the content designer would need to learn both, so perhaps the simple format is in fact just adding complexity? Should I change all the data files to a uniform "exact dates" style?

jackscotti:
:+1: for uniforming all the data files!

tadast:
I've unified the data files format now, see the latest commit.

dsingleton:
:+1: :tada:

PR #1561, opened 2015-04-20
Marriage abroad: Removal of "Irish partner" response

chrisroos:
PT: https://www.pivotaltracker.com/story/show/91098778

@floehopper and I are fairly confident that we've removed the 'Irish partner' response in this pull request.

One thing we're unsure about is whether we should also be making these changes to marriage-abroad-v2 as well?


tadast:
As per off-GH discussion, this needs more V2 sync dancing with other changes :dancer: so please do not merge this :)

PR #1562, opened 2015-04-21
Update poltergeist to work with latest version of phantomjs.

floehopper:
I was seeing test failures like this one:

    Cliver::Dependency::VersionMismatch: Could not find an executable 'phantomjs' that matched the requirements '~> 1.8', '>= 1.8.1'. Found versions were {"/usr/local/bin/phantomjs"=>"2.0.0"}.
        test/integration/engine/report_a_problem_test.rb:8:in `block (3 levels) in <class:ReportAProblemTest>'

This incompatibility was [fixed in poltergeist v1.6.0][1].

@chrisroos has confirmed that this change is backwardly compatible with
phantomjs v1.9.8, so I'm hopeful it should work with other older versions.
The Continuous Integration build for this branch should tell us if it isn't.

[1]: https://github.com/teampoltergeist/poltergeist/blob/master/CHANGELOG.md#160

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -58,7 +58,7 @@ group :test do
   gem 'webmock', '1.20.4', :require => false
   gem 'simplecov', '~> 0.6.4', :require => false
   gem 'simplecov-rcov', '~> 0.2.3', :require => false
-  gem 'poltergeist', '1.5.1'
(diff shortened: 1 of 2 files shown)


tadast:
LGTM :+1:

floehopper:
Thanks.

PR #1563, opened 2015-04-21
Update check-uk-visa: visitor text changes

lutgendorff:


tadast:
I'll fetch those changes in to another PR that will contain other related changes. Thanks Liz :trophy:

PR #1568, opened 2015-04-22
Convert "Inherits someone dies without will" to use next_node_if

chrisroos:
The README states that passing a block to `next_node` has been deprecated.

This branch updates the 'inherits-someone-dies-without-will' Smart Answer to use
`next_node_if` instead.

One benefit of this change is that you can now '/visualise' this Smart Answer in the browser.

Is this still a desirable change and can/should it be merged to master?

### DIFF
--- a/lib/smart_answer_flows/inherits-someone-dies-without-will.rb
+++ b/lib/smart_answer_flows/inherits-someone-dies-without-will.rb
@@ -138,34 +138,46 @@ multiple_choice :siblings? do
 
   save_input_as :siblings
 
-  next_node do |response|
-    case region
-    when "england-and-wales"
-      if partner == "yes"
-        response == "yes" ? :outcome_22 : :outcome_1
-      else
-        response == "yes" ? :outcome_4 : :half_siblings?
+  on_condition(variable_matches(:region, 'england-and-wales')) do
+    on_condition(variable_matches(:partner, 'yes')) do
+      next_node_if(:outcome_22, responded_with('yes'))
+      next_node_if(:outcome_1, responded_with('no'))
+    end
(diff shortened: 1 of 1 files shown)


tadast:
Looking good. The only minor comment I have: the rest of the repo seems to be using symbols for variable names when calling `#variable_matches` e.g. `variable_matches(:var_name, "value")`, it would be nice to keep it consistent.

I think it should go to master for consistency (if not for visualise). It's better to have a bad convention (I'm not saying next_node_if is bad though) than no conventions at all :)

chrisroos:
> The only minor comment I have: the rest of the repo seems to be using symbols for variable names when calling #variable_matches e.g. variable_matches(:var_name, "value"), it would be nice to keep it consistent.

Good spot! I'll make those changes shortly.

chrisroos:
I've added fixup! commits to update strings to symbols in all the places I'm using `variables_matches`.

tadast:
:+1: :shipit:

chrisroos:
Are we confident enough in the tests for me simply to merge this change, or should it be checked more thoroughly by someone else before being merged to master?

tadast:
The tests are quite thorough for this flow I think. I'd check a few edge-cases manually just in case but it LGTM.

If you don't feel confident there are some QA people at GDS who could do some more testing to increase confidence if you provide them with two entry URLs to compare.

In some cases I've used a simple crawler script that checks the outcomes' text. Should be fairly simple to compare two versions (e.g. production v.s. local dev) with smth like this https://gist.github.com/tadast/07524de341d834356ba6 except there'd be more variables that can form the URL.

chrisroos:
I'm going to put my faith in the tests and merge this change to master.

PR #1576, opened 2015-04-23
UK visa checker fixes based on FC feedback

tadast:
## Original Request
```
Tourism, including visiting friends or family
Should not say the following (there is no such thing any more and people visiting family just need a standard visit visa):
“Visiting family

“Apply for a family visit visa if you plan to visit your family or partner (eg spouse) while you’re in the UK.”

Work, academic etc > longer than 6 months
Should include a link to standard visit visa guide for academics doing research on sabbatical who can have a standard visit visa for 12m

Get married
Should include a reference to civil partnership as well as marriage

Under (studying) “for 6 months or less”, the reference should read:

“Apply for a short-term study visa if you are studying for 6 months or less.”

Under (studying) for more than 6 months the bullet “under 18 years old” should be taken out.  The 11 month short-term student route is for 18s and over only.

There is a line that has been taken out from the studying for more than 6 months section that should go back in “Apply for a child study visa if you’re under 18” which should link to Tier 4 (Child).
```

### DIFF
--- a/lib/smart_answer_flows/locales/en/check-uk-visa-v2.yml
+++ b/lib/smart_answer_flows/locales/en/check-uk-visa-v2.yml
@@ -142,7 +142,7 @@ en-GB:
           work: "Work, academic visit or business"
           tourism: "Tourism, including visiting friends or family"
           school: "Visit your child at school"
-          marriage: "Get married"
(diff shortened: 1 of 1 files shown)


BenJanecke:
LGTM :+1:

PR #1577, opened 2015-04-23
WIP: Use ERB templates for Ruby/YAML Smart Answer outcomes

floehopper:
**DO NOT MERGE - FOR DISCUSSION ONLY**

@chrisroos & I think that using ERB templates for outcomes might make the code/content easier to follow as opposed to the current implementation which uses Ruby phrase keys and YAML locale files. 

We think doing something like this might give us many of the benefits of using Smartdown without incurring all the cost of actually converting the Smart Answers into Smartdown.

This branch is a bit of a spike to see (a) how hard it would be to implement this kind of idea and (b) whether the result is indeed any easier to follow.

So far we've only converted the Student Finance Calculator. You can see the resultant outcome templates here:-

* [outcome_uk_full_time_students.txt.erb](https://github.com/alphagov/smart-answers/blob/use-outcome-templates-for-student-finance-calculator/lib/smart_answer_flows/student-finance-calculator/outcome_uk_full_time_students.txt.erb)
* [outcome_uk_all_students.txt.erb](https://github.com/alphagov/smart-answers/blob/use-outcome-templates-for-student-finance-calculator/lib/smart_answer_flows/student-finance-calculator/outcome_uk_all_students.txt.erb)
* [outcome_eu_students.txt.erb](https://github.com/alphagov/smart-answers/blob/use-outcome-templates-for-student-finance-calculator/lib/smart_answer_flows/student-finance-calculator/outcome_eu_students.txt.erb)

One problem with this approach is that (like when using conditional logic `$IF`/`$ENDIF`/etc in Smartdown outcomes), it becomes more awkward to test that various bits of content have been included or not included. This is why the `{{marker: pet-is-a-cat}}` syntax was introduced into Smartdown and we could do something along those lines here, although we're not convinced that's the right approach.

We haven't attempted to address the testing issue in this branch yet, but we thought we'd see what people thought before going any further.

### DIFF
--- a/app/presenters/outcome_presenter.rb
+++ b/app/presenters/outcome_presenter.rb
@@ -1,6 +1,7 @@
 require 'erubis'
 
 class OutcomePresenter < NodePresenter
+  include ActionView::Helpers::NumberHelper
 
   def title
     translate!('title')
@@ -25,7 +26,28 @@ class OutcomePresenter < NodePresenter
     calendar.present?
   end
 
+  def has_body?
+    use_template? || super()
+  end
+
+  def body
+    if use_template?
+      erb_template_path = Rails.root.join("lib/smart_answer_flows/#{@node.flow_name}/#{name}.txt.erb")
+      template = File.read(erb_template_path)
+      govspeak = ERB.new(template).result(binding)
+      GovspeakPresenter.new(govspeak).html
+    else
+      super()
+    end
+  end
+
   private
+
+  attr_reader :state
+
+  def use_template?
+    @node.use_template?
+  end
 
   def render_data_partial(partial, variable_name)
     data = @state.send(variable_name.to_sym)
--- a/lib/smart_answer/flow.rb
+++ b/lib/smart_answer/flow.rb
@@ -84,8 +84,16 @@ module SmartAnswer
       add_node Question::Checkbox.new(name, &block)
     end
 
+    def use_outcome_templates
+      @use_outcome_templates = true
+    end
+
     def outcome(name, options = {}, &block)
-      add_node Outcome.new(name, options, &block)
+      modified_options = options.merge(
+        use_outcome_templates: @use_outcome_templates,
+        flow_name: self.name
+      )
+      add_node Outcome.new(name, modified_options, &block)
     end
 
     def outcomes
--- a/lib/smart_answer/flow_registry.rb
+++ b/lib/smart_answer/flow_registry.rb
@@ -53,8 +53,8 @@ module SmartAnswer
     def build_flow(name)
       absolute_path = @load_path.join("#{name}.rb").to_s
       Flow.new do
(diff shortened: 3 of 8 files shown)


BenJanecke:
:+1:  LGTM when It comes to testing we might want to have a look at what we can do with capybara and it's within blocks.

One question I have is how we are going to work with content shared between multiple outcomes(partials perhaps?)

floehopper:
@BenJanecke: Thanks. Yes, we were imagine having something like Rails partial templates to share content between outcomes.

floehopper:
@tadas, @lutgendorff, @bradleywright, @dsingleton: Do you have any thoughts on the changes in this PR?

floehopper:
We know that this change loses the potential for supporting multiple locales. However, if we needed to do that it could be addressed using something akin to [Rails' Localized Views](http://guides.rubyonrails.org/i18n.html#localized-views).

bradleywright:
I'm not comfortable letting Smartdown templates use ERB unless we strictly lock down what methods are available to it (if that's even a thing). One of the design considerations of Smartdown is security - editors and others shouldn't be able to run arbitrary Ruby code on production machines.

floehopper:
@bradleywright: I thought I'd give a bit more context for this PR.

We can see that some aspects of Smartdown smart answers are clearly better than the Ruby smart answers, but there are some aspects which we're less sure about. Smartdown's use of templating for outcomes (and questions) is clearly an improvement.

However, since we think it would require a lot of effort to convert all Ruby smart answers to Smartdown and we're not convinced about all aspects of Smartdown, we think it might be a good **incremental** step to move in a direction like that demonstrated in this PR.

This change is not intended as a final solution and we wouldn't imagine these templates being directly edited by the content team or at least not without passing through a PR process and being reviewed by a developer.

I guess you might've been thinking in terms of the templates being editable in an admin UI. If we go in that direction, clearly we wouldn't use ERB, but we don't think it would be hard to convert from ERB to Smartdown or any other templating library.

One way to address the security concern would be to use a logic-less templating library like [Liquid][1]. One aspect of Smartdown we're not convinced about is the use of a custom templating engine - it feels a bit like re-inventing the wheel.

In summary, we think that changes like the ones in this PR would have the following benefits:

* More readable, maintainable code & content
* Brings Ruby smart answers incrementally closer to Smartdown smart answers (making any possible later conversion easier)
* Avoids reinventing the wheel for templating

[1]: http://liquidmarkup.org/

bradleywright:
@floehopper sorry! I completely missed that this was just for the Ruby smart answers (I guess I saw the `.txt` extension and made an assumption). In that case fine, please proceed thinking in this direction.

bradleywright:
Thanks for the detailed explanation thought :+1:

floehopper:
I'm going to close this now. I'm planning on making the changes properly and I'll submit a separate PR for that.

PR #1578, opened 2015-04-24
Merge v2 of overseas passport

BenJanecke:
1 Laos has a new process. Delete “and 3 alternative dates and times” from all outcomes.

Current text:
You’ll need to book an appointment by email. Include your first name and last name and 3 alternative dates and times. You will receive an email confirming your appointment.

New text: 
You’ll need to book an appointment by email. Include your first name and last name. You will receive an email confirming your appointment.

Example: https://www.gov.uk/overseas-passports/y/laos/renewing_new/adult

2 The courier fee for Uzbekistan should be £23.01 not £24.72 for all outcomes.
Example: https://www.gov.uk/overseas-passports/y/uzbekistan/renewing_new/child

3 Users can no longer apply for a passport in Yemen.

Change outcome for all application types in Yemen so it looks similar to Libya outcome: 
https://www.gov.uk/overseas-passports/y/libya

But with a link to Yemen travel advice not Libya.

New text:
You can’t apply for a British passport from Yemen at this time.

You should apply in a neighbouring country of your choice.

If you need more information read the Foreign Commonwealth Office travel advice.


tadast:
I would copy over a full description of changes from the V2 PR to the second commit message, so that a dev looking at git history could read why those changes where made to the V1 files. Currently all the useful commit messages are lost together with the V2 and are a lot harder to access :)

Other than that :+1:

PR #1579, opened 2015-04-24
Fix a randomly failing test

tadast:
To reproduce run `rake test TESTOPTS="--seed=56021"`.

smartdown_scenarios_test.rb was modifying registry options that
got cached in a class variable and errored next time it was
being accessed with different options.

It may have some performance degradation, but we must choose 2 out of 3:
1) Being able to run tests in random order
2) Having fast tests
3) Living with intermittent test failures
:)

### DIFF
--- a/test/unit/smartdown_content/smartdown_scenarios_test.rb
+++ b/test/unit/smartdown_content/smartdown_scenarios_test.rb
@@ -38,6 +38,7 @@ class SmartdownScenariosTest < ActiveSupport::TestCase
       end
     end
   end
+  SmartdownAdapter::Registry.reset_instance
 
   def questions_are_asked(flow, answers, question_names)
     assert_nothing_raised("Exception was thrown when running flow #{flow.name} with answers #{answers}") do


chrisroos:
I just ran the tests using `rake test TESTOPTS="--seed=56021"` and see 48 failures all in "test/functional/smart_answers_controller_test.rb". Are they the errors you were seeing, @tadast?

tadast:
I was seeing this exception `https://github.com/alphagov/smart-answers/blob/master/lib/smartdown_adapter/registry.rb#L9` being raised from smart_answers_controller.rb

chrisroos:
These are the test failures I see - https://gist.github.com/chrisroos/4e4ab09df7ea2dc9568d

Could the seed be affecting the tests differently on different machines? I'm on a Mac. Are you using the VM?

chrisroos:
I can confirm that the fix in this branch fixes the failing tests encountered when running with seed 56021.

@tadast I've realised that I was confused by your use of "test" (singular) in the pull request title. I think I'd find it clearer if it said "tests" (plural). Apart from that minor comment, this looks good to me.

PR #1583, opened 2015-04-27
Add exact marker assertion to smartdown runner

BenJanecke:
This matches an exact set of markers whereas the has marker/s syntax only checks for the presence of one or a few markers.


tadast:
LGTM

PR #1585, opened 2015-04-28
Calculate your statutory redundancy pay uprate

BenJanecke:
Uprated max amount for 2015/2016 from 13,920 to 14,250

### DIFF
--- a/lib/data/rates/redundancy_pay_v3.yml
+++ b/lib/data/rates/redundancy_pay_v3.yml
@@ -1,0 +1,17 @@
+---
+- start_date: 2012-01-01
+  end_date: 2013-01-31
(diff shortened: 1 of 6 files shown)


jackscotti:
I am not sure you should keep a v2 not updated with these changes floating around. What if someone merges it into a v1 and you are going to revert these new changes?

BenJanecke:
We could apply these changes to the V2 as well?

jackscotti:
You will then need to go through all the changes made on it... I'd say, kill the v2 in another PR. If they'll get back to us we will revert it?

BenJanecke:
Done see #1586

jackscotti:
:+1:

PR #1586, opened 2015-04-28
Remove v2 of statutory redundancy pay

BenJanecke:
This change was made and never fact checked we are removing this temporarily. 

Once we find out who this change belongs to we will restore it from the git history for fact check.


jackscotti:
thanks Ben!

PR #1587, opened 2015-04-28
Report a lost or stolen passport updates

BenJanecke:
For Canada route, abroad only. Change opening line to: Fill in form LS01 and post it to your nearest British embassy, high commission or consulate. 

https://www.gov.uk/report-a-lost-or-stolen-passport/y/abroad/canada


jackscotti:
oooh... gotta love exceptions... :+1:

PR #1588, opened 2015-04-28
Allow deploying to Heroku

chrisroos:
The current mechanism for making changes to Smart Answers isn't ideal: We have to duplicate everything about the Smart Answer, make the changes and then deploy to preview for someone to view and test. See "[Making bigger changes](https://github.com/alphagov/smart-answers#making-bigger-changes)" for more information.

We think that using git branches, instead of manually branching in the filesystem, will make our lives easier in this situation. Although we can deploy branches to Preview, the auto-deployment functionality means that we can't guarantee how long they'll live. Allowing us to deploy to Heroku should solve this particular problem.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ To help developers track changes in files easily, it is best if you commit V2 fi
 
 ## Deploying to Heroku
 
-    heroku apps:create
+    heroku apps:create --region eu
 
     heroku config:set GOVUK_APP_DOMAIN=preview.alphagov.co.uk
     heroku config:set PLEK_SERVICE_CONTENTAPI_URI=https://www.gov.uk/api
--- a/README.md
+++ b/README.md
@@ -98,14 +98,8 @@ To help developers track changes in files easily, it is best if you commit V2 fi
 
 ## Deploying to Heroku
 
-    heroku apps:create --region eu
+The 'startup_heroku.sh' shell script will create and configure an app on Heroku, push the __current branch___ and open the marriage-abroad Smart Answer in the browser.
 
-    heroku config:set GOVUK_APP_DOMAIN=preview.alphagov.co.uk
-    heroku config:set PLEK_SERVICE_CONTENTAPI_URI=https://www.gov.uk/api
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -95,3 +95,17 @@ To help developers track changes in files easily, it is best if you commit V2 fi
 ## Detailed documentation
 
 - [How to archive a Smart Answer](doc/archiving.md)
+
+## Deploying to Heroku
+
+    heroku apps:create
+
+    heroku config:set GOVUK_APP_DOMAIN=preview.alphagov.co.uk
```
tadast: Multiple variables can be set in one command, would make things faster.

README.md
```diff
@@ -95,3 +95,17 @@ To help developers track changes in files easily, it is best if you commit V2 fi
 ## Detailed documentation
 
 - [How to archive a Smart Answer](doc/archiving.md)
+
+## Deploying to Heroku
+
+    heroku apps:create
```
tadast: Heroku(Amazon) has servers in Europe, so `heroku apps:create --region eu` would have lower latency

bradleywright:
I am very much :+1: to this workflow improvement.

tadast:
This is looking great <3 Can't wait to try this out

bradleywright:
Worth making a shell script called `./heroku-startup.sh` or similar that contains the env variables? And then update the instructions to use that instead.

README.md
chrisroos: Cool. I didn't know that. Thanks, @tadast.

chrisroos:
> Worth making a shell script called ./heroku-startup.sh or similar that contains the env variables? And then update the instructions to use that instead.

@bradleywright: Absolutely. We figured we'd go with the README initially but then wrap it all in a script if we go ahead with this change.

chrisroos:
I'm going to merge these changes to master.

chrisroos:
Manually merged in 849ff4057ae2db762d93685f2a06396f2675da19.

PR #1589, opened 2015-04-29
DRY up additional-commodity-code Smart Answer

chrisroos:
None of the `calculator`, `commodity_code` or `conditional_result` calculations
were used anywhere except in the single `outcome` node. We can avoid a lot of
duplication by moving them to the `outcome` node as `precalculations`.


tadast:
Good to see those red lines. I check out the branch locally just to be sure and it looks good.
:shipit:

PR #1595, opened 2015-04-29
No need to generate test coverage reports when running locally.

floehopper:
I based this on [how this was done in Whitehall](https://github.com/alphagov/whitehall/commit/59886b6aa9e11ce398fbf5d20f69ad58148eafd1).

I have verified that the test coverage report is still [generated as part of the Jenkins build](https://ci-new.alphagov.co.uk/job/govuk_smartanswers_branches/1925/console).

### DIFF
--- a/jenkins.sh
+++ b/jenkins.sh
@@ -13,6 +13,6 @@ export GOVUK_APP_DOMAIN=dev.gov.uk
 export GOVUK_ASSET_HOST=http://static.dev.gov.uk
 
 export DISPLAY=:99
-RAILS_ENV=test SPEC_REPORTER=true bundle exec rake test
+RAILS_ENV=test SPEC_REPORTER=true TEST_COVERAGE=true bundle exec rake test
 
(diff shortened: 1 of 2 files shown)


tadast:
What's the reason? Does this make the test suite faster? :)

floehopper:
Yeah. Sorry. I find it particularly slow when I'm just running a few tests at a time.

Out of interest, do you look at the coverage reports?

tadast:
Never. I was actually going to investigate how much time does simplecov add to the suite at some point.
Would be curious to see the numbers, but if you say it feels faster without it that's good enough :)

:shipit: :+1:

floehopper:
When running a single smart_answer_flow integration test case (8 tests, 50 assertions), it takes nearly 10 seconds with SimpleCov, but only 2.5 seconds without.

tadast:
(╯°□°）╯︵ ʌoɔǝןdɯıs

PR #1596, opened 2015-04-29
SPIKE: Inlining question text and options

chrisroos:
__DO NOT MERGE.__ This code doesn't work and the build will fail.

In a similar(ish) vein to #1577 I wondered how it would look/feel if we moved the question text from the YAML files to the Ruby DSL.

I _think_ this might make it slightly easier for developers to amend existing Smart Answers but I'm really not convinced about this approach.

Does anyone else have any thoughts? @bradleywright, @tadast, @BenJanecke, @heathd, @dsingleton, @floehopper.


BenJanecke:
Not a big fan of this,  With this syntax markdown formatting is very likely to break.

chrisroos:
@BenJanecke: Can you elaborate on, "With this syntax markdown formatting is very likely to break", please?

BenJanecke:
With the heredoc format 

```
<<-EndBody
This includes:

* talking to someone at the benefits office who explained the decision, either on the phone or face to face
* a written 'statement of reasons' explaining the decision
EndBody
```

you would always have to push content to the far left to avoid indentation issues

tadast:
Pros I can see:

- Easier to trace and change questions and answer options (happens rarely in comparison to outcome changes)
- YML files become outcome text and phrases for interpolation only

Cons I can see:

- Mixing data and logic, might introduce some data duplication (not very likely though)
- Ugly and potentially more error prone indentation for multiline strings
- Would require a lot of work to transform all classic smartanswers or leave us hanging in the limbo state. 

All in all it's a nice change, but given it has some downsides as well I'm not sure ROI is good enough to proceed.

heathd:
I do think that having logic and text copy in one file is advantageous provided that the logic conditions are understandable within the context of the text (appropriately named variables etc). This means that someone reading the file can more-or-less understand the different permutations of outcome text and in which circumstances they would apply, and therefore be able to modify the text confidently.

As far as I know, smartdown goes most of the way towards supporting this. There are some concerns about the work of migrating smart answers to smartdown, but given the [partial automation](https://github.com/alphagov/smartanswer-to-smartdown) of that conversion I think it would be do-able. There may also be issues with the smartdown syntax itself but I think those are also surmountable.

I agree with Brad's comment on the other PR that the disadvantage of ERB is that it doesn't provide a sandboxed language, and therefore could not be used by non-developers.

chrisroos:
> you would always have to push content to the far left to avoid indentation issues

@BenJanecke: I can see that there would be problems if the content wasn't pushed all the way to the left but I'm not sure it's any worse than the current situation where the content needs to be indented correctly in the YAML file.

chrisroos:
> All in all it's a nice change, but given it has some downsides as well I'm not sure ROI is good enough to proceed.

@tadast: I think that's my general feeling. It feels like it's a bit better but not necessarily enough to warrant the change.

chrisroos:
> As far as I know, smartdown goes most of the way towards supporting this.

@heathd: While there are certainly benefits, I guess we're not yet convinced that Smartdown is the right approach. In these sort of pull requests, we're exploring some smaller, incremental changes to Ruby Smart Answers so that we can start improving the majority of the Smart Answers that exist without going for a wholesale conversion to Smartdown.

floehopper:
@heathd: [My comment on the ERB PR](https://github.com/alphagov/smart-answers/pull/1577#issuecomment-97726847) might be relevant too.

chrisroos:
Closing as we're not planning to do anything with this in the short term.

PR #1598, opened 2015-04-30
Revert "Write should_eventually test for error messages"

floehopper:
This reverts commit 8affa3be8c2a08f5cc76e75e0737bd7df9d6b9e5.

This DEFERRED test was introduced over 3 years ago and it still doesn't pass.

I don't see the benefit of keeping it around, especially when it introduces
noise into the test output.

### DIFF
--- a/test/unit/flow_test.rb
+++ b/test/unit/flow_test.rb
@@ -205,11 +205,6 @@ class FlowTest < ActiveSupport::TestCase
       should "truncate path after error" do
         assert_equal [:do_you_like_chocolate?], @flow.path(%w{no bad})
       end
-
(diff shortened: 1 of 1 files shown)


chrisroos:
Assuming this functionality doesn't exist (in which case we might be able to get the test to pass) then I definitely agree that we should get rid of it.

tadast:
I guess the original idea of the author was to annoy the devs with that message enough to implement this functionality. 

Worth moving the idea of custom errors to issues after merging this?

:+1:

floehopper:
As per @tadast's suggestion, I've [added an issue](https://github.com/alphagov/smart-answers/issues/1605) for this, so that I can feel better about removing the test.

PR #1599, opened 2015-04-30
Merge v2 of lost or stolen passport

BenJanecke:
For Canada route, abroad only. Change opening line to: Fill in form LS01 and post it to your nearest British embassy, high commission or consulate.

https://www.gov.uk/report-a-lost-or-stolen-passport/y/abroad/canada


BenJanecke:
@tadast should be published now

PR #1602, opened 2015-05-01
Configure Errbit for preview env on heroku

tadast:
@chrisroos can you take a look please? As far as I can tell the changes have no impact on dev environtment, in other environments they will get overwritten by deploy script anyway.

:heavy_check_mark: Tested with `heroku run rake airbrake:test` and it works


config/initializers/airbrake.rb
```diff
@@ -1,7 +1,12 @@
...
 Airbrake.configure do |config|
+  config.api_key = ENV["ERRBIT_API_KEY"]
+  config.host    = ENV["ERRBIT_HOST"]
+  config.environment_name = ENV["ERRBIT_ENV"]
+  config.secure  = ENV["ERRBIT_API_KEY"].present?
+
   # Adding "production" to the development environments causes Airbrake not
   # to attempt to send notifications.
-  config.development_environments << "production"
+  config.development_environments << "production" unless ENV["ERRBIT_API_KEY"].present?
```
chrisroos: @tadast: Do you understand why this line was there in the first place? The original commit (727c3ad9049171ba8fdcb7906c194e63cbbcaa59) doesn't offer any clues. My guess would be that it's to prevent Airbrake from sending exceptions even if the environment is set to `production` on a developer machine.

I wonder whether we should be checking for `RUNNING_ON_HEROKU` as well as `ERRBIT_API_KEY`?

chrisroos:
I think you should mention that airbrake.rb is overwritten in the commit note (at the moment it's just in this pull request). It'd be great if you were able to link to the code that actually overwrites airbrake.rb too.

startup_heroku.sh
```diff
@@ -29,5 +30,6 @@ export SMART_ANSWER_TO_OPEN="marriage-abroad"
 open $HEROKU_URL$SMART_ANSWER_TO_OPEN
 
 echo
+echo "*Set ERRBIT_API_KEY and ERRBIT_HOST manually to enable error reporting*"
 echo "All done"
```
chrisroos: Should this tell you to set `ERRBIT_ENV` as well as `ERRBIT_API_KEY` and `ERRBIT_HOST`?

startup_heroku.sh
```diff
@@ -29,5 +30,6 @@ export SMART_ANSWER_TO_OPEN="marriage-abroad"
 open $HEROKU_URL$SMART_ANSWER_TO_OPEN
 
 echo
+echo "*Set ERRBIT_API_KEY and ERRBIT_HOST manually to enable error reporting*"
```
chrisroos: Can we print some information about where to find the `ERRBIT_API_KEY` and `ERRBIT_HOST`? I wouldn't know where to find those at the moment.

chrisroos:
Even with these changes, it'd still be easy to deploy the app to Heroku without realising that Errbit isn't configured. Maybe that's not such a big deal, though.

config/initializers/airbrake.rb
tadast: > My guess would be that it's to prevent Airbrake from sending exceptions even if the environment is set to production on a developer machine.

This is my understanding too. @alext is this correct (I saw you added it [here](https://github.com/alphagov/router-api/blame/master/config/initializers/airbrake.rb))?

> I wonder whether we should be checking for RUNNING_ON_HEROKU as well as ERRBIT_API_KEY?

I think if you have this ENV var set it's pretty clear that you are expecting to communicate with Airbrake. Heroku is out of scope in this file, we would assume too much by adding it here imho.

alext: Prior to https://github.com/airbrake/airbrake/pull/268 this was necessary to prevent the airbrake gem attempting to send error notices to airbrake when not configured (configuration was done by overwriting this file).  If this is using airbrake 3.1.17 or later, this should no longer be necessary.

tadast:
> I think you should mention that airbrake.rb is overwritten in the commit note

:heavy_check_mark: 

> It'd be great if you were able to link to the code that actually overwrites airbrake.rb too.

It's not public and I don't have access to it :|

> Even with these changes, it'd still be easy to deploy the app to Heroku without realising that Errbit isn't configured. Maybe that's not such a big deal, though.

I agree, it's not a big deal.

Anything else @chrisroos :) ?

chrisroos:
All looks good to me, @tadast. Good job!

PR #1604, opened 2015-05-05
Pass `response` into `calculate` blocks.

floehopper:
* Brings such blocks into line with other types of blocks e.g. `validate`,
`define_predicate`, `next_node`.

* Reduces duplication and makes code slightly terser.

* I've only declared the block argument where it is actually needed i.e. there
are still many `calculate` blocks which do not have a `response` argument
declared.

* This may seem like a small change, but I think it will make some other
refactoring easier.


tadast:
Looks good to me :+1:

floehopper:
@tadast: Thanks. I've just merged it.

PR #1606, opened 2015-05-05
Set 2010/2011 rate for statutory sick pay

BenJanecke:


tadast:
@BenJanecke don't forget to mark the issue on errbit as resolved once deployed :+1:

PR #1608, opened 2015-05-07
Allow parsing of date question responses into date objects instead of strings

floehopper:
This PR add an optional `parse` option to `date_question` which when `true`
causes the response to be a `Date` object and not just a date `String`.

Having parsed the response into a `Date` object it makes no sense to me to then
convert it back into a `String`. There seems to be a lot of code in the Ruby DSL
flows (and their calculators) which has to re-parse the date
strings. Also some date "variables" are `Date` objects and some are date strings
in these smart answers which is really confusing.

The plan is ultimately to **always** supply the response to a
`SmartAnswer::Question::Date` as a Date object and to only use `Date` objects
within the Ruby DSL flows and associated calculators. However, to get there
incrementally I'm introducing this option so we can convert one question or
smart answer flow at a time.

This PR includes changing the `appeal-a-benefits-decision` smart answer to use
the new option as an example of how it works. We plan to introduce similar
changes to other smart answers directly into `master` in order to reduce the
possibility of merge problems.

We think that it will be worth adopting this pattern for all question types,
but we have to start somewhere. The thinking is that the Ruby DSL flows
and their associated calculators should operate on richer Ruby objects where
possible rather than just strings.

### DIFF
--- a/app/presenters/date_question_presenter.rb
+++ b/app/presenters/date_question_presenter.rb
@@ -1,6 +1,6 @@
 class DateQuestionPresenter < QuestionPresenter
   def response_label(value)
-    I18n.localize(Date.parse(value), format: :long)
+    I18n.localize(value.to_date, format: :long)
   end
 
   def start_date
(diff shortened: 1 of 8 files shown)


chrisroos:
This looks good to me.

PR #1614, opened 2015-05-08
Notify about missing phrases via Airbrake

tadast:
In https://github.com/alphagov/smart-answers/commit/3e9c2909758c480bfb4fd2897f20748d051739e6 a change was introduced to silently fall back to a phrase key if the phrase is not defined in the YML file. 

I think developers should know about cases like this. Hence I'm adding an airbrake notification, but leaving the UX as it was at least initially. Perhaps later we can just raise those exceptions so we know about it in dev/test environments as well as staging/production.

Thoughts?

### DIFF
--- a/app/presenters/node_presenter.rb
+++ b/app/presenters/node_presenter.rb
@@ -32,7 +32,12 @@ class NodePresenter
     when ::SmartAnswer::PhraseList then
       if nested == false
         value.phrase_keys.map do |phrase_key|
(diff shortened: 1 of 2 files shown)


floehopper:
This change is definitely a good step in the right direction. :+1:

In an ideal world, this kind of problem would be discovered when the tests were run or (in some even more hypothetical future) at some kind of "validation" time when the content was created.

tadast:
Totally agree about early feedback.

PR #1615, opened 2015-05-08
Phraselist and test cleanup for marriage abroad

tadast:
I've extracted all the keys from marriage abroad yml and matched them with the ruby file. As expected, there where a bunch of unused phrases. Some of them where harder to find as keys are interpolated. I'm fairly confident I haven't removed anything that isn't used, but a pair of extra :eyes: would be helpful.

Also, I did [this](https://github.com/alphagov/smart-answers/pull/1614) to increase confidence.

I've also noticed that some of the tests where not run. This is due to top-level context descriptions not being unique (only the last non-unique context is executed).

### DIFF
--- a/lib/smart_answer_flows/locales/en/marriage-abroad.yml
+++ b/lib/smart_answer_flows/locales/en/marriage-abroad.yml
@@ -207,9 +207,6 @@ en-GB:
         complete_affidavit: |
           You’ll need to complete an ‘Affidavit for marriage’ form that will be used for your affidavit. Take the completed form with you to your appointment.
 
-        complete_affidavit_with_download_link: |
-          You’ll need to complete an [‘Affidavit for marriage’ form](/government/publications/marriage-in-%{country_name_lowercase_prefix}) that will be used for your affidavit. Take the completed form with you to your appointment.
-
         download_affidavit: |
           $D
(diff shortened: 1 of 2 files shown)


floehopper:
LGTM. I've only checked a subset of the phrase key changes, but they were all as expected. :+1:

I wonder whether it would be better to add explicit entries in the locale file for the interpolated keys. I worry slightly that after the introduction of #1614, we'll be inundated with missing key exceptions. By adding explicit keys we could trade-off some "duplication" against simpler behaviour (always expect key to be present) and increased explicitness.

tadast:
> I wonder whether it would be better to add explicit entries in the locale file for the interpolated keys.

I'm not sure I follow. Interpolated keys are in the locale file; they are [interpolated in the ruby flow file](https://github.com/alphagov/smart-answers/blob/master/lib/smart_answer_flows/marriage-abroad.rb#L1358). We could add comments to the ruby files of some interpolated variants to help with easier discovery using global search, but they will get out of sync quickly.

Are we talking about different things? Sorry if I didn't make it clear in the description :)

floehopper:
@tadast: Sorry. I did get the wrong end of the stick. Please ignore me.

PR #1617, opened 2015-05-11
Remove can-i-get-a-british-passport Smart Answer

chrisroos:
* Added on 7th Mar 2012 - ae39fa485477592718df060b49f054c691788c41
* Last updated on 7th Mar 2012 - 1f27b1d19dad7e86ed161bcce49c7fb321930cfe
* Never published
* No associated test

Gwen and Alex have confirmed that this can be removed.


chrisroos:
My motivation for opening this pull request is to run the tests and to see whether people spot anything else that can be removed in this branch.

floehopper:
LGTM. I can't see anything else which can be removed.

lib/smart_answer_flows/can-i-get-a-british-passport.rb
```diff
@@ -1,75 +0,0 @@
-status :draft
-satisfies_need "100983"
```
tadast: This makes me think: do they need too update anything in Maslow (user needs tracker)? Mind checking with Gwen?

tadast:
:shipit:

lib/smart_answer_flows/can-i-get-a-british-passport.rb
chrisroos: Good question. I'll find out.

chrisroos: @tadast: Liz says, "I think we're fine - at least 3 of those would have content besides the smart answer, so if there are any needs in maslow, they'll be covered."

PR #1622, opened 2015-05-11
Publish changes to minimum vage calculators

tadast:
1) Update helpline numbers

2) Add apprentice rate
The apprentice rate is for apprentices aged 16 to 18 and those
aged 19 or over who are in their first year. All other apprentices
are entitled to the National Minimum Wage for their age.

@jackscotti you should be the best person to review this as you're the author of some of those changes :) Thanks


jackscotti:
:+1:  LGTM

PR #1625, opened 2015-05-11
Log missing phrase keys

tadast:
in 3e9c290 a change was introduced
to silently fall back to a phrase key if the phrase is not defined
in the YML file. I think developers should know about cases like
this. Hence I'm logging it, but leaving the UX as it was at least
initially. Perhaps later we can just raise the exception so we know
about it in dev/test environments as well as staging/production.

This overwrites the changes in https://github.com/alphagov/smart-answers/pull/1614 because they generated too much noise in staging and messed up metrics and alerts

/cc @jennyd @floehopper

### DIFF
--- a/test/unit/node_presenter_test.rb
+++ b/test/unit/node_presenter_test.rb
@@ -74,7 +74,7 @@ module SmartAnswer
       state.phrases = PhraseList.new(:four, :one, :two, :three)
       presenter = NodePresenter.new("flow.test", outcome, state)
 
-      Airbrake.expects(:notify_or_ignore).once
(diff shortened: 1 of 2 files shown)


test/unit/node_presenter_test.rb
```diff
@@ -74,7 +74,7 @@ def teardown
       state.phrases = PhraseList.new(:four, :one, :two, :three)
       presenter = NodePresenter.new("flow.test", outcome, state)
 
-      Airbrake.expects(:notify_or_ignore).once
+      Rails.logger.expects(:warn).with("[Missing phrase] The phrase being rendered is not present: flow.test.phrases.four. Responses: ").once
```
tadast: This is a bit of a bad example, but in real life this will log the path that leads to the outcome that has missing phrases.

PR #1627, opened 2015-05-12
Improve missing keys logging

tadast:
Only log when a phrase is missing in an outcome. Otherwise there is too much noise as all the phrases are being rendered in each step, even if the nested interpolation keys are missing.

Use the translation key from the error message: more accurate than what we had before.

A realistic log message in real life would look like this:

```
[Missing phrase] The phrase being rendered is not present: flow.maternity-paternity-calculator.phrases.maternity_leave_table	Responses: maternity/2015-08-11/yes/2015-05-26/yes/yes/2015-04-29/2015-02-28/monthly/4000.0/usual_paydates/last_working_day_of_the_month/1,2,3,4,5
```

/cc @floehopper

### DIFF
--- a/app/presenters/node_presenter.rb
+++ b/app/presenters/node_presenter.rb
@@ -32,11 +32,10 @@ class NodePresenter
     when ::SmartAnswer::PhraseList then
       if nested == false
         value.phrase_keys.map do |phrase_key|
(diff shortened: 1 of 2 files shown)


floehopper:
LGTM

PR #1629, opened 2015-05-12
Allow parsing of value_question responses into Integers instead of Strings

floehopper:
This PR adds an _optional_ `parse` option to `value_question` which when set to
`Integer` causes the response to be an `Integer` object and not just a `String`.

There seems to be a lot of code in the Ruby DSL flows which has to parse the
response from a `value_question` into an `Integer`. It feels as if it would be
simpler if the Ruby DSL flows always dealt with richer non-string value types
and the parsing was dealt with in one place in
`SmartAnswer::Question::Value#parse_input`.

I've chosen to specify the `Integer` type as the `parse` option value, because it
looks as if we'll need the flexibility to handle genuine `String` or even `Float`
responses.

Making the option optional means that I'll be able incrementally convert one
question or smart answer flow at a time to use the new parsing mechanism.

This PR includes changing the `calculate-agricultural-holiday-entitlement`
smart answer to use the new option as an example of how it works. I plan
to introduce similar changes to other smart answers directly into `master` in
order to reduce the possibility of merge problems.

This is part of a plan to adopt this kind of mechanism for all question types.

c.f.  #1608 where we did a similar thing for `date_question`.

### DIFF
--- a/lib/smart_answer/flow.rb
+++ b/lib/smart_answer/flow.rb
@@ -64,8 +64,8 @@ module SmartAnswer
       add_node Question::Date.new(name, &block)
     end
 
-    def value_question(name, &block)
-      add_node Question::Value.new(name, &block)
+    def value_question(name, options = {}, &block)
+      add_node Question::Value.new(name, options, &block)
     end
 
     def money_question(name, &block)
--- a/lib/smart_answer/question/value.rb
+++ b/lib/smart_answer/question/value.rb
@@ -1,6 +1,18 @@
 module SmartAnswer
   module Question
     class Value < Base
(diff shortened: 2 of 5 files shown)


chrisroos:
Did you consider introducing an `integer_question` rather than adding an option to `value_question`? Are the pros/cons of each approach?

chrisroos:
I've added a couple of questions but this looks good to me irrespective of the answers.

floehopper:
@chrisroos: I did consider adding an `integer_question`, but adding the option seemed like a "smaller" change. I don't rule out the idea of introducing such a question type in the future.

PR #1631, opened 2015-05-13
Add :to_i parse option to value_questions

floehopper:
Following on from #1629, this PR provides a less strict integer parsing option
for `value_question`.

Some `value_questions` are using `String#to_i` in `calculate` blocks to parse
string responses into integers. Ideally we'd be consistently using the same
parsing mechanism for all value questions expecting an integer response, but
it's difficult to determine whether the differences are by accident or design.
So for the moment, providing two mechanisms i.e. using `Integer()` or
`String#to_i` feels like a small improvement and at least makes the difference
more explicit. We can decide whether to try to combine these alternative
mechanisms later.

It's worth noting that using the new `:to_i` parse option instead of converting
the string response within a `calculate` block in the flow will result in a
few small changes in behaviour. If the user enters a non-integer string
response, the **parsed** `Integer` will be used in (a) the URL path segment;
(b) the "Previous Answers" section; and (c) the pre-filled value if the user
clicks on the "Change" link for the relevant previous question.

I think this new behaviour, while different, is actually better. Also the old
URL paths and query strings will still work sensibly. So I'm happy to go with
it.

This PR includes an example of using the new `:to_i` option, but I plan to
introduce similar future changes directly into `master` to minimise merge
problems.

### DIFF
--- a/lib/smart_answer/question/value.rb
+++ b/lib/smart_answer/question/value.rb
@@ -9,6 +9,8 @@ module SmartAnswer
       def parse_input(raw_input)
         if Integer == @parse
           Integer(raw_input)
+        elsif :to_i == @parse
+          raw_input.to_i
         else
           super
         end
--- a/lib/smart_answer_flows/shared_logic/maternity-calculator.rb
+++ b/lib/smart_answer_flows/shared_logic/maternity-calculator.rb
@@ -193,9 +193,9 @@ multiple_choice :when_in_the_month_is_the_employee_paid? do
 end
 
 ## QM13
-value_question :what_specific_date_each_month_is_the_employee_paid? do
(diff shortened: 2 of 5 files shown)


chrisroos:
> It's worth noting that using the new :to_i parse option instead of converting...

I wonder whether it'd be worth adding an example of these changes? Something like:

## Before

* User responds with 1.1
* URL contains "/1.1"
* "Previous Answers" contains "1.1"
* Textbox contains "1.1" after clicking "Change" within "Previous Answers"

## After

* User responds with 1.1
* URL contains "/1"
* "Previous Answers" contains "1"
* Textbox contains "1" after clicking "Change" within "Previous Answers"

floehopper:
@chrisroos: Thanks. I'll add that to the commit note now.

chrisroos:
Looks good to me.

PR #1635, opened 2015-05-14
Add St Martin and Saint-Barthélemy to marriage abroad logic

tadast:
Those two countries were recently added to the worldwide API
but had no corresponding logic to handle them. This change leads
to "no British consular facilities" outcome for both countries,
as advised by FCO.

Some paths affected:

/marriage-abroad/y/st-martin/other/st-martin/partner_british/same_sex

/marriage-abroad/y/saint-barthelemy/other/afghanistan/partner_british/opposite_sex


tadast:
FCO got back to us, I've amended the commit and the message. I had to introduce an exception for those two countries, explanation in the commit message :|

tadast:
@floehopper I've updated the commit message to explain the `_os_` reference.

floehopper:
LGTM :+1:

PR #1636, opened 2015-05-14
Serve static assets when running on heroku

tadast:
This explains it https://github.com/heroku/rails_12factor#rails-4-serve-static-assets

@chrisroos

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -11,7 +11,8 @@ SmartAnswers::Application.configure do
   config.action_controller.perform_caching = true
 
   # Disable Rails's static asset server (Apache or nginx will already do this)
-  config.serve_static_files = false
+  # unless running on Heroku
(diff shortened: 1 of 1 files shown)


config/environments/production.rb
```diff
@@ -11,7 +11,7 @@
   config.action_controller.perform_caching = true
 
   # Disable Rails's static asset server (Apache or nginx will already do this)
-  config.serve_static_files = false
+  config.serve_static_files = !!ENV['RUNNING_ON_HEROKU']
```
floehopper: I'm not a big fan of the double-bang - I don't think it's very readable. How about using `ENV['RUNNING_ON_HEROKU'].present?` instead?

It would also be nice to make this and the [other occurrence](https://github.com/alphagov/smart-answers/blob/206a55d6b1c66a99212e7eec049065a6f7735863/config/environments/production.rb#L68) consistent.

floehopper:
Other than my small comment, this looks good to me. :+1:

PR #1637, opened 2015-05-14
Add overseas passport logic for St Martin

tadast:
We have been told the procedure is the same as for the Dutch side of the same island, St Maarten. So it should be safe to merge without FC.

This will stop 500'ing for users in St Maarten.

We're still trying to find out the procedure in Saint-Barthélemy.

### DIFF
--- a/lib/data/passport_data.yml
+++ b/lib/data/passport_data.yml
@@ -1711,6 +1711,15 @@ st-maarten:
   renewing_old: 8_weeks
   applying: 8_weeks
   replacing: 8_weeks
+st-martin:
+  type: ips_application_1
+  group: ips_documents_group_1
+  app_form: hmpo_2_application_form
+  online_application: true
+  renewing_new: 6_weeks
+  renewing_old: 8_weeks
+  applying: 8_weeks
+  replacing: 8_weeks
 st-pierre-and-miquelon:
   type: ips_application_1
   group: ips_documents_group_1
--- a/test/integration/smart_answer_flows/overseas_passports_test.rb
+++ b/test/integration/smart_answer_flows/overseas_passports_test.rb
@@ -8,7 +8,7 @@ class OverseasPassportsTest < ActiveSupport::TestCase
   include GdsApi::TestHelpers::Worldwide
 
   setup do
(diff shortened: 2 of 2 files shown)


lib/data/passport_data.yml
```diff
@@ -1711,6 +1711,15 @@ st-maarten:
...
   replacing: 8_weeks
+st-martin:
+  type: ips_application_1
+  group: ips_documents_group_1
+  app_form: hmpo_2_application_form
+  online_application: true
+  renewing_new: 6_weeks
+  renewing_old: 8_weeks
+  applying: 8_weeks
+  replacing: 8_weeks
```
floehopper: Did you consider using a YAML anchor & alias to make it clearer that this data is intended to be an exact copy of the data for `st-maarten`? I suspect we're not doing that anywhere else, so perhaps it's a bad idea.

test/integration/smart_answer_flows/overseas_passports_test.rb
```diff
@@ -1311,4 +1311,16 @@ class OverseasPassportsTest < ActiveSupport::TestCase
...
+      worldwide_api_has_no_organisations_for_location('st-martin')
+      add_response 'st-martin'
+      add_response 'renewing_new'
+      add_response 'adult'
+
+      assert_current_node :ips_application_result_online
+      assert_phrase_list :how_long_it_takes, [:how_long_6_weeks, :how_long_additional_info_renewing_new, :how_long_additional_time_online]
+    end
+  end
 end
```
floehopper: I can't see an equivalent test for "St Maarten". Is that expected?

floehopper:
Other than my small comments, this looks good to me. :+1:

test/integration/smart_answer_flows/overseas_passports_test.rb
tadast: It wasn't tested before, because it was not an exceptional case perhaps. I'll add a test and make sure it's clear behaviour has to be the same for both sides of the island :)

lib/data/passport_data.yml
tadast: There are a lot of duplicated settings, some of them could be shared between 60 or so countries. I do not now if this is incidental or if there's anything unifying those countries. I also don't know how likely are the rules on the two sides of the island to change (e.g. marriage rules are different already). I'll leave it as it is for now.

PR #1638, opened 2015-05-14
[Don't merge] Make test coverage data available for smart answer flows

floehopper:
## DO NOT MERGE

This is a "spike" which demonstrates how to obtain test coverage data for
the `additional-commodity-code` flow. This data was not [previously
available][1], because the flow files were being read and eval'ed rather than
simply required. The latter is [necessary for SimpleCov to record coverage
data][2] for the file.

By extracting the existing DSL code into a `define` method on a subclass of
`SmartAnswer::Flow`, we can safely require the file separately from
instantiating the flow. This feels like a better state of affairs in general,
but it specifically addresses the test coverage problem above.

We'd really like to have this coverage data available so we can have confidence
that we're not breaking anything as we continue to refactor the app.

Note that I've intentionally not fixed the indentation in the
`additional-commodity-code` flow file so that it's easier to see my changes.

I think it would be simple enough to make this change for all smart answer
flows and then the code in `SmartAnswer::FlowRegistry#build_flow` could be
considerably simplified.

[1]: https://ci-new.alphagov.co.uk/job/govuk_smart_answers/2396/rcov/
[2]: https://github.com/colszowka/simplecov/issues/38

### DIFF
--- a/lib/smart_answer/flow_registry.rb
+++ b/lib/smart_answer/flow_registry.rb
@@ -1,3 +1,5 @@
+require "smart_answer_flows/additional-commodity-code"
+
 module SmartAnswer
   class FlowRegistry
     class NotFound < StandardError; end
@@ -51,10 +53,16 @@ module SmartAnswer
     end
 
     def build_flow(name)
-      absolute_path = @load_path.join("#{name}.rb").to_s
-      Flow.new do
-        eval(File.read(absolute_path), binding, absolute_path)
-        name(name)
+      if name == 'additional-commodity-code'
+        flow = AdditionalCommodityCode.new
+        flow.define
+        flow
+      else
+        absolute_path = @load_path.join("#{name}.rb").to_s
+        Flow.new do
+          eval(File.read(absolute_path), binding, absolute_path)
+          name(name)
+        end
       end
     end
 
--- a/lib/smart_answer_flows/additional-commodity-code.rb
+++ b/lib/smart_answer_flows/additional-commodity-code.rb
@@ -1,3 +1,10 @@
+module SmartAnswer
+
+class AdditionalCommodityCode < Flow
+
+  def define
+name 'additional-commodity-code'
+
 status :published
 satisfies_need "100233"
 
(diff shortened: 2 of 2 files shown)


tadast:
I don't mind if all the flows become ruby classes and we get rid of `eval` :) The biggest problem I see is it will make it a bit more difficult to look at the git history before this change, since we'll need to indent all the code in flows.

Is SimpleCov smart enough to measure test coverage in a  DSL like ours? If so and if you trust those stats, then it's a worthwhile change perhaps :)

floehopper:
@tadast:

* I share your concern about the indentation affecting the git history, but I wonder whether we can get around that by using [`git diff --ignore-space-change`][1] and/or [appending `?w=1` to GitHub diff URLs][2].

* Yes, SimpleCov can measure test coverage for the smart answer DSL - at the end of the day, it's just Ruby code! Using the changes in this PR I was able to run the integration test for the `additional-commodity-code` flow and generate a coverage report [including coverage for the DSL part of the code][3].

[1]: https://www.kernel.org/pub/software/scm/git/docs/git-diff.html#_options
[2]: https://github.com/blog/967-github-secrets
[3]: https://dl.dropboxusercontent.com/u/1584612/lib_smart_answer_flows_additional-commodity-code.rb.html

chrisroos:
I'm very much in favour of this!

We're hoping to introduce some fairly comprehensive regression tests around the existing smart answers. Simplecov helps us by allowing us to build up the smallest set of inputs that exercise most/all of the flows.

tadast:
LGTM then, an extra step when browsing git history is a small price to pay for increased confidence :) :+1:

floehopper:
I've just made these changes in a separate branch and merged it in [this commit][1].

Note that I have purposefully not yet changed the following:

* calculate-employee-redundancy-pay
* calculate-your-redundancy-pay
* marriage-abroad

Closing this PR which was just a spike.

[1]: 2910747663879e0b8d42fc32ff9fa41649ab49bf

PR #1639, opened 2015-05-15
Uprate adoption calculator

BenJanecke:
* Check the due date and show the lel at the beginning of that period
* Check if due_date is in a specific range and show the rate for that period
* Check the match_date and show the lel at the beginning of that period
* Check if match_date is in a specific range and show the rate for that period

To test visit 
/pay-leave-for-parents-adoption/y/yes/no/2014-1-1/employee/employee/yes/yes/400-week/yes/yes/yes/400-week/yes

You should see rates for both 2013/2014 and 2014/2015


tadast:
Other than one comment it's good to merge. 

It's not LGTM though, we should consider rewriting it in ruby as complexity of changes like this does not seem to be getting better and the next person to work on this thing next is going to require a lot of time and patience to understand what's going on here.

tadast:
:shipit:

PR #1640, opened 2015-05-18
Respond with 404 not 500 when format is unknown

floehopper:
This addresses a set of exceptions which appear regularly in Errbit e.g. [these ones][1].

I've been able to reproduce the `500 Internal Server Error` locally and to use
a `rescue_from` statement in `ApplicationController` to respond with a
`404 Not Found` instead. The latter seems to be more appropriate and will
reduce the noise we're seeing in the Errbit reports.

Note that there appears to be some slightly odd behaviour when the unknown
format is `.js` in that the text "404 error" rendered by
`ApplicationController#error` is rendered **without** any layout. This appears
to be behaviour provided by `slimmer`. Although this behaviour seems odd to me
it feels out of scope of this change and in any case, this commit should still
make things better than they are at the moment.

[1]: https://errbit.production.alphagov.co.uk/apps/533c35ae0da1159384044f5f/problems/55583e4765786355f0919a00

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -6,6 +6,7 @@ class ApplicationController < ActionController::Base
   before_action :set_analytics_headers
 
   rescue_from GdsApi::TimedOutException, with: :error_503
+  rescue_from ActionController::UnknownFormat, with: :error_404
 
 protected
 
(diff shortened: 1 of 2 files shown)


tadast:
LGTM :+1:

chrisroos:
Looks good to me.

PR #1642, opened 2015-05-18
Add regression tests for student finance calculator

chrisroos:
We want tests that give us confidence to make some larger changes to the existing Smart Answers (e.g. switching to ERB templates).

The current integration tests aren't good enough as they don't exercise the flows completely, and they rely on using the internals of the flows (e.g. tests that assert things about the phrase lists).

These regression tests aim to comprehensively test the Smart Answers and only rely on the generated html, which should allow us to change the implementation without worrying about breaking the tests.

Testing Smart Answers in this way involves:

1. Creating a questions-and-responses file that contains a set of responses that should be used for each question (see scripts/generate-questions-and-responses-for-smart-answer.rb).

2. Creating a responses-and-expected-results file that contains each combination of results from the questions-and-responses file, along with the node name and whether it's an outcome node (see scripts/generate-responses-and-expected-results-for-smart-answer.rb).

3. Running the smart_answers_regression_test to test the smart answer using each combination of responses from the responses-and-expected-results file.

Running the smart_answers_regression_test in this branch takes around 100 seconds, which isn't an insignificant amount of time. I have some ways of optimising the tests so that they only run when changes to the flows or input responses mean that the saved outcomes might be different. I didn't want to add them to this pull request as I'm more interested in what people about the approach in general.

We don't imagine these tests being very long lived. We'd hope to use them while we're making changes to the Smart Answers and for them to be replaced with more focussed tests over time.

Any questions and/or thoughts about this approach?

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,46 @@ Test a single Smartdown flow by running:
 
 [smartdown-scenarios]: https://github.com/alphagov/smartdown/blob/master/doc/scenarios.md
 
+### Adding regression tests to Smart Answers
+
+1. Generate a set of responses for the flow that you want to add regression tests to.
+
+        $ rails r script/generate-questions-and-responses-for-smart-answer.rb <name-of-smart-answer>
+
+2. Commit the generated questions-and-responses.yml file (in test/data) to git.
+
+3. Change the file by adding/removing and changing the responses:
+
+  * Add responses for any of the TODO items in the file.
+
+  * Remove responses that you don't think cause the code to follow different branches, e.g. it might be useful to remove all but one (or a small number) of countries to avoid a combinatorial explosion of input responses.
+
+  * Combine responses for checkbox questions where the effect of combining them doesn't affect the number of branches of the code that are exercised.
+
+4. Commit the updated questions-and-responses.yml file to git.
+
+5. Generate a set of input responses and expected results for the Smart Answer.
+
+        $ rails r script/generate-responses-and-expected-results-for-smart-answer.rb <name-of-smart-answer>
+
+6. Commit the generated responses-and-expected-results.yml file (in test/data) to git.
+
+7. Run the regression test to generate the HTML of each outcome reached by the set of input responses.
+
+        $ TEST_COVERAGE=true ruby test/regression/smart_answers_regression_test.rb
+
+8. Commit the generated outcome HTML files (in test/artefacts) to git.
+
+9. Inspect the code coverage report for the Smart Answer under test (`open coverage/rcov/index.html` and find the smart answer under test).
+
+  * If all the branches in the flow have been exercisde then you don't need to do anything else at this time.
+
+  * If there are branches in the flow that haven't been exercised then:
+
+      * Determine the responses required to exercise those branches.
+
+      * Go to Step 3, add the new responses and continue through the steps up to Step 9.
+
 ## Issues/todos
 
 Please see the [github issues](https://github.com/alphagov/smart-answers/issues) page.
--- a/lib/smart_answer_test_helper.rb
+++ b/lib/smart_answer_test_helper.rb
@@ -1,0 +1,30 @@
+class SmartAnswerTestHelper
+  def self.data_path
+    Rails.root.join('test', 'data')
+  end
+
+  def initialize(flow_name)
+    @flow_name = flow_name
+  end
+
+  def question_and_responses_path
+    data_path.join(question_and_responses_filename)
+  end
+
+  def write_questions_and_responses(questions_and_responses)
+    FileUtils.mkdir_p(data_path)
+    File.open(question_and_responses_path, 'w') do |file|
+      file.puts(questions_and_responses.to_yaml)
+    end
+  end
+
+  private
+
+  def question_and_responses_filename
+    "#{@flow_name}-questions-and-responses.yml"
+  end
+
+  def data_path
+    self.class.data_path
+  end
+end
--- a/lib/smart_answer_test_helper.rb
+++ b/lib/smart_answer_test_helper.rb
@@ -18,10 +18,24 @@ class SmartAnswerTestHelper
     end
   end
 
+  def responses_and_expected_results_path
+    data_path.join(responses_and_expected_results_filename)
+  end
+
+  def write_responses_and_expected_results(data)
+    File.open(responses_and_expected_results_path, 'w') do |file|
+      file.puts(data.to_yaml)
+    end
+  end
+
   private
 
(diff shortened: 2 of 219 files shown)


tadast:
This will help immensely when refactoring smartanswers. Great commits with descriptive messages too :trophy: 

The generated HTML files are quite big, if you suggest committing them the smartanswers repo size will grow significantly. It will also pollute commit messages, because most of the changes we normally do change outcomes/flows legitimately. Did you commit those just for the demo purposes? If so then ignore this paragraph :)

Other than that it looks like a brilliant tool that will encourage me to refactor hairy parts in smartanswers :) In the past I've written brittle ad-hoc scripts to crawl the URLs I expected to not change to achieve what this PR does.

Also, it will need a nice writeup in README :+1:

chrisroos:
> The generated HTML files are quite big, if you suggest committing them the smartanswers repo size will grow significantly. It will also pollute commit messages, because most of the changes we normally do change outcomes/flows legitimately. Did you commit those just for the demo purposes? If so then ignore this paragraph :)

I was imagining that we *would* commit the HTML files to the repo. I think this is the only way we'll be able to have tests that fail when something's changed. I don't feel entirely comfortable about this approach and went back and forth between having a tool that runs outside of tests versus one that's part of the build process.

We might be able to avoid the commits becoming polluted by updating the YAML and HTML files in a separate commit on the branch that makes a change to the flow/s. This isn't ideal because it means we have a commit with failing tests, but it might be a suitable compromise at this point in time?

In the longer term, I'm hoping that we'll be able to reduce these sort of tests and therefore remove a lot of the HTML from the repo.

Do my comments address your concern about storing the HTML files in the repo, or do you think we should revisit that part of the idea?

floehopper:
@tadas:

I'm not sure it's obvious from Chris' description, but our plan in the short term is to **replace** the existing integration tests with tests like these. For the moment, Chris' tool will only replace the `assert_phrase_list` aspect of the tests, but we envisage extending it to cover asserting against the question nodes too. Hopefully knowing that these tests are intended to replace the existing integration tests makes it more obvious why we're thinking of committing the outcome HTML to the repo.

Do you have a specific concern about the repo size getting bigger?

I can see how including the outcome HTML in commits might make them more confusing, but I think it should be manageable.

tadast:
Ok, I somehow missed the intention to completely replace integration tests :) Then my concerns are:

1) how would one test-drive changes to the flow? By modifying the yaml files?
2) test descriptions are often a good medium to write up exceptional behaviour and reasons why the behaviour is the way it is, how would we do it with the new approach?
3) what we do most of the time is change phrases/flows in smart answers. I'm afraid this would cause a lot more failures. i.e. for a smartanswer with a countries drop down list a single change could result in 200+ failures. I'm afraid whoever is responsible for reviewing them and making sure the failures are caused by intended changes would just blindly accept the diff or spend hours verifying the failures are legit. Test driving a change like this would require some serious automation as well I imagine.

All in all I understand how this can be useful when refactoring without any externally visible changes, but I have doubts this would be useful for day-to-day changes Ben and I usually make. I would be curious to see if my concerns are relevant in practice, but the thing I'm working on (and will be for the next couple of weeks) is not suitable for trying it out. Would it be possible to have it as an optional tool first, see how useful it is in practice and then if we think it's good, replace flow integration tests with it.

> In the longer term, I'm hoping that we'll be able to reduce these sort of tests and therefore remove a lot of the HTML from the repo.

How would we do that? 

> Do you have a specific concern about the repo size getting bigger?

Cloning the repo would take longer, editors would need specific setup to exclude HTML dirs from global search paths to make search faster etc.

chrisroos:
I've tried to answer your questions below. Doing so has made us realise that @floehopper and I haven't done a particular good job of keeping you informed about how our thinking has changed over time. Can we get together tomorrow (Weds) and chat through this in person, @tadast and @BenJanecke?

> 1) how would one test-drive changes to the flow? By modifying the yaml files?

Modifying the YAML files should be possible, yes. But I can also imagine being in a place where we don't test-drive changes to the flow. If, for example, we end up with some kind of interface to allow non-developers to create/edit flows then I don't think they'd be test-driven. Instead, I think we'd want tools that give us information about the flows to ensure that they're not "broken" (for some definition of "broken") and to allow us to visualise them in some way to ensure they're working the way we expect.

> 2) test descriptions are often a good medium to write up exceptional behaviour and reasons why the behaviour is the way it is, how would we do it with the new approach?

I'd like to see us move to having the logic of the policies encapsulated in well tested objects that better describe the sort of behaviour you're talking about. At the moment, we have lots of Smart Answers that mix both the flow and policy logic, making it difficult to test one without the other. Splitting flow and policy logic should allow us to retain descriptive tests without having to have those tests exercise the flow itself.

> 3) what we do most of the time is change phrases/flows in smart answers. I'm afraid this would cause a lot more failures. i.e. for a smartanswer with a countries drop down list a single change could result in 200+ failures. I'm afraid whoever is responsible for reviewing them and making sure the failures are caused by intended changes would just blindly accept the diff or spend hours verifying the failures are legit. Test driving a change like this would require some serious automation as well I imagine.

This is related to question 1 in that I hope some new tools can help. I'd hope that rather than diff'ing the HTML, it'd be possible to at least diff the YAML (an intermediate representation of the flow) and ideally diff a visual representation of the flow to ensure the changes have the effect you want.

> Would it be possible to have it as an optional tool first, see how useful it is in practice and then if we think it's good, replace flow integration tests with it.

I certainly think this be possible. Even in its current form, we can selectively add these tests for the Smart Answers that we convert to ERB templates, i.e. hopefully not disrupt your work.

>> In the longer term, I'm hoping that we'll be able to reduce these sort of tests and therefore remove a lot of the HTML from the repo.

> How would we do that?

I think I've covered this in my answers to the previous questions. Hopefully, though, a combination of well-tested smaller objects describing the policy logic and flow logic would make the flows themselves much easier to test, either by writing actual tests or by using tools that verify they're behaving as expected.

> Cloning the repo would take longer, editors would need specific setup to exclude HTML dirs from global search paths to make search faster etc.

Certainly valid concerns.

tadast:
Thanks for the answer, it's starting to make more sense now! :) Let's talk tomorrow

chrisroos:
@floehopper, @tadast, @BenJanecke and I discussed this in some detail today.

We've agreed to merge this to master and to continue to review it as we start to add more of these regression tests. We're certain that the increase in test duration is going to cause us pain and we'll almost certainly tackle that soon. We think that the very tests themselves _might_ cause us some pain but we've agreed to tackle that as and when it arises.

PR #1644, opened 2015-05-19
[FOR DISCUSSION ONLY] Fix rubocop violations

chrisroos:
What do people think about Rubocop?

I'm in favour of getting it into our build process in order to enforce a consistent coding style. I care less about which particular style we use, and more about the consistency.

This branch demonstrates how we can update Rubocop, automatically generate some config files and then start to fix violations of the Ruby Style Guide.

We can get it into our build process fairly easily by using the .rubocop_todo.yml file, as that disables all the rules we're in violation of. As we go through and remove violations we can regenerate the .rubocop_todo.yml file which will ensure that future violations of the same rule will raise an error.

What do y'all think?


floehopper:
I think this would be a good idea. It'll be a bit of a challenge making sure we keep ratcheting up the number of checks we're doing, but as soon as we have a few checks, it will mean we can't make things worse. Consistency is definitely a worthwhile goal in terms of making the app more maintainable - there is a *lot* of inconsistency in the app at the moment.

tadast:
I have rubocop integrated with my editor already and always use it on greenfield projects. I wouldn't mind agreeing on a styleguide and following it eg by updating the file to match the styleguide before you make any changes to it. Having to do that would a bit of a distraction from the actual problem though. It's also distracting when a CI build fails because of it.

All in all, I think it's a good tool devs can use, but perhaps we shouldn't be enforcing it?

Also, there is this https://github.com/alphagov/styleguides/blob/master/using-rubocop.md

chrisroos:
> All in all, I think it's a good tool devs can use, but perhaps we shouldn't be enforcing it?

I think we probably have to enforce it for it to become useful. If it's not enforced then I can imagine a situation where in a given team of people, only a subset of them care enough to use the tool and fix any violations.

> Also, there is this https://github.com/alphagov/styleguides/blob/master/using-rubocop.md

Awesome. I hadn't seen that before and it looks like it contains useful information.

chrisroos:
I'm going to close this for now. I'll look at the styleguide page @tadast linked to and decide what to do with this.

PR #1645, opened 2015-05-19
Use DSL validate blocks for validation in Ruby smart answer flows

floehopper:
This PR illustrates another type of change we're planning to make across the Ruby smart answers.

There are many places where `calculate` blocks are used to conditionally raise a `SmartAnswer:: InvalidResponse` exception depending on whether the user's response is valid. However, the Ruby DSL already provides for `validate` blocks which are already used in a _few_ places.

Although in some ways it feels as if using the `validate` blocks ties us more into the DSL, we think that it's useful in making all the validation more explicit and separates it from calculation logic.

Again this isn't intended as an end goal, but another small step on the way to separating all the concerns in the Ruby smart answers into a more maintainable form.

### DIFF
--- a/lib/smart_answer_flows/appeal-a-benefits-decision.rb
+++ b/lib/smart_answer_flows/appeal-a-benefits-decision.rb
@@ -93,11 +93,14 @@ module SmartAnswer
         from { 5.years.ago }
         to { Date.today }
 
+        validate do |response|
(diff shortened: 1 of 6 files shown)


tadast:
LGTM :+1:

chrisroos:
Looks good to me, too.

PR #1646, opened 2015-05-19
Ignore commas in numeric responses to value questions

floehopper:
There are 3 places in the `simplified-expenses-checker` flow (and in the `v2`
flow of the same name) which explicitly ignore commas in numeric responses:

* https://github.com/alphagov/smart-answers/blob/2014c530db5d5179ba1734039ee00933c528e71a/lib/smart_answer_flows/simplified-expenses-checker.rb#L182
* https://github.com/alphagov/smart-answers/blob/2014c530db5d5179ba1734039ee00933c528e71a/lib/smart_answer_flows/simplified-expenses-checker.rb#L206
* https://github.com/alphagov/smart-answers/blob/2014c530db5d5179ba1734039ee00933c528e71a/lib/smart_answer_flows/simplified-expenses-checker.rb#L223

This seems a reasonable thing to do, but I feel as if it would be better to do
it consistently and centrally for all numeric responses to value questions.

### DIFF
--- a/lib/smart_answer/question/value.rb
+++ b/lib/smart_answer/question/value.rb
@@ -8,16 +8,22 @@ module SmartAnswer
 
       def parse_input(raw_input)
         if Integer == @parse
-          Integer(raw_input)
+          Integer(without_commas(raw_input))
         elsif :to_i == @parse
-          raw_input.to_i
+          without_commas(raw_input).to_i
         elsif Float == @parse
-          Float(raw_input)
+          Float(without_commas(raw_input))
         elsif :to_f == @parse
-          raw_input.to_f
+          without_commas(raw_input).to_f
         else
           super
         end
+      end
+
+      private
+
+      def without_commas(raw_input)
+        raw_input.delete(',')
       end
     end
   end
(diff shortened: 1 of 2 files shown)


tadast:
I guess most of the calculators are targeting local users, so it should be alright, but just out of curiosity I googled around and found this:
screenshot 2015-05-20 10 24 55

LGTM :shipit:

floehopper:
Yeah. I did consider that issue, but, as you say, most users will be from the UK. However, perhaps the safer approach would be not to allow commas anywhere in numeric responses.

chrisroos:
I think it'd be useful to describe the difference in behaviour before and after this change, i.e. describe how the `:to_i` and `Integer` `value_question`s behave when supplied with commas before and after this change.

floehopper:
I'm now wondering whether this is such a good idea as some of the smart answers are aimed at people outside the UK.

floehopper:
@chrisroos has suggested that we look at the logs to work out what values people are entering e.g. are people using commas in numeric values, are they using european-style separators, etc. I've requested access to the logs, but this PR will have to wait for now.

chrisroos:
@floehopper: Is it worth closing this until we have some more data to help us decide what to do?

floehopper:
@chrisroos: Yes. Closing.

Issue #1546, opened 2015-04-13
Refactor FLOW_REGISTRY_OPTIONS


tadast:
This constant is used to keep default configuration options for the smart answer registries. The registry instances are cached in class variables and [complain](https://github.com/alphagov/smart-answers/blob/34b2c115c87227ebdd97e1ca27a6d96bcdc60390/lib/smartdown_adapter/registry.rb#L8-L10) if a different instance is requested. We end up [modifying the constant](https://github.com/alphagov/smart-answers/blob/34b2c115c87227ebdd97e1ca27a6d96bcdc60390/test/helpers/test_fixtures_helper.rb#L3-L5) in tests to work around this issue.

This can cause random test failures depending on the test scenario order if not reset to old values in teardown stage correctly.

I suggest encapsulating the constant in a class method eg `FlowRegistry.config` that gets initiated [here](https://github.com/alphagov/smart-answers/blob/34b2c115c87227ebdd97e1ca27a6d96bcdc60390/config/initializers/flow_preview.rb#L4). This would allow stubbing out the test configuration on a per-case basis without worrying of changing the state globally.

We could also encapsulate the stubbing itself in a test helper that would take care of other related things that get cached in class variables (e.g. plugin load paths), which would help with issues like [this](https://github.com/alphagov/smart-answers/commit/34b2c115c87227ebdd97e1ca27a6d96bcdc60390) too.

Other suggestions welcome :)

chrisroos:
I've moved this to Trello in an attempt to end up with a single place containing Smart Answers stories/tasks/bugs.

See the [Trello card](https://trello.com/c/WEGW7pBh/143-refactor-flow-registry-options) (not public).

Closing this issue.

PR #1647, opened 2015-05-19
Speed up DuplicateFlowTest

floehopper:
This test takes about 15 seconds to run on my machine. Most of the time appears
to be taken in loading the Smartdown flows, notably the `pay-leave-for-parents`
and the `pay-leave-for-parents-adoption` flows. These are being (re-)loaded,
because the singleton instance is being reset. Unfortunately I couldn't tell
from the [commit note](aea5d2aa56076c33962dc10093d47f9ae6199abe) why this reset is necessary, so I didn't feel comfortable
just removing it.

However, I noticed that the `Smartdown::Registry#available_methods` method being
tested returns a list of flow names which are **only** based on filenames - it
does **not** rely on instantiating the flows themselves. So it seems as if it
should be safe to temporarily disable flow preloading for the duration of just
this test.

With this change in place, the test takes much less than 1 second to run.

### DIFF
--- a/test/functional/duplicate_flow_test.rb
+++ b/test/functional/duplicate_flow_test.rb
@@ -4,6 +4,14 @@ require_relative '../../lib/smartdown_adapter/registry.rb'
 require_relative '../../lib/smart_answer/flow_registry.rb'
 
 class DuplicateFlowTest < Minitest::Test
+  def setup
+    @original = FLOW_REGISTRY_OPTIONS[:preload_flows]
+    FLOW_REGISTRY_OPTIONS[:preload_flows] = false
+  end
+
+  def teardown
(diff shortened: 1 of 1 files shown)


tadast:
>  Unfortunately I couldn't tell from the commit note why this reset is necessary, so I didn't feel comfortable just removing it.

Perhaps something similar to https://github.com/alphagov/smart-answers/pull/1579 ?

I do not feel good about how we solve this problem in smartanswers ([see here](https://github.com/alphagov/smart-answers/issues/1546)) but in this particular case I don't see how it could cause any problems :+1:

floehopper:
@tadast: I have some ideas about how to improve the whole registry singleton/options thing, but I think I'd still prefer to make this change now. Thanks for the feedback.

PR #1648, opened 2015-05-20
Use outcome templates for student finance calculator

chrisroos:
This builds upon the work discussed in #1577.

It replaces the traditional `PhraseList` outcomes in the student-finance-calculator with ERB templates.

I've checked the output of the ERB templates using the regression tests in #1642 and am confident that they're identical.

You'll notice that we've removed a number of assertions that were checking things about the PhraseLists, and therefore the logic around what gets displayed in outcomes. The removal of these assertions means that we've lost some test coverage. This reduction in test coverage is one of our main motivations for wanting to add the regression tests in #1642.

Are people still happy to make the switch to ERB templates, and does anyone have any thoughts about how we might test the logic in the templates if we don't end up adding the proposed regression tests?


chrisroos:
I should mention that there is a precedent for testing against the content of Smart Answer outputs in a number of tests:

* [marriage-abroad-test](https://github.com/alphagov/smart-answers/blob/master/test/integration/smart_answer_flows/marriage_abroad_test.rb#L2308)
* [overseas-passport-test](https://github.com/alphagov/smart-answers/blob/master/test/integration/smart_answer_flows/overseas_passports_test.rb#L126)
* [report-a-lost-or-stolen-passport-test](https://github.com/alphagov/smart-answers/blob/master/test/integration/smart_answer_flows/report_a_lost_or_stolen_passport_test.rb#L49)

These tests use the [`outcome_body` helper method][outcome_body] to assert things about the outcome.

We could potentially use this technique to retain the same level of coverage that we had before I removed the assertions in this branch.

[outcome_body]: https://github.com/alphagov/smart-answers/blob/master/test/integration/smart_answer_flows/flow_test_helper.rb#L21

floehopper:
LGTM. As we've discussed previously, it would be good to have a partial template mechanism to remove some of the duplication from the new outcome templates.

chrisroos:
@floehopper, @tadast, @BenJanecke and I discussed this in some detail today, as part of [our discussion about adding regression tests](https://github.com/alphagov/smart-answers/pull/1642#issuecomment-103909315).

I'm going to get this merged to master shortly.

PR #1649, opened 2015-05-20
Use standard Rails testing Railtie

floehopper:
The main motivation for doing this is to make the app more like a standard
Rails app. However, it also has the benefit that we can remove the code which
was defining the `test:all` rake task.

Other side benefits are:

* The ability to [run a subset of tests][1] by filename and/or test method name.
* A bunch of [pre-defined rake tasks][2] for running subsets of tests
  e.g. `test:units`. Annoyingly these rake tasks don't have descriptions
  and so they don't show up when you execute `rake -T`.

[1]: http://guides.rubyonrails.org/v4.2.1/testing.html#running-tests
[2]: http://guides.rubyonrails.org/v4.2.1/testing.html#rake-tasks-for-running-your-tests


tadast:
:confetti_ball: :confetti_ball: :confetti_ball: :confetti_ball: :confetti_ball: :confetti_ball: :confetti_ball: :confetti_ball: :confetti_ball: :confetti_ball: :confetti_ball: :confetti_ball: :confetti_ball: :confetti_ball: :confetti_ball: :confetti_ball: :confetti_ball: :confetti_ball: :confetti_ball: :confetti_ball: 
> The ability to run a subset of tests by filename and/or test method name.

LGTM

PR #1651, opened 2015-05-20
Remove unused code

chrisroos:
Can anyone see a reason not to remove these files?


tadast:
Can this go too https://github.com/alphagov/smart-answers/blob/master/lib/smartdown_plugins/shared/data_partial.rb? I couldn't find any refs to it apart from docs and tests

chrisroos:
> Can this go too https://github.com/alphagov/smart-answers/blob/master/lib/smartdown_plugins/shared/data_partial.rb? I couldn't find any refs to it apart from docs and tests

I don't think we can get rid of that DataPartial just at the moment. [It's being used in the Smartdown Animal example](https://github.com/alphagov/smart-answers/blob/master/lib/smartdown_plugins/animal-example-simple/render_time.rb#L3).

Issue #1654, opened 2015-05-21
TypeError: can't convert nil into Integer


chrisroos:
See the [Errbit report][] for more information.

This exception is raised when the response to a `value_question :name, parse: Integer` is nil. I can't see that this is possible through the web interface but we should protect against it nonetheless.

@floehopper mentioned that this might become a problem in commit 28785dd848d206e80e82c6c147b0db719258862a.

You can see the error by visiting http://smartanswers.dev.gov.uk/calculate-your-child-maintenance/y/pay/1_child/no/400.0?next=1. The key is the lack of 'response=' in that URL.

All of the exceptions we've seen in Errbit came from a single IP address and the requests don't contain a user-agent which suggests to me that they might be from a bot.

[Errbit report]: https://errbit.production.alphagov.co.uk/apps/533c35ae0da1159384044f5f/problems/555dc6746578635c07b60300

PR #1655, opened 2015-05-21
Avoid errors when requests exclude the response param

chrisroos:
This is one way of addressing #1654, although @floehopper discusses an
alternative approach in 28785dd848d206e80e82c6c147b0db719258862a.

The problem we were seeing was caused by a combination of:

* The `parse: Integer` option being set on a `value_question`.

* The presence of a `nil` response in the `SmartAnswerPresenter#all_responses`
array

This fix feels OK to me as the current implementation implies that the presence
of `:next` in the params should mean that `:response` is also present.

The effect of this change is for the
`SmartAnswersController#redirect_response_to_canonical_url` behaviour to kick
in and redirect the user back to the URL without the `next` parameter.

### DIFF
--- a/app/presenters/smart_answer_presenter.rb
+++ b/app/presenters/smart_answer_presenter.rb
@@ -152,7 +152,7 @@ class SmartAnswerPresenter
 
   def all_responses
     normalize_responses_param.dup.tap do |responses|
-      if params[:next]
+      if params[:next] && params[:response]
         responses << params[:response]
       end
     end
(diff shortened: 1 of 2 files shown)


chrisroos:
@floehopper: Can you see any advantage/disadvantage to this approach over your suggestion in 28785dd848d206e80e82c6c147b0db719258862a?

chrisroos:
I've answered my own question by causing the build to fail! I'm going to close this pull request and pick this up again next week.

PR #1656, opened 2015-05-26
Use erb templates for additional commodity code

chrisroos:
I've introduced the ability to control whitespace trimming in our ERB templates, so that I could add ERB templates for this Smart Answer without having to update all the regression test HTML files.

This Smart Answer's outcome also has a title that is still in the YAML file. We don't currently support ERB templates for titles so I plan to address that separately.


tadast:
LGTM

PR #1657, opened 2015-05-26
Avoid using SmartAnswer::FlowRegistry in integration tests

floehopper:
The motivation behind this is to reduce our dependency on the `SmartAnswer::FlowRegistry` in the hope that this will make it possible to simplify or even remove the `FLOW_REGISTRY_OPTIONS` mechanism for configuring the registry singleton. I also think this change makes the tests simpler and more explicit.

I've converted the integration tests for all of the smart answer flows, **except** the following which are currently being worked on by @tadast & @BenJanecke:

* calculate-employee-redundancy-pay
* calculate-your-redundancy-pay
* marriage-abroad

Once these ones have been made, we should be able to simplify the code in `FlowTestHelper#setup_for_testing_flow`.

### DIFF
--- a/lib/smart_answer/flow.rb
+++ b/lib/smart_answer/flow.rb
@@ -5,6 +5,12 @@ module SmartAnswer
     attr_reader :nodes, :outcomes
     attr_accessor :state, :status, :need_id
 
+    def self.build
+      new.tap do |flow|
+        flow.define
+      end
+    end
+
     def initialize(&block)
       @nodes = []
       @state = nil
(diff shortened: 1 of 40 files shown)


tadast:
Nice, I like the solution. So much simpler. 

I don't see why it would, but does it increase the test suite runtime duration?

floehopper:
@tadast: The difference is probably in the noise, but tests run slightly quicker with these changes on my local machine: 214 seconds on `master` vs 207 seconds on this branch.

PR #1658, opened 2015-05-26
Use erb templates for additional commodity code outcome titles

chrisroos:
I've introduced support for using ERB templates to render Outcome titles, and updated the additional-commodity-code Smart Answer to use an ERB template.

There's currently some duplication between `OutcomePresenter#title` and `OutcomePresenter#body` but I plan to address that separately.

### DIFF
--- a/app/presenters/outcome_presenter.rb
+++ b/app/presenters/outcome_presenter.rb
@@ -9,7 +9,14 @@ class OutcomePresenter < NodePresenter
   end
 
   def title
-    translate!('title')
+    if use_template? && title_erb_template_exists?
+      view_context = @state.dup
+      safe_level, trim_mode = nil, '-'
+      title = ERB.new(title_erb_template_from_file, safe_level, trim_mode).result(view_context.instance_eval { binding })
+      title.chomp
+    else
+      translate!('title')
+    end
   end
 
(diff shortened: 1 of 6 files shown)


floehopper:
Did you consider implementing this using something like [`ActionView::Helpers::CaptureHelper#content_for`](http://api.rubyonrails.org/classes/ActionView/Helpers/CaptureHelper.html#method-i-content_for)?

chrisroos:
> Did you consider implementing this using something like ActionView::Helpers::CaptureHelper#content_for?

I certainly considered it but assumed it would be more work than the approach I've gone with at the moment. Do you think it's OK to go with my approach for now and consider a switch to something like `content_for` later?

chrisroos:
> I found the commit title confusing. I would've written "Add support for ERB templates for Outcome titles".

I'll update it.

floehopper:
Yes, I think your approach is fine for now. I just thought it would be good to record if we'd discarded the idea.

floehopper:
LGTM. :+1:

floehopper:
We just spoke about this, but I thought it was worth recording that the motivation behind making this change was so that we could remove **all** the outcome-related keys from the locale file.

PR #1659, opened 2015-05-26
Improvements to outcome presenter

chrisroos:
This builds on the use-erb-templates-for-additional-commodity-code-outcome-titles branch so please ignore the first three commits. They'll hopefully be merged as part of #1658.

Most of the changes are simple renames.

I've changed the behaviour of rendering an ERB template that doesn't exist. I'm now returning nil, instead of raising an exception, to retain the behaviour that exists in the YAML templates.

I've introduced a `ViewContext` wrapper around `State` objects that allows me to restrict the methods that can be called from the ERB templates.

### DIFF
--- a/app/presenters/outcome_presenter.rb
+++ b/app/presenters/outcome_presenter.rb
@@ -83,7 +83,7 @@ class OutcomePresenter < NodePresenter
   end
 
   def default_title_erb_template_path
-    Rails.root.join("lib/smart_answer_flows/#{@node.flow_name}/#{name}_title.txt.erb")
+    template_directory.join("#{name}_title.txt.erb")
   end
 
   def body_erb_template_from_file
@@ -95,10 +95,14 @@ class OutcomePresenter < NodePresenter
   end
 
   def default_body_erb_template_path
-    Rails.root.join("lib/smart_answer_flows/#{@node.flow_name}/#{name}_body.govspeak.erb")
+    template_directory.join("#{name}_body.govspeak.erb")
   end
 
   private
+
+  def template_directory
+    Rails.root.join('lib', 'smart_answer_flows', @node.flow_name)
+  end
 
   def render_erb_template(template, view_context)
     safe_level, trim_mode = nil, '-'


app/presenters/outcome_presenter.rb
```diff
@@ -1,7 +1,25 @@
...
 class OutcomePresenter < NodePresenter
-  class OutcomeTemplateMissing < StandardError; end
+  class ViewContext
+    def initialize(state)
+      @state = state
+      state_methods = state.methods - SmartAnswer::State.new(nil).methods
+      @state_reader_methods = state_methods.reject { |m| m =~ /=$/ }
+    end
+
+    def method_missing(method, *args, &block)
```
tadast: Would it be possible to use [this](http://apidock.com/rails/Module/delegate)?

app/presenters/outcome_presenter.rb
```diff
@@ -31,40 +55,60 @@ def has_calendar?
     calendar.present?
   end
 
-  def has_body?
-    use_template? || super()
-  end
-
   def body
-    if use_template?
-      view_context = @state.dup
```
tadast: Why was this `.dup` here? Do we not need this anymore?

app/presenters/outcome_presenter.rb
```diff
@@ -1,7 +1,25 @@
...
+
+    def method_missing(method, *args, &block)
+      if @state_reader_methods.include?(method)
+        @state.send(method, *args, &block)
+      else
+        super
+      end
+    end
+
+    def get_binding
```
tadast: I think it's not very common in ruby to prefix methods with `get_`, can we leave it as `binding` or will it conflict with anything?

app/presenters/outcome_presenter.rb
```diff
@@ -1,7 +1,25 @@
 require 'erubis'
 
 class OutcomePresenter < NodePresenter
-  class OutcomeTemplateMissing < StandardError; end
+  class ViewContext
+    def initialize(state)
+      @state = state
+      state_methods = state.methods - SmartAnswer::State.new(nil).methods
+      @state_reader_methods = state_methods.reject { |m| m =~ /=$/ }
```
tadast: Minor: I think http://apidock.com/ruby/String/end_with%3F reads better :)

app/presenters/outcome_presenter.rb
chrisroos: I'm not sure I can. I need to inspect the `@state` object to determine which methods to delegate to it and I'm currently doing that in the initializer. Unless I'm missing a different way of using `delegate`?

app/presenters/outcome_presenter.rb
chrisroos: I've not come across `#end_with?` before. Nice!

chrisroos: Hmm, I've actually got an array of symbols so I can't use `#end_with?` without first turning them to strings. Either of these work but I wonder whether they're actually any clearer now. What do you think?

    @state_reader_methods = state_methods.reject { |m| m.to_s.end_with?('=') }

    or

    @state_reader_methods = state_methods.map(&:to_s).reject { |m| m.end_with?('=') }.map(&:to_sym)

app/presenters/outcome_presenter.rb
tadast: Oh right. I was thinking we could call `self.class.delegate *state_reader_methods, to: :state`, but I overlooked that it depends on the instance protocol rather than the class.

I guess we'll have to stick to method_missing, but I would suggest overriding [`respond_to?`](https://robots.thoughtbot.com/always-define-respond-to-missing-when-overriding) as well.

app/presenters/outcome_presenter.rb
chrisroos: I went with `#get_binding` as that's what's used in a number of examples in the Ruby docs. See the [`Kernel#binding` docs](http://ruby-doc.org/core-2.2.2/Kernel.html#method-i-binding), for example.

We can't name it `binding` as that results in an infinite loop of `#binding` trying to call itself.

app/presenters/outcome_presenter.rb
tadast: I like `state_methods.reject { |m| m.to_s.end_with?('=') }` as it's more explicit than `state_methods.reject { |m| m =~ /=$/ }`. But again, it's a minor thing, a matter of taste rather than a readability concern :)

app/presenters/outcome_presenter.rb
chrisroos: Hmm, good question. I think it was to avoid us changing the state object but I don't remember for sure. Let me dig into this to work out why we added it!

app/presenters/outcome_presenter.rb
tadast: That's cool. I was referring to this by the way https://github.com/bbatsov/rubocop/blob/master/lib/rubocop/cop/style/accessor_method_name.rb#L15-L19

app/presenters/outcome_presenter.rb
floehopper: I suspect we wanted to make sure the state object couldn't be modified. It might be possible to achieve that another way e.g. hiding the "setter" methods on the instance of `State`.

app/presenters/outcome_presenter.rb
floehopper: I suspect @chrisroos has taken this from the [example in the Ruby docs](http://ruby-doc.org/core-2.2.2/Binding.html). I think it might be confusing or even cause a problem to use `#binding`.

app/presenters/outcome_presenter.rb
tadast: sounds like [Object#freeze](http://ruby-doc.org/core-2.2.2/Object.html#method-i-freeze) except that it will raise an error if the state is being modified.

chrisroos: I checked out the commit that introduced `.dup`, removed it, ran all the tests and nothing failed!

I'm happy for it to go particularly as I'm (admittedly crudely) preventing setter methods from being called on the state object from the `ViewContext`. We can always bring it back if/when we realise we're unintentionally changing the state of the `state` object.

app/presenters/outcome_presenter.rb
chrisroos: Changed in f137a549f9be9369bc462182613dc027435aa49e.

chrisroos:
> Have you checked that the `NoMethodException` exception bubbles all the way up so that it would be detected by the regression tests?

I have now :-) They do indeed bubble up and cause the regression tests to fail.

chrisroos:
> Commit note typos...

Thanks! I've fixed them.

PR #1660, opened 2015-05-26
Move all outcome templates to a single directory

chrisroos:
I find this slightly easier to browse than having the outcome templates mixed
alongside everything else in lib/smart_answer_flows.


floehopper:
:+1:

chrisroos:
@floehopper: You mentioned yesterday that you're less sure about this approach. Would you be able to explain why in here?

floehopper:
I find it more logical that the flow file and it's corresponding directory of outcome templates are in the same directory. I think what's actually confusing is having the `data_partials`, `locales` & `shared_logic` directories in the `smart_answer_flows` directory as well as the flow-related directories, but hopefully in time these will go away. I think it's a bit of a subjective decision, so I'm happy for you to move them if it makes more sense to you.

@tadast, @BenJanecke: Do you have any preference?

tadast:
I agree with @floehopper that keeping all flow-related files bundled together makes sense. At the same time I like this approach, because it's closer to the conventional rails' app/ [views/controllers/models] dir structure. 

I'd say a weak :+1: to this PR

chrisroos:
I agree with @floehopper that it's the mix of Smart Answer specific files/folders (i.e. flow files and outcome directories) and shared folders (i.e. data_partials, locales and shared_logic) that I find slightly confusing.

I'm less convinced about this change so I'm going to close this pull request.

PR #1661, opened 2015-05-27
Check for missing outcome artefacts in regression tests

floehopper:
A change to a flow (and its associated `questions-and-responses` & `responses-and-expected-results` files) can mean that one or more outcome HTML artefacts are no longer generated when the regression tests are run. We want the regression tests to fail in this scenario and that's the behaviour I've added in this PR.

I've also taken the opportunity to remove the parts of the old-style integration tests which cover outcome nodes, because these are now covered by the regression tests. I've belatedly realised that these may not belong in this PR, but let me know what you think.

### DIFF
--- a/lib/smart_answer_test_helper.rb
+++ b/lib/smart_answer_test_helper.rb
@@ -41,9 +41,13 @@ class SmartAnswerTestHelper
     YAML.load(responses_and_expected_results_yaml)
   end
 
+  def path_to_outputs_for_flow
+    artefacts_path.join(@flow_name)
+  end
+
   def save_output(responses, response)
     filename = [responses.join('-'), 'html'].join('.')
-    path = artefacts_path.join(@flow_name)
+    path = path_to_outputs_for_flow
     FileUtils.mkdir_p(path)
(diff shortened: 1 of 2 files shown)


tadast:
Re test removal: did we not say we'll keep integration tests while we fully transition (only removing the phraselist assertions)? I'm cool if they go if you're confident this testing approach is going to work and we won't need to revert all integration tests. I'm not confident yet, because I have not tried using regression tests with big smartanswers yet :)

floehopper:
I've just (force) pushed my [alternative solution](70d55fa10fdcda77fcffc76bb59cdc1e581288eb) which retains the one-assertion-per-test style of the regression tests. It's not pretty, but I think it's ok.

floehopper:
@tadast: Regarding the integration tests, my thinking was along the same lines as for the removal of the phrase list assertions i.e. since we now have coverage in the regression tests, I thought it would be ok to remove the assertions against outcome nodes. @chrisroos: do you have any thoughts? For now I'm going to split those integration test changes into their own PR, so I can merge the regression test change.

PR #1662, opened 2015-05-27
Remove unused calendar functionality

floehopper:
This functionality was only used in the `calendars-sample.rb` fixture which
was in turn only exercised in the `CalendarExportTest` integration test.

Since it was not being used in any production code, it feels as if it's just
adding needless complexity to the app and slowing the tests down.

To this latter point, the reason I started investigating this in the first
place was that the `should "not store the calendar in memory"` test in
`CalendarExportTest` was by far the slowest test.

I was sure I'd already merged this, but it appears that I hadn't.


tadast:
LGTM

PR #1664, opened 2015-05-27
Hack-fix failing autoloading of flows.

tadast:
Now that most of the smart answer flows are classes
as of [this commit](https://github.com/alphagov/smart-answers/commit/0800f52b001f3d0d4bed9a497b7ab0aa2ce60f9d), in development environment
we get uninitialized constant SmartAnswer::<FlowName> error after changing the flow files and need to restart the server to reload the flow code correctly.

This is because the flows dir structure and namespaces do not follow rails conventions and autoloading fails. This is a rather hackish solution that force-reloads the class every time it is being loaded in dev environment.

Thoughts? 
@floehopper

### DIFF
--- a/lib/smart_answer/flow_registry.rb
+++ b/lib/smart_answer/flow_registry.rb
@@ -95,6 +95,7 @@ module SmartAnswer
     def build_flow(name)
       if SMART_ANSWER_FLOW_NAMES.include?(name)
         class_prefix = name.gsub("-", "_").camelize
+        load "lib/smart_answer_flows/#{name}.rb" if Rails.env.development?
         namespaced_class = "SmartAnswer::#{class_prefix}Flow".constantize
         namespaced_class.build
       else


floehopper:
I'm not convinced this is a good idea.

I had a quick play with trying to get Rails autoloading & reloading working with the files as they are, but it turns out to be a bit of a nightmare.

So instead I think I'd prefer to change the namespace for the flows to be `SmartAnswerFlows` to match the `lib/smart_answer_flows` directory and rename the flow files to match their new class names e.g.

`lib/smart_answer_flows/additional-commodity-code.rb` ->
`lib/smart_answer_flows/additional_commodity_code_flow.rb`

i.e. replace hyphens with underscores and add `_flow` suffix.

I've pushed a demo of this up in this branch: https://github.com/alphagov/smart-answers/compare/rename-smart-answer-flow-files-and-namespace-to-match-flow-classes-and-directory

This means that the directory & file names match the module & class names and Rails does it's magic so we don't need to require the flows any more. I have checked and it fixes the problem that @tadast describes above.

Renaming the flow files is not ideal from a git history point of view, but git is showing the one in the demo branch as a rename and I have verified that it is possible to follow the history using a command like this:

`git log -p --follow lib/smart_answer_flows/additional_commodity_code_flow.rb`

What do people think...?

chrisroos:
I think I prefer the idea of renaming the files to bring us in line with the Rails naming convention.

tadast:
Totally agree, I intended this as a temporary fix before we renamespace all the flows. If we do this soon enough I'm happy to close this.

Another thing to be aware of is [this rake task](https://github.com/alphagov/smart-answers/blob/2166edf1bdd737d517ea1cf21641aa3dd8c9824d/lib/tasks/version.rake), it may still be helpful with the existing V2s. Perhaps we should chase content designers to find out if we can deploy/remove those V2s.

floehopper:
I think I should be able to do the renaming a bit later this afternoon. Let's leave this open until I do in case I run into unexpected problems. Thanks for the pointer to the rake task - hadn't seen that.

floehopper:
I had a go at this, but ran into problems, because of all the constants in the `SmartAnswer` namespace which needed to be fully qualified in order to continue to work once the flows were in a new `SmartAnswerFlows` namespace. I want to come up with a better plan, but in the meantime, I think we should use the solution that @tadast has proposed in this PR.

tadast:
:( I hope we can revert this PR soon.

PR #1666, opened 2015-05-27
Remove integration tests asserting against outcome nodes

floehopper:
This PR removes the parts of three old-style integration tests which cover outcome nodes, because these are now covered by the corresponding regression tests.

* additional-commodity-code
* student-finance-calculator
* vat-payment-deadlines

These changes have already been discussed a little in the comments on #1661.


chrisroos:
I think it's probably OK to remove these tests given that we've got the regression tests in place.

At the moment, I can't imagine us wanting to bring these sort of tests back so don't think there's much risk in removing them.

floehopper:
I think the only argument for keeping them that I can think of is if there is some intention-revealing text in the test/context names. The only candidates I can see for this are:

* https://github.com/alphagov/smart-answers/pull/1666/files#diff-a010bf8fa41c9fe608ecc540483f3163L81
* https://github.com/alphagov/smart-answers/pull/1666/files#diff-4953b003b4a8789139d9c669b780d113L89

I don't think the comment in the first of the two adds anything intention-revealing over the comments and code in [the implementation](https://github.com/alphagov/smart-answers/blob/remove-integration-tests-asserting-against-outcome-nodes/lib/smart_answer_flows/student-finance-calculator.rb#L104-L119). In the end this logic will be better extracted out into "calculator" class and unit tested with intention-revealing test names.

The second of the two isn't actually very intention-revealing given that it's not possible to enter a date that is not the end of the month, because of a validation check, and in any case one of the tests I've left in place still [tests that validation](https://github.com/alphagov/smart-answers/blob/remove-integration-tests-asserting-against-outcome-nodes/test/integration/smart_answer_flows/vat_payment_deadlines_test.rb#L20-L23).

All the others seem to just repeat the value of the response and/or the name of the outcome.

So I think I'm happy to merge this.

@tadast: Do you have any more thoughts following on from your comments in #1661?

tadast:
I'm neither convinced nor dissuaded, I have not tried this enough to have a strong opinion about future perspectives of this approach, so I will trust your judgment :)

PR #1667, opened 2015-05-28
Reduce time taken for regression tests

chrisroos:
This contains an approach to reducing the time it takes to run the regression tests, by only running them for a smart answer when that smart answer's source files have changed.

I store a hash of the contents of the smart answer source files and check those to determine whether to run the regression tests for that Smart Answer. You can override this by setting the `RUN_REGRESSION_TESTS` environment variable to 'true' to run all regression tests, or '<flow-name>' to run the regression tests for a single flow.

This should help us keep the time of the test suite down as we add more regression tests.

Does this seem like a reasonable approach?

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,6 +133,12 @@ Test a single Smartdown flow by running:
 
       * Go to Step 3, add the new responses and continue through the steps up to Step 9.
 
+10. Generate a yaml file containing the set of source files that this Smart Answer depends upon. The script will automatically take the ruby flow file, locale file and erb templates into account. You just need to supply it with the location of any additional files required by the Smart Answer (e.g. calculators and data files). This data is used to determine whether to run the regression tests based on whether the source files have changed.
+
+        $ rails r script/generate-checksums-for-smart-answer.rb <name-of-smart-answer> <path/to/additional/files>
+
+11. Commit the generated yaml file to git.
+
 ## Issues/todos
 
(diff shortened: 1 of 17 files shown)


floehopper:
I haven't gone through all the code changes in detail, but I've chatted to @chrisroos about his approach and (since there are no changes to production code) I'm happy that this gets merged and we try it out.

I think it's important that we merge this asap, because both @chrisroos & I are now adding regression tests at a rate of knots and the build time is climbing.

@tadast & @BenJanecke do you have any thoughts?

tadast:
I love the fact that it allows to run only the tests that need to run. I don't like the fact that it introduces some more non-conventional manual steps required to make it work. The only damage here is:

1) Devs will need to learn how to do this
2) Devs will find out about broken tests from the CI if they forget to add/remove dependent calculator/query class.

All in all I think it makes things better :+1:. Ideally we would reduce the amount of manual work to do (e.g. by detecting external classes used in code and adding them as dependencies) but this is for the future.

floehopper:
@tadast: Regarding your point 2 above, we're thinking that the tests would run conditionally like this on both the local development build and on the CI server. The idea would be that a developer would set `RUN_REGRESSION_TESTS` and run *all* the tests if they changed something not specific to an individual flow. While this isn't ideal, because it relies on individual developers remembering to do this, we think it represents the best compromise between having thorough regression test coverage and a build which runs fast enough.

PR #1668, opened 2015-05-28
Add regression test to ensure all nodes are reached

chrisroos:
By storing both the `current_node` and `next_node` in the responses-and-expected-results YAML file, we can compare the nodes that have been exercised with all the nodes in the Flow. I'd added a test that fails if the two sets of nodes don't match.

This is the test that @floehopper mentioned in 53e67d850500421fc26c226a004c13d02429cee8.

### DIFF
--- a/script/generate-responses-and-expected-results-for-smart-answer.rb
+++ b/script/generate-responses-and-expected-results-for-smart-answer.rb
@@ -23,17 +23,18 @@ def answer_question(flow, state)
   existing_responses = state.responses
 
   QUESTIONS_AND_RESPONSES[question_name].each do |response|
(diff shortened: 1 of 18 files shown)


floehopper:
Did all the existing regression tests just pass once you'd regenerated their responses-and-expected-results? I assume they did.

LGTM :+1:

chrisroos:
> Did all the existing regression tests just pass once you'd regenerated their responses-and-expected-results? I assume they did.

They did indeed!

chrisroos:
I've updated this branch to include the updated responses-and-expected-outputs and checksum data for the regression tests that have been added since last week. I'm going to get this merged to master.

PR #1670, opened 2015-05-28
Downgrade google-api-client gem to fix analytics rake task

floehopper:
The `analytics:analyse` rake task makes use of the `google-api` command-line
tool which is only available in pre-v0.7 versions of the `google-api-client`
gem.

I also had to explicitly restrict the version of the `jwt` gem to ~> 0.1.4 to
get the google-api CLI tool to run.

I still haven't got the rake task to work completely, because I'm waiting on
some permission changes to my Google account. However, I'm confident that we
need these changes if the task is ever going to work and they don't _seem_ as if
they should affect any production code.

Also I thought it was better to get this into master rather than have it languishing
on my local machine.


floehopper:
Does anyone have any thoughts on this. More specifically do you think that any of the gem downgrades might be risky?

chrisroos:
> I also had to explicitly restrict the version of the jwt gem to ~> 0.1.4 to get the google-api CLI tool to run.

Out of interest, what was the error when using a newer version of jwt?

chrisroos:
It looks as though the `Google::APIClient` is only used by the analysis.rake task so I don't see any harm in downgrading it. I guess it's possible that downgrading some of the related gems might have some unintended side effects but hopefully our test suite will catch those.

floehopper:
I thought the `jwt` thing was a version dependency problem, but looking at the `Gemfile.lock` now, it's not obvious that that is the case. I'll check again before merging.

floehopper:
I'm going to close this for now until we actually might need it.

PR #1671, opened 2015-05-29
Bump slimmer for rendering app meta tag

edds:


chrisroos:
Yo @edds. Can you add some more information about why we need this change, and whether it needs to be merged by a certain date (e.g. because the old version will stop working) please?

edds:
The only change this has is an extra meta tag will be added to the page. This will be read by the analytics JavaScript if it is there (and give us the ablity to segment performance metrics per rendering application amongst other things), it doesn't matter if it isn't present though so nothing will break with/without this. There is no dependancy or urgency on merging or deploying.

floehopper:
I imagine the same change should be made to the [calculators app](https://github.com/alphagov/calculators).

edds:
@floehopper indeed it will, I current making my way through our 17 frontend applications.

chrisroos:
You beat me to it, @tadast. I was planning to rewrite the commit to include @edds comment before merging.

tadast:
Oh sorry. This would have been a good thing to do. I'll be more careful next time :)

floehopper:
I notice that this branch (along with a number of other branches) have been merged into master without rebasing first. I think this makes the git history really hard to follow. I thought the plan was to always rebase against master before merging - have I got the wrong end of the stick?

edds:
We have no policy on rebasing on master before merging that I am aware of. Only that you should rebase to make sure your commits make logical sense.

floehopper:
@edds: Thanks. That's helpful to know. My comment was really addressed to @tadas who pushed the merge button. Sorry for the confusion.

tadast:
I only tend to rebase when I need to tweak commits. I'm fine with rebasing before merging, but then we should write it up in the README and submit a PR where we can discuss merging practices. I'm also not a big fan of committing directly to master, so we could discuss it there too :)

rboulton:
Individual projects shouldn't have special instructions for PRs and merging unless there's a very good reason, because that makes it harder for people to start working on the project.  The styleguides (in particular https://github.com/alphagov/styleguides/blob/master/git.md#branchingmerging-conventions) are the place for documenting our practices.

rboulton:
One good reason for *not* rebasing before merging is that if comments have been made in github on the individual commits, rebasing will stop those comments being shown on the pull request (because the commit will no longer be part of the pull request).

floehopper:
Is there a better forum for this discussion? This PR doesn't seem appropriate.

PR #1672, opened 2015-06-01
Remove redundant code from pip-checker

floehopper:
The changeover date of 3rd February, 2014 has long since passed, so this code
is only ever being executed in the integration test using `Timecop.travel` to
set the current date in the past.


tadast:
LGTM :+1:

PR #1674, opened 2015-06-02
Update checksum for towing-rules

BenJanecke:
A content change was made to towing-rules in 8f816ac


floehopper:
Can you explain **why** you need to do this in the commit note. Presumably because you updated some content...?

Also if this commit is intended to fix the build, I'd probably put it straight into master.

floehopper:
Also I think the commit title is inaccurate. It's not a "date checksum", just a "checksum".

PR #1676, opened 2015-06-03
Validate date ranges

tadast:
It all started with a few errbit 500 errors where a user was tweaking the url to provide a date of birth that was out of the range of dates covered by logic. We decided that validating input in such cases is the best solution. Then I realised there are more questions affected.

This PR adds an optional `validate_in_range` behaviour, that makes sure user provided date input is within the allowed range defined by `#from` and `#to` methods.

It also adds some more years in DOB selectors to make sure people of all ages can use the tool.

### DIFF
--- a/lib/smart_answer/question/date.rb
+++ b/lib/smart_answer/question/date.rb
@@ -7,6 +7,10 @@ module SmartAnswer
         super
       end
 
+      def validate_in_range
+        @validate_in_range = true
+      end
+
       def from(from = nil, &block)
         if block_given?
           @from_func = block
@@ -105,6 +109,7 @@ module SmartAnswer
           else
            raise InvalidResponse, "Bad date", caller
           end
+        validate_input(date) if @validate_in_range
         date
       rescue
         raise InvalidResponse, "Bad date: #{input.inspect}", caller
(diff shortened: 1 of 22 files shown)


tadast:
Thanks for your earlier comments @chrisroos @floehopper - very useful. I've addressed your comments, added more integration tests, and uncovered a new issue (see https://github.com/alphagov/smart-answers/commit/6047bdd9dfcafcf8d672ae7984585ab450a9a143). I have not added the sorting option though, I would like to do it in a separate PR.

I had to tweak the commits, rebase and force-push, which unfortunately hides all out previous discussions.

floehopper:
I think @chrisroos mentioned something about this previously, but it would be good to remove the duplication and reveal the intention of the `122.years.ago` thing. Extracting the `122.years` into an intention-revealing constant would be a small, but useful step.

tadast:
@chrisroos @floehopper this PR changes a dropdown for many questions. Does it mean I now have to regenerate regression test HTML artefacts for all of them? :(

```
> git diff master --stat | grep flows
 lib/smart_answer_flows/calculate-state-pension.rb  |  6 +-
 lib/smart_answer_flows/energy-grants-calculator.rb |  3 +-
 lib/smart_answer_flows/pip-checker.rb              |  4 +-
 lib/smart_answer_flows/register-a-birth.rb         |  2 +-
 lib/smart_answer_flows/state-pension-topup.rb      |  3 +-
 .../smart_answer_flows/pip_checker_test.rb         | 12 ++++
```

tadast:
Ignore me, it only keeps the outcomes, so I think I don't need to regenerate anything. Any more comments or should I merge this?

floehopper:
LGTM

PR #1677, opened 2015-06-03
Bump slimmer for new rendering app meta tag

edds:
This now uses the GOVUK_APP_NAME env variable rather than the
application class name so it can be cross referenced to other things in
the stack.

This contains https://github.com/alphagov/slimmer/pull/128


dsingleton:
Tests failed on an unrelated error, i've restarted the build.

PR #1678, opened 2015-06-04
Remove unreachable outcome from overseas-passports flow

floehopper:
The `SmartAnswer::Calculators::PassportAndEmbassyDataQuery#fco_application?`
predicate never returns `true`, because none of the entries in
`lib/data/passport_data.yml` have `type: pretoria_south_africa`.

The history of the latter file is a bit confusing, but it appears as if most
of the occurrences of `type: pretoria_south_africa` were removed in
[this commit][1] and the last occurence in [this commit][2].

[1]: f26262ffeae2558d1e36042b638eed9099340fc4
[2]: 3a86b8ecd1c4686885dcbd9466da43f504999df9

### DIFF
--- a/lib/smart_answer/calculators/passport_and_embassy_data_query.rb
+++ b/lib/smart_answer/calculators/passport_and_embassy_data_query.rb
@@ -16,10 +16,6 @@ module SmartAnswer::Calculators
         "IPS")
     end
 
-    def fco_application?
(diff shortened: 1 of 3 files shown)


tadast:
LGTM :+1:

However, would it not be easier to look at git history if it was all a single commit?

floehopper:
I think there are arguments either way, but in this case I've combined them all into a single commit as you suggest. Merging...

PR #1679, opened 2015-06-05
Fix a bug with answer changing in smartdown.

tadast:
## Problem as reported

Three issues with the Parental Leave & Pay calculator: two minor, one serious, all very likely related.

From the results page at:

https://www.gov.uk/pay-leave-for-parents/y/no/2015-6-8/employee/yes/yes/160-week/yes

* clicking [change] for mother's employment details.

has no effect - should return us to the correct form.

Similarly,

https://www.gov.uk/pay-leave-for-parents/y/yes/2015-5-5/employee/employee/yes/yes/200-week/yes/yes/yes/150-week/yes

* clicking [change] for partner's employment details

has no effect - as above

But

* clicking [change] for mother's employment details

does take us back to the correct form. If we then

* change the mother's earnings
* click [next_step]

we get back to the results page, but it's the *partner's* earnings that are updated,
not the mother's.

### Bug description

When a change link was opened, responses to the question that's being changed
(previous_response_* query params) were added to the current state, which is
incorrect. The current state is being calculated based purely on those answers,
therefore we were displaying the question that goes after the one that is being
changed.

This only affects smartdown answers.

[The commit](https://github.com/alphagov/smart-answers/commit/cafeee0e66d807b896a7eb161b79c2eeb551567b) that has introduced this code has no useful description nor tests. We have added an integration test and have done some manual testing, but it should still be reviewed thoroughly.

### DIFF
--- a/lib/smartdown_adapter/presenter.rb
+++ b/lib/smartdown_adapter/presenter.rb
@@ -126,21 +126,14 @@ module SmartdownAdapter
 
     def responses_from_query_params(request)
       responses = []
-      # get form submission request: one response
       if request[:response]
         responses << request[:response]
       end
 
-      #get form submission request: for multiple responses
       if request[:next]
         (@previous_smartdown_state.current_node.questions.count - responses.count).times do |index|
           responses << request.query_parameters["response_#{index+1}"] || nil
         end
-      elsif !request.query_parameters.select { |key| key.to_s.match(/^previous_response_\d+/) }.empty?
-        @previous_smartdown_state.current_node.questions.count.times do |index|
-          responses << request["previous_response_#{index+1}"] || nil
-        end
-        responses
       end
       responses
     end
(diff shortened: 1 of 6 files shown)


chrisroos:
The integration test feels quite removed from the fix in smartdown_adapter/presenter.rb. Did you consider adding a unit test around presenter.rb to ensure that we don't re-introduce the regression?

I presume the integration test fails without the change to presenter.rb?

chrisroos:
> But
>
> clicking [change] for mother's employment details
> does take us back to the correct form. If we then
> 
> change the mother's earnings
> click [next_step]
> we get back to the results page, but it's the partner's earnings that are updated,
> not the mother's.

Has this been fixed? There doesn't appear to be a test for it.

BenJanecke:
change the mother's earnings
click [next_step]
we get back to the results page, but it's the partner's earnings that are updated,
not the mother's.

Is the same bug as the rest. 
It boils down to:

Given a page with multiple questions
When I want to change the answers for that page 
It takes me to the next page with the answers for the previous page pre-filled  

While it should be 

Given a page with multiple questions
When I want to change the answers for that page 
It takes me to that page with it's values pre-filled

chrisroos:
Thanks for the clarification, @BenJanecke.

PR #1682, opened 2015-06-08
Use Erubis instead of ERB for outcome templates

floehopper:
The ERB trim mode chosen in [this commit][1] was intended to replicate how
Rails handles whitespace trimming. However, in fact [Rails uses Erubis][2]
in its default trim mode.

This commit uses Erubis in its default trim mode instead of ERB to render
outcome templates.

This change means there is no longer any need to use `-%>` tags on lines which
(other than the ERB tag) only contain whitespace. This is because the
suppression of leading & trailing whitespace and the trailing newline is the
default behaviour in Erubis.

I had to add an explicit newline to one of the outcome templates in order to
generate the appropriate paragraph HTML tag to match the regression test
outcome HTML.

This change will make it easier to switch to Rails templating in the future
which is part of our plan.

[1]: d0520fcce0529f40a103853ecc3f7a09d287976f
[2]: http://api.rubyonrails.org/classes/ActionView/Base.html

### DIFF
--- a/app/presenters/outcome_presenter.rb
+++ b/app/presenters/outcome_presenter.rb
@@ -95,8 +95,7 @@ class OutcomePresenter < NodePresenter
   end
 
   def render_erb_template(template, view_context)
-    safe_level, trim_mode = nil, '-'
-    ERB.new(template, safe_level, trim_mode).result(view_context.get_binding)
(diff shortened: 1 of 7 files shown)


tadast:
LGTM

PR #1683, opened 2015-06-08
Use arrays where data should be an array

tadast:
It looks like due to typos we had some data that was supposed to be
arrays of countries, but was a space separated string instead.

It kept working because active support's #include? matched a substring
provided. So this should not change any behaviour, just make things
more consistent and less surprising.


floehopper:
LGTM

PR #1686, opened 2015-06-09
Use ActionView instead of Erubis to render outcome templates

floehopper:
In general this moves us closer to being a standard Rails app which is one of
our longer-term goals. More specifically:

* By supplying the state in the `locals` Hash, we don't have to worry about the
view modifying the state via writer methods.
* Strings marked as HTML-unsafe (e.g. user input) are automatically escaped.
I've added an integration test to demonstrate this.
* We don't have to do mess about with a binding object.
* Note that `ActionView` uses Erubis in its implementation which is what we
were using previously, so this reduces the scope of the change.
* A different exception is raised if a non-existent state variable is accessed:
`ActionView::Template::Error` instead of `NoMethodError`.
* It should now be much easier to introduce "partial" templates for sharing
content between outcome templates.

### DIFF
--- a/app/presenters/outcome_presenter.rb
+++ b/app/presenters/outcome_presenter.rb
@@ -2,12 +2,12 @@ class OutcomePresenter < NodePresenter
   def initialize(i18n_prefix, node, state = nil, options = {})
     @options = options
     super(i18n_prefix, node, state)
+    @view = ActionView::Base.new(["/"])
   end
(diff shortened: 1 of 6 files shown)


floehopper:
Closing to investigate an unexpected issue.

floehopper:
This has been superseded by #1689.

PR #1687, opened 2015-06-09
Improve "Next node undefined" error message

chrisroos:
I ran into this error when testing the uk-visa-checker with st-martin selected
as my passport.

The previous error message was:

> Next node undefined (staying_for_how_long?(six_months_or_less))

This doesn't give enough information to recreate the error (i.e. it doesn't
contain all the responses).

The new message for the same error mentioned above is:

> Next node undefined. Node: staying_for_how_long?. Responses: ["st-martin",
"work", "six_months_or_less"]

While I'm sure the copy can be improved further, I think this is enough of an
improvement for now.


tadast:
LGTM :+1:

floehopper:
LGTM

PR #1689, opened 2015-06-09
Use ActionView instead of Erubis to render outcome templates

floehopper:
In general this moves us closer to being a standard Rails app which is one of
our longer-term goals. More specifically:

* By supplying the state in the `locals` Hash, we don't have to worry about the
view modifying the state via writer methods.
* Strings marked as HTML-unsafe (e.g. user input) are automatically escaped.
I've added an integration test to demonstrate this.
* We don't have to do mess about with a binding object.
* ActionView helper modules (e.g. `ActionView::Helpers::NumberHelper`) are
automatically made available.

Notes:

* `ActionView` uses Erubis in its implementation which is what we
were using previously, so this reduces the scope of the change.
* A different exception is raised if a non-existent state variable is accessed:
`ActionView::Template::Error` instead of `NoMethodError`.
* It should now be much easier to introduce "partial" templates for sharing
content between outcome templates.

Supersedes #1686.

### DIFF
--- a/app/presenters/outcome_presenter.rb
+++ b/app/presenters/outcome_presenter.rb
@@ -2,12 +2,12 @@ class OutcomePresenter < NodePresenter
   def initialize(i18n_prefix, node, state = nil, options = {})
     @options = options
     super(i18n_prefix, node, state)
-    @view = ActionView::Base.new(["/"])
+    @view = ActionView::Base.new([template_directory])
   end
 
   def title
     if use_template? && title_erb_template_exists?
-      title = @view.render(template: title_erb_template_path, locals: @state.to_hash)
+      title = @view.render(template: title_erb_template_name, locals: @state.to_hash)
       title.chomp
     else
       translate!('title')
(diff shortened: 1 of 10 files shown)


tadast:
LGTM :+1:

app/presenters/outcome_presenter.rb
```diff
@@ -1,41 +1,13 @@
...
-
-    def method_can_be_delegated_to_state?(method)
-      @state.respond_to?(method) && !method.to_s.end_with?('=')
-    end
-  end
-
   def initialize(i18n_prefix, node, state = nil, options = {})
     @options = options
     super(i18n_prefix, node, state)
+    @view = ActionView::Base.new(["/"])
```
chrisroos: I think you mentioned not being completely happy with having to set the load path to `"/"`. Are you now satisfied that this isn't actually a problem?

floehopper: Good point. I'd forgotten about it. Thanks. I'll have a look now.

floehopper: I've addressed this in 221cfd4908b3e4430af81038434e6f1946fcacf5.

floehopper:
I've just pushed a bunch more commits to this branch. The main change to production code is the one addressing @chrisroos' point about the view paths (see comment above). Otherwise the changes are mainly refactoring the `OutcomePresenterTest`.

I should've said in the original PR description, but I've run the regression tests for the three flows which have outcome ERB templates. And I've run them successfully again with these latest changes.

chrisroos:
Looks good to me.

PR #1692, opened 2015-06-10
Remove unused methods from childcare cost calculator

chrisroos:
I came across these unused methods by looking at the coverage report when adding regression tests for childcare-costs-for-tax-credits Smart Answer.


tadast:
LGTM :+1:

PR #1693, opened 2015-06-10
Standardise birth registrations in Gibraltar

tadast:
Affected URL /register-a-birth/y/gibraltar

It use to refer to Spain for birth registrations in Gibraltar.
As per original ticket:

>the outcome for Gibraltar - https://www.gov.uk/register-a-birth/y/gibraltar - needs to be brought into line with the other 'Outcome Commonwealth' outcomes (like Australia, etc) This means that the 2 references to Spain just have to change to 'Gibraltar'.

As a side note, @floehopper, do you think we should add some countries [from this file](https://github.com/alphagov/smart-answers/blob/master/lib/data/registrations.yml) to the regression test answer list for this flow?

### DIFF
--- a/lib/data/registrations.yml
+++ b/lib/data/registrations.yml
@@ -23,7 +23,6 @@ registration_country:
   french-guiana: france
   french-polynesia: france
   gabon: cameroon
-  gibraltar: spain
   grenada: barbados
   guadeloupe: france
   guinea-bissau: senegal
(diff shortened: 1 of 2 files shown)


floehopper:
@tadast: I agree with your suggestion that we should include some countries from the `registrations.yml` file.

BenJanecke:
LGTM :+1:

tadast:
@floehopper I'll address this in a separate PR

PR #1694, opened 2015-06-10
Remove unreachable "result" outcome from overseas-passports (part 2)

floehopper:
There are only two questions from which it's theoretically possible to reach
the `:result` outcome:

* `:child_or_adult_passport?`
* `:country_of_birth?`

In both these questions there are a bunch of `next_node_if` statements and a
`next_node` statement inside an `on_condition` block, all followed by a final
`next_node` statement which points at the `:result` outcome.

In both cases the final `next_node` statement within the `on_condition` block
points at the `:ips_application_result` outcome i.e. not the `:result` outcome.

Thus the `:result` outcome is only reachable if the
`SmartAnswer::Calculators::PassportAndEmbassyDataQuery#ips_application?`
predicate returns `false`.

However since `st-helena-ascension-and-tristan-da-cunha` was switched to `type:
ips_application_3` in [this commit][1] the predicate only returns `false` when the
response to the `:which_country_are_you_in?` question is one of the following:

* `syria`
  * However, in this case the `PassportAndEmbassyDataQuery#ineligible_country?`
  predicate returns `true` for a response of "syria" and the user is taken
  immediately to the `:cannot_apply` outcome.
* `the-occupied-palestinian-territories`
  * However, in this case the user is subsequently asked to choose either
  `gaza` or `jerusalem-or-westbank` and either of these responses cause the
  `PassportAndEmbassyDataQuery#ips_application?` to return `true`.
* `united-kingdom`
  * However, in this case this response is not included in the
  `:which_country_are_you_in?` question's country select list (it is not
  included in such questions by default; only if the `:include_uk` option is
  set to `true`). So it wouldn't be possible for a user to give this response
  in production without hacking the URL.

Thus I'm pretty convinced the `:result` outcome is unreachable given the
current logic and data. Obviously if the data changed, it might be possible
for the outcome to become reachable again, but I don't think the overhead of
keeping all this code hanging around is justified.

I've elected to remove the two `next_node(:result)` statements which (although
perhaps slightly controversial) should mean we will see a `RuntimeError`
exception with the message "Next node undefined" if none of the other
conditions are satisfied by a response from the user.

I've also tried to remove as much code, content & data which the removal of
the outcome has rendered redundant.

[1]: 3a86b8ecd1c4686885dcbd9466da43f504999df9

### DIFF
--- a/lib/data/passport_data.yml
+++ b/lib/data/passport_data.yml
@@ -191,7 +191,6 @@ belize:
 benin:
   type: ips_application_1
   group: ips_documents_group_3
-  helpline: pretoria_south_africa
   app_form: hmpo_1_application_form
   address: durham
   renewing_new: 6_weeks
@@ -316,7 +315,6 @@ burma:
 burundi:
   type: ips_application_3
   group: ips_documents_group_3
-  helpline: pretoria_south_africa
   app_form: hmpo_1_application_form
   renewing_new: 6_weeks
   renewing_old: 10_weeks
@@ -333,7 +331,6 @@ cambodia:
 cameroon:
   type: ips_application_1
   group: ips_documents_group_3
-  helpline: pretoria_south_africa
   app_form: hmpo_1_application_form
   address: durham
   renewing_new: 6_weeks
@@ -576,7 +573,6 @@ equatorial-guinea:
 eritrea:
   type: ips_application_1
   group: ips_documents_group_3
-  helpline: pretoria_south_africa
   app_form: hmpo_1_application_form
   address: durham
   renewing_new: 6_weeks
@@ -595,7 +591,6 @@ estonia:
 ethiopia:
   type: ips_application_1
   group: ips_documents_group_3
-  helpline: pretoria_south_africa
   app_form: hmpo_1_application_form
   address: durham
   renewing_new: 8_weeks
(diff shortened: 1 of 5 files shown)


floehopper:
I'm a bit less sure about removing this code than I was. I've come across some `else` statements in `precalculate` blocks in the rest of the `overseas-passports` flow which are similarly unreachable given the current dataset.

I think there's a difference between removing code that is fundamentally unreachable versus code that is currently unreachable *because of the current dataset*.

Having said that, it also feels wrong to maintain a bunch of code "just in case" the data changes at some time in the future.

@tadast, @BenJanecke: Do you have any thoughts on this?

tadast:
I've had the very same issue in the marriage abroad flow. After looking at the data I just made an educated guess that it's not likely to change in a way that leaving the code would be beneficial. 

Unless the dataset has changed often historically, or the responsible content designer (Abigail in this case) thinks it will change soon, I'd remove it. Most of the times we get requests that are quite concrete i.e. given inputs X the result should be <smartdown document>. So the worst that can happen is a developer re-implementing the missing code without knowing it can be recovered from git.

BenJanecke:
I think we should take this on a case per case basis, but unless you can replicate the flow that actually executes that code I would suggest removing it.

floehopper:
Thanks, folks. I've started leaning back towards removing it, because having an unreachable outcome breaks one of the fundamental assertions in the regression tests.

Also I've now run the tests with *all* combinations of countries & other responses and I have confirmed that I cannot reach the outcome.

PR #1696, opened 2015-06-11
Update Simplecov gem

tadast:
You can see the changelog here https://github.com/colszowka/simplecov/blob/master/CHANGELOG.md.

Nothing concrete we need, but it's always better to use the latest and the greatest :)


floehopper:
I can't see anything in the changelog that looks scary, so I reckon this is worth doing. Is `simplecov-rcov` also at the most recent version? If not I think it would be worth updating it at the same time.

tadast:
Yes, we're on 0.2.3 and it's currently the latest version https://rubygems.org/gems/simplecov-rcov.

I've used the newest version of simplecov to add regresion tests to marriage abroad and it worked well :)

floehopper:
:+1:

PR #1697, opened 2015-06-11
Regression tests for marriage abroad

tadast:
As suggested by @chrisroos [here](https://github.com/alphagov/smart-answers/pull/1688) I'm raising a separate PR for adding regression tests for marriage abroad branch that is being fact checked.

PR #1688, opened 2015-06-09
Add regression tests to marriage abroad

tadast:
This is a PR for another WIP branch that is being fact-checked. 

- Update SimpleCov gem version
- Add yaml with questions and answers for marriage abroad. This includes countries that were in integrations tests + some extra to increase test coverage
- Remove :skull: code branches (based on simplecov and further investigation, see commit messages)
- add HTMl artefacts

Best reviewed commit-by-commit (except for the one that adds html artefacts)

PR #1698, opened 2015-06-11
Remove unused code branches

tadast:
I'm submitting this PR separately from #1697 as suggested by @chrisroos [here](https://github.com/alphagov/smart-answers/pull/1688).

Using the regression test setup + simplecov I have uncovered some code branches in marriage abroad that were not used anymore. I have investigated each case making sure they are really unreachable. Each removal is written up in a commit message.


BenJanecke:
This is a lot of stuff being taken out at once, the commits are at least nice and atomic so we can roll back if necessary. As far as I can see it LGTM :+1:

tadast:
Merged in https://github.com/alphagov/smart-answers/commit/47ee035ff1d0a60f8082bcb9e9e956c513581714

PR #1699, opened 2015-06-11
Add script to add/update an organisations fixture for a worldwide location

floehopper:
I've found myself needing to do this quite a few times when adding regression tests for smart answer flows. I've updated the README - I think the old instructions were quite out-of-date.


tadast:
This is going to be very helpful. LGTM :+1:

PR #1700, opened 2015-06-11
SPIKE: Add GA virtual pageview to outcome pages

chrisroos:
This results in us recording two page views for the outcome pages.

Assuming the following outcome URL:

/marriage-abroad/y/afghanistan/uk/uk_england/partner_british/opposite_sex

We'll record the following two page views in GA:

/marriage-abroad/y/afghanistan/uk/uk_england/partner_british/opposite_sex
/outcome/marriage-abroad/y/afghanistan/uk/uk_england/partner_british/opposite_sex

This commit is intended to be used to start a discussion about this approach.

### DIFF
--- a/app/views/smart_answers/_result.html.erb
+++ b/app/views/smart_answers/_result.html.erb
@@ -22,5 +22,7 @@
 </div>
 
 <script type="text/javascript">
-  GOVUK.analytics.trackPageview('/outcome' + window.location.pathname);
+  var flowName = '<%= @name %>';
+  var outcomeName = '<%= @presenter.current_node.name %>';
+  GOVUK.analytics.trackPageview('/' + flowName + '/outcome-reached/' + outcomeName);
 </script>
--- a/app/views/smart_answers/_result.html.erb
+++ b/app/views/smart_answers/_result.html.erb
@@ -20,3 +20,7 @@
     <% end %>
   </article>
 </div>
+
+<script type="text/javascript">
+  GOVUK.analytics.trackPageview('/outcome' + window.location.pathname);
+</script>


tadast:
Technically it looks good. I do not feel like I'm knowledgable enough to comment about impact of this to analytics, but some things to think about are:
- will this skew our pageview stats?
- will this be confusing as the url tracked does not really represent any resource (if so, could we record a pageview with a query parameter `?outcome=true` or similar?)

edds:
Is the need for this you want to be able to see how often people get to the end of a smart answer? 

Could we use a custom dimension for this instead? It doesn't feel right adding this as another page view. That would still allow you to track how many people see it and wouldn't count extra pageviews that aren't happening.

chrisroos:
> Is the need for this you want to be able to see how often people get to the end of a smart answer?

@edds: Yup, it certainly is.

> Could we use a custom dimension for this instead?

Very possibly. I've gone down the virtual pageview route as that's what the Analytics chap suggested. I'll mention a custom dimension to him.

chrisroos:
> will this be confusing as the url tracked does not really represent any resource (if so, could we record a pageview with a query parameter ?outcome=true or similar?)

@tadast: I can see how this would've been quite nice ('?outcome=true' or '?outcome=<outcome-name>') but it seems that our Analytics is currently configured to ignore everything in the Smart Answer path after `/y/` so I don't think this would help.

dsingleton:
:+1: to the overall idea. The nature of SA urls makes analytics around them a real pain.

But +1 to @edds on custom dimension vs virtual page view. We should be annotating state to the existing page view, rather than generating additional ones, which may throw off over metrics accidentally. 

If you go the dimension route then tracking a state that could be used for more than just boolean outcome would be good. So we'd have the flexibility to track drop-off by question type (or similar).

For a bit of historical context, SA completion was tracked via an event previously, see:
- https://github.com/alphagov/static/blob/167382a605654505a57431e0dfc6e47b33a4675f/app/assets/javascripts/analytics/tracking.js#L103-L109
- https://github.com/alphagov/static/blob/167382a605654505a57431e0dfc6e47b33a4675f/app/assets/javascripts/analytics/tracking-strategies.js#L65-L71

It wasn't found to be very useful by the Analyst working on SA at the time, and it broke for a couple of months without anyone noticing, so we removed the whole block of track/strategies JS. It was an event because thats what the rest of generic-tracking code did, which IMO wasn't a good fit for SA.

chrisroos:
Closing this fow now as everything's gone a bit quiet at this end.

PR #1702, opened 2015-06-11
Remove redundant simplified-expenses-checker-v2

floehopper:
This draft flow was created in [this commit][1].

Other than the fact that the v2 version is in the `draft` status, the only
difference between v1 and v2 is explained by [this change][2] to v1. Thus the
v1 version is more up-to-date than the v2 version.

No changes have been made to the v2 version that haven't also been made to the
v1 version.

So I think it's safe to delete this flow and its associated files.

[1]: dfbb2bd55fd73d2f5553f49925de4a472a530ff7
[2]: a71b941ba23a72b161764bb6ce20ea8c3fba3876


tadast:
:shipit:

PR #1705, opened 2015-06-12
Update towing rules: text change A6

lutgendorff:


tadast:
@lutgendorff thanks for making the change. It would be amazing if you could give a sample of an affected URL for your future PRs :)

PR #1707, opened 2015-06-15
Marriage abroad ticket 14

tadast:
Sorry, this is a MASSIVE change. The most prominent changes are:

- Removes ‘Irish’ option from “What is your partner’s nationality?”. The rules for Irish nationals are the same as for nationals of any other country now.
- Removes ability to post CNI/swear affidavit in a third country. Therefore the second country dropdown asking where the user is residing was replaced with a three option question: UK/country of ceremony/other. For Third country resident CNI outcomes we just point to two other answers.
- Updates a list of CNI countries that require 3 day residency before posting notice.
- Standardises the use of phrases for travel & legal advices, embassy API usage etc
- Renames phrase keys to better reveal their purpose
- Standardises the use of conditional logic (it used to use `and`, `&` and `&&` interchangeably)
- Converts marriage-abroad flow to a class so we can use simplecov
- Adds regression tests (mostly to test-drive how well/badly it feels on a complex flow like this)

Some of the changes were reviewed as separate commits or PRs to this branch. I did not manage to submit granular PRs for each change, because the project management process was very chaotic - things were changing often and I kept rebasing. However, this has been fact-checked and FCO have no comments remaining.


BenJanecke:
This is a lot of stuff and was reviewed on a rolling basis during development by the team, this has also been thoroughly fact checked by FCO and regression tests have been added. 

http://images.fandango.com/images/spotlight/nux-chrome.gif
Witness me! as I merge this.

PR #1710, opened 2015-06-15
Raise exceptions when marriage abroad is misused

tadast:
I added a couple of errors to inform developers when they're doing things in a manner that is unexpected by the existing logic.

Detailed explanations can be found in the commit messages.

Pro tip: add `?w=1` to the url param to ignore withespace changes when reviewing.

### DIFF
--- a/lib/smart_answer/question/base.rb
+++ b/lib/smart_answer/question/base.rb
@@ -2,6 +2,7 @@ module SmartAnswer
   module Question
     class Base < Node
       attr_reader :next_node_function_chain
+      class NextNodeUndefined < StandardError; end
 
       def initialize(name, options = {}, &block)
         @save_input_as = nil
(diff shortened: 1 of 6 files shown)


BenJanecke:
This needs a rebase

chrisroos:
I've left a couple of comments but overall I think this looks good to me.

chrisroos:
> The whole reason for the change is basically to make it obvious that this outcome does not have a default set of phrases.
>
> Having said it all I do not feel strongly about it, but I believe it's more useful than harmful.

That sounds entirely sensible to me. I hadn't considered the case you've demonstrated with the multiple `if ... end`'s versus the `if ... elsif ... else ... end`'s.

tadast:
Merged manually: 554aacf9ae2871293e8d12b5f81304fb78e8558f aa8fc2dc4d2dbfb9e8340a9e22e576790b02bee3 0d714cf79d13377cd3bca63f58328d88bec67331

PR #1712, opened 2015-06-17
Update Rails and rack to address CVE-2015-3225, CVE-2015-3226 and CVE-2015-3227

tadast:


chrisroos:
I think it'd be useful to update the commit note with a little more information. I'd love to see links to the CVEs, an explanation as to which CVE is fixed by which version change and ideally the command you ran to update the Gemfile.lock.

bradleywright:
Those are all good suggestions @chrisroos, but in this case I'm merging because we've patched most other apps and I want this code secure.

PR #1713, opened 2015-06-17
Marriage abroad ticket 15. Estonia

tadast:
Add a callout for Estonia's outcomes when residing in the country.

There is a CNI exception for long-time residents.

### DIFF
--- a/lib/smart_answer_flows/locales/en/marriage-abroad.yml
+++ b/lib/smart_answer_flows/locales/en/marriage-abroad.yml
@@ -295,6 +295,8 @@ en-GB:
           Your partner will probably need to get an equivalent document from their national authorities.
         partner_declaration: |
           ^Your partner will need to make a declaration as well.^
+        cni_exception_for_permanent_residents_estonia: |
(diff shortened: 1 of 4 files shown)


chrisroos:
Should this also contain updates to the regression test html outcomes in test/artefacts?

chrisroos:
Other than my question about html outcomes this looks good to me.

tadast:
@chrisroos ideally it would, but I'm planning to do that on a tt_ma_ticket_15 -> master PR (note that this is a PR for tt_ma_ticket_15 branch).

I'm not doing that simply because it takes 1h to generate HTML artefacts and I will have 20 or so PRs for this batch of changes.

However, if I could somehow restrict the set of answers to use for a specific question when generating HTML artefacts e.g. `ceremony_country=estonia,latvia` it would improve the experience for me :)

chrisroos:
> I'm not doing that simply because it takes 1h to generate HTML artefacts and I will have 20 or so PRs for this batch of changes.

Bah, sorry. I didn't realise it took so long to generate the output. We'll have a think about whether there's anything we could do to speed it up.

tadast:
Merged manually https://github.com/alphagov/smart-answers/commit/34d1ae59098475b82a43c3edac7ba86c415d95f2

PR #1714, opened 2015-06-17
[In Fact Check - DO NOT MERGE] Updates to calculate-state-pension

BenJanecke:
[Pivotal Tracker story](https://www.pivotaltracker.com/story/show/94781974).

It is not possible to give accurate estimations for people over the age of 55 so we provide them with only the date and age and then send them to a helpline for further assistance.

Url to test 
/calculate-state-pension/y/amount/male/1958-01-02


tadast:
Run regression tests and integration tests with simplecov and see if there are untested code branches or additional code to remove.

Also, consider using relative dates instead of stubbing time for the whole test (unless this is scheduled to be replaced by another tool, as you've mentioned).

chrisroos:
It looks like you're missing the updated "calculate-state-pension-responses-and-expected-results.yml" file: It needs to be regenerated after making changes to the questions-and-responses.yml file. You can generate it with `rails r script/generate-responses-and-expected-results-for-smart-answer.rb calculate-state-pension`. You'll then need to update the checksum file and regression tests artefacts.

chrisroos:
I'm not too keen on stubbing time for the whole test. I'd prefer for each test to set-up what it needs in order to make the assertions it's concerned with. You already seem to be doing (e.g. cd6860fb21b2ed6137fe503aa8ec15045cb3cd15) so I think that stubbing time for the whole test may be overkill.

chrisroos:
> Run regression tests and integration tests with simplecov and see if there are untested code branches or additional code to remove.

I agree with @tadast. You've removed a lot of tests but none of the corresponding code. It looks like there must be quite a lot we can get rid of as part of this commit.

tadast:
@chrisroos this has come back from FC and is ready to go live. Is it ok if you handle merging this to master, as you've been working on it already?

chrisroos:
@tadast: Sure thing. I'll get this merged to master.

chrisroos:
OK, so this still needs more work. These changes to the Smart Answer have resulted in unreachable code that we should remove (found by using the generate-responses-and-expected-outcomes script with code coverage enabled).

I'm not going to be able to do this until Tuesday next week so if it's urgent (i.e. needs to go live before Tuesday) then I'd suggest either @floehopper or @tadast picking it up.

tadast:
@chrisroos I've done some cleanup, but I didn't find much code to remove. I'll wait for you to double-check this when you're back.

chrisroos:
Thanks @tadast. I'm going to take a look at this now.

chrisroos:
This looks good to me. I can see that there a few lines that still aren't covered by code coverage but it looks like they may have been redundant before this change anyway - See the notes at the bottom of the commit note for 086a7ca0be520d0ef04e5a50f98465e29c01aa5e.

I think it'd be good to reorganise the last two commits a bit. I think I'd go for three commits:

* Update questions-and-responses (currently in the last commit)
* Update responses-and-expected-results (currently in the last commit)
* Update rendered outcomes in test/artefacts (current in the last two commits)

Can I leave this with you @tadast, or would you still like me to merge it?

tadast:
I can sort it out tomorrow morning. Feel free to do it today if you want to reduce the number of open PRs :)

tadast:
@chrisroos is this what you meant? ^

chrisroos:
Yup, that looks good to me @tadast :-)

PR #1715, opened 2015-06-17
Remove unused outcomes from inherits someone dies without will

chrisroos:
I discovered this unused code when adding regression tests for inherits-someone-dies-without-will.

I need to remove these unused outcomes before I can merge the regression tests as I currently get a failing test because not all nodes are being exercised.

### DIFF
--- a/lib/smart_answer_flows/inherits-someone-dies-without-will.rb
+++ b/lib/smart_answer_flows/inherits-someone-dies-without-will.rb
@@ -167,10 +167,6 @@ module SmartAnswer
           end
         end
         on_condition(variable_matches(:region, 'northern-ireland')) do
-          on_condition(variable_matches(:partner, 'yes')) do
-            next_node_if(:outcome_64, responded_with('yes'))
-            next_node_if(:outcome_65, responded_with('no'))
-          end
           on_condition(variable_matches(:partner, 'no')) do
             next_node_if(:outcome_4, responded_with('yes'))
             next_node_if(:grandparents?, responded_with('no'))
--- a/lib/smart_answer_flows/inherits-someone-dies-without-will.rb
+++ b/lib/smart_answer_flows/inherits-someone-dies-without-will.rb
@@ -111,10 +111,6 @@ module SmartAnswer
         save_input_as :parents
 
         on_condition(variable_matches(:region, 'england-and-wales')) do
-          on_condition(variable_matches(:partner, 'yes')) do
-            next_node_if(:outcome_21, responded_with('yes'))
-            next_node_if(:siblings?, responded_with('no'))
-          end
           on_condition(variable_matches(:partner, 'no')) do
             next_node_if(:outcome_3, responded_with('yes'))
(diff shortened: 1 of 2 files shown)


chrisroos:
The build failure on Jenkins was caused by our intermittent Poltergeist error. I've restarted the build.

lib/smart_answer_flows/inherits-someone-dies-without-will.rb
```diff
@@ -111,10 +111,6 @@ def define
         save_input_as :parents
 
         on_condition(variable_matches(:region, 'england-and-wales')) do
-          on_condition(variable_matches(:partner, 'yes')) do
-            next_node_if(:outcome_21, responded_with('yes'))
-            next_node_if(:siblings?, responded_with('no'))
-          end
           on_condition(variable_matches(:partner, 'no')) do
```
tadast: Does it make sense to keep `on_condition(variable_matches(:partner, 'no'))` if no other cases lead to this node?

floehopper: @tadast: [This commit note](871d09ada36367b41eb7e9461e1ee61f9f870bc2) suggests that it is **still** possible to reach the `:parents?` and `:siblings?` questions if `partner` is set to "no", so I think we do still need to keep the `on_condition` statement which you mention. Or have I misunderstood?

floehopper:
LGTM :+1:

lib/smart_answer_flows/inherits-someone-dies-without-will.rb
tadast: No, you're right. :+1:

chrisroos: I'm a bit confused. I actually think we probably can remove the use of `on_condition`, but leave the inner `next_node_if`'s, because the only way to get to this section of code is by responding 'no' to `:partner`.

That would mean that this:

```
on_condition(variable_matches(:partner, 'no')) do
  next_node_if(:outcome_3, responded_with('yes'))
  next_node_if(:siblings?, responded_with('no'))
end
```

would become:

```
next_node_if(:outcome_3, responded_with('yes'))
next_node_if(:siblings?, responded_with('no'))
```

Is that what you meant, @tadast?

@floehopper: Do you see a problem with removing these out `on_condition`'s?

floehopper: Ah. I see. I had misunderstood. I don't see a problem with removing the `on_condition` as you've described.

tadast: That's what I meant originally, but then I got confused. I guess the tests should fail if unwrapping those two lines from `on_condition` changes anything :)

chrisroos:
Thanks for the feedback, folks. I've made the changes we talked about (removing the outer `on_condition`'s) and merged to master in 01034fa85cef94f4cf84f82555dda7220b1d4463.

PR #1716, opened 2015-06-18
Customise Kuwait outcomes for local residents

tadast:
1) American express cards are not accepted for payments
2) Users must check with Notary public if they need a CNI (not an embassy)


BenJanecke:
Regression tests for this will be generated once the entire ticket 15 has been finished.

PR #1717, opened 2015-06-18
Customise Montenegro outcomes

tadast:
1) Advise users to check with Notary Public if they need a CNI.
   This is to avoid users contacting embassy, as they can't help
2) Add appointment booking link to OS outcomes
3) Add appointment booking link for SS outcomes where both partners
   are british


floehopper:
Other than one small comment which can probably be addressed in the commit note, this LGTM.

PR #1719, opened 2015-06-18
Marriage abroad ticket 15. Greece

tadast:
Customise Greece outcome to include Greece-specific list of supporting document requirements. 
Add CNI collection instructions

### DIFF
--- a/lib/smart_answer_flows/marriage-abroad.rb
+++ b/lib/smart_answer_flows/marriage-abroad.rb
@@ -742,8 +742,8 @@ module SmartAnswer
             end
           end
 
-          if ceremony_country == 'italy' && resident_of != 'uk'
-            phrases << :required_supporting_documents_italy
+          if resident_of != 'uk' && data_query.phrase_exists?("required_supporting_documents_#{ceremony_country}")
+            phrases << :"required_supporting_documents_#{ceremony_country}"
           elsif resident_of == 'ceremony_country' && %w(germany japan spain).exclude?(ceremony_country)
             if cni_notary_public_countries.include?(ceremony_country) || %w(japan macedonia spain).include?(ceremony_country)
               phrases << :required_supporting_documents_notary_public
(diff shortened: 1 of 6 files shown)


lib/smart_answer_flows/marriage-abroad.rb
```diff
@@ -733,8 +733,8 @@ def define
             end
           end
 
-          if ceremony_country == 'italy' && resident_of != 'uk'
-            phrases << :required_supporting_documents_italy
+          if resident_of != 'uk' && data_query.phrase_exists?("required_supporting_documents_#{ceremony_country}")
```
tadast: I'm testing a new approach here. Instead of relying on a hardcoded list of countries that have exceptions and trusting those phrases are in the yml file, this relies purely on the exceptional phrase being present in the yml file.

What do you think about this approach?

floehopper: I think this is something you mentioned the other day. It seems reasonable given the current state of the code, but more generally I do wonder whether it's adding yet another place where such logic exists.

It feels to me as if the code would be easier to follow if all the characteristics of a given country were represented in a single place, rather than having list of countries and exceptions. But that's a job for another day.

floehopper:
Other than my small comment about the name of a test, this LGTM.

lib/smart_answer_flows/marriage-abroad.rb
tadast: Do you mean something like this: https://github.com/alphagov/smart-answers/blob/efe056e4f6b67efb2677e0e868f0b50b88774a2d/lib/smart_answer_flows/marriage-abroad.rb#L373-L475?

If so it would make it easier to make changes that affect a single country, but it would also introduce some duplication in logic that would make it more difficult to make changes that affect multiple countries. In this flow both things happen quite often.

floehopper: I suppose I do mean something like that, but I'm thinking further ahead when this kind of thing would be extracted out of the flow class.

I take your point about potentially making changes affecting multiple countries harder, but I think the data should be modelled in a way that makes it easy to use & understand within the application rather than in a way that makes it easy to update. If necessary, we can then provide tools to bulk-edit multiple countries. Anyway, this is probably a discussion for another day and shouldn't affect this PR.

PR #1721, opened 2015-06-18
Remove unused code from energy grants calculator

chrisroos:
I discovered this unused code while adding regression tests for this Smart Answer.

I wanted to remove the unused outcome nodes so that I could get the regression tests to pass, and I ended up removing the rest of the code where it was obvious that it wasn't reachable.


floehopper:
Apart from my one small comment, this LGTM :+1:

chrisroos:
I've amended the commit as you've suggested @floehopper.

Manually merged in 87ebcf50575df58ab0544b0ab70a51b30be9999f.

PR #1722, opened 2015-06-19
Change payment method for Russia

tadast:
Users can pay only by MasterCard and Visa credit cards in both consulates

### DIFF
--- a/lib/smart_answer_flows/locales/en/marriage-abroad.yml
+++ b/lib/smart_answer_flows/locales/en/marriage-abroad.yml
@@ -144,6 +144,8 @@ en-GB:
         embassies_data: |
           +[data_partial:overseas_passports_embassies:overseas_passports_embassies]
 # payment methods
+        pay_by_mastercard_or_visa: |
+          You can only by MasterCard and Visa credit cards.
         payment_methods_japan: |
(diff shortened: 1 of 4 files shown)


chrisroos:
This pull request says it contains 28 commits. Is that right, @tadast?

tadast:
@chrisroos I've rebased tt_ma_ticket_15 on master after this commit was pushed. Needed to rebase tt_ma_ticket_15_russia as well. Fixed now

lib/smart_answer_flows/locales/en/marriage-abroad.yml
```diff
@@ -144,6 +144,8 @@ en-GB:
         embassies_data: |
           +[data_partial:overseas_passports_embassies:overseas_passports_embassies]
 # payment methods
+        pay_by_master_or_visa: |
```
chrisroos: I think I'd prefer this to be named `pay_by_mastercard_or_visa`. While `master` may imply `mastercard` in the context of `master_or_visa` my preference would always be for explicitness.

chrisroos:
1 commit is much better than 28 :-)

Apart from my minor comment, this looks good to me.

tadast:
I've addressed the comment and merged in https://github.com/alphagov/smart-answers/commit/84471086a30c194051a94e119b138f2ee57a8f7e

PR #1723, opened 2015-06-19
Convert apply-tier-4-visa to ERB

chrisroos:


tadast:
Nice! Have you regenerated HTML artefacts for regression tests? I'm pleasantly surprised not even whitespace changes were introduced.

:+1:

chrisroos:
> Have you regenerated HTML artefacts for regression tests? I'm pleasantly surprised not even whitespace changes were introduced.

Yup, and me too :-) I think the template could be made slightly nicer with a little more whitespace but for now I'm mostly interested in keeping the output identical.

floehopper:
The "Update apply-tier-4-visa-files.yml" commit looks as if it's at the wrong point in the git history i.e. I would've expected it to be last. Otherwise this looks good to me.

chrisroos:
> The "Update apply-tier-4-visa-files.yml" commit looks as if it's at the wrong point in the git history i.e. I would've expected it to be last.

I reordered the commits and merged this branch to master in f0081a48c31345397619304d6ce9fa8f724567e2.

PR #1724, opened 2015-06-19
Seychelles now offer same sex marriages

tadast:
The British high commission in the Seychelles is now offering same-sex
consular marriages. The advice should mirror same outcomes we have for
Bolivia same-sex marriage.


chrisroos:
You've mentioned that the outcome for Seychelles should match that of Bolivia. I can see that you've added a test for Seychelles but the lack of an equivalent test for Bolivia would make it easy for the two to drift apart in future. Is that something we're concerned about?

I'd find it useful if the commit note included the command used to generate seychelles_organisations.json.

Apart from that one question and minor comment, this all looks good to me.

tadast:
> You've mentioned that the outcome for Seychelles should match that of Bolivia.

That's was just the easiest way to define requirements for FCO. Those two countries are not related and are not expected to stay in sync.

> I'd find it useful if the commit note included the command used to generate seychelles_organisations.json.

I'll add it

PR #1726, opened 2015-06-19
Convert calculate-statutory-sick-pay flow to class

floehopper:
Converts the flow to a class and directly instantiates it in the corresponding integration test to bring it into line with the majority of other smart answer flows.

Although I need to convert some other flows, I decided to keep this in its own branch, because I'm going to need to rebase the corresponding fact-checking branch(es) and it feels as if it will be easier to manage that process with this as a separate branch.


chrisroos:
Looks good to me.

PR #1727, opened 2015-06-19
Marriage abroad amendments for Italy

tadast:
1) Remove and change duplicated required supporting documents list
2) Remove references to notary public, as it is not involved in the process there
3) Update translator links
4) Explain how CNIs are issued in Italy
5) Add and appointment booking link


jackscotti:
:it:

PR #1728, opened 2015-06-19
Remove draft benefits-abroad flow

floehopper:
@BenJanecke says that the canonical flow is `uk-benefits-abroad` and that this
was one of a number of (failed) attempts at building it. This is corroborated
by the fact that the [last changes][1] of any substance were on 1st May, 2013.

As far as I can see it doesn't have any locale file associated with it, but I
have removed the corresponding integration test.

I've also removed it from the list of flows in the `SmartAnswer::FlowRegistry`.

[1]: 9dff2215b33a62e14ad816f2b3ec4ebe05762bfc


chrisroos:
Looks good to me.

PR #1729, opened 2015-06-19
Convert redundancy-pay-related flows to classes

floehopper:
There are two redundancy-pay-related flows which share the same logic, differing only in terms of their locale files:

* `calculate-employee-redundancy-pay`
* `calculate-your-redundancy-pay`

This PR converts the two flows into classes and directly instantiates the `SmartAnswer::CalculateEmployeeRedundancyPayFlow` one in the relevant integration test to bring things into line with the majority of other smart answer flows.

Although I need to convert some other flows, I decided to keep these in their own branch, because I'm going to need to rebase the corresponding fact-checking branch(es) and it feels as if it will be easier to manage that process with these changes as a separate branch.

c.f. #1726


chrisroos:
Looks good to me.

PR #1734, opened 2015-06-22
Remove appeal-a-benefits-decision draft flow

floehopper:
This draft flow hasn't had any logic/content-related changes since [Nov 16, 2012][1]
(over 2.5 years ago) when it was moved from `published` to `draft` and
[Nov 6, 2012][2] when it was first (?) published.

Given that there's a significant cost in terms of developer effort in keeping
it around, we've decided to archive it for now. If we find out in the near
future that it is needed, we should be able to bring it back to life without
too much difficulty.

[1]: 915e10a
[2]: 1cc8903


chrisroos:
Looks good to me.

Issue #1624, opened 2015-05-11
Test failure on Jenkins - Capybara::Poltergeist::JavascriptError


chrisroos:
I saw this error in [Jenkins build 2378][jenkins-2378]. It appears to be intermittent as the following build succeeded.

The tests were run with seed 53125.

[jenkins-2378]: https://ci-new.alphagov.co.uk/job/govuk_smart_answers/2378/consoleFull

```
ReportAProblemTest
  test_: with javascript smart-answer start page should have current url in report-a-problem. ERROR (0.11s)
Capybara::Poltergeist::JavascriptError:         Capybara::Poltergeist::JavascriptError: One or more errors were raised in the Javascript code on the page. If you don't care about these errors, you can ignore them by setting js_errors: false in your Poltergeist configuration (see documentation for details).
        
        TypeError: 'null' is not an object (evaluating 'data['title']')
        TypeError: 'null' is not an object (evaluating 'data['title']')
            at http://127.0.0.1:50084/smartanswers/smart-answers.js:97 in addToHistory
            at http://127.0.0.1:50084/smartanswers/smart-answers.js:89
            at https://assets.digital.cabinet-office.gov.uk/static/libs/jquery/jquery-1.7.2.js:16
            at https://assets.digital.cabinet-office.gov.uk/static/libs/jquery/jquery-1.7.2.js:16
            at https://assets.digital.cabinet-office.gov.uk/static/libs/jquery/jquery-1.7.2.js:18 in S
            at https://assets.digital.cabinet-office.gov.uk/static/libs/jquery/jquery-1.7.2.js:18
            /home/jenkins/bundles/govuk_smart_answers/ruby/2.1.0/gems/poltergeist-1.6.0/lib/capybara/poltergeist/browser.rb:323:in `command'
            /home/jenkins/bundles/govuk_smart_answers/ruby/2.1.0/gems/poltergeist-1.6.0/lib/capybara/poltergeist/browser.rb:31:in `visit'
            /home/jenkins/bundles/govuk_smart_answers/ruby/2.1.0/gems/poltergeist-1.6.0/lib/capybara/poltergeist/driver.rb:95:in `visit'
            /home/jenkins/bundles/govuk_smart_answers/ruby/2.1.0/gems/capybara-2.1.0/lib/capybara/session.rb:193:in `visit'
[...]

PR #1735, opened 2015-06-23
Disable session store

floehopper:
As requested by @tekin as a precautionary measure for all "front-end" apps.

The Smart Answers app is designed to be stateless - this change makes
that explicit.

Apparently the Varnish cache strips session cookies on the way in and out
anyway, so cookies aren't currently making it through to users, but at some
point such filters may be relaxed in order to implement authentication.
This change ensures that cookies will never make it through to users.

Also our cookie policy doesn't describe any of the cookies we might set, so
we'd be breaking our own policy by enabling sessions.

c.f. alphagov/government-frontend#48

Although the configuration line was commented out in [this commit][1], this
isn't sufficient to disable session cookies; you need to actively set the session
store to `:disabled`.

Note that I was able to see the session cookie in the browser when the line was
just commented out, but only when I set a value in the session.

[1]: 84ba7be

### DIFF
--- a/config/initializers/session_store.rb
+++ b/config/initializers/session_store.rb
@@ -1,6 +1,6 @@
 # Be sure to restart your server when you modify this file.
 
-# SmartAnswers::Application.config.session_store :cookie_store, key: '_smart-answers_session'
+SmartAnswers::Application.config.session_store :disabled
 
 # Use the database for sessions instead of the cookie-based default,
 # which shouldn't be used to store highly confidential information
--- a/test/unit/session_store_test.rb
+++ b/test/unit/session_store_test.rb
@@ -1,0 +1,7 @@
+require_relative '../test_helper'
+
+class SessionStoreTest < ActiveSupport::TestCase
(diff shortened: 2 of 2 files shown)


chrisroos:
The build failure on Jenkins was caused by our recurring Poltergeist problems - #1624. I've kicked the build.

chrisroos:
Looks good to me. I can confirm that test fails if the cookie store is re-enabled.

I think it'd be worth including something along the lines of @bradleywright's comment in the commit note:

> our cookie policy doesn’t describe any of the cookies we might set, so we’d be breaking our own policy by enabling sessions. They’re also designed to be stateless so we should make that explicit.

floehopper:
@chrisroos: Thanks. I've just been amending the commit note to include @tekin's explanation from Slack, but I'll update it again to include the wording you suggest.

floehopper:
I've just amended the commit note again and updated the PR description accordingly. I'll wait for the CI build to pass and then merge this.

PR #1736, opened 2015-06-24
Convert uk-benefits-abroad to use ERB templates for all outcomes

floehopper:
I've converted one outcome at a time with each conversion in its own commit. This means that there are a lot of commits, but most of them are very straightforward.


chrisroos:
It looks like you can remove the now-unused `already_abroad_text_two` calculate block and associated phrases.

chrisroos:
This looks like a mammoth task. Good job!

All looks good to me.

floehopper:
@chrisroos: Thanks for the feedback. I believe `already_abroad_text_two` is still needed, because it's interpolated inside the [`esa_more_than_a_year ` option][1] for the `esa_how_long_abroad?` question. Or am I missing something?

I did consider renaming it (i.e. removing the `_two` suffix), but I thought that might also be confusing.

[1]: https://github.com/alphagov/smart-answers/blob/1a96852a3186254d03767f22b15754e5830200ac/lib/smart_answer_flows/locales/en/uk-benefits-abroad.yml#L266

chrisroos:
Ah, no, you're quite right. I completely overlooked the use of `already_abroad_text_two` in `esa_how_long_abroad?` question.

tadast:
Regression tests are invaluable for this kind of changes. LGTM :+1:

PR #1737, opened 2015-06-24
Use single ERB template per outcome

chrisroos:
Using `content_for` blocks allow us to combine our title and body templates
into a single template per outcome. We think this is probably easier to manage
right now, and hope that it'll make it easier to add support for `next_steps`.


app/presenters/outcome_presenter.rb
```diff
@@ -74,4 +73,19 @@ def render_data_partial(partial, variable_name)
     partial_path = ::SmartAnswer::FlowRegistry.instance.load_path.join("data_partials", "_#{partial}")
     ApplicationController.new.render_to_string(file: partial_path.to_s, layout: false, locals: {variable_name.to_sym => data})
   end
+
+  def render_erb_template
+    unless @rendered_erb_template
```
tadast: Is there any chance that `@state` could change between calling `#title` and `#body`? If so, there would be a possibility of rendering the template prematurely thus making it really hard to debug an issue. Could be worth raising an error if `@state` has changed

tadast:
It sucks that we can't indent erb blocks properly. Is the change to generated HTML visible to users or is it only white space that is causing regression tests to fail?

chrisroos:
> It sucks that we can't indent erb blocks properly. Is the change to generated HTML visible to users or is it only white space that is causing regression tests to fail?

I agree that it sucks.

Indenting the title results in additional whitespace in the rendered outcomes, which isn't really a big deal. 

Indenting the body isn't possible/easy as the leading whitespace screws with govspeak/markdown. We could do something nasty like strip the first two spaces but that doesn't seem ideal.

floehopper:
I agree with @tadast that the lack of indenting isn't great, but I think that's a problem that is independent of this PR i.e. the problem is the mixing of markdown/govspeak with ERB.

floehopper:
Other than my small comments, this LGTM :+1:

app/presenters/outcome_presenter.rb
chrisroos: I've added another commit to this branch ("Avoid instantiating OutcomePresenter multiple times") that should address your concern. The combination of `SmartAnswerPresenter#current_state` being memoized and by now passing the instance of `OutcomePresenter` to the '_result.html.erb' partial, we can be sure that the `state` object within the `OutcomePresenter` won't change.

app/presenters/outcome_presenter.rb
```diff
@@ -74,4 +73,19 @@ def render_data_partial(partial, variable_name)
...
+      @view.render(template: erb_template_name, locals: @state.to_hash)
+      @rendered_erb_template = true
+    end
+  end
+
+  def has_content_for_body?
```
tadast: I believe a common convention in ruby is to avoid `has_`, `is_` prefixes for methods that return a boolean value. A `?` is enough to indicate that.

As per http://xaviershay.github.io/writing/docs/ruby_style_guide.html

>Use ? suffix for methods rather than has_ or is_ prefix.↩
```
def valid?; "right" end
def is_valid?; "wrong" end
def has_valid?; "wrong" end
```

app/presenters/outcome_presenter.rb
tadast: This is great :+1:

app/presenters/outcome_presenter.rb
chrisroos: You're right! I think I probably went with `has_nnn?` as that seems to be the convention in [node_presenter.rb](https://github.com/alphagov/smart-answers/blob/master/app/presenters/node_presenter.rb#L79).

I'm going to leave these as they are for now (as I want to get this merged) but accept that they could be more rubyish.

chrisroos:
Manually merged in 3391fc3eaf9030f5b502672e32baf37086844f30.

PR #1739, opened 2015-06-24
Marriage abroad ticket 15. Remove birth certificate requirement

tadast:
> Users no longer need to provide a full birth certificate or naturalisation
certificate as supporting documents in the following CNI countries
(NOTE: this is not all CNI countries):

> Albania, Algeria, Angola, Armenia, Austria, Azerbaijan, Bahrain, Belarus,
Bolivia, Bosnia & Herzegovina, Bulgaria, Chile, Croatia, Cuba, Denmark,
Dominican Republic, El Salvador, Estonia, Ethiopia, Georgia, Greece, Guatemala,
Honduras, Hungary, Iceland, Kazakhstan, Kosovo, Kuwait, Kyrgyzstan, Latvia,
Lithuania, Luxembourg, Macedonia, Mexico, Moldova, Montenegro, Nepal, Norway,
Panama, Poland, Romania, Russia, Serbia, Slovenia, Spain, Sudan, Sweden, Tajikistan,
Tunisia, Turkmenistan, Ukraine, Uzbekistan, Venezuela, Democratic Republic of Congo

Note to the reviewer: this leaves us with 4 "standard" phrases:

* List without birth cert
* List without birth cert in a country with notary public
* List with birth cert
* List with birth cert in a country with notary public

I have considered interpolating the two variables above, but the whole outcome is within a precalculate block already, which is miles long, so adding another two blocks above the current one would seem out of context.

I have also considered having a list that requires birth certificate (CNI countries - countries that don't need birth cert), but I decided to leave it as it is in this PR as this is what FCO gave us, so most likely they will send updates in the same form in future.

Any suggestions how to improve this PR are very welcome


chrisroos:
When you mention interpolation, do you mean something like:

    # In marriage-abroad.rb
    precalculate :birth_certificate_requirement do
      if resident_of == 'ceremony_country' && %w(germany italy japan spain).exclude?(ceremony_country)
        if no_birth_cert_requirement.exclude?(ceremony_country)
          Phraselist.new(:birth_certificate_requirement)
        end
      end
    end

    # In marriage-abroad.yml
    birth_certificate_requirement: |
      - your [full birth certificate](/order-copy-birth-death-marriage-certificate/) or [naturalisation certificate](/get-replacement-citizenship-certificate)
    
    required_supporting_documents: |
      You’ll need to provide supporting documents, including:

      - your passport
      %{birth_certificate_requirement}
      - proof of residence, such as a residence certificate - check with the %{embassy_or_consulate_ceremony_country} to find out what you need
      - equivalent documents for your partner
    
    required_supporting_documents_incl_birth_cert_notary_public: |
      You’ll need to provide supporting documents, including:

      - your passport
      %{birth_certificate_requirement}
      - proof of residence, such as a residence certificate - check with the %{embassy_or_consulate_ceremony_country} or notary public to find out what you need
      - equivalent documents for your partner

If so, I agree that I'm not sure it's worth doing. While it removes some duplication from the phrase lists, it adds some duplication to the Smart Answer.

I can't really think of any other ways to improve what you've got so this all looks good to me.

chrisroos:
The latest commit looks good to me, @tadas, although I'd prefer the update of marriage-abroad-files.yml to be separate, i.e. move the "Update checksums for regression tests" commit after "Unify countries that require 3 day residency and don't need birth certs".

PR #1740, opened 2015-06-24
Convert calculate-married-couples-allowance to ERB

chrisroos:
__NOTE.__ This branch has been taken from use-single-erb-template-per-outcome. I'm intending to get that branch merged to master first as it makes adding support for `next_steps` (required for this SmartAnswer) easier than if we were still using multiple templates per outcome.

### DIFF
--- a/app/presenters/outcome_presenter.rb
+++ b/app/presenters/outcome_presenter.rb
@@ -37,6 +37,16 @@ class OutcomePresenter < NodePresenter
     end
   end
 
+  def next_steps
+    if use_template? && next_steps_erb_template_exists?
+      render_erb_template
+      govspeak = @view.content_for(:next_steps) || ''
+      GovspeakPresenter.new(govspeak.to_str).html
+    else
+      super
+    end
+  end
+
   def erb_template_path
     template_directory.join(erb_template_name)
   end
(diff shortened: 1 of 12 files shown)


floehopper:
I think it would be better to DRY up the calls to `number_to_currency(x, precision: ((x.to_f == x.to_f.round) ? 0 : 2))` into a helper method, e.g. `#format_money`. This code appears in these two commits:

* 4a71ca1d72e2ce1e6d8262be51512647c69bfbde
* e32566be1361a79716ab5328de7c56998b16f759

floehopper:
Other than my small comments, this LGTM :+1:

chrisroos:
> I think it would be better to DRY up the calls to `number_to_currency(x, precision: ((x.to_f == x.to_f.round) ? 0 : 2))` into a helper method, e.g. `#format_money`.

I certainly thought about DRYing it up but was put off because of uncertainty about where/how to add such a helper method: I think I'm right in saying that we haven't done that elsewhere yet.

I'll take a look in the morning to see what I can come up with.

chrisroos:
I've added fixup commits to address your comments about indenting `<%= render partial: ... %>`.

floehopper:
I think it would be fine to merge this PR and DRY these methods up later, but I'd want to ensure that happens soon. I don't think we have done it elsewhere yet.

chrisroos:
@floehopper: I've added the `format_money` helper method you suggested. Can you take a look at the approach and let me know if you think it's OK.

chrisroos:
Manually merged in 0c7b68b95ac6579e0ba8c1c8d0d884768f0ebcd4.

PR #1741, opened 2015-06-24
Marriage Abroad ticket 15. Philippines

tadast:
FCO has [a pdf](https://www.gov.uk/government/uploads/system/uploads/attachment_data/file/422903/Checklist_for_marriage_in_the_Philippines.pdf) that lists all the requirements for documentation in Philippines. Migrate the missing bits to the smartanswers outcome so they can stop directing users to the pdf.


chrisroos:
Looks good to me.

PR #1742, opened 2015-06-24
Add Google Analytics Event to outcome pages

chrisroos:
We hope that recording an event when someone reaches an outcome will help us
understand more about how Smart Answers are used.

We initially investigated using Virtual Pageviews (see PR #1700 for that
discussion) before settling on Events. This approach was suggested by Chris
Russell.

I tested this locally by configuring my local static app to send data to my
personal Google Analytics account and confirming that the events appear as
expected.

### DIFF
--- a/app/views/smart_answers/_result.html.erb
+++ b/app/views/smart_answers/_result.html.erb
@@ -20,3 +20,8 @@
     <% end %>
   </article>
 </div>
+
+<script type="text/javascript">
+  var flowName = '<%= @name %>';
+  GOVUK.analytics.trackEvent('Smart Answer', 'Completed', {label: flowName, nonInteraction: true});
+</script>


floehopper:
LGTM :+1:

tadast:
LGTM :+1: 

Unrelated to this PR: looks like we're using a lot of instance variables in this partial. A good practice in Rails is to pass them in as locals. Is it worth creating an issue for that?

edds:
What was the reasoning for not using a dimension for this? Adding some context to a pageview feels more like what dimensions are for rather than events.

chrisroos:
> What was the reasoning for not using a dimension for this? Adding some context to a pageview feels more like what dimensions are for rather than events.

@edds: We arrived at using an Event after discussion with Chris Russell. Chris said that recording these Events would give him the data required to generate the reports he's interested in. I did pass on your suggestion of using Custom Dimensions but Chris said, "I'd not come across using Custom Dimensions as goals before and realised that they aren't an option when we're setting them up within reports.". I'm afraid I don't understand Google Analytics enough to have an opinion on the approach. Do you feel strongly that we should be using Custom Dimensions?

chrisroos:
> Unrelated to this PR: looks like we're using a lot of instance variables in this partial. A good practice in Rails is to pass them in as locals. Is it worth creating an issue for that?

I did actually make this change in another branch and then threw it away as it wasn't necessary for the changes I wanted to make. Feel free to either open an issue, or make the change :-)

edds:
@chrisroos ooo, yer, trying to set up a goal using a dimension probably isn't so easy. Thanks for checking for me. :+1:

chrisroos:
Manually merged in 5d09931ce02c65ce1ca255f74960e1ee4ec2681e.

PR #1743, opened 2015-06-24
Fix New Zealand same sex marriage outcome

tadast:
The same-sex marriage outcomes for New Zealand stated embassy (even though
it’s a high commission) can perform same-sex marriages and customers can
contact them for advice. This is incorrect. The high commission cannot perform
same-sex marriages and customers should contact the relevant local authorities
in the New Zealand to find out about local laws, including what documents
you’ll need.

CP_CNI_NOT_REQUIRED_COUNTRIES was sorted alphabetically and new-zealand
was added.

There are currently no integration tests to cover New Zealand and I'm not sure it's worth adding it, as it pretty much follows one of the standard outcomes.


chrisroos:
Looks good to me.

PR #1745, opened 2015-06-25
Update pip checker

lutgendorff:
Big rework to remove the postcode area info.
Replaced with new text and new postcode areas, divided by pre/post 2017.
Made the child outcome independent of the postcode info because the way the text needs to work to explain how it all now works.


chrisroos:
Just a reminder for whoever picks this up: We'll have to regenerate the checksum data and run/re-generate the test artefacts.

tadast:
Closing in favour of #1747

PR #1748, opened 2015-06-25
Run only regression tests on Jenkins if RUN_REGRESSION_TESTS set

floehopper:
This is so that we can create a new Jenkins job to run all the regression tests
every night. They are pretty exhaustive tests and they currently take about an
hour to run on my local machine, so it's unrealistic to run them as part of the
standard CI build.

We do have a mechanism in place which will run regression tests for a specific
flow if it looks like some of its files have changed, but this isn't perfect.
So it would be good to have a nightly build which runs all the regression tests
as a safety net.

The idea would be that the standard CI build `govuk_smart_answers` would stay
as it is i.e. without `RUN_REGRESSION_TESTS` being set and triggered to run
when a change is pushed to GitHub.

A new Jenkins job, `govuk_smart_answers_regression`, would be created with
`RUN_REGRESSION_TESTS` set to `true` and with the build set to run periodically
every night.

### DIFF
--- a/jenkins.sh
+++ b/jenkins.sh
@@ -13,6 +13,11 @@ export GOVUK_APP_DOMAIN=dev.gov.uk
 export GOVUK_ASSET_HOST=http://static.dev.gov.uk
 
 export DISPLAY=:99
-RAILS_ENV=test TEST_COVERAGE=true bundle exec rake test
 
-bundle exec rake assets:precompile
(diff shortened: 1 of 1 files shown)


chrisroos:
Looks good to me.

tadast:
This was merged to master via https://github.com/alphagov/smart-answers/commit/de667a17dc6ac92c0b526ab771e470261e3266bc

PR #1753, opened 2015-06-29
Marriage Abroad ticket 15. Kazakhstan and Kyrgyzstan

tadast:
For all opposite sex marriage outcomes in Kazakhstan and Kyrgyzstan, where the users needs to post notice, they can book an appointment with the embassy in Almaty (based in Kazakhstan) using the following link: https://www.consular-appointments.service.gov.uk/fco/#!/british-embassy-office-almaty/certificate-of-no-impediment/slot_picker

Take out all embassy (API) addresses from all outcomes.


chrisroos:
I can see that you've updated marriage-abroad-questions-and-responses.yml but not marriage-abroad-responses-and-expected-results.yml. Is that intentional? I'd expect to see the responses-and-expected-results file updated when the questions-and-responses file is changed.

chrisroos:
Aside from my question this all looks good to me.

PR #1754, opened 2015-06-29
Marriage Abroad ticket 15. Norway and Belgium

tadast:
These are some major changes to the process in Norway accompanied by changes in Belgium (explained below).

The Embassy in Norway no longer issues CNIs for opposite sex marriages, instead users can now swear an affirmation or affidavit. Users will need to bring a copy of their partner’s passport. The process for UK residents remains unchanged. That means that Norway is a CNI or affirmation country depending on the user's location.

Same sex marriages in Norway and Belgium need affirmation, so a new outcome is created for that (hence two countries in one PR).

Some more changes for opposite sex marriages in Belgium needed to be done (in addition to https://github.com/alphagov/smart-answers/pull/1711) so I snuk them in here to make review/rebase/regenerate-sha process less laborious.


chrisroos:
I've added a few comments but in general it all looks good to me.

PR #1755, opened 2015-06-29
Convert state-pension-through-partner outcomes to use ERB templates

floehopper:
This was a somewhat painful process, because the flow used an unusual style. In particular there was originally only a single outcome, but that outcome consisted mainly of a single PhraseList element which was constructed completely separately in three different places.

To make things simpler, I decided to split the single outcome into three before converting the outcomes to use ERB templates.

Having converted these three outcomes to use ERB templates, I noticed that many of the partial templates I'd introduced were really outcomes in their own right (and were even referenced as such in some of the comments and tests). I therefore decided to further split the three outcomes into eight outcomes and move some of the logic that had ended up in the templates back into the flow.

New outcomes:

* age_dependent_pension
* current_rules_national_insurance_no_state_pension
* current_rules_no_additional_pension
* impossibility_due_to_divorce
* impossibility_to_increase_pension
* married_woman_and_state_pension
* married_woman_no_state_pension
* widow_and_old_pension

I decided to stick with the `next_node_if` style of flow control to match the style used in the rest of the flow. I'm sure there are a lot more improvements that could be made to this smart answer, notably I think there is a lot of duplication (or near duplication, intended to be duplication) in the outcome templates, but that can wait for another time.

### DIFF
--- a/test/data/state-pension-through-partner-files.yml
+++ b/test/data/state-pension-through-partner-files.yml
@@ -1,6 +1,15 @@
 --- 
-lib/smart_answer_flows/state-pension-through-partner.rb: fde6fc823fc6f523af9ffa6be793f0cc
-lib/smart_answer_flows/locales/en/state-pension-through-partner.yml: 741b1a6be2aeed88b3745e9351020ff0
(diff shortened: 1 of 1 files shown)


tadast:
I've reviewed it quite thoroughly and have nothing to add. 

LGTM :100:

PR #1758, opened 2015-06-29
Convert simplified-expenses-checker outcomes to use ERB templates

floehopper:
The only thing of any interest in this PR is that when I converted the `you_can_use_result` outcome to use an ERB template, I found that I had to add some `precalculate` blocks to ensure that some variables existed with a default value of `nil`. This is because of a difference in the way that state variables are interpolated in ERB templates versus in the YAML/I18n/NodePresenter mechanism. See the commit entitled "Convert you_can_use_result outcome" for further details.


floehopper:
The Jenkins build failed because of the recurring intermittent `Capybara::Poltergeist::JavascriptError` [problem](https://github.com/alphagov/smart-answers/issues/1624#issuecomment-116733309). I've kicked the build.

chrisroos:
Looks good to me.

PR #1759, opened 2015-06-29
Convert report-a-lost-or-stolen-passport outcomes to use ERB templates

floehopper:
The main thing of interest in this PR is that it's the first one where we've need to use a data partial. For the moment we've implemented this by simply using a standard call to `render partial: 'foo'`, but prefixing the template name with a relative path to the shared data partial directory. See the commit entitled "Convert contact_the_embassy outcome" for further details.


chrisroos:
Looks good to me.

PR #1764, opened 2015-06-30
Marriage Abroad ticket 15. Remove Channel Islands (CI) and Isle of Man (IOM) exceptions

tadast:
Remove the `residency_uk?` question that asks where in the UK does the user live. Also remove `outcome_os_iom_ci ` outcome and related phrases, update links for CNI third country outcomes and update a paragraph about CNI validity to include IOM and CI

This has been coordinated with FCO. We decided that regular, non Isle of Man/Channel Island outcomes are more informative and useful.


floehopper:
I haven't looked into it in detail, but with the removal of one of the questions, I would've expected some of the integration tests to have become essentially duplicate and therefore redundant, but I didn't notice any of them being removed. If this is correct, I'd probably remove them in a separate commit for clarity.

floehopper:
Other than my question above, this LGTM :+1:.

tadast:
@floehopper you're right, one of them had slipped through the cracks https://github.com/alphagov/smart-answers/commit/b6df0d8f95f635f3529fabea43a0cebc15335e28 :) That code path was not very well tested...

PR #1765, opened 2015-06-30
Add a custom outcome for Hong Kong

tadast:
As users in Macao are now directed to Hong Kong, we decided to update the
HK outcome too. Even though local consular facilities do not process
british marriages, it can be done via Immigration Department in HK.


floehopper:
LGTM :+1:

I hope you're still feeling in control of all these marriage-abroad branches!

PR #1766, opened 2015-06-30
Convert register-a-death outcomes to use ERB templates

floehopper:
The most significant thing to note in this PR is that I chose to replicate an [existing bug][1] rather than fix it. This was because the bug affected both this flow and the `register-a-birth` flow and the fix was non-obvious. Other things worth noting:

* I introduced a couple of `precalculate` blocks to make local variables in the flow available to the template.
* I added calls to `String#html_safe` for strings created within the flow in order to avoid HTML-escaping. I would've preferred to have moved these strings into the template, but doing that would've been awkward, because they're used in quite a few places.
* I used `case` statements to replace dynamically constructed `PhraseList` keys.
* Removed `assert_phrase_list` statements from the integration test, but as per usual I didn't consider whether in doing this any of the tests have become superfluous/duplicates.
* This conversion has revealed a bunch of unused phrases in the local YAML file. I plan to remove them as part of a separate PR, because I'm concerned they might contain important content which ought to be displayed.

[1]: https://www.pivotaltracker.com/story/show/98101814


chrisroos:
@floehopper and I just discussed the indenting of ERB tags in the templates.

We've agreed to avoid indenting the outer level of conditionals within the `content_for` blocks. Other conditionals within the top-level should be indented.

That means that we use:

```
<% content_for :body do %>
<% if foo %>
Foo text
  <% if bar %>
Bar text
  <% end %>
<% end %>
<% end %>
```

Instead of:

```
<% content_for :body do %>
  <% if foo %>
Foo text
    <% if bar %>
Bar text
    <% end %>
  <% end %>
<% end %>
```

tadast:
Should we add this to README then?

chrisroos:
Aside from the indenting issue mentioned above, this all looks good to me.

chrisroos:
> Should we add this to README then?

Probably :-) I'll add a card to Trello to prompt us to update it.

PR #1767, opened 2015-06-30
Remove unused phrases from register-a-death

floehopper:
These phrase keys don't appear to be used anywhere and the regression tests
all pass without them, so I think it's safe to remove them. However, I thought
I'd put this in a pull request to get a second opinion.

I've updated the checksums to indicate that I'm happy that the removal of these
phrases have had no effect on the behaviour of the system.


tadast:
I couldn't find any usages of those phrases either :+1:

Out of curiosity, did you follow a similar process [as here](https://github.com/alphagov/smart-answers/commit/661abef5036079cf44bcbeb6fcacd12b3f8ff902) or did something different?

floehopper:
No, I wasn't quite as rigorous. I relied on (a) not being able to find either whole or part matches for the keys; and (b) the fact that the regression tests still pass after removing them. However, I think that gives me enough confidence to merge this.

PR #1768, opened 2015-07-01
Remove unreachable code paths

tadast:
Same sex marriages in New Zealand lead to outcome_cp_no_cni. Remove
all code branches and references that are New Zealand-specific in
the outcome_cp_commonwealth_countries outcome that was used for NZ
before the change.

It was uncovered using this command:
```
        $ rm -rf coverage && \
          TEST_COVERAGE=true \
          rails r script/generate-responses-and-expected-results-for-smart-answer.rb \
          marriage-abroad
```

### DIFF
--- a/lib/smart_answer_flows/marriage-abroad.rb
+++ b/lib/smart_answer_flows/marriage-abroad.rb
@@ -270,7 +270,7 @@ module SmartAnswer
           })
 
           next_node_if(:outcome_cp_commonwealth_countries, -> {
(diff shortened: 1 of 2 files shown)


floehopper:
Was this the result of a change you've made recently or just something that's been around a while? Particularly if it was the former, then it would be good to reference the commit or PR in which the change was made.

Has this commit made any of the phrase list keys redundant?

Otherwise this LGTM.

PR #1772, opened 2015-07-01
Convert register-a-birth outcomes to use ERB templates

floehopper:
This follows the pattern of similar previous pull requests, particularly the one for register-a-death.

I've followed the [indentation rules](https://github.com/alphagov/smart-answers/pull/1766#issuecomment-117579072) for the ERB template which we agreed earlier today.


floehopper:
The build failure was due to the usual [Capybara::Poltergeist::JavascriptError](https://github.com/alphagov/smart-answers/issues/1624). I've kicked the build.

floehopper:
This has been kicking around for nearly a week. I'd be really grateful if someone could review it so I can get it merged. Thanks.

lib/smart_answer_flows/register-a-birth/embassy_result.govspeak.erb
```diff
@@ -0,0 +1,265 @@
+<% content_for :body do %>
+$!You can apply to register the birth with the <%= embassy_high_commission_or_consulate %> in <%= registration_country_name_lowercase_prefix %>.$!
+
+<% if %w(usa).include?(country_of_birth) %>
```
tadast: Minor: the contents are USA-specific, so it's not likely to have another country use this paragraph, therefore it would be nicer to use `country_of_birth == 'usa'`

floehopper: Yes. I've seen this pattern in a lot of places. I think I'd prefer to keep it like this in this PR to make it more obvious where the condition code as come from.

tadast:
LGTM :shipit:

PR #1773, opened 2015-07-01
Fix a minimum wage calculation bug

tadast:
Looks like 658b9f0 has introduced a bug
in minimum wage caluclators that stopped uplifting the amount underpaid
by an employer to an employee. The commit message does not clarify what
the intention of the change was. Based on feedback from BIS it has introduced
incorrect behaviour.

Correct behaviour is explained here: http://www.hmrc.gov.uk/manuals/nmwmanual/NMWM13160.htm#IDANFFGH
Amount due = total of underpaid / Original NMW rate [* Current NMW rate]
```
Example taken from HMRC’s 2011/12 case:
(Total Underpaid) 43.20 / (Original rate) 6.08 * (Current rate) 6.08 = £43.20

In this example, the 2011 rate has been used as the current rate….rather than the 2015 rate.

It should be: 43.20 / 6.08 * 6.50 = £46.19
```

The change in the commit `(underpayment / minimum_hourly_rate * per_hour_minimum_wage(@date)).round(2)`
meant that `minimum_hourly_rate` (which is the rate at the date provided by the user) and
`per_hour_minimum_wage(@date)` (where @date is the user provided date) meant both methods
returned the same date, therefore underpayment was always divided by 1, thus not uplifting
the owed amount.

In addition to reverting the changes in aforementioned commit, this adds a new test that is using the example provided by the department as a way to replicate the bug.


floehopper:
This looks ok to me as far as I can see, but I haven't been able to access the Zendesk ticket. It would be good it any important info from there ended up in commit messages or PR descriptions.

tadast:
@floehopper thanks for your comments. I've updated the commit message and the PR description. Let me know if it makes more sense now

floehopper:
LGTM :+1:

floehopper:
@tadast: I think it was the changes in this PR which caused the regressions CI build to start failing - https://ci-new.alphagov.co.uk/job/govuk_smart_answers_regressions/10/.

It looks like you updated the HTML artefacts for `am-i-getting-minimum-wage` but not the ones for `minimum-wage-calculator-employers` which share the same logic.

floehopper:
@tadast: I think this is fixed in 47618f57493f2b62a2509bcaf20b6d6bf81bd403.

PR #1774, opened 2015-07-02
Fix Kyrgyzstan payment callout

tadast:
This is addressing internal FC.

Users in Kyrgyzstan need to use institutions in Kazakhstan, therefore
the payments are done there. Make sure they match.

I am not generating HTML as this is one of quite a few changes I need to address today and time is pressing. I'll check HTML changes once all of the feedback from internal FC is addressed.


floehopper:
LGTM :+1:

tadast:
https://github.com/alphagov/smart-answers/commit/f64a9488f702ab02ffd97322c4914e5b372d54e7

PR #1781, opened 2015-07-03
Bump GDS API adapters to 20.1.1

elliotcm:
This changes the user agent to include the app
name in API requests.

https://trello.com/c/qkHdtob4/247-bump-gds-api-adapters-everywhere-it-s-used-to-include-user-agent-information


floehopper:
While I'd hope the automated tests would've caught any problems that this might have introduced, it's worth noting that this PR upgrades `gds-api-adapters` from v10.16.0 to v20.1.1.

Assuming `gds-api-adapters` is using semantic versioning (and perhaps it doesn't), this corresponds to a jump of 10 in major version number. It seems surprising to me that a change of that magnitude has required no changes to the Smart Answers application code.

As well as including the user agent change described in the commit note, this also includes a bunch of other changes. See the `gds-api-adapters` [CHANGELOG](https://github.com/alphagov/gds-api-adapters/blob/master/CHANGELOG.md).

elliotcm:
Hi @floehopper 

A fair concern.  The reason 10 major bumps hasn't broken the code is that the `gds-api-adapters` gem contains quite a few adapters and test helpers, and breaking changes to endpoints not used by smartanswers cause a major version bump.

As well as the automated test I did a review of the API usages in the app and as best as I could tell nothing had changed.  The endpoints used by smartanswers are not the ones under heavy (or even moderate) development.

floehopper:
@elliotcm: That's good to know. Thank you!

PR #1784, opened 2015-07-06
Apply tier 4: remove downtime message

tadast:
The downtime is in the past now so the users do not need to see the temporary downtime message introduced in 6ca437a3e052379fe24219cc0233eb33c9219c19 anymore.

Closes https://github.com/alphagov/smart-answers/pull/1782


floehopper:
LGTM :+1:

PR #1785, opened 2015-07-06
Add instruction for merging content team pull requests to README

floehopper:


tadast:
I think those instructions apply to more than the Content team members e.g. it could be a GOV.UK developer who does not have write access to this repo, a dev from a different department and so on. 

Having said that, I think that rephrasing it to cover a more generic case would make it less useful to a new developer skimming through the README as it would be less likely to catch their eye.

I don't feel strongly about this. What do you think?


Another important reason why it needs to be checked out locally and re-submitted as a PR is that Ci builds don't currently run against 'external' PRs.

chrisroos:
What's the status of this? Can we merge it?

floehopper:
@chrisroos: I'm looking at it right now.

floehopper:
@tadast: I addressed your second point by changing point 10 in the [instructions](https://github.com/alphagov/smart-answers#instructions) to suggest submitting a new PR. I decided not to make the instructions more generic for the reasons you mentioned.

PR #1789, opened 2015-07-07
Convert overseas-passports to use ERB templates for outcomes

floehopper:
This was pretty gnarly mainly due to the complexity of the logic constructing phase lists e.g. deeply nested conditionals, conditionals with many branches, and dynamically constructed phrase keys (sometimes dynamic in multiple dimensions).

I chose to only introduce partial templates when a phrase was actually used in more than one place. It might've been better to introduce partials templates for structural reasons (i.e. to split the template up in to manageable chunks) and to give a one-to-one mapping between phrases and partial templates.

The latter might've meant it was easier to relate the changes in this PR with the original logic in the outcome precalculate blocks.

I'm not 100% convinced that the result is "better" than how it was before, however it should be noted that this was a very naive translation of the outcome logic and its phrases. A lot more refactoring will be required to make this more maintainable, but I _think_ this is still the right direction.

In the longer term, it might be worth considering slicing things differently by having an outcome/partial template for each country and then reducing duplication using partial templates as appropriate.


tadast:
Doesn't look too bad in comparison to the previous version. Let's see what it feels like to make changes to it!

LGTM :shipit:

test/integration/smart_answer_flows/overseas_passports_test.rb
```diff
@@ -289,16 +244,12 @@ class OverseasPassportsTest < ActiveSupport::TestCase
       add_response 'renewing_new'
       add_response 'adult'
       assert_current_node :ips_application_result_online
-      assert_phrase_list :how_long_it_takes, [:how_long_6_weeks, :how_long_additional_info_renewing_new, :how_long_additional_time_online]
-      assert_phrase_list :cost, [:passport_courier_costs_ips1, :adult_passport_costs_ips1]
-      assert_phrase_list :how_to_apply, [:how_to_apply_online, :how_to_apply_online_prerequisites_renewing, :how_to_apply_online_guidance_doc_group_3, :how_to_apply_online_guidance_doc_outro]
     end
     should "use the document group of the country of birth - Spain (which is 1)" do
```
tadast: I've belatedly noticed that this test description has become incorrect (and perhaps the whole test became irrelevant) as we are no longer testing the content in this test. There may be more cases of this.

floehopper: Yes. Initially when doing these conversions I tried to change the tests to keep them making sense, but having chatted to @chrisroos about it we decided that since we imagine these tests going away that it probably wasn't worth the time it would take. So since then I've been doing the bare minimum to keep the tests passing which mainly means removing `assert_phrase_list` statements.

PR #1790, opened 2015-07-07
Remove redundant phrases from overseas-passports

floehopper:
* I came across these while converting the outcomes to use ERB templates - these
phrases were left over at the end.

* I did a text search in the project for the keys and for likely looking part
keys and didn't find any of them.

* All the regression tests pass when these are removed in combination with the
above, this means that I'm confident that the phrases are no longer needed.


tadast:
LGTM :+1:

PR #1792, opened 2015-07-09
Update tenant-nationality

lutgendorff:
From factcheck:
- removing Channel Islands route (will be caught by later questions)
- changing outcome for non-eea family members to a later outcome (because they won't have the other documents)


tadast:
https://github.com/alphagov/smart-answers/pull/1795

PR #1794, opened 2015-07-09
[Spike - DO NOT MERGE] Generate text (Govspeak) output from regression tests

floehopper:
This is something we've talked about before, but I decided to do a quick spike after we ran into the problem of having to update **all** the output HTML [1] because of a change to the Google Analytics code for the second time.

These changes mean that you can add `.txt` to the end of any outcome page URL and see a Govspeak version of the page. Currently I've added Kramdown comments to delimit the `title`, `body` & `next_steps` sections, but these are probably overkill.

We think that having the outputs stored in text form will make any diffs easier to understand.

One notable downside of this approach is that we would no longer have coverage of the previous question section of the HTML, but I think I can live with that. It might even make sense to be able to put the regression tests into HTML mode temporarily in order to make a specific change.

I had hoped that the change might make the regression tests run quicker, but they took about 50 mins on my machine which isn't markedly quicker than before.

What do people think?

[1]: 5b3dc0c4085cc8b4dfc2e2741bc0d43837779d2e


tadast:
From my perspective (someone who makes changes to flows v.s. converts formats without any external impact) I think govspeak is a lot easier to read and compare than HTML and it's a big plus.

A possible downside would be a possibility of missing important formatting issues e.g. whitespaces at the end of the line could mean a linebreak in HTML. However, if there is a diff detected a developer would check the output in the browser anyway.

All in all I give it a big fat :+1:

chrisroos:
I like it!

I'm not sure about the Kramdown comments. They look a bit noisy when browsing the outcomes, but maybe they'd come in useful if you were trying to work out why/how a change had been introduced?

If we did allow the regression tests to be run in both text and html modes, then I wonder whether it'd be worth running the full suite in html mode on Jenkins.

floehopper:
@tadast, @chrisroos: Thanks for the feedback. I think I'll get rid of the Kramdown comments. Do either of you think it's a problem to expose the `.txt` versions in production? Obviously it won't be possible to navigate to them, but it will be possible to "hack" the URLs to access them.

tadast:
I think Content team would benefit from seeing govspeak as they use this
format to represent changes in tickets e.g. text before/text after (all
govspeak). I don't know how they currently do the conversion, my guess
would be manually.

chrisroos:
> Do either of you think it's a problem to expose the .txt versions in production?

I think it might be better to avoid exposing them in production. I can imagine it being confusing to people that accidentally stumble on the text representation without realising that they exist.

I do actually wonder whether they'd even show up in production without us doing something with panopticon/the router. On the off chance they don't show up in production then I'd rather they were explicitly blocked in the code.

floehopper:
@tadast: When you say the Content Team would benefit from seeing Govspeak, do you mean seeing it in production? Given @chrisroos's point, I think I'm only going to enable it in development & test environments for now.

tadast:
> When you say the Content Team would benefit from seeing Govspeak, do you mean seeing it in production?

Yes

<hr/>

I'm starting to think it would be more accurate to use a new mime type
for those things e.g. `text/govspeak; charset=UTF-8`. What these endpoints return is not plain text, it's a markup language, very much like HTML, XML or markdown, so it would make sense to use a govspeak-specific mime type.

This way you'd have to explicitly add .govspeak to the URL or pass it in as a request header, thus reducing the chance of accidentally stumbling upon this feature. This addresses @chrisroos' point below.

>  I can imagine it being confusing to people that accidentally stumble on the text representation without realising that they exist.

Even if we don't go with my suggestion above, I think it's ok to make it accessible in production as users can already accidentally stumble upon json representation of outcomes.

We can ask around the content peeps if they'd find it useful and if not, then we can disable it in production.

floehopper:
@tadast: Thanks for the feedback. I did consider adding a new MIME type, but I think I'd feel uncomfortable introducing it without discussion across GOV.UK / GDS. I feel as if making the govspeak text available in production is a separate issue and a bit of a distraction from the motivation behind this PR, so I'm going to make the change just for the regression tests for now. We can talk about the production side of things separately.

floehopper:
Closing in favour of #1815.

PR #1793, opened 2015-07-09
Update can rent 12 months: fixed typos

lutgendorff:

PR #1795, opened 2015-07-09
Landlord immigration check FC feedback

tadast:
Closes #1792 and #1793.

In addition to the above, this PR removes a no longer used `named_person_of_eea_switzerland_person` node and adds a test for the modified flow.

Note: this smartanswer is in draft state, so it's safe to merge to master.

### Original PR description

From factcheck:

- removing Channel Islands route (will be caught by later questions)
- changing outcome for non-eea family members to a later outcome (because they won't have the other documents)
- fixing a typo


floehopper:
Other than my suggestion about splitting the first commit, this LGTM.

PR #1796, opened 2015-07-09
Convert minimum-wage-calculator-employers to use ERB templates for outcomes

floehopper:
This was all pretty straightforward, particularly because @chrisroos had already converted the other minimum wage flow, `am-i-getting-minimum-wage`, which uses the same shared flow logic.

The only thing of any real note was that I've chosen to introduce a `shared` directory to share a partial template between the two minimum wage flows.


chrisroos:
Looks good to me.

PR #1797, opened 2015-07-09
Overseas passports: all users may need to attend an interview

tadast:
All users may need to attend an interview

This is an older request that has slipped through the cracks for some countries.

Original request from zendesk:
- any applicant type can be invited for interview, not just first time - amend sentence and move it further down in the section.

The 'You may need to attend interview' has been moved to its own paragraph in all
cases as per content designer advice.


chrisroos:
Looks good to me.

PR #1756, opened 2015-06-29
Convert calculate-state-pension to ERB

chrisroos:
The amount_result.govspeak.erb template is pretty lengthy and can almost certainly be DRYed up. I did try extracting the four main branches to partials but ran into problems as I needed to pass `locals` in to each of the partials and it wasn't obvious to me that the result was going to be much better than what I have here.

At the moment I'm planning to leave the template as it is unless anyone has any insights as to how to effectively DRY it up.

PR #1801, opened 2015-07-09
Convert calculate-state-pension to ERB

chrisroos:
This is an updated version of #1756 now that Ben's changes to calculate-state-pension have been merged to master.

The amount_result.govspeak.erb template is pretty lengthy and can almost certainly be DRYed up. I did try extracting the four main branches to partials but ran into problems as I needed to pass `locals` in to each of the partials and it wasn't obvious to me that the result was going to be much better than what I have here.

At the moment I'm planning to leave the template as it is unless anyone has any insights as to how to effectively DRY it up.


floehopper:
The build failure looks legitimate - because the checksums aren't up-to-date.

floehopper:
Other than my small comments and the build failure, this looks good to me :+1:

chrisroos:
I made amendments in my local branch, rebased on master and manually merged in 63746e319e7ae87e4dc8325451fa770c2ae73244.

PR #1802, opened 2015-07-09
Convert inherits-someone-dies-without-will to use ERB templates for outcomes

floehopper:
This conversion was pretty straightforward. The only thing of note was that a `PhraseList` was constructed in the first question and conditionally modified in another question (which not everyone had to answer). I dealt with this by constructing an equivalent array of symbols matching the phrase list keys and then use this in ERB tags within the `next_step_links` partial template.


chrisroos:
Apart from my comment about the '_next_step_links.govspeak.erb' partial, this all looks good to me.

Issue #1626, opened 2015-05-11
Failing tests caused by duplicate context names


chrisroos:
We appear to have a number of failing tests that aren't currently running because their names clash with other tests in the repo. The problem is caused by multiple contexts with the same name, containing tests with the same names.

I've created a script that attempts to give all contexts unique names to reveal these failing tests - https://gist.github.com/chrisroos/e7df2b0ded95a0098f1a

I've not yet attempted to fix any of these failures.

@tadast has previously fixed some similar failing tests in 4dfff6356ee30f6a7907a1998874622fdbf017ef.

tadast:
Your version that runs all tests: 3548 runs, 18021 assertions

Currently on master: 3525 tests, 17897 assertions

chrisroos:
I've just run the [rename-shoulda-contexts.rb] script in my local copy of Smart Answers, ran `rake` and didn't see any test failures. I can only imagine we've gradually fixed these problems while making other changes in the code/tests.

Closing.

PR #1803, opened 2015-07-09
Fix hidden and failing tests

chrisroos:
I was bitten by the duplicate context/test issue (documented in #1626) while converting energy-grants-calculator to use ERB templates: My conversion caused tests to become invalid, but they didn't fail because of the duplicate name problem.

I used the code from the gist referenced in that issue to give each context a unique name and found these two failing tests.

There are still more hidden tests but I figured these failing ones were a priority.

### DIFF
--- a/test/integration/smart_answer_flows/am_i_getting_minimum_wage_test.rb
+++ b/test/integration/smart_answer_flows/am_i_getting_minimum_wage_test.rb
@@ -283,8 +283,8 @@ class AmIGettingMinimumWageTest < ActiveSupport::TestCase
         end
         should "make outcome calculations" do
           assert_state_variable "total_hours", 45
-          assert_state_variable "minimum_hourly_rate", 6.08
-          assert_state_variable "total_hourly_rate", 6.07
+          assert_state_variable "minimum_hourly_rate", 6.5
+          assert_state_variable "total_hourly_rate", "6.12"
           assert_state_variable "above_minimum_wage", false
         end
       end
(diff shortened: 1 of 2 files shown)


floehopper:
LGTM :+1:

chrisroos:
Manually merged in 743e14311e0cc0369c099b2704e9e6ed7183e655.

test/integration/smart_answer_flows/am_i_getting_minimum_wage_test.rb
```diff
@@ -283,8 +283,8 @@ class AmIGettingMinimumWageTest < ActiveSupport::TestCase
         end
         should "make outcome calculations" do
           assert_state_variable "total_hours", 45
-          assert_state_variable "minimum_hourly_rate", 6.08
-          assert_state_variable "total_hourly_rate", 6.07
+          assert_state_variable "minimum_hourly_rate", 6.5
+          assert_state_variable "total_hourly_rate", "6.12"
```
tadast: It's not a fault of this commit, but I've noticed that this test doesn't seem to stub out current time. This means once we change the rates next year this test will pass at first, but when the new fiscal year starts, it will start failing. :watch: :bomb:

chrisroos: Yeah. There are some similar assertions in this file that have been commented out because of the data being date dependent. The comment suggests that the commented out assertions are now in the calculator test. If we think that's a useful approach then I can imagine doing a similar thing with these values.

tadast: I can see that `total_hourly_rate` calculation is [covered](https://github.com/alphagov/smart-answers/blob/2620dc0ab2f4d477fc0f9d01f1e02bc63c57ceed/test/unit/calculators/minimum_wage_calculator_test.rb#L85) in unit tests. Then the only thing that it's testing here is that those values are exposed as state variables, which is indirectly covered by regression tests I believe. Therefore I think it should be OK get rid of this test.

PR #1804, opened 2015-07-10
Log more info about partials rendering

tadast:
With the new smartanswers format the `?debug=1` is not useful in showing where
the content is coming from anymore. The log shows which partials are rendered
but it does not show what code calls it. In some cases the same partial can
be rendered from 10 or more places (e.g. overseas-passports). This adds a line
to the log that shows what code has invoked partial rendering.

Before:
```
Rendered lib/smart_answer_flows/overseas-passports/_passport_costs_ips.govspeak.erb (0.3ms)
```
After:
```
Rendered lib/smart_answer_flows/overseas-passports/_passport_costs_ips.govspeak.erb (0.3ms) from /smart_answer_flows/overseas-passports/ips_application_result.govspeak.erb:120
```

This feature is only invoked when `LOG_PARTIAL_SOURCE` environment variable is set

This is up for discussion, alternative solution suggestions are very welcome

### DIFF
--- a/config/initializers/log_subscriber.rb
+++ b/config/initializers/log_subscriber.rb
@@ -1,0 +1,14 @@
+if ENV['LOG_PARTIAL_SOURCE']
+  module ActionView
+    class LogSubscriber
+      def render_partial(event)
+        info do
+          message = " Rendered #{from_rails_root(event.payload[:identifier])}"
+          message << " within #{from_rails_root(event.payload[:layout])}" if event.payload[:layout]
+          message << " (#{event.duration.round(1)}ms)"
+          message << " from #{caller.find { |line| line =~ /\.erb/}.to_s.split(':in').first.split('/lib').last}"
+        end
+      end
+    end
+  end
+end


chrisroos:
1. Can you explain why it can be necessary to see where a partial is being rendered from? I'm not disputing that it is useful, I just want to understand the use case.

2. Can you explain how this works? Are you monkey patching `ActionView::LogSubscriber#render_partial`, or is Rails doing something magic by looking for `*Subscriber` classes?

tadast:
> Can you explain why it can be necessary to see where a partial is being rendered from? I'm not disputing that it is useful, I just want to understand the use case.

Sorry, I should have explained this in the commit note.

I needed certain countries to have outcomes with a certain text, say text that I found in `send_application_address.govspeak.erb` by matching content in it with the one that was requested. If you look at https://github.com/alphagov/smart-answers/blob/2620dc0ab2f4d477fc0f9d01f1e02bc63c57ceed/lib/smart_answer_flows/overseas-passports/ips_application_result.govspeak.erb this partial can be rendered from 5 different places. Ideally I would understand all conditional branches in that file and deduct the right place to make the change, however it is nearly impossible as the logic is intermingled with content in a rather large file. Therefore I fall back to rendering an outcome for another country outcome which was given to me as a point of reference as an example of an outcome to follow. Looking at the log output I can see which partials are rendered, therefore I can trace back the conditional branches that are being executed, therefore understanding the logic to follow and finding the right place to make the change. I know, it doesn't sound great :) Happy to explain in person if it doesn't make sense.

> Can you explain how this works? 

I am monkeypatching `ActionView::LogSubscriber#render_partial` which is not great. I will try to find a better solution if we all decide this is the right tool for the problem in the short term :)

chrisroos:
@floehopper, @tadast and I discussed this in person last week. We agreed that we're happy for the additional logging to be added but that we'd prefer to see it added in such a way that didn't involve monkey patching the existing logger.

tadast:
I'm closing this for now with an intention to revisit once I need something like this again.

PR #1805, opened 2015-07-10
Overseas passports process changes in Burma

tadast:
## Burma changes - moves from embassy to Visa Application Centre

Summary of changes:

1. Under "Cost" - text change for where passport sent to, the Visa Application Centre.

2. Under "Making your application" (renewing_new) changes: - new address and email address

3. Under "Making your application" (applying, renewing_old, replacing) changes: - new address and email address

4. Under "Getting your passport" (renewing_new) - text change for where get passport from.

5. Under "Getting your passport" (applying, renewing_old, replacing) - text change for where get passport from.


chrisroos:
It's not immediately obvious how the changes in these two commits have resulted in all the changes to the rendered outcome HTML. That makes it quite hard to see the effect of these changes in isolation.

If some of the other changes come from another branch then it's probably worth updating the rendered outcomes there and then rebasing this branch on that.

tadast:
@chrisroos I believe this is confusing because the text changes only slightly, but this is the result of adding Burma to a few country lists that then causes different paragraphs to be rendered (via partials or via executing different conditional branches). This is also a rather big pain when making changes as there seem to be a lot of duplicated or nearly duplicated text e.g. try running a global search for "Include your first name and last name and 3 alternative dates and times." and see text around it :)

PR #1806, opened 2015-07-10
Overseas passports. Application centre address change in China

tadast:
The address of the application centre in Shanghai has changed.


chrisroos:
Looks good to me.

PR #1807, opened 2015-07-10
Overseas passports application centre address change in Nepal

tadast:
The application centre address and opening hours in Nepal have changed.


chrisroos:
Looks good to me.

PR #1808, opened 2015-07-10
Overseas passports process changes in Vietnam

tadast:
The process in Vietnam has changed from apply online to apply in person. 

We've been advised to mimic the process in Laos.

```
Change - from online to apply in person at a UK VAC.

- Please use the same text, and logic as for Laos: https://www.gov.uk/overseas-passports/y/laos

- Include 2 UK VAC addresses for Vietnam - see new addresses in "Address Details" column. 
```


chrisroos:
The changes look OK to me, but I don't really feel as though I'm in a position to be able to check the changes in the rendered outcomes.

PR #1809, opened 2015-07-13
Live has been removed use on instead

edds:
Live is removed in the next version of jQuery, use on instead which has
the same behaviour.

http://jquery.com/upgrade-guide/1.9/#live-removed

### DIFF
--- a/app/assets/javascripts/smart-answers.js
+++ b/app/assets/javascripts/smart-answers.js
@@ -15,7 +15,7 @@ $(document).ready(function() {
 
     // events
     // get new questions on submit
-    $(formSelector).live('submit', function(event) {
+    $('#content').on('submit', formSelector, function(event) {
       $('input[type=submit]', this).attr('disabled', 'disabled');
       var form = $(this);
       var postData = form.serializeArray();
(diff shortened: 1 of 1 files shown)


tadast:
Have you tested that this works as intended? 

It looks correct to me, but according to the docs you've linked to the preferred syntax is ` $(document).on('submit', formSelector, function(event){...}`. I believe this syntax also works in case `formSelector` content is added via ajax/pjax.

edds:
@tadast you are indeed correct, in my head I thought I was changing `bind` to `on`. Have updated it to properly delegate rather than bind the events.

tadast:
:cool: Looks good to me now :+1:

chrisroos:
@edds: I don't suppose you'd be able to add some JS tests around this functionality? I realise this would be a pain because there aren't any similar tests at the moment.

edds:
@chrisroos without rewriting all the JavaScript, no. There is no way to tie into the JavaScript as it is all just run on load rather than following our styleguide of being encapsulated methods which can then be unit tested.

chrisroos:
@edds: Fair enough!

PR #1810, opened 2015-07-13
Simplify `SmartAnswer::FlowRegistry`

floehopper:
This PR converts all the remaining (test fixture) flows into classes and simplifies the `SmartAnswer::FlowRegistry` code since we no longer need to do the old-style `eval`-ing of flow files.


chrisroos:
Looks good to me!

tadast:
:confetti_ball:

PR #1811, opened 2015-07-13
Convert legalisation-document-checker to use ERB templates for outcomes

floehopper:
The only thing of note is that I've ended up with a `case` statement inside a loop, which isn't ideal. When I tried extracting each `when` statement into a partial, I had to pass in a couple of locals which made me decide it wasn't worth doing.


chrisroos:
Looks good to me.

PR #1813, opened 2015-07-13
Convert estimate-self-assessment-penalties to use ERB templates for outcomes

floehopper:
There's some slightly silly logic for formatting zero in different ways in the commit titled "Convert late outcome" which I've naively reproduced exactly. However, I question whether zero should be being formatted in these three different ways. Otherwise everything else was all pretty standard.


chrisroos:
Looks good to me.

floehopper:
Manually merged in 485c0bc36514f87c778908d891f8fa5b385d9258.

PR #1814, opened 2015-07-14
Overseas passports process changes

tadast:
This is a combination of already reviewed PRs that:

- change application waiting time in some countries
- removes Dual National notification from the start page
- changes wording for application waiting time in some countries
- changes application centre addresses in Nepal and China
- changes application processes in Vietnam and Burma

For more detailed changes please read the commit messages.

If everything is alright here I will revert [this](https://github.com/alphagov/smart-answers/blob/05fcaf50f3a0fd8d37806add3d2cce0930fbba7a/test/data/overseas-passports-questions-and-responses.yml#L26-L30) before merging.


chrisroos:
Looks good to me.

I think we spoke about this yesterday. Did you say that you'd get rid of the interim html artefact generation commits, and do it all in one go at the end?

tadast:
Thanks for reviewing!

> I think we spoke about this yesterday. Did you say that you'd get rid of the interim html artefact generation commits, and do it all in one go at the end?

Yes, I'll do it once we get a :+1: from Fact Check.

tadast:
This has been merged manually in ee2172e

I have removed all commits related to HTML generation, as master has changed to generating govspeak since (and regenerating govspeak for each change would have been a pain). I have also removed those four countries that have changed from the regression test inputs, as the regression tests where only useful when making changes for them.

floehopper:
Does the removal of the four countries still leave us with sufficient test coverage?

tadast:
@floehopper none of these changes have introduced any new code paths so we're as good as we were before. 

Actually, perhaps I should add lib/smart_answer/overseas_passports_helper.rb as a file dependency...

floehopper:
:+1:

PR #1815, opened 2015-07-14
Generate govspeak text output in regression tests instead of HTML

floehopper:
This is a non-spike version of #1794 which I will now close in favour of this one.

The changes in this PR mean that you can add `.txt` to the end of any outcome page URL in development/test and see a Govspeak version of the page. The title, body & next_steps are output in that order with a blank line between them.

We think that having the outputs stored in text form will make any diffs smaller and easier to understand.

One notable downside of this approach is that we would no longer have coverage of the previous question section of the HTML, but we've decided that we can live with that.

I've updated all the artefacts as part of the PR.


floehopper:
Manually merged in ac05978c64d2b57d64d8af361d110e95ce352bba.

PR #1818, opened 2015-07-15
Birth certificates are not required for marriage in Belgium

tadast:
Received as a part of the Fact Check feedback.


chrisroos:
Looks good to me.

PR #1820, opened 2015-07-16
Marriage abroad FC feedback. Italy

tadast:
Summary:

- Italian embassy in the UK does not charge for anything: remove all sentences claiming otherwise for uk residence outcomes
- Show only fees relevant to Italy in Italian residency outcomes
- For non-british partners only their ID document is required in Italy
- "No need to stay after posting notice" goes up by one paragraph
- Rework statutory declaration requirements in Italy as they differ for British and foreign partners
- Rename a few phrases to better reflect their purpose


chrisroos:
Looks good to me.

PR #1777, opened 2015-07-02
[In Fact Check - DO NOT MERGE] Updates to calculate-statutory-sick-pay (re-worked)

floehopper:
This is a re-worked version of #1634. Assuming all is well with this PR, I plan to close #1634 in favour of this one. Ideally we'd also deploy this branch (which has been rebased against a recent version of `master`) to Heroku for Fact Checking.

* c93b15096c9f628d1c83b6ffadf2ff325bf61ea8 was applied directly to `master` in 55d84d95f1e7321491f9038727c7f00483674e66
* f935adac60e9e1f0abd4f712cef7344652705e65 was applied directly to `master` in 30415653b3c06bb9eec51c39df1a1e63dbf2a6f4
* 3f805b803b992756e0999e414fc42717177b00bd was applied directly to `master` in 257a36f756cea354e9816188f291a9d5d913fc1a
* I also extracted a couple of other small unrelated changes into commits on `master`:
  * 341b5688b25019e60162c5faf06419d92ef0ef03
  * 7d474468e278bb7c8c34d66519ad681de8b8ffc9
* A rebased version of 757c4663462d3aedb30de7f5e6467c7dd2743870 (i.e. "Move LEL check to last question") has been included as the first commit in this PR.
* The remaining 5 commits have been squashed into the second commit in this PR.

This resulted in a branch with more atomic commits with all the tests passing after each commit. This in turn made it easier to rebase the branch against `master` despite it being very out-of-date and the flow code having been moved into a class in the meantime.

The test URL appears to work ok on my local machine with these changes in place.

The original PR description (by @BenJanecke) is included below for completeness:

I believe this is the related [Pivotal Tracker story](https://www.pivotaltracker.com/story/show/93691714).

We now ask for previous periods of sickness earlier so we can use the correct dates when asking for and calculating AWE

Test url for deploy 

http://smartanswers.dev.gov.uk/calculate-statutory-sick-pay/y/none/yes/no/2014-08-10/2014-08-20/yes/2014-05-01/2014-08-20/eight_weeks_more/monthly/2013-04-01/2013-02-03/4000.0/1,2,3,4,5

PR #1821, opened 2015-07-16
Add regression tests for calculate-statutory-sick-pay

floehopper:
Note that this branch is based on `statutory-sick-pay-updates-with-squashed-commits-each-with-passing-tests` and shouldn't be merged until #1777 is merged.


chrisroos:
#1777 has now been merged so I believe we should be able to rebase and merge these changes too.

One thing I've just noticed is that it looks like @tadast added another 4 commits to #1777 after this PR was opened. I guess that might mean that the regression tests added in this branch have to be updated.

chrisroos:
I've rebased this branch on master and updated the title to reflect the fact that it can now be merged.

There were two conflicts during the rebase. One in the commit with subject "TEMP: Use next_node with block for all questions", and then again in the commit that reverts that change. The conflicts were easy enough to resolve.

To satisfy myself that the regression tests were still valid, I checked out the commit with subject "Generate expected results for calculate-statutory-sick-pay" and re-generated the responses and expected results with the `TEST_COVERAGE=true` environment variable set. I was able to see that the uncovered lines of code matched those documented in the note for commit with subject "Update responses for calculate-statutory-sick-pay".

I'm going to get this merged to master.

floehopper:
@chrisroos: Thanks.

PR #1823, opened 2015-07-16
Marriage abroad FC feedback. Thailand

tadast:
Clarify affirmation/affidavit requirements in Thailand

Either of affirmation/affidavit document is accepted, so avoid confusion and make sure both are mentioned in the relevant sentences.

This is addressing FC feedback

### DIFF
--- a/lib/smart_answer_flows/locales/en/marriage-abroad.yml
+++ b/lib/smart_answer_flows/locales/en/marriage-abroad.yml
@@ -178,6 +178,10 @@ en-GB:
 # "what you need to do" phrases
         what_you_need_to_do: |
           ##What you need to do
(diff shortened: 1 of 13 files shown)


tadast:
To my knowledge China is the only other country that accepts both, but it's handled by China-specific phrases.

chrisroos:
Presumably the text is different enough that we need both special cases (for China and Taiwan)?

tadast:
Yes, the Chinese outcome has custom links to affirmation and affidavit documents https://www.gov.uk/marriage-abroad/y/china/ceremony_country/partner_british/opposite_sex

chrisroos:
Fair enough :-)

PR #1825, opened 2015-07-17
Update to ruby 2.2.2

tadast:
~~Ruby 2.1.5 has a security vulnerability [1]. It is not clear whether it can
be exploited in smartanswers, but perhaps it's easier to upgrade the ruby
version than to investigate the ways to exploit the app.~~ 

A new PR to address the security vulnerability https://github.com/alphagov/smart-answers/pull/1846

I also had to switch to psych for YAML parsing as syck does not seem to be compatible with Ruby 2.2.2 [at the moment](https://github.com/tenderlove/syck/issues/9). It's been deprecated for a long time anyway.

Minor syntax changes were required to make smartanswers work with psych,
namely, interpolation can only happen if the string value is surrounded by
quotes. I've ran all regression tests after making the change and the outcomes
have not changed so I'm confident no more changes are required.

[1]- https://www.ruby-lang.org/en/news/2015/04/13/ruby-openssl-hostname-matching-vulnerability/


Gemfile.lock
```diff
@@ -322,10 +321,12 @@ DEPENDENCIES
...
-  syck (= 1.0.5)
   therubyracer (~> 0.12.1)
   tilt (= 1.4.1)
   timecop
   uglifier
   unicorn (= 4.8.3)
   webmock (= 1.20.4)
+
+BUNDLED WITH
+   1.10.5
```
floehopper: Is this actually needed as pat of these changes? Or is it just because you're using a recent version of Bundler locally?

floehopper:
Is there a reason you chose to upgrade to Ruby 2.2.2 rather than 2.1.6? I'd have been inclined to upgrade to 2.1.6 to resolve the potential vulnerability and then separately upgrade to 2.2.x.

Gemfile.lock
tadast: I just installed ruby 2.2.2 locally and got the newest bundler version with it. It seems that bundler introduced this feature as they're planning to introduce breaking changes in version 2. This number is only increasing, never decreasing, so there shouldn't be too many issues. 

I think we have a few options here:

- roll with whatever the bundler team thinks is the best
- patch local bundler versions to avoid adding those lines
- remove it from this commit and open a separate PR to discuss what we want to do
- use pre 1.10 version of bundler locally

tadast: To add, this does not break any currently officially supported bundler versions https://github.com/bundler/bundler/issues/3697#issuecomment-116246436

floehopper: I've heard of people having problems with different projects having different values of `BUNDLED WITH`, but I may be worrying about nothing.

When you say you "got the newest bundler version with it", Bundler isn't part of Ruby is it? i.e. presumably you got the latest version because you did `gem install bundler` in your freshly installed Ruby version. If so, this still feels like a separate change to me. You *could* prescribe the same version of Bundler as you were using in the earlier install of Ruby when you install the gem.

I prefer to keep changes like this as small as possible, because then if there's a problem it's easier to to work out where it's been introduced.

tadast: I'm using rbenv with rbenv-default-gems plugin that installs me bundler and some other gems automatically when new ruby versions are installed.

Ok, I'll downgrade and remove this change :)

tadast:
> Is there a reason you chose to upgrade to Ruby 2.2.2 rather than 2.1.6?

I was upgrading anyway so why not go with the latest and greatest? :) Based on the commits notes using Syck was just technical debt, so it's good removing it too.

floehopper:
I don't have an issue with upgrading to 2.2.2, but I think you're doing that for a different reason and I prefer to make changes like this as small as possible. Could we not have one PR to address the vulnerability which upgrades to 2.1.6 and another which is effectively the upgrade from 2.1 to 2.2?

tadast:
> Could we not have one PR to address the vulnerability which upgrades to 2.1.6 and another which is effectively the upgrade from 2.1 to 2.2?

What's the benefit of doing that after the fact?

chrisroos:
I agree with @floehopper. I'd personally upgrade to 2.1.6 separately to 2.2.2.

Upgrading to 2.1.6 should protect us from any potential vulnerabilities found in the older version and I'd hope the upgrade itself would be uncontroversial and very easy.

The upgrade to 2.2.2 is then a further improvement that's unrelated to any known security vulnerabilities. It also requires other changes in the codebase that I can imagine have the potential to cause us other problems. Should we discover such problems then it'll be useful to be able to rollback to a known good commit that uses 2.1.6.

chrisroos:
Is changing the value in '.ruby-version' good enough to start using the new Ruby in CI, preview, staging and production, or do we have to change some puppet/chef recipes somewhere?

tadast:
Ok, I'll open a new PR with 2.1.6. I'll also investigate in more depth if there are any other required changes for syck -> psych migration. I was told ruby 2.2.2 is already installed on all servers.

Thanks for all comments :)

chrisroos:
> I was told ruby 2.2.2 is already installed on all servers.

Do you know if we're using rbenv on all the servers too? Presumably we'll have to be for this change to have the desired effect.

tadast:
Yes we use rbenv and both 2.2.2 and 2.1.6 are currently in use :)

tadast:
I've opened a PR for 2.1.6 https://github.com/alphagov/smart-answers/pull/1846
and have manually checked some smartanswers that interpolate in question titles and bodies, all appears to look unchanged.

floehopper:
LGTM. It would be good to explain that the motivation behind this upgrade is to use the latest & greatest version of Ruby and not to address the security vulnerability which was done in #1846.

The fact that all the regression tests pass is comforting, but I guess there might still be hidden issues with gems which aren't compatible with Ruby 2.2. Hopefully any serious ones would come to light during the staging deployment.

floehopper:
@tadast: Are you planning to merge this?

tadast:
My plan was to merge and deploy it after all important things that need to go live have been deployed. This is to reduce the risk of rolling back due to unexpected bugs in the new ruby version. 

I will try to deploy this tomorrow if there's anyone from the 2nd line in the office.

floehopper:
@tadast: Sounds good.

PR #1826, opened 2015-07-20
Fix an error where a template can not access 'lived_or_worked_abroad'

tadast:
This addresses an errbit report from production. `lived_or_worked_abroad`
was not exposed to erb templates but was referred to. Here's a sample affected path: `/calculate-state-pension/y/amount/male/1991-08-12/4`

I've added a new DoB input to the regression test input that replicates the error scenario. This does not improve code coverage according to simplecov with the setup we're using. I think this means that we can't rely on the coverage reports as much as we do, because some views are logic-heavy and they are not included in the coverage reports atm.


chrisroos:
Looks good to me.

PR #1827, opened 2015-07-20
Always expose cost_change_4_weeks to views

tadast:
This is addressing an error from errbit.
Sample affected path: /childcare-costs-for-tax-credits/y/yes/yes/monthly_diff_amount/2400.0/127.0

cost_change_4_weeks was only set if it's true, otherwise erb views could not
see it as a defined variable. There was also no input set for regression tests
that covers a case where `ten_or_more` is true, therefore this code branch
in erb was never exected.

The issue is fixed by explicitly exposing cost_change_4_weeks for the outcome
that may need it and setting it to false unless it's already set.

Add a regression test input to reproduce the issue and some extra assertions
to highlight the importance of cost_change_4_weeks.

This is very similar to https://github.com/alphagov/smart-answers/pull/1826

### DIFF
--- a/lib/smart_answer_flows/childcare-costs-for-tax-credits.rb
+++ b/lib/smart_answer_flows/childcare-costs-for-tax-credits.rb
@@ -242,6 +242,10 @@ module SmartAnswer
           weekly_difference_abs >= 10
         end
 
+        precalculate :cost_change_4_weeks do
+          cost_change_4_weeks || false
+        end
+
         precalculate :title_change_text do
           weekly_difference >= 10 ? "increased" : "decreased"
         end
--- a/test/artefacts/childcare-costs-for-tax-credits/yes/yes/monthly_diff_amount/2400.0/10.txt
+++ b/test/artefacts/childcare-costs-for-tax-credits/yes/yes/monthly_diff_amount/2400.0/10.txt
@@ -1,0 +1,4 @@
+Your average weekly childcare costs have increased by £37.
+
+Tell the Tax Credit Office so they can adjust your tax credit.
+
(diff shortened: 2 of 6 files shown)


chrisroos:
Do you need to optionally set `cost_change_4_weeks` to false? I think you should just be able to add the `precalculate` block and return `cost_change_4_weeks`. If it's not already set then it'll return nil which should be good enough to satisfy the view.

chrisroos:
Aside from my comment, this looks good to me.

tadast:
> Do you need to optionally set cost_change_4_weeks to false? 

I wanted to make it more explicit for two reasons. 

1) To avoid confusion. In current master you can access any 'variable' on a `State` because it inherits from `OpenStruct`. However when we pass values to the view we only consider variables that are explicitly set. This makes it a bit confusing.

2) Reliable testing. Explicitly setting the variable to false allows for a more reliable testing, see [here](https://github.com/alphagov/smart-answers/blob/tt_fix_childcare-costs-for-tax-credits/test/integration/smart_answer_flows/childcare_costs_for_tax_credits_test.rb#L251). If I were to assert for nil the tests would pass on master, which is not what I want.

chrisroos:
> 1) To avoid confusion. In current master you can access any 'variable' on a State because it inherits from OpenStruct. However when we pass values to the view we only consider variables that are explicitly set. This makes it a bit confusing.

I agree that this is confusing. The reason I asked about the `|| false` part of the `precalculate` block is that we've added these sort of `precalculate` blocks elsewhere but in every other case we're simply returning the same variable (i.e. we're not conditionally setting it to something else).

> 2) Reliable testing. Explicitly setting the variable to false allows for a more reliable testing, see here. If I were to assert for nil the tests would pass on master, which is not what I want.

That makes sense. I'd like to think we'll get to a point where testing the state variables directly isn't necessary but we're not there yet so this seems like a good approach for now.

PR #1830, opened 2015-07-20
Marriage abroad FC feedback. Macao 

tadast:
At least one person must be a resident in Macao to marry there.

The requirement is listed in the required documents section already,
but we have been asked to make it clearer. At least one person from a
couple must live in Macao to have an opposite sex marriage there.

We do not show this sentence for users who answer that they live
in Macao.


chrisroos:
Looks good to me.

PR #1832, opened 2015-07-21
Remove reference to scanned documents pre-verification

tadast:
ORU no longer provide a service where they check scanned document copies for users before they send the originals via post. 

We were not told why this service is being discontinued.


chrisroos:
Looks good to me.

PR #1834, opened 2015-07-21
Births and deaths fees table changes

tadast:
There are a couple changes to the document returns fees tables in births and deaths. I've extracted a shared partial with the fees before making the changes, as those fees are identical. As a side effect the register-a-death fees tables started using `<th>` element for headers, this makes it more consistent with the rest of the tables.

Summary: 
1) Clarification of what qualifies as UK and Europe regions
2) Fee changes that will kick in on 1st of August 2015

I've used the mechanism we use for calculators to store the rate values that change on a certain date. I guess this could work for all the fee tables if we thing this would help reduce the cost to making changes.

Also worth noting that this will break the regression tests on 1st August as the rates in the views will change. I don't think there is an elegant solution to this so I'll just resolve it manually when it happens.

### DIFF
--- a/lib/data/rates/births_and_deaths_document_return_fees.yml
+++ b/lib/data/rates/births_and_deaths_document_return_fees.yml
@@ -1,0 +1,13 @@
+---
+- start_date: 2014-01-01 # This is an arbitrary date, it's not know when's the last time the rates changed
+  end_date: 2015-07-31
+  post_to_uk: £4.50
+  post_to_europe: £12.50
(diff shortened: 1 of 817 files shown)


chrisroos:
This all looks great to me.

We currently freeze time to 1st Jan 2015 when running the regression tests, so I don't think we have to worry about them failing after the 1st Aug 2015.

tadast:
Ok, I didn't know about that! Thanks

PR #1836, opened 2015-07-22
Register a birth. DNA test changes

tadast:
- Clarify that users are to pay any costs related to DNA testing
- Add Somalia as the second country (Libya is the first) that may require DNA testing


chrisroos:
Looks good to me.

It's great to see the switch from `if country_of_birth == 'libya'` to the much more descriptive `if reg_data_query.may_require_dna_tests?(country_of_birth)`.

PR #1837, opened 2015-07-22
Update contact info for advice for unmarried couples.

tadast:
Affects all scenarios where the child was born before 1 July 2006 and father is unmarried,
for example: /register-a-birth/y/austria/father/no/2006-06-30

The tool used to say: 'Contact your local British embassy...' This has changed.
Users can contact the Overseas Registration Unit for advice at
birthregistrationenquiries@fco.gov.uk


chrisroos:
Looks good to me.

PR #1838, opened 2015-07-22
Add pitcairn-island, st-martin and saint-barthelemy to ORU outcomes

tadast:
Pitcairn Island had slipped through the cracks when transitioning all countries.

St Martin and Saint Barthelemy were added to the country drop down and were
leading to the default embassy outcome. 

More changes to follow in another PR that will change the default path to the ORU outcome and clean up no longer used code paths.


tadast:
If I get :+1: perhaps I should remove `pitcairn-island` from the regression test responses and regenerate artefacts as it is no longer an exceptional case. Should shave off some time

chrisroos:
Looks good to me.

> perhaps I should remove pitcairn-island from the regression test responses and regenerate artefacts as it is no longer an exceptional case

Presumably we can remove it without losing any coverage? If that's the case then it sounds good to me.

tadast:
> Presumably we can remove it without losing any coverage?
This is correct 7fdbd5f3a48f2c14e41fbc5a2cc2ef70c7a90054 :)

PR #1849, opened 2015-07-24
Births and deaths post-ORU-transition cleanup

tadast:
Reopening for merging into another branch as the old PR https://github.com/alphagov/smart-answers/pull/1844 is outdated now

This is an attempt to clean up residues of the 5-phase transition to handle most of the registrations via Overseas Registration unit (ORU). Now this has become a default way to handle registrations, but this fact was not well reflected in the code. I believe the series of commits in this PR make the fact clearer and make the code easier to follow.

This is a refactoring, therefore I do not expect there to be any externally visible impact to birth and death registration smartanswers.


chrisroos:
Looks good to me.

I didn't completely follow the changes in "Default to the ORU outcome for births registration" or "Default to the ORU outcome for deaths registration abroad" but I'm confident that the regression tests would've told us if something was wrong.

PR #1851, opened 2015-07-27
Update links to student finance forms

tadast:
We've been asked to update the links to student forms as the stakeholder (slc.co.uk) is changing the CMS that is serving the forms.

This has been fact-checked and needs to go live on 28th.


chrisroos:
Looks good to me.

Issue #1828, opened 2015-07-20
[BUG] Handling 29th of Feb in calculate-your-holiday-entitlement


tadast:
path to reproduce: /calculate-your-holiday-entitlement/y/days-worked-per-week/leaving/2015-05-01/2016-02-29/1

It fails with an unhelpful error message `undefined local variable or method `holiday_entitlement_days'`. This is because the block calculating the value raises an error, but it disappears (we should look at that separately).

The real issue is handling 29th Feb, which only exists in leap years. [This method](https://github.com/alphagov/smart-answers/blob/master/lib/smart_answer/calculators/holiday_entitlement.rb#L135-L137) fails
 as it's not taking into consideration leap year dates (funnily enough that's the main reason it was written).

I'm leaving it here for now as it's an edge case and I have other time-sensitive things at the moment.

floehopper:
It happened again today: https://errbit.publishing.service.gov.uk/apps/533c35ae0da1159384044f5f/problems/560ef3cc65786306396e1000

Since it keeps happening, it would be good to fix this soon.

I'd hope that using the `YearRange` class might make it easier to fix it.

chrisroos:
This has happened again today - https://errbit.publishing.service.gov.uk/apps/533c35ae0da1159384044f5f/problems/5632289c657863063bd71d00

It's worth noting that the error is now "ArgumentError: Invalid date" now that we're no longer rescuing `ArgumentError`s raised in the `next_node` blocks (see PR #2032 for more).

chrisroos:
I think @erik-eide has fixed this problem in PR #2041.

erik-eide:
Closing this, seems to have been resolved with #2041 with the failing [URL](https://www.gov.uk/calculate-your-holiday-entitlement/y/days-worked-per-week/leaving/2015-05-01/2016-02-29/1) now ok

PR #1855, opened 2015-07-29
Holiday Entitlement calculator cleanup

tadast:
I started this cleanup with hope to address https://github.com/alphagov/smart-answers/issues/1828, but it's taking me too long to understand how the code is supposed to behave in all those edge-cases so I'll try to attack it another time. 

For now it's just a cleanup that does not alter any behaviour


chrisroos:
Looks good to me.

PR #1856, opened 2015-07-29
Refactor minimum-wage Smart Answers

chrisroos:
I've made a number of changes to the minimum-wage shared logic that's used by both am-i-getting-minimum-wage and minimum-wage-calculator-employers.

There are quite a few commits but they're all fairly small and I hope the commit message explains the motivation.

In broad terms, these are the changes I've made:

* Remove unused code:

  * From unused `precalculate` blocks.
  * From unused `save_input_as` values.
  * From unused custom validation where the validation is already provided by the `money_question`.
  * From parsing that's already handled by questions that use (e.g.) the `parse: Float` option.

* Move formatting to templates, e.g. use `OutcomeHelper#format_money`.

* Use calculator values in templates and remove no-longer necessary calculate blocks.

* Use `validate` blocks instead of custom validation in calculate blocks. Add tests where they didn't exist previously.

* Set calculator attributes in `next_node` blocks rather than `calculate` blocks whose return values aren't used.

* Consistently use `next_node` blocks rather than the shortcuts available to `multiple_choice` questions.

* Instantiate the calculator early on in the flow and then add each response to it as they come in, rather than instantiating the object in one go later on in the flow. I had to add accessors to the calculator to make this possible. Setting attributes directly on the calculator means that I don't have to pollute the state object in order to share variables within the flow.

* Move validation logic out of the flow and into the calculator.

* Change some tests that were asserting against state variables so that they now assert against values on the calculator. I started out deleting the tests that were asserting against state variables that no longer existed but then realised that it's possible I'd be losing some useful tests. These integration tests that now assert against the calculator should move to the calculator tests if they're deemed useful.

__NOTE__. All but the first two questions appear twice, once for determining minimum wage for the current period and once for a previous period.

Having now gone through the process I can see a slightly better ordering of commits but trying to reorder the commits in this branch would be a bit painful and probably not worth doing.


chrisroos:
I've added a few more commits.

One first commit moves the unit tests from test/integration/smart_answer_flows to test/unit/smart_answer_flows.

The next three commits move policy logic from the flow to the Calculator.

The final two commits update the checksum data for am-i-getting-minimum-wage and minimum-wage-calculator-employers.

chrisroos:
> "elligible" -> "eligible"

Thanks @floehopper. Corrected in fixup commit 98f7d55c87ad9491fca95788002a615adc8eda91.

floehopper:
I might be wrong, but I think the `save_input_as :current_or_past_payments` statement in `what_would_you_like_to_check?` is now redundant.

floehopper:
Do you know whether the `calculate :accommodation_charge` blocks are still needed in `:current_accommodation_charge?` & `:past_accommodation_charge?`? Also do you know why we need to convert the money object into a `Float` in these cases?

floehopper:
It might be nice to extract `response.to_i == 0` from `:how_many_hours_overtime_do_you_work?` & `:how_many_hours_overtime_did_you_work?` into a method on the calculator e.g. `#any_overtime_worked?`. Also I'm pretty sure the `to_i` is redundant, because `0.0 == 0 # => true`.

floehopper:
This is great! A heroic effort! :smile: 

It makes me a lot more confident that this approach is the right one - the flow is now so much easier to follow and the code & behaviour is so much more consistent.

All my comments are pretty nit-picky except for the one about the validation range being accidentally changed in 676b611bb330e95e1f060d77cc99229e8884d7ed.

:+1:

chrisroos:
> I might be wrong, but I think the save_input_as :current_or_past_payments statement in what_would_you_like_to_check? is now redundant.

I think you're right. I'll get rid of it.

chrisroos:
> Do you know whether the `calculate :accommodation_charge` blocks are still needed in `:current_accommodation_charge?` & `:past_accommodation_charge?`? Also do you know why we need to convert the money object into a Float in these cases?

The `accommodation_charge` is passed to the `#accommodation_adjustment` method in the `:current/past_accommodation_usage?` questions.

Having just looked at the implementation of `#accommodation_adjustment` I can see that it converts the parameter to a `Float` to I think I can probably get away with using `save_input_as` instead of the `calculate :accommodation_charge` blocks.

chrisroos:
> It might be nice to extract response.to_i == 0 from :how_many_hours_overtime_do_you_work? & :how_many_hours_overtime_did_you_work? into a method on the calculator e.g. #any_overtime_worked?. Also I'm pretty sure the to_i is redundant, because 0.0 == 0 # => true.

Yup, that sounds good. I must've missed this when extracting other things to the calculator.

chrisroos:
I've made a number of changes suggested by @floehopper, rebased on master and force pushed to this branch. I'll double check that the tests pass, including regression tests, and then manually merge to master.

chrisroos:
Manually merged in af48943f3dfeff6789e48fc623e25d0af459a7a3.

floehopper:
Yay! :smile: :star: :star: :star:

PR #1857, opened 2015-07-29
Update parser Gem

chrisroos:
This avoids the following warning that I've been seeing since we upgraded to
Ruby 2.1.6:

    warning: parser/current is loading parser/ruby21, which recognizes
    warning: 2.1.5-compliant syntax, but you are running 2.1.6.

I updated the Gem using:

    $ bundle update parser


floehopper:
LGTM :+1:

tadast:
10 x :+1:

PR #1858, opened 2015-07-29
Overseas passports cleanup

tadast:
This morning I was showing some erb style smartanswers to content designers and have noticed that there are several phrases that are repeatedly used in different overseas passports cases.

This is a series of commits that reduce duplication. See each commit note.

Test artefacts are a bit noisy, but I'm ok with this noise as it's helps to have better looking erb templates/partials without odd newlines.


chrisroos:
Looks good to me. I like seeing lots of red in the diffs :-)

PR #1859, opened 2015-07-29
[DO NOT MERGE] Sketch of new calculate-your-part-year-profit smart answer

floehopper:
This is the first iteration of this new smart answer - just dealing with the "happy path" for now.

I've opened this PR to garner feedback on the general approach. The approach has a lot in common with the work @chrisroos has been doing in #1856.

It's just one giant commit at the moment, because I've been playing around with things quite a bit, but I plan to split the changes up into multiple commits shortly.

Some things worth noting:

* I'm using dependency injection to supply a dummy calculator object to the flow in the tests. It's not great that I have to assign the instance variable to a local variable in order to make it in scope for the `next_node` blocks, but it does have the advantage that the calculator is only added to the state in the outcome.

* I'm using an instance of `OpenStruct` as the dummy calculator which isn't great, but it's [awkward](https://github.com/freerange/mocha/issues/44#issuecomment-4368689) to stub an attribute setter method in the setup and then expect it in one of the tests, but not the others.

* I'm unit testing the flow class rather than using the usual integration test or regression test. I've added a "view test" at the end of this unit test to check that various values are rendered in the outcome. This might be better in its own file/sub-directory.

* The unit test for the flow class doesn't check that the question pages are rendered successfully, but I'm not sure the integration/regression tests do this at the moment, do they?

* I've had to add a couple of `precalculate` blocks in the `:what_is_your_taxable_profit?` question, because the title in the YAML needs to be interpolated with these values. This is possibly a small motivation to move to using ERB for question pages.

* I've implemented `DateRange` and `TaxYear` classes in an attempt to encapsulate common concepts in the domain of this smart answer (including more complex logic that will be needed in the more exceptional paths). Hopefully they also make the code more readable.

* Also I haven't added any validation yet.

Please let me know asap if you can see problems with this approach. Also let me know if you think there are other tests that I should add.

### DIFF
--- a/lib/smart_answer/calculators/part_year_profit_calculator.rb
+++ b/lib/smart_answer/calculators/part_year_profit_calculator.rb
@@ -1,0 +1,25 @@
+module SmartAnswer
+  module Calculators
+    class PartYearProfitCalculator
+      include ActiveModel::Model
+
+      attr_accessor :tax_credits_awarded_on, :accounts_start_on, :profit_for_current_period
+
+      def basis_period
+        DateRange.new(begins_on: accounts_start_on, duration: 1.year)
+      end
+
+      def award_period
+        DateRange.new(begins_on: TaxYear.current.begins_on, ends_on: tax_credits_awarded_on)
+      end
+
+      def profit_per_day
+        (profit_for_current_period / basis_period.number_of_days).floor(2)
+      end
+
+      def part_year_profit
+        (profit_per_day * award_period.number_of_days).floor
+      end
+    end
+  end
+end
--- a/lib/smart_answer/date_range.rb
+++ b/lib/smart_answer/date_range.rb
@@ -1,0 +1,38 @@
+module SmartAnswer
+  class DateRange
+    class ::Date::Infinity
+      def to_date
+        self
+      end
+    end
+
+    LATEST_DATE = Date::Infinity.new
+
+    attr_reader :begins_on, :ends_on
+
+    def initialize(begins_on:, ends_on: LATEST_DATE, duration: nil)
+      @begins_on = begins_on.to_date
+      @ends_on = duration.present? ? (begins_on + duration - 1) : ends_on.to_date
+    end
+
+    def include?(date)
+      (date >= @begins_on) && (date <= @ends_on)
+    end
+
+    def number_of_days
+      (@ends_on == LATEST_DATE) ? Float::INFINITY : (@ends_on - @begins_on).to_i + 1
+    end
+
+    def ==(other)
+      eql?(other)
+    end
+
+    def eql?(other)
+      (self.class == other.class) && ([begins_on, ends_on] == [other.begins_on, other.ends_on])
+    end
+
+    def hash
+      self.class.hash ^ [begins_on, ends_on].hash
+    end
+  end
+end
--- a/config/application.rb
+++ b/config/application.rb
@@ -2,6 +2,7 @@ require File.expand_path('../boot', __FILE__)
 
 # Don't include all of rails, we don't need activerecord
 require "action_controller/railtie"
+require "active_model/railtie"
 require "action_mailer/railtie"
 require "sprockets/railtie"
 require "rails/test_unit/railtie"
--- a/lib/smart_answer/flow.rb
+++ b/lib/smart_answer/flow.rb
@@ -5,8 +5,8 @@ module SmartAnswer
     attr_reader :nodes, :outcomes
     attr_accessor :state, :status, :need_id
 
-    def self.build
-      new.tap do |flow|
+    def self.build(*args)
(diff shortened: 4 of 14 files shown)


tadast:
> The unit test for the flow class doesn't check that the question pages are rendered successfully, but I'm not sure the integration/regression tests do this at the moment, do they?

They don't

lib/smart_answer/calculators/part_year_profit_calculator.rb
```diff
@@ -0,0 +1,29 @@
...
+      def basis_period
+        DateRange.new(begins_on: accounts_start_on, duration: 1.year)
+      end
+
+      def award_period
+        DateRange.new(begins_on: TaxYear.current.begins_on, ends_on: tax_credits_awarded_on)
+      end
+
+      def profit_per_day
+        (profit_for_current_period / basis_period.number_of_days).floor(2)
```
tadast: Are we sure that `profit_for_current_period` is never going to be a Fixnum?

lib/smart_answer/date_range.rb
```diff
@@ -0,0 +1,32 @@
...
+
+    def ==(other)
+      eql?(other)
+    end
+
+    def eql?(other)
+      (self.class == other.class) && ([begins_on, ends_on] == [other.begins_on, other.ends_on])
+    end
+
+    def hash
```
tadast: What's the practical purpose of this method? I found [this](http://ruby-doc.org/core-2.2.2/Object.html#method-i-hash) but was curious how it's helpful in this particular case.

tadast:
`DateRange` and `TaxYear` classes do make tests easier to read :+1:

View tests are cool too, might not be as thorough as regression tests, but a lot less brittle.

I think generally it's looking good :+1:

lib/smart_answer/calculators/part_year_profit_calculator.rb
floehopper: It think it should always be an instance of `Money`, because it's the response from a `money_question`, but you're right that the `Calculator` doesn't enforce this. I'll have a ponder.

chrisroos:
> I'm using dependency injection to supply a dummy calculator object to the flow in the tests. It's not great that I have to assign the instance variable to a local variable in order to make it in scope for the next_node blocks, but it does have the advantage that the calculator is only added to the state in the outcome.

@floehopper and I just discussed this approach, particularly versus setting the calculator in a `next_node_calculation` block in the first question. James mentioned running into problems trying to test the use of the `next_node_calculation` and that the dependency injection route feels better as it moves us closer to a more Ruby implementation.

We talked about this working with multiple outcomes and decided that duplicating the `precalculate` block in each outcome isn't too bad for now. If it becomes a problem then we can remove that duplication in other ways.

lib/smart_answer/date_range.rb
floehopper: That method probably isn't being used at the moment, but since I've overridden `#eql?`, I think I'm duty-bound to override `#hash` to make it consistent. I want `DateRange` to be a proper immutable [ValueObject](http://c2.com/cgi/wiki?ValueObject) and this feels like the right thing to do to avoid surprises later. The implementation of these equality-related methods was taken from the [Values gem](https://github.com/tcrayford/Values).

chrisroos:
I think this is looking great :-)

lib/smart_answer/calculators/part_year_profit_calculator.rb
floehopper: I've added a [Trello card](https://trello.com/c/8AXxpSvp/42-enforce-type-of-attributes-on-calculator-policy-objects) about this explaining a bit about how I think we should approach this problem. I'm not planning to address it at the moment.

tadast: That's cool. What about doing something like this in the meantime?

```ruby
def initializer...
  ...
  @profit_for_current_period = Money.new(args[:profit_for_current_period])
end
```

floehopper: If we were to do that, it feels as if we should also do it for the other (`Date`) attributes. Also we should do it for the `attr_writer` methods. I prefer to leave this as is it for now and rely on (a) the calculator unit test to pass in the correct types of objects; and (b) the flow unit test to verify that passing in a string to the `money_question` results in an instance of `Money` being set on the calculator. I hope that makes sense.

tadast: It does, that's cool :+1:

floehopper:
I've force pushed my updated version of this branch. Sorry if this messes up the GH comments.

floehopper:
I'm closing this for now, because I've continued to spike on this locally and (more importantly) there is ongoing discussion with HMRC about how the logic should work.

PR #1860, opened 2015-07-30
Update _scheme_postcodes.govspeak.erb

lutgendorff:
Added more postcode areas for claiming PIP

Needs to be added as a separate group as they announce them advance and need to appear on the tool.

Deploy for Aug 3, no preview needed


tadast:
Closing in favour of https://github.com/alphagov/smart-answers/pull/1862

PR #1864, opened 2015-07-30
Marriage in Spain changes

tadast:
Marriage abroad tool outcomes in Spain are getting a brand new custom outcome.

Technically Spain is a CNI country, however I have decided to create a new outcome for Spain as the majority of the phrases are Spain-specific. It used to be the case before as well, thus making the CNI outcome a bit convoluted (the other two most prominent culprits are Italy and Germany).

This is currently out for fact check.


tadast:
This is now back from FC, could someone review the latest changes that address FC feedback (commit named "Remove references to sending fee in Spanish outcomes" and onwards)? The older commits have been reviewed in separate branches.

chrisroos:
Looks good to me.

PR #1866, opened 2015-07-30
Configure Mocha to prevent stubbing non-existent methods

floehopper:
I wanted to enable [this option](http://gofreerange.com/mocha/docs/Mocha/Configuration.html) in the tests I'm adding for a new smart answer, but it seems more sensible to enable it for *all* tests rather than just the ones I'm working on.

The commits in this PR fix a few tests that failed when I enabled the option. The final commit enables the option.

### DIFF
--- a/lib/smart_answer/calculators/holiday_entitlement.rb
+++ b/lib/smart_answer/calculators/holiday_entitlement.rb
@@ -85,16 +85,25 @@ module SmartAnswer::Calculators
       number.to_s.sub(/\.0+$/, '')
     end
 
-    def method_missing(*args)
+    def method_missing(symbol, *args)
       # formatted_foo calls format_number on foo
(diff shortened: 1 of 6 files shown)


lib/smart_answer/calculators/holiday_entitlement.rb
```diff
@@ -85,17 +85,26 @@ def strip_zeros(number)
...
-      if args.first.to_s =~ /\Aformatted_(.*)\z/
-        format_number(self.send($1.to_sym), args[1] || 1)
+      if formatting_method = formatting_method(symbol)
+        format_number(send(formatting_method), args.first || 1)
       else
         super
       end
     end
 
+    def respond_to?(symbol, include_all = false)
```
tadast: :100:

tadast:
This is great :+1:

PR #1867, opened 2015-07-30
fixed hash syntax to Ruby 1.9

askl56:


floehopper:
@askl56: Thanks for the PR. Unless I'm missing something, it looks as if your changes only address _some_ of the old-style hash syntax. If we're going to make a change like this, I'd prefer to change all of them. Also I'd prefer to introduce Rubocop (or similar) to enforce the use of the new syntax as part of the same set of changes.

Out of interest, do you have a specific motivation for making these changes? Are you using this code yourself?

askl56:
I've added a rudimentary Rubocop sweep which should have caught any outliers. No motivation, beyond trying to help provide consistency and help where I can.

chrisroos:
Hey @askl56. Thanks for trying to help out.

I can see that you've added a separate commit containing rubocop fixes. Unfortunately, this also contains fixes for a number of other Rubocop violations (e.g. it switches us to use `do ... end` instead of `{ ... }` for multiline blocks) instead of just changing the hash syntax. If we're to consider merging this PR then it'll need to be focussed on just changing the hash syntax.

In addition: The commit message should be written in the present tense as per our [Git Styleguide][] (i.e. "Fix rather than Fixed"), and I think the commit message should mention our [Ruby Styleguide][] as that's what tells us to "Use Ruby 1.9 syntax for symbolic hash keys".

[Git Styleguide]: https://github.com/alphagov/styleguides/blob/master/git.md#structure
[Ruby Styleguide]: https://github.com/alphagov/styleguides/blob/master/ruby.md

floehopper:
@askl56: Hi Andrew. Thanks again for the PR. @chrisroos recently setup the project to use [govuk-lint](https://github.com/alphagov/govuk-lint) in #1901 & made it run on our CI server in #1911. I've just used Rubocop's auto-correct functionality to change the style across the project - see 4cb96a4d66521af1ec0ff130e30c1b08476634db. I think that makes this PR redundant and I'm going to close it. Thanks again for your time and I hope it's some consolation that you've encouraged us to fix this!

PR #1868, opened 2015-08-03
Fix formatting of the Passport Adviceline contacts block

tadast:
Preserve linebreaks for the Passport Adviceline contacts. A `\\` adds an
explicit linebreak (`<br>`) in generated HTML.


floehopper:
LGTM :+1:

I'd prefer to fix all of these as described in [this govspeak issue](https://github.com/alphagov/govspeak/issues/55), but this is fine for now given that we've done the same thing elsewhere.

PR #1869, opened 2015-08-03
Update application email addresses

tadast:
Update overseas passports application email addresses and fix formatting - opening hours used to show on the same line as email addresses.


floehopper:
LGTM :1:

As per [comment elsewhere](https://github.com/alphagov/smart-answers/pull/1868#issuecomment-127196686) it would be good to fix the line breaks more globally in govspeak, but this is fine for now.

PR #1870, opened 2015-08-03
Add 6 week deadline for sending supporting documents

tadast:
The deadline used to apply to a certain list of countries, now it applies to all countries, except for the ones where you can't use the service. There are handled by non IPS outcomes.


floehopper:
Other than my small comment, this looks good to me, especially as it represents a simplification in the logic! :+1:

tadast:
Thanks for reviewing. Merged manually https://github.com/alphagov/smart-answers/commit/7c1ae6f1d2518cc6b64e1f9fd0a93df45d616130

PR #1873, opened 2015-08-03
Change the way Certificate of freesale can be legalised

tadast:
BIS have informed us that these certificates are no longer signed but must bear the “Signature Valid” text along with a “Tick” on the certificate.

Rules have changed for this document only, so I had to create a new "group" for it. Not sure it's the least confusing way of making this change.


tadast:
This has now been fact-checked and is ready for review.

floehopper:
LGTM :+1:

PR #1876, opened 2015-08-04
Preload entire app before forking Unicorn workers.

joshmyers:
smart-answers is CPU hungry, especially during app
startup. It's preloading the answers and some of
them take a long time to load. We have seen an
issue where application was slow to start.

Setting this to true reduces the start up time for
starting up the Unicorn worker processes. This
uses CoW to preload the application before forking
other worker processes. However, there is a big
gotcha. We must take special care that any sockets
(such as database connections) are properly closed
and reopened. We do this using before_fork and
after_fork.

### DIFF
--- a/config/unicorn.rb
+++ b/config/unicorn.rb
@@ -4,3 +4,6 @@ end
 load_file_if_exists(self, "/etc/govuk/unicorn.rb")
 working_directory File.dirname(File.dirname(__FILE__))
 worker_processes 4
+
+# Load application in master process before forking workers
+preload_app true


config/unicorn.rb
```diff
@@ -4,3 +4,21 @@ def load_file_if_exists(config, file)
...
 working_directory File.dirname(File.dirname(__FILE__))
 worker_processes 4
+
+# Load application in master process before forking workers
+preload_app true
+
+before_fork do |server, worker|
+  # The following is highly recomended for Rails + "preload_app true"
+  # as there's no need for the master process to hold a connection.
+  defined?(ActiveRecord::Base) and
```
tadast: SmartAnswers does not use any database, so we should be ok without these :)

floehopper:
@joshmyers: Am I right in thinking that this is a less than ideal solution? i.e. it would be better if we could reduce the start-up time of the app?

I have an idea about how to do that, but it's not going to be a trivial change and we're unlikely to get to it for a while. However, it would be good to know whether or not ideally we'd revert the changes in this commit if we do manage to improve the start-up time in the future.

Do you have a sense of what start-up time would be acceptable?

joshmyers:
@floehopper it's not great, but it shouldn't be bad either. We already do this in Whitehall and should not effect the overall startup time of the app negatively. The preload_app directive loads the application before forking workers, so it can share any loaded data structures. I don't think this will need to be reverted pending more work to improve the application startup time.

floehopper:
@joshmyers: Thanks for the explanation.

PR #1877, opened 2015-08-04
Simplify FlowTestHelper#setup_for_testing_flow method

floehopper:
This finishes off the conversion to directly instantiate flows in all integration tests rather than looking them up in the registry.


tadast:
:confetti_ball: :+1:

PR #1878, opened 2015-08-10
Fix a typo in appointment booking link for marriages in Romania

tadast:
We were provided with an incorrect booking link first. This fixes it.

@chrisroos unfortunately this is rather urgent so you'll need to rebase your conversion work once this is merged :|


floehopper:
Budapest <-> Bucharest. A pretty significant typo!

I assume there's no regression test coverage for this which is why the test artefacts haven't changed...?

tadast:
> I assume there's no regression test coverage for this which is why the test artefacts haven't changed...?

That's correct

floehopper:
LGTM :+1:

PR #1879, opened 2015-08-11
UK-issued CNI is now accepted by Greek town halls

tadast:
The Greek Ministry has now agreed for Greek Town Halls to accept the Domestic
CNI issued by the UK Superintendent Registrar for marriage in Greece without
it having to be exchanged by FCO.

There is a very high volume of British nationals who marry in Greece over the
summer. This is a significant change and users should be notified that the
extra financial cost of exchanging the CNI is no longer needed.

@chrisroos unfortunately this will cause you some merge conflicts :(


chrisroos:
Looks good to me.

PR #1880, opened 2015-08-11
Consistently use synonyms_of_cp_in_nnn phrase

chrisroos:
The naming of multi-word countries in the `synonyms_of_cp_in_NNN` phrases was
inconsistent. Some used hyphens and one ('south-africa') used an underscore. In
addition, three outcomes were doing `:"synonyms_of_cp_in_#{ceremony_country}"`,
while one was converting hyphens to underscores
(`"synonyms_of_cp_in_#{ceremony_country}".gsub('-', '_').to_sym`).

I've renamed `synonyms_of_cp_in_south_africa` to
`synonyms_of_cp_in_south-africa` in the YAML file and removed the special case
of replacing hyphens with underscores in the
`outcome_cp_commonwealth_countries` outcome.

### DIFF
--- a/lib/smart_answer_flows/marriage-abroad.rb
+++ b/lib/smart_answer_flows/marriage-abroad.rb
@@ -1348,7 +1348,7 @@ module SmartAnswer
 
         precalculate :commonwealth_countries_cp_outcome do
           phrases = PhraseList.new
-          phrases << "synonyms_of_cp_in_#{ceremony_country}".gsub('-', '_').to_sym
+          phrases << :"synonyms_of_cp_in_#{ceremony_country}"
 
           if resident_of == 'uk'
             phrases << :contact_high_comission_of_ceremony_country_in_uk_cp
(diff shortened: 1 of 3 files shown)


lib/smart_answer_flows/marriage-abroad.rb
```diff
@@ -1298,7 +1298,7 @@ def define
 
         precalculate :commonwealth_countries_cp_outcome do
           phrases = PhraseList.new
-          phrases << "synonyms_of_cp_in_#{ceremony_country}".gsub('-', '_').to_sym
+          phrases << :"synonyms_of_cp_in_#{ceremony_country}"
```
floehopper: Not a big deal, but I think `.to_sym` is clearer than the colon prefix.

floehopper:
LGTM :+1:

lib/smart_answer_flows/marriage-abroad.rb
chrisroos: I agree. I've gone with the colon prefix as that's what we're using elsewhere.

PR #1881, opened 2015-08-13
Further disambiguation of marriage in Spain process

tadast:
In addition to fb5ba0d more references to 'sending fees' need to be removed as
it may be confusing, and in addition to 3c0b933 'supporting' documents need to
be replaced with 'original' documents to make it clear what needs to be sent.

Unfortunately the mentioned commits are in master already, so I can't rebase.


chrisroos:
Looks good to me.

PR #1884, opened 2015-08-17
Convert student finance forms from Smartdown to the standard ruby flows format

tadast:
Add regression test artefacts for student-finance-forms, then convert it to the standard ruby/ERB flow format. I tried to make the commits granular and easy to follow (e.g. by adding commented out Smartdown logic and converting it to the standard format one block at a time.

I believe that due to a restrictive nature of Smartdown there are a lot of outcomes. There should be a way reduce the number of outcomes, but I'd like to the this 1:1 conversion merged first.


chrisroos:
In general this looks great! Thanks for taking this on.

My only concern is the use of `next_node_if`. I think we want to standardise on using `next_node {}` and to add '/visualise' support to it. I realise it would be a pain to edit these commits so I think it's fine to get this merged and then to start moving away from `next_node_if` as a separate exercise.

PR #1886, opened 2015-08-17
Add regression tests for landlord immigration check

floehopper:
I haven't done any investigation into test coverage.

@tadast: Did you do any when you converted `student-finance-forms`?


test/data/landlord-immigration-check-questions-and-responses.yml
```diff
@@ -0,0 +1,50 @@
...
+- 'no'
+has_other_documents:
+- 'yes'
+- 'no'
+immigration_application:
+- 'yes'
+- 'no'
+property:
+- 'WC2B 6SE'
+- 'B1 1PW'
```
tadast: I assume you took those values from `test/unit/smartdown_plugins/landlord-immigration-check_test.rb`, I think it would be useful to add a comment saying why those particular values were chosen as it is not obvious here.

tadast:
@floehopper I did not, because all questions in `student-finance-forms` are multiple choice, hence all values there are predefined. In this case there's the postcode value input, but looking at [the method](https://github.com/alphagov/smart-answers/blob/fa6977830b9f3853ff2dbf90dd69484f536a299e/lib/smartdown_plugins/landlord-immigration-check/render_time.rb#L21-L26) that uses this input it looks like you have covered both scenarios. (Unrelated to this PR: perhaps this method should end with a `?` as it returns a boolean value)

LGTM :+1:

test/data/landlord-immigration-check-questions-and-responses.yml
floehopper: Yes, that's correct. I did explain that in the [commit note](7f839d1138d3e8a753b0a638f05d6ccada216df9). I'd rather not add the explanation as a comment, because it's only really relevant for this change. In particular as soon as the scenarios go away the comment would no longer make sense.

floehopper:
@tadast: Thanks. Yes, that `valid_postcode` method isn't great. I don't really understand why it needs to deal with the HTTP status code and/or raw JSON when the Content API uses RestClient to handle that stuff elsewhere.

PR #1887, opened 2015-08-18
Fix same-sex marriage in Portugal title

tadast:
It was missing the word 'marriage'. I've confirmed this with a content designer


floehopper:
LGTM :+1:

PR #1890, opened 2015-08-18
Fix a bug in rendering student-finance-forms Yes/No questions

tadast:
When rendering Yes/No questions in student-finance-forms the 'Yes/No' values in
YAML were treated as boolean 'true/false' values, which is not intended. Add
quotes around those values to ensure YAML treats them as strings.

Add regression tests for the two questions affected by this bug.


floehopper:
LGTM :+1:

I think we could improve these flow view "unit" tests further by instantiating a `SmartAnswerPresenter` for the flow and calling the `presenter_for(node)` method on it. That way we can avoid having to e.g. construct the `i18n_prefix`. But I think this is fine for now.

PR #1895, opened 2015-08-19
Remove support for non ERB outcome templates

chrisroos:
__NOTE.__ This is based on #1894 so I'll want to make sure that's merged to master before this.

This removes the ability to render Outcomes using data stored in the YAML locale files.

This looks like a lot of commits but they're all really quite small; particularly the 35 commits that remove `use_outcome_templates` from each of the Smart Answer flows. I committed those separately so that I could commit the removal of that line and the change to the related checksum YAML file.

### DIFF
--- a/app/presenters/outcome_presenter.rb
+++ b/app/presenters/outcome_presenter.rb
@@ -14,8 +14,6 @@ class OutcomePresenter < NodePresenter
       render_erb_template
       title = @view.content_for(:title) || ''
       strip_leading_spaces(title.chomp)
-    else
-      translate!('title')
     end
   end
 
--- a/app/presenters/outcome_presenter.rb
+++ b/app/presenters/outcome_presenter.rb
@@ -34,8 +34,6 @@ class OutcomePresenter < NodePresenter
       govspeak = @view.content_for(:body) || ''
       govspeak = strip_leading_spaces(govspeak.to_str)
       html ? GovspeakPresenter.new(govspeak).html : govspeak
-    else
-      super
     end
   end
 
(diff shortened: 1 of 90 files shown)


app/presenters/outcome_presenter.rb
```diff
@@ -10,45 +10,28 @@ def initialize(i18n_prefix, node, state = nil, options = {})
...
-
   def body(html: true)
-    if use_template? && body_erb_template_exists?
+    if body_erb_template_exists?
       render_erb_template
       govspeak = @view.content_for(:body) || ''
       govspeak = strip_leading_spaces(govspeak.to_str)
       html ? GovspeakPresenter.new(govspeak).html : govspeak
-    else
-      super
```
tadast: Should we raise an exception if there's no body? I can't think of a scenario where an empty body would be an expected behaviour.

chrisroos: I don't think we can raise an exception as none of title, body or next steps are required to be present in the Outcomes. The [calculate-agricultural-holiday-entitlement done outcome][done-outcome] is an example of one that only has a title and no body.

[done-outcome]: https://github.com/alphagov/smart-answers/blob/master/lib/smart_answer_flows/calculate-agricultural-holiday-entitlement/done.govspeak.erb

floehopper: I think it's fine to not raise an exception for now, but I do question whether that use case is actually sensible.

tadast: :cool:

tadast:
Looks good to me :shipit:

chrisroos:
Awesome. Thanks chaps.

PR #1896, opened 2015-08-19
Upgrade govuk_frontend_toolkit

chrisroos:
I've upgraded the version of the govuk_frontend_toolkit so that we can use the `GOVUK.analytics.trackEvent` function instead of using the `ga` function directly.

A couple of things I'm unsure about:

* Should I have upgraded to the most recent version of govuk_frontend_toolkit? I've upgraded to the earliest version that adds the support I need.

* How should I test that the upgrade hasn't broken anything? I'd normally rely on test coverage but I'm not entirely convinced that that's good enough in this case.


chrisroos:
I'm closing this while I carry out some more thorough investigation into the changes in govuk_frontend_toolkit.

PR #1897, opened 2015-08-19
Don't use "if" at the start of the bullets for LIC break clause outcome

issyl0:
- "You'll have to carry out a right to rent check *if*" specifies that
  both of the bullets are "if"s.


floehopper:
LGTM :+1:

issyl0:
Content change approved by @keithiopia. :sunny:

PR #1898, opened 2015-08-20
Marriage abroad changes requested in July

tadast:
A batch of changes requested by FCO in July. 

This is a diverse batch of changes, but they are going to be fact-checked at once, hence a single PR. In summary:

- address the outcomes of countries that have inaccurate or slightly incorrect advice. Because of that posts in some of those countries have come up with their own guidance, which can be confusing to users.

- add a clarification that marriage registration in the UK is usually not required if the process is followed correctly. This was often complained about on feedex.

- change appointment booking links to follow the Govspeak style guide.

- move some countries from "no consular CNI" outcome to the new "marriage via local authorities" outcome. This was inaccurate because those countries have consular facilities but some paragraphs in "no consular CNI" suggest otherwise.


chrisroos:
I've left a number of comments about pretty minor things but in general this all looks good to me.

tadast:
Thanks a lot for reviewing this massive PR @chrisroos! :bow:

PR #1899, opened 2015-08-20
Upgrade govuk_frontend_toolkit

chrisroos:
This replaces #1896.

I've upgraded the version of the [govuk_frontend_toolkit][] from 3.1.0 to 4.2.1 so that we can use the `GOVUK.analytics.trackEvent` function instead of using the `ga` function directly.

I've been through the govuk_frontend_toolkit [changelog][] and each of the commits following the one that bumped the version to 3.1.0 (see [dbba96e][]).

The most significant change from 3.1.0 appears to be the move of `GOVUK.analytics` from the [static app][] to govuk_frontend_toolkit. It was added to govuk_frontend_toolkit on 10th Mar 2015 (see [PR 162][] for more), and removed from static on 12th Mar 2015 (see [PR 557][] for more).

We're using `GOVUK.analytics` to track:

* An event when a user clicks "Change" or "Start again"
* An event when a user reaches an Outcome page
* Pageviews as users navigate through the Smart Answer (we call `.trackPageView()` because the Smart Answer questions are loaded into the page using JavaScript).

For each of these cases I used my local Smart Answers and the [Google Analytics Debugger][] to observe the data being sent to GA before and after this branch. I'm confident that the data remains the same and that this upgrade shouldn't affect our Google Analytics tracking.

[changelog]: https://github.com/alphagov/govuk_frontend_toolkit/blob/master/CHANGELOG.md
[govuk_frontend_toolkit]: https://github.com/alphagov/govuk_frontend_toolkit
[dbba96e]: https://github.com/alphagov/govuk_frontend_toolkit/commit/dbba96e194b20ddc25e207e62fb7356536e54264
[static app]: https://github.com/alphagov/static
[PR 162]: https://github.com/alphagov/govuk_frontend_toolkit/pull/162
[PR 557]: https://github.com/alphagov/static/pull/557
[Google Analytics Debugger]: https://chrome.google.com/webstore/detail/google-analytics-debugger/jnkmfdileelhofjcijamephohjechhna?hl=en


app/views/smart_answers/_result.html.erb
```diff
@@ -23,15 +23,11 @@
...
-    hitType: 'event',
-    eventCategory: 'Smart Answer',
-    eventAction: 'Completed',
-    eventLabel: flowName,
-    nonInteraction: 1,
-    page: document.location.pathname
-  };
-  if (typeof window.ga === "function") {
-    window.ga('send', event);
+  var page = document.location.pathname;
```
tadast: Not sure if it's worth worrying about, but this makes `page` and `options` variables available globally. They're rather generic names and if I'm correct it could in theory cause problems with other JS that may rely on `if(options === undefined)` or similar code. Putting it in a closure would avoid this problem e.g.:
```JavaScript
(function(){
  var flowName = '<%= @name %>';
  var page = document.location.pathname;
  var options = {
    label: flowName,
    nonInteraction: true,
    page: page
  };
  GOVUK.analytics.trackEvent('Smart Answer', 'Completed', options);
})();
```

chrisroos: I think you're right. I'll make the change and update this pull request.

chrisroos:
In working on this branch again this morning I've learnt that the Analytics JavaScript we're using actually comes from the version of govuk_frontend_toolkit in Static, and not the one in our app. Which means that I don't need to upgrade govuk_frontend_toolkit in our app.

I'm going to close this, update the branch and open a new pull request.

PR #1900, opened 2015-08-20
Convert landlord-immigration-check from Smartdown to Smart Answer

floehopper:
I took the following approach:

* Introduce `postcode_question` functionality for Smart Answer flows
* Fix a bug in pre-filling of postcode form field which had been present in Smartdown
* Copy Smartdown plugin code for the flow onto a new Calculator class
* Implement a tool to convert a Smartdown flow to a Smart Answer flow
* Run the tool to do the conversion
* Implement custom next node logic for postcode question
* Replicate postcode question validation messages from Smartdown
* Add question-marks to end of question node keys
* Update regression test checksums
* Put question node definitions into a more logical order

Note that the conversion tool has only been "tested" with this flow and is unlikely to work unchanged for any other flows. It should raise exceptions when it thinks the flow is doing something that it doesn't support, but I suspect there may be some things like this which I've overlooked e.g. next steps, `$IF` logic, markers, etc where it's likely that no meaningful exception is raised.

My hope was that the tool does a reasonable job of "documenting" how I've done the conversions and _might_ be of some use when we come to convert other flows.


lib/smart_answer/calculators/landlord_immigration_check_calculator.rb
```diff
@@ -19,7 +18,7 @@ module LandlordImmigrationCheck
     )
 
     def self.valid_postcode(postcode)
```
tadast: You've been adding a lot of question marks in other commits, can I request a bonus feature to add a `?` to the end of this method as it always returns a boolean value? :)

floehopper: I have it on my todo list (along with other improvements to that method's implementation), but I'd prefer to make those changes outside this PR.

PR #1902, opened 2015-08-20
Normalize responses to postcode_questions

floehopper:
This feels like a nice thing to do and is built-in to the `uk_postcode` gem
that we're already using. What's more it means that URLs for the same postcode
in a different format are now the same.

Note that making the change here means that the postcode is displayed in the
normalized form in the responses to previous questions. Also when you click one
of the related "Change" links the postcode text field is pre-filled with the
normalized form of the original value.


tadast:
Nice! :+1:

floehopper:
I forgot to say that I'm happy to make this change without worrying about "old" URLs that might be in the wild, because the only flow that includes a `postcode_question` is the `landlord-immigration-check` one which has not yet been published.

PR #1904, opened 2015-08-21
Updated postcodes for PIP

HollyRC:
Update to outcome: https://www.gov.uk/pip-checker/y/yes/1985-01-01

I've combined the 2 sets of postcodes that went live on 13 July and 17 August. And added a new list of postcodes to be included from 1 September.


tadast:
:+1:

Closing in favour of #1905

PR #1903, opened 2015-08-21
Updated postcodes for PIP

HollyRC:
Amending this outcome: https://www.gov.uk/pip-checker/y/yes/2001-01-01

I've combined the 2 sets of postcodes that went live on 13 July and 17 August. And created a new list of postcodes to be includes from 1 September. 

Relates to ticket: https://govuk.zendesk.com/agent/tickets/1118011

PR #1905, opened 2015-08-24
Update postcodes for PIP checker

tadast:
Needs to go live on on Wednesday 26th.

Updates to two outcomes:
https://www.gov.uk/pip-checker/y/yes/1985-01-01
https://www.gov.uk/pip-checker/y/yes/2001-01-01

Combines the 2 sets of postcodes that went live on 13 July and 17 August.
Adds a new list of postcodes to be included from 1 September.

Relates to ticket: https://govuk.zendesk.com/agent/tickets/1118011

Closes #1903 #1904

### DIFF
--- a/lib/smart_answer_flows/pip-checker/_scheme_postcodes.govspeak.erb
+++ b/lib/smart_answer_flows/pip-checker/_scheme_postcodes.govspeak.erb
@@ -1,7 +1,7 @@
 
 ###Your DLA ends after September 2017 or your DLA award has no end date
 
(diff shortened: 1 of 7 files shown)


floehopper:
Although I have access to Zendesk, I can't see the ticket referenced in the commit note, presumably because of a permissions issue. In case this affects other developers, I think it _might_ be worth including a summary of that ticket in the commit note if there's anything important to add.

tadast:
I believe it's a regular thing that's done and unfortunately there isn't anything particularly useful on zendesk.

floehopper:
OK. No probs.

PR #1908, opened 2015-08-24
Enable regression tests for landing pages

tadast:
Allow rendering landing pages as govspeak

This is going to be useful when converting landing pages to erb. 
We'll want regression tests to ensure smooth transition to the new format, for which we'll need to render and save govspeak output of the landing pages the same way we currently do for outcomes.

I've added a sample of a generated landing page artefact, it has a title, body and post_body parts.
If you're happy with implementation and the `<flow_name>/landing_page.txt` naming convention I'll generate landing pages for each smart answer.

Edit: As voted by the majority, we've decided to go with the  `<flow_name>/<flow_name>.txt` naming convention

### DIFF
--- a/app/controllers/smart_answers_controller.rb
+++ b/app/controllers/smart_answers_controller.rb
@@ -20,8 +20,10 @@ class SmartAnswersController < ApplicationController
           title: @presenter.current_node.title
         }
       }
-      if @presenter.finished? && !Rails.env.production?
-        format.text { render }
+      if !Rails.env.production? && @presenter.render_txt?
+        format.text {
+          render
+        }
       end
     end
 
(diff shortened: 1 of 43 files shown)


floehopper:
Apart from my minor comments, this looks good to me. :+1:

tadast:
I was considering <flow-name.txt> too but figured `<flow-name>/<flow-name.txt>` is less informative than the current alternative. Happy to rename this to `landing-page.txt`. Let's wait for Chris' thoughts :)

floehopper:
I've just rebased this against master in preparation for merging it.

floehopper:
I've fixed a govuk-lint-ruby violation in the first commit and force pushed again.

PR #1909, opened 2015-08-25
Reduce content duplication in "Making your application" in Overseas Passports

tadast:
The aim of this change is to make it easier to understand what needs changing when a change request comes in. Duplicated paragraphs make it difficult to trace the source of the content rendered on the outcome page.

"Making your application" section has been extracted to a partial, with all its (sub)partials nested under a new "overseas_passports/making_your_application" dir. This helps to isolate partials related to this section only, because otherwise it can get really confusing to deal with many partials in the main overseas-passports directory. Also, the ips_application_result file is rather big, reducing the overall number of lines in it makes it slightly easier to work with it.

Some cleanup commits have been made before the last and the most important change. These commits have helped me detect duplicated paragraphs.


chrisroos:
This looks great to me. Good work!

PR #1910, opened 2015-08-25
Fix overseas passports in Saint Barthélemy

tadast:
FCO have confirmed the process is identical to St Martin (online application).
Copy configuration in passports-data.yml from st-martin.

### Some history

Saint Barthélemy was added to the WorldwideApi as per FCO request.
This caused it to show up in all Smart Answers' country question drop downs. Most of the Smart Answers, including overseas passports, did not have logic for Saint Barthélemy and were failing because of that.

### DIFF
--- a/lib/data/passport_data.yml
+++ b/lib/data/passport_data.yml
@@ -1660,6 +1660,16 @@ rwanda:
   applying: 10 weeks
   replacing: 8 weeks
   optimistic_processing_time?: false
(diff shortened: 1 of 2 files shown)


chrisroos:
It might be worth mentioning in the commit note that you haven't had to update any regression test artefacts because we don't have regression tests for saint-barthelemy.

Aside from that minor comment, this looks good to me.

PR #1911, opened 2015-08-25
Add Rubocop to Jenkins

chrisroos:
This branch adds Rubocop to our Jenkins test suite.

Rubocop is supplied by the govuk-lint Gem along with a set of rules that correspond to the GOV.UK styleguides.

Rubocop only checks for violations in the diff introduced in the current branch. This should help us avoid introducing new violations of the styleguide without first having to fix all existing violations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,7 +189,7 @@ files relating to the regression tests e.g. file checksums, HTML artefacts, etc.
 
 ## Instructions
 
-1. Check out the branch from the forked repo onto your local machine. Note that `<github-username>` refers to the owner 
+1. Check out the branch from the forked repo onto your local machine. Note that `<github-username>` refers to the owner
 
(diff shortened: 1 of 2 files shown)


tadast:
I'm not a fan of breaking a build due to style violations, because in my experience automated tools are not perfect at detecting them. However, if I'm right gov-uk-lint will only execute in the build that runs in the background and builds regression tests, so it's a nice solution. Perhaps it's worth clarifying that in the README?

Looks good to me :+1:

chrisroos:
> However, if I'm right gov-uk-lint will only execute in the build that runs in the background and builds regression tests, so it's a nice solution.

Nope :-) This configuration actually runs `gov-uk-lint` before running `rake test` on Jenkins, not before running the regression tests.

My preference would still be to add this to Jenkins and to monitor it. If it starts getting painful then we should re-evaluate.

PR #1912, opened 2015-08-26
Bump slimmer and explicitly set slimmer template

edds:
The new version of slimmer changes the default template that gets used.
Bump slimmer to the newest version and ensure that the wrapper template
is still used.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -330,3 +330,6 @@ DEPENDENCIES
   uk_postcode (~> 1.0.1)
   unicorn (= 4.8.3)
   webmock (= 1.20.4)
+
+BUNDLED WITH
+   1.10.6
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -236,7 +236,7 @@ GEM
     simplecov-html (0.10.0)
     simplecov-rcov (0.2.3)
       simplecov (>= 0.4.1)
-    slimmer (8.2.1)
+    slimmer (9.0.0)
       activesupport
       json
       nokogiri (>= 1.5.0, < 1.7.0)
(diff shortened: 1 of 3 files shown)


Gemfile.lock
```diff
@@ -330,3 +330,6 @@ DEPENDENCIES
   uk_postcode (~> 1.0.1)
   unicorn (= 4.8.3)
   webmock (= 1.20.4)
+
+BUNDLED WITH
+   1.10.6
```
tadast: We had a discussion about this in another PR and decided not to commit it as part of other, unrelated changes 
https://github.com/alphagov/smart-answers/pull/1825#discussion_r34886514

edds: This is being generated by the version of bundler and ruby on the development VM. If you want I can remove it, but I can only see more and more people adding it as they bundle from within the VM.

chrisroos:
Did you consider adding a test to assert that we're using the "wrapper" template, and describe why we're using it?

chrisroos:
Was the upgrade of slimmer required for any other reason, or are you upgrading it now to avoid someone else doing it in future without realising that they also needed to make the other changes you've made in this commit?

edds:
Yes this was just preventative to ensure someone didn't miss it if they updated it.

I can try adding a test if you would like me to, though the vector for a developer to change this without realising is fairly small as the page does look quite broken.

Gemfile.lock
chrisroos: I think we're going to have to live with the `BUNDLED WITH` line being added to Gemfile.lock sooner or later, so we may as well have it in this commit.

I did just try running `bundle install` in master to see whether that would add `BUNDLED WITH` to the Gemfile.lock, but it didn't. Presumably it only gets added when something actually changes, i.e. a Gem version bump like this.

chrisroos:
> I can try adding a test if you would like me to, though the vector for a developer to change this without realising is fairly small as the page does look quite broken.

Nah, I don't think you need to add a test. If it's obviously broken without that line then I think that's good enough for now.

tadast:
Ok, let's merge this :shipit:

PR #1914, opened 2015-08-26
Update README for Smart Answer flows

chrisroos:
* Add `save_input_as` to the list of methods that can be used to store data for
use later in the flow.

* Move `save_input_as` further down in the example flow to illustrate when it's
actually called.


tadast:
LGTM

PR #1915, opened 2015-08-27
Energy grants calculator: Universal credit and Affordable Warmth Obligation

tadast:
As of 2015 April the new Universal Credit replaces the following:

- Jobseeker's Allowance
- Housing Benefit
- Working Tax Credit
- Child Tax Credit
- Employment and Support Allowance
- Income Support

This adds Universal Credit as an option, however it is still too early to remove
the other options as there are still users claiming those older types of credits.

If user selects Universal Credit, the behaviour is exactly the same as for
the process when the user receives Jobseeker's allowance, but the three possible outcomes must
have an extra paragraph about Affordable Warmth Obligation.

This has been fact-checked.


chrisroos:
Aside from my one comment, this looks good to me.

PR #1916, opened 2015-08-27
Overseas Passports. IPS application result sections as partials

tadast:
Continuing work in https://github.com/alphagov/smart-answers/pull/1909, I've extracted three more sections to their own partials in Overseas Passports. This makes it slightly easier on the eye to deal with this outcome.

It also draws clearer boundaries for local variables with lists of countries used in logic, and exposes what data each section depends on (locals passed to those partials).


chrisroos:
Presumably the regression test artefacts didn't change at all after extracting these partials? I think it'd be explain that in the three commit notes, or at least the merge commit note.

chrisroos:
All looks good to me.

PR #1918, opened 2015-08-28
Extract OS marriage in Italy to its own outcome

tadast:
This is up for discussion.

Opposite-sex marriages in Italy have plenty of exceptions. That means a lot of extra ifs in the generic CNI outcome. Splitting Italy out duplicates some of the CNI logic, but also removes those extra ifs in the already big CNI outcome. Similar thing has been done to other countries eg Spain.

### DIFF
--- a/lib/smart_answer_flows/marriage-abroad.rb
+++ b/lib/smart_answer_flows/marriage-abroad.rb
@@ -165,6 +165,10 @@ module SmartAnswer
           data_query.os_marriage_via_local_authorities?(ceremony_country)
         }
 
+        define_predicate(:marriage_in_italy) {
+          ceremony_country == 'italy'
+        }
+
         next_node_if(:outcome_brazil_not_living_in_the_uk, ceremony_in_brazil_not_resident_in_the_uk)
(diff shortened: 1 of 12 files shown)


chrisroos:
There are a number of changes in marriage-abroad-responses-and-expected-results.yml where `next_node` has changed from `outcome_os_consular_cni` to `outcome_marriage_via_local_authorities`. These don't look as though they should be part of this commit.

chrisroos:
This all looks good to me.

Do you think that this points toward country-specific outcomes being a sensible approach? Or is it too early to tell (i.e. you'd need to convert more to see whether it made things easier)?

tadast:
> There are a number of changes in marriage-abroad-responses-and-expected-results.yml

I don't know how this happened, but here's a fix https://github.com/alphagov/smart-answers/commit/58f978c55711515736944e3346b853d4b06ffa67

> Do you think that this points toward country-specific outcomes being a sensible approach?

I think it makes sense in this particular case, because Italy has a lot of exceptions and there seem to be more changes that introduce exceptions at a per-country level rather than a group of countries (e.g. all CNI outcomes). The trade off here is if all CNI countries need to change, a developer will need to know to change the consular_cni_outcome, outcome_spain, outcome_os_italy etc.

PR #1920, opened 2015-09-01
Update minimum_wage_data

lutgendorff:
For October 1

Added National minimum wage rates for the next  year


tadast:
That's great, thanks! I've updated tests and opened a new PR with your changes https://github.com/alphagov/smart-answers/pull/1922

PR #1921, opened 2015-09-01
Update minimum wage

lutgendorff:
Update for Oct 1

Adding date range for last year's national minimum wage rates.


tadast:
That's great, thanks! I've updated tests and opened a new PR with your changes #1922

PR #1922, opened 2015-09-01
Update minimum wage rates

tadast:
Add National minimum wage rates for the next year.

:warning: This needs to be deployed on 1st of October 2015


chrisroos:
Looks good to me.

Out of interest, what's the problem with deploying this before 1st Oct?

lutgendorff:
I'm assuming that it would change the current rate as the post-October 1 rate (which would be incorrect for all of September). But if it doesn't, we could deploy.

tadast:
@lutgendorff for the current rate calculation it will keep using correct rates and then automatically swap over on 1st Oct (for this reason we can't really fact-check this change). 

The only change that would become visible immediately is the new option in the past rate calculations for "Oct 2014 - Sep 2015", which I guess is ok to deploy before 1st Oct? This can be fact-checked, basically, by matching the current rate calculations (before 1st Oct) with outcomes for "Oct 2014 - Sep 2015". Should I create a preview link?

chrisroos:
@lutgendorff: Given @tadast's answers above, can we deploy this before 1 October?

chrisroos:
In reviewing this PR again I noticed that we also needed to update the text of question 1 in both am-i-getting-minimum-wage and minimum-wage-calculator-employers. I've made that change and also taken the opportunity to split the last commit into a number of smaller commits.

I've also rebased on master and force pushed.

chrisroos:
I've rebased and force pushed this branch in preparation for merging it to master.

PR #1923, opened 2015-09-01
Extract Marriage Abroad Affirmation outcome sections as partials

tadast:
Continuing the work I started by extracting Overseas Passports  sections,
I'm doing the same for this outcome. This doesn't gain much from the
Computer-science point of view, but it does reduce the mental load a developer
has to deal with when looking at the affirmation outcome template.

This also makes it clearer which sections of the outcome depend on what pieces
of data exposed by the flow file.

This change results in no changes to regression test artefacts.


chrisroos:
Looks good to me mate.

PR #1924, opened 2015-09-01
Stop referencing the fees table in the text

tadast:
We do not do this with any other sections, so keep it consistent.
Suggested by a content designer.


floehopper:
LGTM :+1:

PR #1925, opened 2015-09-02
Update Overseas Passports process in Afghanistan

tadast:
As of 7th Sept 2015 users in Afghanistan will need to contact the visa
application centre in Pakistan with passport-related queries.

This has been fact-checked and is ready to go live.


chrisroos:
The change looks good to me, although I'm not sure you want the third commit to be a `fixup!` commit. If you use an interactive rebase to merge the first and third commits then you'll have some sorting out to do to get the second commit to apply as it currently amends files in the first commit that will no longer exist. If you're not planning to do an interactive rebase then I think the use of `fixup!` in the third commit message is confusing. I think it'd be better to say that it essentially reverts the first commit.

Does that make sense?

tadast:
I think it will be easiest if I just cherry pick the changes and regenerate the checksums in master.
Thanks for reviewing!

PR #1926, opened 2015-09-03
Marriage in Vietnam changes

tadast:
The process of getting married in Vietnam has changed. See details in each commit message.


floehopper:
Apart from minor comments this LGTM :+1:

PR #1927, opened 2015-09-03
Marriage in Poland changes

tadast:
Honestly, I don't know the exact reason why these things are changing as the content designer who wrote up the changes is unavailable. I've made some educated guesses when writing up commit messages.


chrisroos:
Looks good to me.

PR #1929, opened 2015-09-04
Marriage abroad generic copy improvements

tadast:
These globally applicable improvements were extracted from the latest batch of country-specific changes. They are applied globally with the contend designer's blessing :)


floehopper:
LGTM :+1:

PR #1930, opened 2015-09-04
Changes to marriage in UAE

tadast:


floehopper:
Apart from minor comments, this LGTM :+1:

tadast:
I've addressed the comments and merged manually via 7a1f9c3

PR #1932, opened 2015-09-09
Allow exposing pages as govspeak based on an env variable

tadast:
To simplify coordination of content change requests between content designers
and developers we're going to expose Smart Answer outcomes as govspeak by
appending `.txt` to Smart Answer outcome URLs. This will allow content designers
to copy, paste and modify (in their text editor) govspeak. Currently they are
'reverse-engineering' it from HTML outputs.

We do not want to expose this feature to the wide public, but we want to enable
it on preview. The main concerns for not exposing it in production are:

- a fear of needing to support it in a long term
- it is an internal 'tool' and it feels clunky to have it available in production

This also sets `EXPOSE_GOVSPEAK=true` on heroku instances and documents it in README.

NB.: I've submitted a PR to the private repo that sets this env variable to true on dev VMs and in the preview environment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,11 +24,34 @@ For more information, please go to the [Smartdown SmartAnswer README](lib/smartd
 Smart answers are by default expected to be in Ruby/YAML style.
 To transition a smart answer from Ruby/YML to Smartdown style, register it in the smartdown registry (`lib/smartdown/registry.rb`).
 
-## Debugging current state
+## Developing
+
+### Installing and running
+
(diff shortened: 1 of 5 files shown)


floehopper:
Other than my small suggestions, this LGTM :+1:

tadast:
Thanks for feedback. I've made some tweaks, could you have another look please?

floehopper:
LGTM :+1:

PR #1935, opened 2015-09-10
Simplify date question

chrisroos:
This all came out of the work on the new HMRC Smart Answer (calculate-your-part-year-profits). I found the implementation of `Question::Date` to be somewhat more complicated than it needed to be. This pull request aims to simplify the class.

In summary, I've:

* Removed the unused `Question::Date#default` method.

* Simplified the implementations of `Question::Date#from`, `#to`, `#default_day`, `#default_month`, `#default_year` so that they can _only_ be set by providing a block rather than by providing a block or a value argument.

* Removed `Question::Date#defaulted_day?`, `#defaulted_month?` and `#defaulted_year?` and replaced their use with `Date#default_day`, `#default_month` and `#default_year`.

* DRYed up calls to the blocks stored in `@from_func` and `@to_func`.

* Restricted the type of exceptions we rescue in `Question::Date#parse_input` and `#to_response`. We were previously rescuing everything that inherits from `StandardError`, which isn't ideal as it means we might end up swallowing genuine exceptions.

* Simplified the implementation of `Question::Date#parse_input` when it's parsing a date supplied as a hash.

### DIFF
--- a/app/views/smart_answers/_date_question.html.erb
+++ b/app/views/smart_answers/_date_question.html.erb
@@ -10,7 +10,7 @@
       <%= select_month default_for_date(prefill_value_for(question, number, :month)), prefix: response_prefix, prompt: '' %>
     </label>
   <% end %>
-  <% unless question.defaulted_year? %>
+  <% unless question.default_year %>
     <label for="<%=response_prefix%>_year">Year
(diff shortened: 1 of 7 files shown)


floehopper:
Other than my minor comments/suggestions, this LGTM :+1:

chrisroos:
I've just rebased this on master and force pushed to this remote branch.

chrisroos:
I've force pushed again now that I've fixed the failures in date_question_test.rb.

PR #1938, opened 2015-09-10
Remove unused acronyms from marriage abroad content

tadast:
I've noticed that GRO and FCO acronyms are no longer used in most of the pages that have explanation for it (that usually activates when a user hovers over an acronym in text). This PR removes them for pages that don't use them.


floehopper:
LGTM :+1:

PR #1939, opened 2015-09-10
Halve the time it takes to run regression tests

tadast:
Shelling out and running git diff for each test artefact was consuming more than
half of the time to run regression tests. Avoiding that results in:
```
Before:
	marriage-abroad: 10:45
	register-a-birth: 3:53
After:
	marriage-abroad: 5:45
	register-a-birth: 1:34
```

I still left an option to enable early feedback by setting `EARLY_FEEDBACK=true`. I mostly hesitated to remove the diff completely because then regression tests would look very odd to someone unfamiliar with this performance issue :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,6 +160,9 @@ Test a single Smartdown flow by running:
         $ RUN_REGRESSION_TESTS=<name-of-smart-answer> \
           ruby test/regression/smart_answers_regression_test.rb
 
+If you want individual tests to fail early when differences are detected, set `ASSERT_EACH_ARTEFACT=true`.
+Note that this more than doubles the time it takes to run regression tests.
+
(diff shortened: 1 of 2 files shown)


floehopper:
Other than my comments, this look great to me! :+1:

tadast:
> Subjective: Part of me would prefer the environment variable to include REGRESSION_TEST or something similar in its name, but I can't really explain/justify it. I guess I just feel a bit uncomfortable that a "global variable" has a very generic name...

I've renamed the env variable from `EARLY_FEEDBACK` to `ASSERT_EACH_ARTEFACT` which should be regression-test-specific-enough :)

floehopper:
:+1:

PR #1941, opened 2015-09-10
Convert landing pages in Smart Answer flows to use ERB templates c.f. outcomes

floehopper:
I'm sorry there are a lot of commits, but the majority of them are small preparatory refactorings to avoid duplicating behaviour already implemented in `OutcomePresenter` in the newly added `StartNodePresenter`.

Note that for the moment I haven't gone as far as introducing a `StartNode`. This is because it would mean a bunch of extra work just so that the remaining Smartdown flows weren't broken. Since we're planning to migrate these to Smart Answer flows, I didn't think it was worth it at this point.

The more significant commits are:

* "Generate ERB templates for landing pages for smart answer flows"
* "Remove landing page content from locale YAML files"
* "Render landing pages from ERB templates vs locale YAML files"


tadast:
I have really enjoyed reviewing this PR commit by commit, it was like reading a fluent story :100:

I left some minor comments, but it looks great to me :+1:

floehopper:
@tadas: Thanks for reviewing. I'm going to force push & merge now.

PR #1943, opened 2015-09-11
[Deploy on 2015-09-21] Switch Cambodia to use UK Application Centre

tadast:
- Colour copy of documents is required when applying
- New courier fee is 23.01 (which means it's an IPS 3 group I guess)

This should go live on 21st Sep 2015

Update: fact-check process found no issues, :+1: for deployment on 21st


floehopper:
LGTM :+1:

PR #1944, opened 2015-09-11
[Don't merge. FC] Explicit Bus Pass age calculation

tadast:
Currently state pension age outcomes from from gov.uk/calculate-state-pension show a paragraph
about bus pass eligibility age. However it is not very easy to find as it is
at the bottom and only one of many paragraphs. A lot of users find it confusing.

A lot of local council and other authority websites link to this calculator
for bus pass age calculation (that's why we rejected an idea of having a
completely separate calculator), therefore making this option explicit should
make it easier for users. The new outcome has only information about the bus
pass age and a link back to State Pension age calculation (ideally we'd link to
an outcome, but we don't know user's gender).

I opted in for some duplication to introduce a bus_pass_dob? question. I could
have reused age_dob? question, but inelegant hacks are needed to workaround the
fact that an instance of StatePensionAmountCalculator is initialized there
and it requires gender to be passed in.

### DIFF
--- a/lib/data/state_pension_date_query.rb
+++ b/lib/data/state_pension_date_query.rb
@@ -12,7 +12,7 @@ class StatePensionDate < Struct.new(:gender, :start_date, :end_date, :pension_da
   end
 end
 
-class StatePensionQuery < Struct.new(:dob, :gender)
+class StatePensionDateQuery < Struct.new(:dob, :gender)
   def self.find(dob, gender)
     state_pension_query = new(dob, gender)
     state_pension_query.find_date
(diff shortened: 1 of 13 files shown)


floehopper:
Ideally we'd move this flow closer to the new style before making these changes, but I can see that that would involve quite a bit of work. If we did do that, it would be easier to add unit tests for e.g. the validation of the date of birth.

floehopper:
It doesn't feel great that there are now three versions of the question asking for date of birth. Ideally it would be good to try to remove this duplication somehow.

floehopper:
Other than my comments, this LGTM. :+1:

tadast:
> It doesn't feel great that there are now three versions of the question asking for date of birth. Ideally it would be good to try to remove this duplication somehow.

I've tried, but I found the DSL very unsupportive for such use cases. Perhaps this is better explained in person.

I've addressed all other comments and am planning to ship this tomorrow. Let's talk about further improvements when you're back.

floehopper:
:+1:

PR #1945, opened 2015-09-17
Update pip_checker.rb

lutgendorff:
Changed result 4 to result 7 as we no longer need it. 

Will have to get Tadas to help me remove that result and the postcode info from the checker as I can't.


tadast:

PR #1947, opened 2015-09-17
Update result_7

lutgendorff:
The national roll out of PIP is now happening and so the post code areas are no longer needed.

I've removed the postcode info and added the post-Sept 2017 here.

There is a bit of clean up to remove the postcode and result_4 from github once this change is live as they are no longer needed.


tadast:

PR #1949, opened 2015-09-17
Fix panopticon integration

tadast:
As of 57493beeda38f35e3a9763bd3761ab3779f305fc Smart Answer titles and meta
descriptions are no longer stored in YAML files, however when flows are
registered with panopticon it is trying to look up those values in YAML
and falls back to humanised titles (that still have dashes in them) and empty
meta descriptions.

This change uses the new mechanism to fetch titles and meta descriptions from
ERB files. The fallback code for missing title has been removed, because
StartNodePresenter inherits NodePresenter#title that has this fallback
implemented.

I've noticed that there are more code that relies on titles and bodies being in YAML, but it's less important  e.g. [here](https://github.com/alphagov/smart-answers/blob/05f855cb21edd96637a25b89c4238f2fd3f9aef1/lib/graph_presenter.rb#L98), [here]( https://github.com/alphagov/smart-answers/blob/05f855cb21edd96637a25b89c4238f2fd3f9aef1/test/fixtures/graph_presenter_test/graph.yml#L7)

/cc @floehopper @alext

### DIFF
--- a/app/presenters/flow_registration_presenter.rb
+++ b/app/presenters/flow_registration_presenter.rb
@@ -14,7 +14,7 @@ class FlowRegistrationPresenter
   end
 
   def title
-    lookup_translation("title") || @flow.name.to_s.humanize
+    start_node.title
   end
 
   def paths
@@ -26,14 +26,14 @@ class FlowRegistrationPresenter
   end
 
   def description
-    lookup_translation("meta.description")
+    start_node.meta_description
   end
 
   NODE_PRESENTER_METHODS = [:title, :body, :hint]
 
   def indexable_content
     HTMLEntities.new.decode(
-      text = @flow.nodes.inject([lookup_translation("body")]) { |acc, node|
+      text = @flow.nodes.inject([start_node.body]) { |acc, node|
         pres = NodePresenter.new(@i18n_prefix, node)
         acc.concat(NODE_PRESENTER_METHODS.map { |method|
           begin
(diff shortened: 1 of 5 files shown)


floehopper:
Sorry I missed this when I made the original change.

This all looks good to me. :+1:

floehopper:
@tadast: I'm going to have a look at fixing the remaining places which suffer from this problem.

tadast:
:100: thanks

floehopper:
@tadast: I've removed the remaining landing page content from i18n YAML files in [this commit](fa0186dce2fb4978055001499b9824a785688dbf), one of which you linked to in your comment above.

As far as I can tell, the methods on `GraphPresenter` (e.g. `#node_title`) are not used to obtain content for landing pages; instead the standard `SmartAnswerPresenter` is used which will obtain content from ERB templates as appropriate.

However, since we moved outcome content into ERB templates, the `GraphPresenter` methods will not be finding outcome content in i18n YAML files. It looks as if they're using the typical rescue-exception-and-fallback pattern to a simple default value which is presumably why we haven't been seeing any exceptions. Either that or nobody is using the visualisations :wink:.

Fixing the latter problem doesn't seem like a high priority, so I've decided just to create a [Trello card](https://trello.com/c/Qx4t6lnE) in our backlog.

tadast:
:cool: 
I actually found the graph very useful working on a change that involves rearranging questions. I did not notice the absence of information in outcomes though :)

PR #1950, opened 2015-09-18
Require parent contact details form in three countries

tadast:
Bangladesh, India and Pakistan now require parents applying for their child's
*first* passport to fill in the parent contact details form.

Note that the numbering in the bullet points is off by one in some cases in
Pakistan. This could by avoided by introducing more conditional branches, but
I opted in to simply ignore it, because when markdown renders ordered lists
it simply strips provided numbers and uses HTML's \<ol\> element that shows the
correct numbers to users.


chrisroos:
Looks good to me.

tadast:
This came back positive from the fact check, I'll merge when master is stable :)

PR #1951, opened 2015-09-21
UK visa checker urgent changes

tadast:
Three somewhat urgent changes (users are given wrong advice). This has been through FC and this can go live.

## Differentiate between Israeli travel documents

We've had letters of complaint from customers that the Visa Checker tool is not
clear for holders of Israeli travel documents in lieu of passports. Israeli
passport holders are non-visa nationals, so in general only require entry
clearance if they are coming here for longer than 6 months. Israeli-issued
travel documents are issued to non-Israeli citizens (like Palestinians) and
the holders are visa nationals so require a visa to come to the UK for any
reason.

Add an Israel-specific question that collects information about the user's
travel document type and show the corresponding outcome.

## Update visa advice to Venezuelans in transit

It has been reported by the department that Venezuelans in transit need a visa
and the current advice on GOV.UK is incorrect.

## Update visa advice for Taiwanese users

Outcomes were incorrectly linking to the short term study visa when users are
actually required to apply for a standard visa.


chrisroos:
Looks good to me. Presumably the tests have failed because the changes in the final 'fixup' commit need the regression test artefacts to be regenerated again?

tadast:
Thanks for reviewing.

I haven't updated the checksums because content designers are not sure how best to describe the provisional passport. I'll update this PR once we hear back from the department :)

tadast:
@chrisroos I've added two more commits to this PR, would you mind checking them as well?

tadast:
Update: apparently there were two people fact-checking this and one of them thinks a correction is needed, so please hold off reviewing for now

chrisroos:
We've now received confirmation that this change can go live. I'm going to double check it and get it merged.

PR #1954, opened 2015-09-21
Add regression tests for question pages

floehopper:
This is in preparation for refactoring the code around question pages.

Unlike the code for outcome pages, question pages have little (?) or no
response-related logic, and so, even if we only render the page for each
question node *once*, this should give us decent regression test coverage.

Rendering the page for each question just once means that time taken to run
the regression tests is only very slightly increased.

Unlike the landing and outcome pages, there is currently no easy way to
render a simplified (e.g. Govspeak) version of a question page, so we're
rendering the full HTML version of the page for these nodes.

### DIFF
--- a/lib/smart_answer_test_helper.rb
+++ b/lib/smart_answer_test_helper.rb
@@ -76,8 +76,8 @@ class SmartAnswerTestHelper
     artefacts_path.join(@flow_name)
   end
 
-  def save_output(responses, response)
-    filename = responses.pop + '.txt'
+  def save_output(responses, response, extension: 'txt')
+    filename = "#{responses.pop}.#{extension}"
     path_to_output_directory = path_to_outputs_for_flow.join(*responses)
     FileUtils.mkdir_p(path_to_output_directory)
     path_to_output_file = path_to_output_directory.join(filename)
--- a/test/artefacts/additional-commodity-code/0.html
+++ b/test/artefacts/additional-commodity-code/0.html
@@ -1,0 +1,134 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <link rel="stylesheet" media="all" href="/smartanswers/application.css" />
+    <link rel="stylesheet" media="print" href="/smartanswers/print.css" />
+    <title>Look up Meursing code - GOV.UK</title>
+    <script src="/smartanswers/smart-answers.js" defer="defer"></script>
+    <meta name="robots" content="noindex">
+
+  </head>
+<body class="mainstream">
+
+  
+
+  <div id="wrapper" class="answer smart_answer">
+    
+
+<main id="content" role="main">
+  <div id="js-replaceable"  class="smart-answer-questions group">
+      
+
+  <header class="page-header group">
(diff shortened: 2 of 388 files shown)


chrisroos:
It doesn't look like we're saving the very first question page. For example, the initial question page for additional-commodity-code is at https://www.gov.uk/additional-commodity-code/y, while the first question page we're storing is from https://www.gov.uk/additional-commodity-code/y/0.

Do you think we need to store the initial question page too?

tadast:
Rendering HTML means we will need to re-generate artefacts if things like google analytics code, HTML in the header or "Last updated:" text changes. I wonder if it's worth the effort now to only persist the relevant part of HTML e.g. "#js-replaceable" div?

floehopper:
@chrisroos: Good catch. I've just force-pushed a fixed version. You can now see the first question page of additional-commodity-code here: https://github.com/alphagov/smart-answers/commit/2f1fb2bad9ca396951853a36e89616f2e4e70cce#diff-babfe1bd80dc97f0f788402edbd3145a

@tadast: I'd prefer to live with the brittleness initially and only improve on it if we find it's a problem.

chrisroos:
I've just noticed in the commit titled "Question page HTML for marriage-abroad" that there are 4 question page artefacts generated. There are 5 questions in marriage-abroad so I think I'd expect to see 5 artefacts generated. Having just had a real quick look with @floehopper it looks as though we're not generating the HTML for the `:marriage_or_pacs?` question.

floehopper:
@chrisroos: I've just fixed this and force pushed again. You should be able to see that there are now 5 question page artefacts for marriage abroad in 5fd20581eaaf59386439a11c4913958d84eeebd8 vs 4 in the old version of the same commit 7a311b06d841087458f1212e201e166e904b9010. The fix to the regression test was to record `next_node` in the `visited_nodes` set rather than `current_node`.

chrisroos:
I just had a brief chat with @floehopper about generating regression tests for questions in Smartdown flows. James was slightly concerned that we might be generating artefacts incorrectly for those flows. I said I'd investigate what was happening.

We only have two Smartdown flows (pay-leave-for-parents and pay-leave-for-parents-adoption) and we've not yet generated regression tests for those, so I don't think we have anything to worry about.

I also compared the number of Smart Answers for which HTML test artefacts for question pages have been generated in this branch (`find test/artefacts -name "*.html" | cut -d"/" -f3 | sort | uniq | wc -l`) to the number of Smart Answers (`ls -l lib/smart_answer_flows/*.rb | wc -l`). We have 37 of each so I'm happy that we're generating question page artefacts for each of the Ruby Smart Answers.

chrisroos:
Looks good to me!

floehopper:
I've addressed @chrisroos' [latest comment](https://github.com/alphagov/smart-answers/commit/97270e4ffa31931383dd5c204718a5cf668bd613#commitcomment-13412318), rebased against master and force-pushed. I plan to merge this after lunch assuming all the tests still pass...

PR #1955, opened 2015-09-22
Fix flow registration (indexable content for outcome nodes)

floehopper:
The main change is in the first commit: "Fix indexable content for outcome nodes in flow registration".

When we switched over to use only ERB templates for outcomes indexable content for outcome nodes was inadvertently "broken" in that no content from outcome nodes will have been included. Unfortunately due to an unfortunate combination of a couple of things, the unit test did not highlight the problem.

The second commit just fixes a typo in the path for a sample flow.


chrisroos:
It's outside of the scope of this PR but I wonder whether the `NodePresenter` could raise an exception if initialized with an `Outcome` node (and potentially other nodes that it shouldn't be used for)?

floehopper:
I think having a factory method (see above) and using it consistently everywhere would probably be good enough rather than raising an exception, but I'll keep it in mind.

chrisroos:
> I think having a factory method (see above) and using it consistently everywhere would probably be good enough rather than raising an exception, but I'll keep it in mind.

Sure. I guess I was wondering whether failing fast might help catch any remaining problems similar to the one addressed in this PR. I can see how consistently using a factory method would avoid these sort of problems too, though.

floehopper:
You're right about `%{i_dont_exist}`. Well spotted. I had moved it onto the question, but then somehow lost the change. I'll sort it out.

floehopper:
I've force-pushed this branch with the various fixes.

PR #1956, opened 2015-09-22
Send `content_id` to panopticon

tijmenb:
We are building a new tagging infrastructure. The API for tagging will work by content-id.  Therefore, every piece of content needs to have a content-id in the content store, in order for it to be able to be tagged.

This PR ensures that we all smart-answers have a `content_id` and that we are sending it to panopticon (when running `rake panopticon:register`). This will aid us in the transition.

A PR to send smart-answers to content-store will be next.

Best reviewed per-commit.

Trello: https://trello.com/c/K87fWrYx (part of https://trello.com/c/ReoZKx9S).

Previous PR on smartdown: https://github.com/alphagov/smartdown/pull/147

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -24,7 +24,7 @@ gem 'uglifier'
 if ENV['SMARTDOWN_DEV']
   gem 'smartdown', path: '../smartdown'
 else
-  gem 'smartdown', '~> 0.15.1'
+  gem 'smartdown', '~> 0.16.0'
 end
 
 if ENV['API_DEV']
--- a/Gemfile
+++ b/Gemfile
@@ -30,8 +30,9 @@ end
 if ENV['API_DEV']
   gem 'gds-api-adapters', path: '../gds-api-adapters'
 else
-  gem 'gds-api-adapters', '20.1.1'
+  gem 'gds-api-adapters', '~> 24.4.0'
 end
+
 gem 'htmlentities', '~> 4'
 
 gem 'extlib', '0.9.16'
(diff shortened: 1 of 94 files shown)


alext:
You'll also need to update gds-api-adapters.  Including the content_id was only added in version 24.2.0 - https://github.com/alphagov/gds-api-adapters/blob/master/CHANGELOG.md#2420

tijmenb:
The tests for this branch seem to be hanging. Investigating.

tijmenb:
@alext I've extracted and refactored the `panopticon:register` rake task to allow isolated testing (and re-use for content-store sending). 

This enabled us to add a test for the data that is sent to panopticon. https://github.com/alphagov/smart-answers/commit/5630b7be7696c064b9a74b9b16300d8e514ca041 indeed failed before updating the gem, and passed after.

chrisroos:
What are the implications of a Smart Answer not having a content-id? Should we fail fast during registration with panopticon if content-id is missing?

rboulton:
@chrisroos Things which don't have content-ids will be unable to be tagged to topics or mainstream browse pages, in the reasonably near.  I think it would be reasonable to fail fast.

chrisroos:
@tijmenb: What do you think about failing fast as per @rboulton's comment?

tijmenb:
With the addition of https://github.com/alphagov/smart-answers/commit/2ecc24a29006390ee7fbffa6f6d94abdd0934567 I think I've addressed all the comments.

As for the failing-fast: yes, but I'd like to wait until we've implemented sending the pages to the content-store, because that would need the same logic and I don't want to duplicate it.

chrisroos:
Looks good to me.

PR #1958, opened 2015-09-23
Merge variable paragraphs into one in overseas passports

tadast:
We used to have three nearly identical paragraphs for countries with different
IPS configuration. I've suggested a content designer the meaning of those
variations is the same and they advised always using the first variation.


floehopper:
LGTM :+1:

PR #1960, opened 2015-09-23
Support graphing of flows that use next_node with a block

chrisroos:
Commit 98c598287955ec56238d1729537ed29099cc1067 added support for
defining `permitted_next_nodes`. As per the conversation on that commit
the intention was that this could be used to graph flows that hadn’t
switched to using `next_node(:name)` or `next_node_if`. This commit
builds on that earlier commit by taking account of
`permitted_next_nodes` in the `GraphPresenter`.

I’ve updated the example `GraphFlow` so that it has two question nodes:
one using the predicate logic and the other using
`permitted_next_nodes`. I’ve also updated the related graph.yml file
for that Smart Answer. Although not strictly necessary, I’ve also
updated the graph_missing_outcome_label.yml file that’s used as an
alternative file for the `GraphFlow`.

I double checked this change by adding `permitted_next_nodes` to the
relevant nodes in the student-finance-calculator flow and confirmed
that the visualisation rendered as expected.


floehopper:
Other than my one comment, this looks good to me.

chrisroos:
> I don't think there should be any YAML entries for the two outcomes - the labels for these are probably just being generated from the node names (assuming there's no ERB templates for them).

You're quite right! And you made me realise that the `GraphPresenter` was still trying to load/read the Outcome title and body directly from the YAML file. I've changed this in the "Always display outcome node name in graph visualisation" commit.

@floehopper: Would you be able to take another look to make sure it's good to merge, please?

floehopper:
This looks good to me. :+1:

PR #1961, opened 2015-09-24
Don't set `GdsApi::Base.logger`

tijmenb:
We're currently setting `GdsApi::Base.logger` when sending something to panopticon. There's no real reason to use this. This code with extra logging was presumably copied from another repo - we don't need it here.

We still need the `puts` statements to log the output of the panopticon registration.

Removing the logger prevents noisy output during the tests and cleans up some special casing in the tests.


floehopper:
This looks good to me. Thanks. :+1:

PR #1962, opened 2015-09-24
Refactor node presenters

floehopper:
My original motivation for doing this was that the generic `NodePresenter` contained a lot of behaviour that was only used in `QuestionPresenter` and its subclasses. In particular it seemed potentially quite confusing that all the i18n translate code was in `NodePresenter`. We're also keen to ditch the i18n translation code soon, so separating it out a bit feels as if it will help with that.

However, while doing this refactoring, I've ended up doing a few other things that became obvious along the way. I'm afraid there are quite a few commits, but I promise most of them are very small!

Here's a bit of a summary of things I've done:

* Push question-specific methods on `NodePresenter` down into `QuestionPresenter`
* Push corresponding unit tests in `NodePresenterTest` down into `QuestionPresenterTest`
* Extract `I18nRenderer` from `QuestionPresenter` to separate most of the i18n translation code
* Remove `has_xxx?` methods from node presenters
* Remove `method_missing` from `NodePresenter` and replace with explicit delegation in relevant node presenters

There's lots more I'd like to do, but I'm keen to get these changes into `master` as soon as possible and our priority has shifted onto writing documentation, so I want to free myself up to do that. Here are some ideas for further work:

* Push all remaining i18n-related code into the `I18nRenderer`
* Simplify public API of `I18nRenderer`
* Add unit tests for `I18nRenderer` (possibly moved or extracted from `QuestionPresenterTest`)
* Simplify node presenter constructor signature if possible


chrisroos:
This all looks good to me.

floehopper:
I've rebased this in preparation for merging.

PR #1964, opened 2015-09-24
Send smart answers to the content-store

tijmenb:
Team Finding Things is on a mission to tag all content on GOV.UK. Part of that mission is a new tagging architecture, to allow for much more flexible tagging. The core idea is to move the "source of truth" for tagging into a central place, and allow tags to be applied to all formats in a uniform way (from the [wiki](https://gov-uk.atlassian.net/wiki/display/FS/Tagging+Architecture)). For that, we need all content in the content-store.

This PR adds the rake task `rake publishing_api:publish` that will be run after every deploy, similar to `panopticon:register`. The rake task will send the smart answers to the content-store via the publishing-api.

Trello: https://trello.com/b/Kg3JL5Cd/finding-things
Previously: https://github.com/alphagov/smart-answers/pull/1956 https://github.com/alphagov/smart-answers/pull/1963

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -33,6 +33,8 @@ else
   gem 'gds-api-adapters', '~> 24.4.0'
 end
 
+gem 'govuk-content-schema-test-helpers', '~> 1.3.0'
+
 gem 'htmlentities', '~> 4'
 
 gem 'extlib', '0.9.16'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -104,6 +104,8 @@ GEM
       kramdown (~> 1.4.1)
       nokogiri (~> 1.5)
       sanitize (~> 2.1.0)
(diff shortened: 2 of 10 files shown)


tijmenb:
@floehopper re the `Services` module.

I've been on a little silent crusade to move away from globals and the `AppName.services(:api_name)` pattern (still used by [collections](https://github.com/alphagov/collections/blob/master/config/initializers/services.rb) for example).

It's now used in a bunch of apps:

https://github.com/search?utf8=%E2%9C%93&q=%22self.publishing_api%22+user%3Aalphagov&type=Code&ref=searchresults

Everyone seems to agree, so I say this is on it's way to become the "standard" way.

floehopper:
This looks good to me :+1:. I'll get it rebased against master and merged...

PR #1966, opened 2015-09-24
Fix typo in test class name & make test class name more consistent

floehopper:


chrisroos:
Looks good to me.

floehopper:

PR #1967, opened 2015-09-24
Statutory sick pay calculator bugfix

tadast:
Currently when calculating sickness duration the calculator does not take into account employee's usual work days, it simply subtracts the dates, which is incorrect. In order to fix this issue we first need to move the question that collects information about employees working days pattern earlier in the process.

Unfortunately this used to be the last question of the flow that carried all the logic to decide the outcome node. 

screenshot 2015-09-24 15 19 43
_taken from www.gov.uk/calculate-statutory-sick-pay/visualise_

Removing this question from it's position means this logic will have to be used in three different questions leading to those outcomes. This is an attempt to make the change somewhat nicely and to gather some feedback before continuing with the actual bug fix.

cc @floehopper


floehopper:
I would prefer the 2nd commit to be split into 2, but otherwise this LGTM :+1:

tadast:
@floehopper I've tried to address your comments but I'm in a rush to go to the airport, so I've pushed the changes to a separate branch https://github.com/alphagov/smart-answers/compare/tt_ssp_bugfix_alternative

floehopper:
I've just started reviewing @tadast's "alternative" branch. It's worth noting that his work so far has been about moving the "Which days of the week do they actually work?" question earlier in the flow. He hasn't yet attempted to fix the calculation to take account of the response. If we're going to do this before @tadast is back from holiday, we're probably going to need some help on understanding the changes to the calculation. /cc @lutgendorff

floehopper:
I'm now addressing my comments on a local version of @tadast's `tt_ssp_bugfix_alternative` new branch. @lutgendorff has explained how the calculation should be changed, so I'll work on that next.

floehopper:
I've addressed my own comments and applied a hacky attempt at solving the actual bug. I've pushed this work up as a [separate branch](https://github.com/alphagov/smart-answers/compare/jm_ssp_bugfix_alternative). I'm planning to deploy this to Heroku for @lutgendorff to check before I do too much more work to make this change properly.

floehopper:
Closing in favour of #1998.

PR #1970, opened 2015-09-24
Use Plek.new instead of Plek.current

floehopper:
In #1964, @tijmenb pointed out that `Plek.new` is an alias of `Plek.current`
and the former is preferred [1].

[1] https://github.com/alphagov/plek/blob/v1.7.0/lib/plek.rb#L50...L54


chrisroos:
Looks good to me.

chrisroos:
I also ran the tests in this branch locally (`rake`) and they all passed. I presume the failure on Jenkins is due to an intermittent test failure.

floehopper:

PR #1971, opened 2015-09-25
State pension top up bugfixes

tadast:
A bug in State Pension Top Up smart answer was reported. Further investigation and discussion to HMRC led us to uncover a few more bugs. Each of them explained in commit notes.


tadast:
This came back as :+1: from fact check. @chrisroos @floehopper could you please review this PR and notify Keith about the deployment time once scheduled. Thanks :)

floehopper:
I'm reviewing this now as per @tadast's request above.

floehopper:
I've now finished reviewing this. Other than my query about the unexpected changes in 2d9730bacf4399f9721e9165ce2314597dc5294b, I'm happy that this could be deployed as it stands. However, since @tadast is away and is not going to be working on Smart Answers when he gets back, I'm going to try to address my own comments as best I can. Hopefully I'll be able to answer my own query (above) once I'm a bit deeper into the code.

floehopper:
I've just force-pushed by local version of this branch having rebased it against `master`.

PR #1973, opened 2015-09-28
over55_result.govspeak.erb

lutgendorff:
Based on Feedex, people are confused why they aren't getting an estimate.

This change is to emphasise that they can't get it because the state pension is changing.

Moved state pension age and date to the bottom of the page with new heading


chrisroos:
This has been superseded by #1974. Closing.

PR #1977, opened 2015-09-29
Convert pay-leave-for-parents to single question per page

chrisroos:
This is in preparation for converting this published Smartdown Smart Answer to Ruby.

I started out by renaming some of the question pages and questions so that they better reflected their purpose. I found this useful to help me understand a bit more about how the Smart Answer worked. Now that I understand more about the Smart Answer I could imagine redoing this pull request without first renaming the question pages and questions in order to limit its scope.

With all the renaming done I started splitting the questions so that we end up with one question per page. I moved one question at a time to a new question page, moving them in reverse order of how they appear in those question pages.

I used the Smartdown scenarios to give me some confidence that I hadn't broken anything, although I'm not entirely sure how comprehensive these tests are.


floehopper:
Apart from my suggested changes to commit notes and a couple of suggested changes to question names, this all looks good to me. :+1:

Before I looked at the actual changes, I was sceptical that it would be sensible to do all the renaming that @chrisroos mentioned in the PR description. However, having looked at the actual changes, I think the renaming is really valuable - many of the original page & question names were really unhelpful - so I would keep them as part of this PR.

Having said that we should be mindful that this means that this `pay-leave-for-parents` flow has diverged even further from the draft `pay-leave-for-parents-adoption` flow, which probably means it will be even harder to convert that to Smart Answers if it turns out we need to.

In the absence of regression tests for this flow, the existing scenarios should give us a fair amount of confidence that the changes in this PR haven't broken anything, but it might be worth seeing how hard/worthwhile it would be to run the [Fact Check table generation](pay-leave-for-parents-adoption) rake task before and after the change to see whether that highlights any differences.

chrisroos:
I've rebased this on master and force pushed to this branch in preparation for merging.

I spent some time investigating whether I could use the `smartdown_generate_factcheck:pay_leave_for_parents` Rake task to compare the before and after of making this change. The Rake task did yield different results before and after but I wasn't able to understand enough of what it was doing to know what the differences meant.

I then realised that I could use the regression tests to check that this branch doesn't break anything. I created a new branch off this one and generated the regression test YAML files and Govspeak artefacts. I then amended that branch and removed the commits that split the Smart Answer to use one question per page. I ran the regression tests again and confirmed that there were no changes to the outcomes. The only changes were to the question pages, which was to be expected given that I was comparing multiple questions per page to single question per page.

I'm going to get this merged so that it can be deployed on Monday.

PR #1978, opened 2015-09-29
Implement QuestionPresenter#post_body

chrisroos:
This is in preparation for converting Smartdown Smart Answers to Ruby.

Smartdown includes the ability to display a `post_body` (rendered as Govspeak)
after the form on a question page. This is displayed by the
"_current_question.html.erb" partial.

This commit adds support for `post_body` to Ruby Smart Answers.

I'm not expecting any regression test artefacts to change because none of the Ruby Smart Answers have a `post_body` in the YAML files.

### DIFF
--- a/app/presenters/question_presenter.rb
+++ b/app/presenters/question_presenter.rb
@@ -52,7 +52,7 @@ class QuestionPresenter < NodePresenter
   end
 
   def post_body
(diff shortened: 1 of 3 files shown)


floehopper:
LGTM :+1:

PR #1979, opened 2015-09-30
Avoid potential bug in StatutorySickPayCalculator

floehopper:
This line in the `#init_payable_days` method doesn't accurately reflect the
intent expressed in the comment in the line above it, presumably due to a
misunderstanding about how references to objects work in Ruby.

`@normal_workdays_missed` is an `Array` of `Date` objects. It looks as if the
intention of assigning the instance variable to a local variable was to avoid
the calls to `Array#shift` mutating the underlying `Array`. However, the
local variable still referenced the same `Array` and the underlying `Array`
**was** in fact mutated.

This was not actually causing a problem, because the only other use of the
the instance variable was in the assignment of `@normal_workdays` in the
calculator constructor. Since this happens **before** the `Array` is mutated,
no harm is done. However, it seems sensible to fix this in case someone makes
use of the instance variable somewhere else.

I have fixed the potential problem by using `Object#dup` to make a copy of the
underlying `Array`.

There are lots of other improvements that could be made to this code, but we
have other priorities at the moment.

### DIFF
--- a/lib/smart_answer/calculators/statutory_sick_pay_calculator.rb
+++ b/lib/smart_answer/calculators/statutory_sick_pay_calculator.rb
@@ -139,7 +139,7 @@ module SmartAnswer::Calculators
 
     def init_payable_days
       # copy not to modify the instance variable we need to keep
-      payable_days_temp = @normal_workdays_missed
+      payable_days_temp = @normal_workdays_missed.dup
       ## 1. remove up to 3 first dates from the array if there are waiting days in this period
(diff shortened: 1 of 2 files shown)


floehopper:
I've just amended the commit to include updated regression test checksums and force-pushed. Hopefully that will fix the Jenkins build.

chrisroos:
Looks good to me.

floehopper:

PR #1980, opened 2015-09-30
Allow Smartdown flows to be rendered as text

chrisroos:
This allows us to render Smartdown landing pages and outcomes as text. This
mirrors the functionality of Smart Answers and allows us to generate regression
tests for Smartdown flows.

* Add the new SmartAnswersControllerForSmartdownFlowsTest test class and copied
  the three main tests from the "format=txt" context from the
  SmartAnswersControllerTest test.

* Add a new Smartdown test fixture named "smartdown-example". I started out by
  re-using one of the existing fixtures but found that the assertions in the
  "render outcome node as text" test weren't as descriptive as I wanted.

* Modify `SmartdownAdapter::NodePresenter#post_body` to accept the `html`
  parameter and mirror the behaviour of `#body` and `#next_steps`.

* Modify `SmartdownAdapter::Presenter#render_txt?` to mirror the behaviour of
  `SmartAnswerPresenter#render_txt?`.

### DIFF
--- a/lib/smartdown_adapter/node_presenter.rb
+++ b/lib/smartdown_adapter/node_presenter.rb
@@ -12,8 +12,8 @@ module SmartdownAdapter
       render(@smartdown_node.body, html: html)
     end
 
-    def post_body
-      @smartdown_node.post_body && markdown_to_html(@smartdown_node.post_body)
+    def post_body(html: true)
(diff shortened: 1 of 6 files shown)


floehopper:
LGTM :+1:

PR #1981, opened 2015-10-01
Freeze time in am_i_getting_minimum_wage_test.rb

chrisroos:
Two tests in this file started failing today (Thurs 1 Oct 2015):

1. "when checking current pay answered 'apprentice over 19 on second year or
onwards' to 'are you an apprentice?' treat the user as a non apprentice' should
show current minimum wage rate"

2. "when checking current pay answered 'no' to 'are you an apprentice?' 25 year
old should make outcome calculations"

This is because the new rates defined in minimum_wage_data.yml come into effect
today.

Freezing the date to 1 Jan 2015 allows us to continue asserting against the pre
1 Oct 2015 rates. An alternative solution would've been to update the tests to
assert against the new rates.

Neither of these feel ideal but I think it's probably OK for now.


floehopper:
LGTM :+1:

PR #1969, opened 2015-09-24
Overseas passports changes

tadast:
Change a bunch for addresses, a courier fee in Tajikistan and move Kyrgyzstan from Visa Application Centre to a country where the process is handled by the embassy.

Currently in fact-check

PR #1982, opened 2015-10-01
Overseas passports changes - Kyrgyzstan, Tajikistan, Turkmenistan, Uzbekistan & Venezuela

floehopper:
This supersedes #1969.

* The UK Visa Application Centre addresses have changed for:
  * Tajikistan,
  * Turkmenistan,
  * Uzbekistan
  * Venezuela
* The courier fee for passport application in Tajikistan has changed.
* In Kyrgyzstan, the embassy is now handling passport applications instead of the UK Visa Application Centre.

The branch in #1969 went out for fact-check and the branch is this PR addresses the two bits of feedback from that process via Abigail Waraker:

> 1. Kyrgyzstan – the “Making your application” section, which should be above the address, is missing – on all application types.
>
> 2. Tajikistan – the new email address isn’t showing as a link, appears to have code (an `<` and `>` either side of the address).

No. 2 was an easy fix - there was a space between the email address and the trailing `>`, but no. 1 was a lot harder. In the end we've gone with a solution which mirrors the "configuration" for Burundi. This may not be correct, but it gives the outcomes which Abigail was expecting.

You will notice a few commits where I temporarily add extra regression test coverage in order to see the diffs. Hopefully they make it clearer what has changed.


chrisroos:
Looks good to me.

PR #1986, opened 2015-10-01
Add regression tests for pay leave for parents

chrisroos:
Now that pay-leave-for-parents has been converted to single-question-per-page we're able to generate regression tests. These tests should give us confidence that we haven't broken anything when we convert from Smartdown to Ruby Smart Answers.

I'm particularly interested in feedback on the first (currently marked as "TEMP") commit. We need to preload the Smartdown flow to avoid the regression tests taking about 20 seconds each. However, we don't always want to preload as that causes the non-Smartdown tests to take longer than necessary. I'm kinda happy with the hacky approach but I think the bit that's missing is resetting the `preload_flows` option in some kind of teardown hook. I haven't looked at whether that might be possible yet.


chrisroos:
This is mainly for discussion at the moment so I'm going to add the "Don't merge" label.

floehopper:
Other than my comments, this looks good to me. :+1:

chrisroos:
I've updated this branch, rebased on master and force pushed.

I've added a number of additional responses to increase the code coverage of this Smart Answer. I'm confident that we're now reaching all the conditional branches in the outcomes/snippets.

I'm not sure about the changes in the last commit (titled, "Only run pay-leave-for-parents tests when required") but I'm hoping it's good enough for now.

floehopper:
Other than my most recent comment, this looks good to me.

chrisroos:
I've rebased on master and force pushed. Merging to master now.

PR #1994, opened 2015-10-12
Add instructions for how to apply for State Pension Top-up

floehopper:
The State Pension Top-up scheme is launched today (12 Oct).

Users will be able to apply from this date, where previously they were only able
to register interest.


chrisroos:
Is this waiting for updated regression test artefacts and checksum data?

floehopper:
@chrisroos: Oops - yes. I'll do it shortly.

floehopper:
I've updated the regression test artefacts & checksums, rebased against master and force-pushed in preparation for merging.

PR #1987, opened 2015-10-02
Fixed 'claim your state pension' link.

keithiopia:
https://govuk.zendesk.com/agent/tickets/1157237

Also removed a couple of bits of 'hover' markdown.

PR #1996, opened 2015-10-12
Fixed 'claim your state pension' link.

floehopper:
Supersedes #1987.

https://govuk.zendesk.com/agent/tickets/1157237

Also removed a couple of bits of 'hover' markdown.

Note from @floehopper:

* Regression test artefacts updated
* Regression test checksums updated
* Commit amended


floehopper:

PR #1998, opened 2015-10-12
Statutory sick pay calculator bugfix

floehopper:
Supersedes #1967.

Currently when calculating sickness duration the calculator does not take into account employee's usual work days, it simply subtracts the dates, which is incorrect. In order to fix this issue we first need to move the question that collects information about employees working days pattern earlier in the process.

Unfortunately this used to be the last question of the flow that carried all the logic to decide the outcome node. 

**Before**

calculate-statutory-sick-pay-before-question-move

**After**

calculate-statutory-sick-pay-after-question-move

Removing this question from it's position means this logic has to be used in three different questions leading to those outcomes. This has been achieved by extracting the decision logic into an inner class, `OutcomeDecision` within `SmartAnswer::CalculateStatutorySickPayFlow`.

Having moved the question, we then need to take the user's working pattern into account when calculating the number of days worked in the linked sickness period. The last few commits are an attempt at doing that.


floehopper:
I've force-pushed up my latest work on this. I still have some uncertainty about the effect of the changes I've made and I've asked @lutgendorff for help by email. I'm going to add the "before" & "after" visualisations of the flow which I've captured to the PR description.

floehopper:
Superseded by #2031.

PR #1999, opened 2015-10-12
Make test definitions consistent

floehopper:
We're using `test 'name' do ... end` in all other tests, so let's make these ones consistent.

/cc @tijmen


chrisroos:
It looks like the majority of our tests use `should 'name' do ... end`. Did you consider converting to that style instead? Irrespective, I think this is still an improvement as it doesn't introduce yet another style of test method.

floehopper:
Oops. I forgot to search for that style! I'll merge this anyway and we can address the wider problem separately. Thanks.

floehopper:

PR #2001, opened 2015-10-12
Add documentation about deploying

chrisroos:
This is a first stab at documenting the deployment process.

I've made a slight change to the process while writing this document. We would previously add details of the changes to the calendar event itself but it seems more appropriate to use GitHub's releases functionality.


floehopper:
Looks great to me! :+1: :+1:

chrisroos:
Rebased and force pushed ready to merge.

PR #2002, opened 2015-10-13
WIP: Define permitted next nodes for all flows

chrisroos:
Trello card - https://trello.com/c/jtjRNOW9/97-update-all-flows-to-use-permitted-next-nodes

I've updated `Question::Base#next_node_for` to raise an exception if the `next_node` isn't found in the list of `permitted_next_nodes`.

I've updated the first three flows to use `permitted_next_nodes` where required to get an idea of how they're going to look.

__NOTE.__ Using `next_node :symbol` or `next_node_if` automatically add the nodes to `permitted_next_nodes` so I've only updated the flows/questions that use `next_node` with a block.

The advantage of using `permitted_next_nodes` is that it allows us to visualise the flows.

## Before

permitted-next-nodes-before

## After

permitted-next-nodes-after


floehopper:
Other than my comments above, this looks good to me. I agree that initially we could just update the flows which use `next_node` with a block.

chrisroos:
@floehopper: Can you take another look at this pull request, please? If you think the result is OK then I plan to rewrite history and open another pull result. Here's the plan:

* Add support for specifying permitted next nodes through the `permitted` option to `next_node`
* Define permitted next nodes for each flow, updating the checksum data each time
* Fail fast if `next_node` is given a block but the `permitted` option isn't set
* Fail fast in `next_node_for` if the next node doesn't exist in the array of `@permitted_next_nodes`

I think this ordering of commits should mean that all the tests pass each time.

floehopper:
Minor: I might be inclined to combine the checksum update commits with their respective commit which changes the flow. I think separating changes to the test artefacts into separate commits can be useful, because there are often a lot of files changing, but I'm not sure it's justified in the case of just a checksum change.

floehopper:
Very minor: I might be inclined to incorporate the Rubocop changes in the commits which triggered them.

floehopper:
Other than my comments, this LGTM :+1:

chrisroos:
> I might be inclined to combine the checksum update commits with their respective commit which changes the flow.

Yup - I'm planning to do that in the new branch.

chrisroos:
> I might be inclined to incorporate the Rubocop changes in the commits which triggered them.

Good idea.

chrisroos:
Closing in favour of PR #2010.

PR #2003, opened 2015-10-13
Make indentation consistent in i18n YAML files

floehopper:
This file was using double the number of spaces compared to other YAML files.


chrisroos:
Looks good to me.

floehopper:

PR #2005, opened 2015-10-14
Always use explicit question specific options in i18n YAML

floehopper:
Previously if an option for a question was not specified at the question-level in the YAML file, the code would first fallback to flow-level options and, failing that, to the string version of the option key.

The first fallback seems to have been used to DRY up the YAML files, but I think the way it works is a bit surprising and definitely confusing in places e.g. when some options are defined at the question level, but others are defined at the flow level. The reduction in duplication is minimal and I think there are probably better ways to DRY up the code (if it's even a good idea).

The second fallback is hardly used and strikes me as being a bit dangerous. Where it's done intentionally, it leaks presentation details into non-presentation code. And where it's done accidentally (i.e. because someone forgot to define the option text), the page will render without error, but probably with unexpected option text.

I seems a lot better and simpler to require that all question options are defined explicitly at the question level and if one is not then to fail fast with an exception. This PR makes that change and updates all the question option definitions accordingly.

Another advantage of this change is that it makes the content for each question more self-contained, which should make it easier if we decide to split the content for each question into its own file.

Although this includes a lot of commits, they are all small and many of the are very similar.

### DIFF
--- a/lib/smart_answer/i18n_renderer.rb
+++ b/lib/smart_answer/i18n_renderer.rb
@@ -15,16 +15,13 @@ module SmartAnswer
     end
 
     def translate_option(option)
-      translate!("options.#{option}") || begin
-        I18n.translate!("#{@i18n_prefix}.options.#{option}", @state.to_hash)
-      rescue I18n::MissingTranslationData
-        option
-      end
+      translate!("options.#{option}", rescue_exception: false)
     end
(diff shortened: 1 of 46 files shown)


chrisroos:
Looks good to me.

floehopper:

PR #2007, opened 2015-10-14
Add docs for CI and deploying to preview

floehopper:


doc/continuous-integration.md
```diff
@@ -0,0 +1,33 @@
+# Continuous integration
+
+## Main
+
+The [govuk_smart_answers CI instance](https://ci-new.alphagov.co.uk/job/govuk_smart_answers/) is triggered by new commits on the `master` branch of the [GitHub repo](https://github.com/alphagov/smart-answers). It runs the [`jenkins.sh`](https://github.com/alphagov/smart-answers/blob/master/jenkins.sh) script with the `RUN_REGRESSION_TESTS` not set. This script does the following:
+
+* Checks out the lastest [govuk-content-schemas](https://github.com/alphagov/govuk-content-schemas) which are used in a few of the tests
+* Installs the bundled gems
+* Runs govuk-lint-ruby (see [Rubocop docs](doc/rubocop.md))
```
chrisroos: I don't think this link should include "doc/" as this continuous-integration.md file is already in the doc folder.

doc/continuous-integration.md
```diff
@@ -0,0 +1,33 @@
...
+* Runs the `test` rake task with `TEST_COVERAGE` enabled
+* Pre-compiles assets
+
+If any of these steps fails, the build fails-fast.
+
+## Branches
+
+The [govuk_smartanswers_branches CI instance](https://ci-new.alphagov.co.uk/job/govuk_smartanswers_branches/) is triggered by new commits on other branches of the [GitHub repo](https://github.com/alphagov/smart-answers). It runs the [`jenkins_branches.sh`](https://github.com/alphagov/smart-answers/blob/master/jenkins_branches.sh) script with the `RUN_REGRESSION_TESTS` not set.
+
+This script does the same as the [main CI instance](#main), but first it tries to merge `master` into the relevant branch in order to work out whether it will merge without conflicts. This merge is always aborted, but if conflicts are detected the script fails fast at this point.
```
chrisroos: I'm not sure this is quite right.

1. It's the jenkins.sh script that merges master into the branch, which just happens to be a no-op if the branch is master.

2. I don't think the merge is always aborted, and I think this is why we've ended up with Jenkins merge commits in our release tags.

floehopper: 1. I meant "the script" in the sense of the combined behaviour of running the `jenkins_branches.sh` script, but I see that that's not obvious.

2. Hmm. I think the command `git merge --abort` is always executed (whether or not `git merge --no-commit origin/master` succeeds or fails), but perhaps the abort has no effect if the merge succeeds. It makes me wonder whether the intention was that the merge should be reverted!

I'll update the docs based on your feedback. Thanks.

chrisroos: 2. I _think_ (but I'm really not sure) that `git merge --abort` will only run if `git merge --no-commit origin/master` fails. If that's the case then I presume the intention is to run the tests as though the branch was already in master, which seems to make sense to me.

floehopper: :+1:

doc/continuous-integration.md
floehopper: Addressed in 025c35b476efb56ceb0c7004396997d15066df68.

doc/continuous-integration.md
floehopper: Addressed in c55aa9807774df8c7e15d413f896c34e16685b0a.

floehopper:
I've just rebased this against master and force-pushed.

floehopper:
I've just incorporated the fixup commits and force-pushed in preparation for merging.

PR #2009, opened 2015-10-14
Fix missing question option translations

floehopper:
These changes should've been made in #2005. They should fix the [broken regression test build](https://ci-new.alphagov.co.uk/job/govuk_smart_answers_regressions/1187).


chrisroos:
Looks good to me.

PR #2010, opened 2015-10-14
Specify permitted next nodes for all flows

chrisroos:
This supersedes PR #2002.

Trello card - https://trello.com/c/jtjRNOW9/97-update-all-flows-to-use-permitted-next-nodes

Based on feedback from @floehopper, I've added the `permitted` named option to `Question::Base#next_node`.

I've updated all flows that use `next_node` with a block to additionally supply the permitted next nodes.

I've updated `Question::Base#next_node` to raise an exception if it's called with a block but the `permitted` next nodes aren't specified.

I've updated `Question::Base#next_node_for` to raise an exception if the `next_node` isn't found in the list of `permitted_next_nodes`.

__NOTE.__ Using `next_node :symbol` or `next_node_if` automatically add the nodes to `permitted_next_nodes` so I've only updated the flows/questions that use `next_node` with a block.

The advantage of using `permitted_next_nodes` is that it allows us to visualise the flows.

## Before

permitted-next-nodes-before

## After

permitted-next-nodes-after


floehopper:
Looks good to me. Maybe we should run all the regression tests before deploying these changes.

chrisroos:
> Maybe we should run all the regression tests before deploying these changes.

Good shout! This caught some problems in my changes to benefit-cap-calculator. I've fixed the changes and force pushed.

chrisroos:
The regression tests have all passed locally so I'm going to get this merged to master!

PR #2011, opened 2015-10-15
Update regression test input YAML to current YAML version

floehopper:
The version of YAML has changed since these files were generated and the
newer version seems to generate slightly different YAML output e.g. it uses
single-quotes instead of double-quotes.

Because of this, when we make a change and regenerate these files, it's
difficult to see the woods for the trees. By loading and saving all the YAML
files in this PR, I hope that it will be easier to see what's changed in
later commits.


chrisroos:
Loading and saving all the files has removed some of the comments that were added to describe why certain values were chosen:

* test/data/estimate-self-assessment-penalties-questions-and-responses.yml
* test/data/help-if-you-are-arrested-abroad-questions-and-responses.yml
* test/data/landlord-immigration-check-questions-and-responses.yml
* test/data/legalisation-document-checker-questions-and-responses.yml
* test/data/maternity-paternity-calculator-questions-and-responses.yml
* test/data/overseas-passports-questions-and-responses.yml
* test/data/register-a-birth-questions-and-responses.yml
* test/data/register-a-death-questions-and-responses.yml
* test/data/simplified-expenses-checker-questions-and-responses.yml
* test/data/state-pension-topup-questions-and-responses.yml
* test/data/uk-benefits-abroad-questions-and-responses.yml

Do you think we should try to retain those?

floehopper:
Oops! Yes. I'll sort that now. Well spotted!

floehopper:
I've updated the middle "Update YAML files using current YAML version" commit to include manually re-adding the comments and updated the commit message to explain how I did it. I've then force-pushed.

chrisroos:
Looks good to me, although presumably you'll need to regenerate all the checksum data too?

floehopper:
Yes, I'll update checksums now and get the PR merged. Thanks.

floehopper:
I've just added a commit which updates the checksums and force-pushed in preparation for merging.

floehopper:

PR #2016, opened 2015-10-15
Updates Rails to latest release 4.2.4

erik-eide:
Bumps version from 4.2.2 and includes changes from 4.2.3, but no CVE specific 
fixes so no real urgency?

Full diff at (https://github.com/rails/rails/compare/v4.2.2...v4.2.4)


floehopper:
I think the release notes/CHANGELOGs are more useful than the full diff:

* http://weblog.rubyonrails.org/2015/6/26/Rails-4-2-3-and-4-1-12-have-been-released/
* http://weblog.rubyonrails.org/2015/8/24/Rails-4-2-4-and-4-1-13-have-been-released/

While the fact that all the tests pass should give us a lot of confidence in this change, since it's not much more effort, I'd be inclined to do this one patch version at a time, ideally deploying each of them separately, but at least running all the regression tests in each case.

I haven't trawled through the changes in details, but these ones looks as if they might be of some relevance: https://github.com/rails/rails/blob/v4.2.3/actionview/CHANGELOG.md#rails-423-june-25-2015

erik-eide:
Ok cool, I can open a separate 4.2.3 pull request.

All the test cases passed with ```rake test```, so not sure what additional tests we run to check a release?

floehopper:
Sounds good. Thanks.

The regression tests don't typically run as part of the rake `test` task and hence they don't run as part of the CI build for branches/PRs. See [this documentation](https://github.com/alphagov/smart-answers/blob/master/doc/regression-tests.md) for an explanation.

You _could_ run all the regression tests locally using `RUN_REGRESSION_TESTS=true ruby test/regression/smart_answers_regression_test.rb`, but that will take a long time and tie up your machine's CPU. Or you could merge the changes into `master` and wait for the [regressions CI instance](https://github.com/alphagov/smart-answers/blob/master/doc/continuous-integration.md) to pass.

PR #2018, opened 2015-10-15
Rename pay-leave-for-parents question pages

chrisroos:
We're planning to migrate pay-leave-for-parents from Smartdown to Ruby in the near future.

We can use the `convert-smartdown-flow-to-smart-answer-flow.rb` to automatically convert some of this Smart Answer, but it won't work for some of the more complicated conditionals. We'll need to convert those by hand and I think that having a one-to-one mapping between question page name (i.e. filename) and question will make that process easier.


floehopper:
LGTM :+1:

chrisroos:

PR #2019, opened 2015-10-16
Update relative links in docs

erik-eide:
Fixes a set of broken links and anchors in the markdown docs, and removes a couple of unnecessary elements in relative paths linking various documents and files.


chrisroos:
Looks good to me.

PR #2020, opened 2015-10-18
Add missing question option translations for PartYearProfitTaxCreditsFlow

floehopper:
This flow was relying on the fallback to the option key which was removed
in #2005. The following exception has just occurred in production [1]:

    I18n::MissingTranslationData: translation missing:
      en-GB.flow.part-year-profit-tax-credits.have_you_stopped_trading?.options.yes

This commit adds unit tests which would've picked up the problem and adds
the missing translations.

[1]: https://errbit.publishing.service.gov.uk/apps/533c35ae0da1159384044f5f/problems/56233f4b6578634c6c971600


tadast:
LGTM :+1:

PR #2021, opened 2015-10-19
Bring docs into line with most recent practice

floehopper:
This is the format @chrisroos used in his most recent release and the one I
copied in the most recent release.


chrisroos:
Looks good to me.

floehopper:

PR #2023, opened 2015-10-19
Move outcome templates into their own sub-directory

floehopper:
## Motivation

We're planning on storing question page content in the flow directory alongside
landing page and outcome page content. It's already a bit confusing that the
landing page content is stored alongside the question page content and it will
become more so when question page content is there too.

So this pull request moves all outcome templates into an `outcomes` sub-directory c.f. smartdown flows.

## User-visible changes

None.

## Technical details

This was all pretty straightforward, if a little tedious!

The first commit breaks the regression tests for all flows by requiring the outcome templates to be in an `outcomes` sub-directory. Each subsequent commit fixes the regression test for an individual flow, updating the checksums accordingly. No changes to the regression test inputs or artefacts were required.

The sub-directories in `marriage-abroad` were moved into the new `outcomes` sub-directory.

The rendering of shared & data partials in various flows uses relative paths outside of the flow directory and so with the outcome templates another level down, these had to be changed from e.g.:

```ruby
<%= render partial: '../shared/partial-name.govspeak.erb' -%>
``` 

to:

```ruby
<%= render partial: '../../shared/partial-name.govspeak.erb' -%>
```

I did think about allowing both the flow directory **and** the new `outcomes` sub-directory, but it made the code quite a bit more complicated and in the end I decided that it's better to be stricter.


erik-eide:
Seems to be a few places where we prefix paths to shared partials with a '../..' path, which is the kind of thing I forget to do and maybe we can do away with by appending it to the Rails view_path?

Otherwise, looks good to me :+1:

floehopper:
I think adding the `shared` and `data_partials` paths to the list of view paths set on the instance of `ActionView::Base` instance used in `SmartAnswer::ErbRenderer` is a good idea.

However, I think it's slightly orthogonal to this PR, although I accept this PR does make things _slightly_ worse. I'll address this concern in a separate PR - I've added [a Trello card](https://trello.com/c/nPxhS9O1) to remind me.

I suspect we might also be able to combine the `shared` and `data_partial` directories since I'm not sure there's much (if any) difference between them now.

floehopper:
I've rebased this branch against master and force-pushed in preparation for merging.

PR #2025, opened 2015-10-20
Move global variables to Services module

floehopper:
## User-visible changes

None

## Motivation

The `Services` module was added relatively recently in commit 34f8f7f4b169c8f10471edc630c7c1088cc205aa. This introduced a different way to "lookup" clients to external APIs. I'm not sure whether there's an official GDS way to do this, but this new way seems a bit better than using global variables. The main thing is that we should only do this *one* way which is what this PR addresses.

## Technical details

Hopefully this is fairly obvious from the diffs.


chrisroos:
Looks good to me.

floehopper:

PR #2027, opened 2015-10-20
Add step to apply for an overseas passport for child in Afghanistan

er-ke-de:
For further details see:
  https://www.pivotaltracker.com/n/projects/1270592/stories/105008654

## Expected changes

* [Example URL](https://www.gov.uk/overseas-passports/y/afghanistan/applying/child/afghanistan)
  * Under How to Apply, a step # 5 with a link to a parent contact details form appears

screen shot 2015-10-20 at 4 11 26 pm

### DIFF
--- a/lib/smart_answer_flows/overseas-passports/outcomes/_how_to_apply.govspeak.erb
+++ b/lib/smart_answer_flows/overseas-passports/outcomes/_how_to_apply.govspeak.erb
@@ -6,7 +6,7 @@
   3. Include 2 identical new photos of you (or your child, if it’s a child passport application). [Follow the rules about passport photos](/photos-for-passports) or your application may be delayed.
   4. Check the guidance notes to find out which supporting documents you must send with your application. You must get documents that aren’t in English - including documents showing an address - fully translated by a professional translator.
 
-  <% if %w(bangladesh india pakistan).include?(current_location) && child_or_adult == 'child' && application_action == 'applying' %>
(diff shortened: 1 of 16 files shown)


chrisroos:
Do you need to update the regression test artefacts for this change too, @erik-eide?

erik-eide:
Yep, I just added a new commit with a response that regresses through the new path

chrisroos:
I hope you don't mind, @erik-eide, but I've had a go at illustrating what I meant when I talked about updating the regression tests to see the effect of this change.

See PR #2028 for a worked example of how we can temporarily update the regression tests to test the effect of a change.  It turns out to be overkill in this instance but the technique can be useful when we're making changes that can affect multiple outcomes; particularly when a change inadvertently affects an outcome that we don't expect it to.

If I was going to suggest a small change to this PR then I think it'd be good to squash the two commits. I don't think there's any real value in adding "afghanistan" to the questions-and-responses file in the first commit but then only generating the regression tests in the second.

All that aside, the actual change in this commit looks good and I apologise that the regression tests have made such a small change feel so painful!

er-ke-de:
I think normally, I would of liked of to have squashed the commits for the reason you had given, but given we had a comment on the PR about them not being updated, I left them as 2 intentionally to show in the history.

Have squashed them now, thanks for the examples and docs on getting it right!

floehopper:
Looks good to me :+1:

PR #2028, opened 2015-10-21
PR 2027 illustration

chrisroos:
This is purely for the purposes of illustration and shouldn't be merged.

I didn't do a very good job of explaining how we could use the regression tests to test the effect of a change and thought this worked example might help.


chrisroos:
I'm closing this as we don't need it open for it to serve its purpose.

PR #2029, opened 2015-10-21
Avoid displaying question page titles for Smartdown flows

chrisroos:
I'm hoping this'll make it easier to convert pay-leave-for-parents from Smartdown to a Ruby Smart Answer.

By not displaying question page titles for Smartdown flows (pay-leave-for-parents is the only Smartdown flow left) I hope we have a set of regression test artefacts that remain constant after the conversion to Ruby.

I think it's OK to remove these question page titles as I don't think they've been adding much value since we converted the last remaining Smartdown flow with multiple questions per page to use single questions per page.

### DIFF
--- a/lib/smartdown_adapter/presenter.rb
+++ b/lib/smartdown_adapter/presenter.rb
@@ -36,7 +36,6 @@ module SmartdownAdapter
     end
 
     def page_title
-      current_node.title
     end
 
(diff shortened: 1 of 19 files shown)


floehopper:
LGTM :+1:

chrisroos:

PR #2031, opened 2015-10-21
Apply working pattern to linked period of sickness in calculate-statuatory-sick-pay

floehopper:
Supersedes #1998.

## Motivation

The calculator is not applying the employee's working pattern to the optional linked sickness period. The working pattern is the days of the week that the employee works e.g. Monday to Friday.

In order to fix this problem we first have to move the question which asks for the employee's usual work days to an earlier point in the process.

After moving the question, we then need to take the working pattern into account when calculating the number of valid sick days in the linked period of sickness i.e. ignoring days on which the employee would not have been working.

### Minor changes

The following came up during the fact-checking process:-

* Extra validation on the "linked sickness end date" question
* Remove the word 'Ordinary' from 'Ordinary Statutory Paternity Pay'

## Technical details

The "usual work days" question is the last question and incorporates all the logic to route the user to one of four outcomes. Unfortunately, moving the "usual work days" question means that there will now be three questions each of which could go to one of four outcomes. Furthermore the outcome routing logic is all contained within the "usual work days" question.

So before moving the question, we extract the logic into a nested `OutcomeDecision` class so we can more easily reuse it across the three new "last" questions.

**Note**: This highlights the fact that the routing logic should probably be modelled as being *between* nodes and not part of a node. But that's a job for another day.

Unfortunately, a number of the changes in this PR require big changes to the huge number of regression test artefacts. The majority of the changes should be simple path renames due to changes in the sequence of responses, but in many cases it looks as if `git`'s rename heuristics aren't correctly identifying all the renames. Hence it's going to be very difficult for someone to review those commits.

To give myself some reassurance, I've generated the visualisation of the flow before and after the change in the order of the questions:

**Before**

calculate-statutory-sick-pay-before-question-move

**After**

calculate-statutory-sick-pay-after-question-move

**Note**: The significant renames mentioned above also correspond to changes in public URLs. Ideally we'd redirect the old URLs to the new ones, but I think this would be a big undertaking and I'm not sure whether there's a precedent for doing for changes like this in Smart Answers.

### Take working pattern into account when calculating days of sick pay paid for linked period of sickness

This is the main change and I believe was the original motivation behind the ZenDesk ticket. In order to do this we had to move the 'Which days of the week do they usually work?' question from being the last question to the 4th question. And then we had to fix the calculation.

Previously the calculator had incorrectly assumed that the employee was paid sick pay for every calendar day of the linked period of sickness. The new version of the calculator applies the employee's working pattern to the linked period of sickness to work out how many days of sick pay were paid.

In the following scenario the linked period of sickness is two days (Saturday & Sunday) and the employee works Wednesdays & Saturdays. Previously it was assumed that sick pay was paid for both these days; whereas now only the Saturday is included.

Thus previously the linked period of sickness accounted for 2 of the 3 waiting days (leaving 1 waiting day); whereas now it only accounts for 1 of the 3 waiting days (leaving 2 waiting days). Therefore where previously the 9 days was reduced by 1 day to 8 days, now the 9 days is reduced by 2 days to 7 days; and the SSP amount for the week ending 4th July 2015 is reduced by half accordingly.

* Old: https://www.gov.uk/calculate-statutory-sick-pay/y/ordinary_statutory_paternity_pay/yes/no/2015-06-27/2015-07-25/yes/2015-05-09/2015-05-10/eight_weeks_more/monthly/2015-04-24/2015-02-27/1600.0/3,6 # => entitled to 8 days of sick pay out of 9
* New: https://www.gov.uk/calculate-statutory-sick-pay/y/statutory_paternity_pay/yes/no/3,6/2015-06-27/2015-07-25/yes/2015-05-09/2015-05-10/eight_weeks_more/monthly/2015-04-24/2015-02-27/1600.0 # => entitled to 7 days sick pay out of 9

### Prevent linked period of sickness overlapping the current period of sickness

During the fact-check process, you pointed out that the linked period of sickness should never overlap the current period of sickness, because then the two periods ought to be considered as a single period of sickness.

We've introduced new validation on the 'Enter the end date for this linked period of sickness?' question such that the date entered must be before the beginning of the current period of sickness. In the following examples, the old version shows a page where an invalid date has been accepted; whereas the new example shows the new error message:

* Old: https://www.gov.uk/calculate-statutory-sick-pay/y/ordinary_statutory_paternity_pay/yes/no/2015-01-07/2015-05-03/yes/2015-01-01/2015-01-07 # => 7 January 2015 appears in previous answers against 'Enter the end date for this linked period of sickness' question
* New: https://www.gov.uk/calculate-statutory-sick-pay/y/statutory_paternity_pay/yes/no/1,2,3,4/2015-01-07/2015-05-03/yes/2015-01-01/2015-01-07 # => Date not accepted and error message displayed: 'You need to enter a date before the start of the current period of sickness or it isn't a separate linked period of sickness'

### Remove the word 'Ordinary' from 'Ordinary Statutory Paternity Pay'

During the fact-checking process, you pointed out that the 'Ordinary' prefix is no longer used. I've removed it from everywhere it occurs in this calculator:

#### In the 'Is your employee getting any of the following?' question

* Old: https://www.gov.uk/calculate-statutory-sick-pay/y
* New: https://www.gov.uk/calculate-statutory-sick-pay/y

#### In the URL when the above option is selected

* Old: https://www.gov.uk/calculate-statutory-sick-pay/y/ordinary_statutory_paternity_pay
* New: https://www.gov.uk/calculate-statutory-sick-pay/y/statutory_paternity_pay

#### On the outcome page when the above options is selected and employee is entitled to sick pay

In the 'What you need to know' section:

* Old: https://www.gov.uk/calculate-statutory-sick-pay/y/additional_statutory_paternity_pay,ordinary_statutory_paternity_pay/no/no/2013-04-02/2013-04-10/yes/2013-01-01/2013-02-15/before_payday/fortnightly/2000.0/30/0,6
* New: https://www.gov.uk/calculate-statutory-sick-pay/y/additional_statutory_paternity_pay,statutory_paternity_pay/no/no/0,6/2013-04-02/2013-04-10/yes/2013-01-01/2013-02-15/before_payday/fortnightly/2000.0/30

### No change

Calculations with no linked period of sickness should be unaffected by the changes in the new version.

* Old: https://www.gov.uk/calculate-statutory-sick-pay/y/ordinary_statutory_paternity_pay/yes/no/2013-01-07/2013-05-03/no/eight_weeks_more/irregularly/2012-12-28/2012-10-26/3000.0/1,2,3,4
* New: https://www.gov.uk/calculate-statutory-sick-pay/y/statutory_paternity_pay/yes/no/1,2,3,4/2013-01-07/2013-05-03/no/eight_weeks_more/irregularly/2012-12-28/2012-10-26/3000.0


floehopper:
This PR is still awaiting the final go-ahead from the fact-check process, so I'm going to mark it as being in fact-check and not-mergeable. However, it would be good if someone can review it, because I'm pretty confident it's ready for publication.

chrisroos:
I've spent quite a while reviewing this PR and chatting to @floehopper about it.

Since reviewing it we've come to the conclusion that the `usual_work_days?` question was moved in order to update/fix the calculation in the `must_be_sick_for_4_days` outcome. That fix hasn't happened yet and may now happen as a separate pull request. Presumably that also means that the movement of the question wasn't actually required for the fixes in this PR but I don't think it's worth trying to untangle it.

We also spoke about manually renaming regression test artefacts in order to make it easier to see the important differences in regression test output. I'm not convinced it's worth the effort of retroactively doing that in this PR but it can be a useful technique when a large number of artefacts are changed.

Aside from the comments, as far as I can tell this all looks good to me.

floehopper:
Superseded by #2051.

PR #2034, opened 2015-10-26
Remove flow auditor code

chrisroos:
The `flow_audit` script would analyse a Smart Answer and inspect the blocks passed to `next_node` in an attempt to list possible next nodes for each question. The output could be used to define `permitted_next_nodes` for each question.

PR #2010 added the equivalent of `permitted_next_nodes` to each question and also ensured that we have to specify permitted next nodes in future. As such, I'm less convinced that this `flow_audit` script is useful.


floehopper:
LGTM :+1:

PR #2037, opened 2015-10-26
Update the gem ordering to an alphabetic sort

erik-eide:
Tidies up the Gemfile so finding dependencies is a bit easier


floehopper:
Other than the change to `Gemfile.lock`, this looks good to me.

PR #2039, opened 2015-10-28
Add ability to read the flow name from within a node

erik-eide:
Would allow flow specific logic to be included in otherwise identical shared logic used by multiple flows.


floehopper:
This looks fine to me :+1:

PR #2040, opened 2015-10-28
Add qualifying week to maternity, paternity and adoption

erik-eide:
## Rationale

This change updates the maternity, paternity and adoption smart answer calculators by asking the user if they were, or will be, employed during the qualifying week - rather than the present date when they are completing the questions.

The qualifying week (QW) is, broadly, the 15th week before the expected week of childbirth or adoption. It's worked out by finding the Sunday before the due or match date (or due day itself if it's a Sunday). The QW is 15 Sundays back from there.

For example, if the due date is 1 January 2016, the QW is 13 September.

For further details see:
  https://www.pivotaltracker.com/n/projects/1270592/stories/104083442

## Expected changes

* [Example Maternity URL](https://www.gov.uk/maternity-paternity-calculator/y/maternity/2016-01-01/yes/2015-12-01/yes)
  * Question should be: "Was the employee (or will they be) on your payroll on 13 September 2015?"

screen shot 2015-10-28 at 4 43 54 pm

* [Example Paternity URL](https://www.gov.uk/maternity-paternity-calculator/y/paternity/yes/2016-01-01/2016-01-01/yes/yes/yes)
  * Question should be: "Was the employee (or will they be) on your payroll on 13 September 2015?"

screen shot 2015-10-28 at 4 47 43 pm

* [Example Adoption URL](https://www.gov.uk/maternity-paternity-calculator/y/adoption/yes/2016-01-01/2016-01-01/yes/yes/yes)
  * Question should be: "Was the employee (or will they be) on your payroll on 13 September 2015?"

screen shot 2015-10-28 at 4 48 56 pm

### DIFF
--- a/lib/smart_answer_flows/locales/en/maternity-paternity-calculator.yml
+++ b/lib/smart_answer_flows/locales/en/maternity-paternity-calculator.yml
@@ -30,7 +30,7 @@ en-GB:
           "yes": "Yes"
       ## QM5
       is_the_employee_on_your_payroll?:
-        title: Is the employee on your payroll?
+        title: Was the employee (or will they be) on your payroll on %{qualifying_week_start}?
         options:
           "no": "No"
           "yes": "Yes"
@@ -143,7 +143,7 @@ en-GB:
           "yes": "Yes"
       ## QP6
       employee_on_payroll_paternity?:
-        title: Is the employee on your payroll?
+        title: Was the employee (or will they be) on your payroll on %{qualifying_week_start}?
         options:
           "no": "No"
           "yes": "Yes"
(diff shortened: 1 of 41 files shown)


tadast:
The changes look good to me, however it would be better to have all the rationale behind the change directly in commit notes and the PR description. People looking at git/GH history in future may not have access to pivotal or we may not use pivotal anymore. Also, we're developing in the public by default, so we should make all information available to the public :) 

Having said that, you can still reference pivotal for internal purposes.

erik-eide:
Thanks @tadast - good points, I've update the description with a rationale section outlining the reason for the change.

tadast:
LGTM after amending the commit note :+1:

PR #2041, opened 2015-10-29
Updates date_of_year function to calculate differences between dates

erik-eide:
The date_of_year function allows you to return a date containing the day and month values of the first argument, with the year of the second, which fails when passing a 29th of Feb value for the date and non-leap value for the year.
    
This change simply subtracts the difference in years from the two arguments, so an earlier year will subtract a positive number of years and a later year subtracts a negative number of years.

In cases where both arguments have equal year values, it avoids the subtraction and simply returns the existing date.

Addresses issue #1828 where Feb 29th values in the first argument cause a failure when the second argument has a non-leap year value.

## Expected changes

[Example URL](https://www.gov.uk/calculate-your-holiday-entitlement/y/days-worked-per-week/leaving/2015-05-01/2016-02-29/1)
 * Should no longer result in an application error "Sorry, we are experiencing technical difficulties", and should calculate the holiday entitlement.

screen shot 2015-10-29 at 5 16 53 pm

### DIFF
--- a/lib/smart_answer/calculators/holiday_entitlement.rb
+++ b/lib/smart_answer/calculators/holiday_entitlement.rb
@@ -107,12 +107,10 @@ module SmartAnswer::Calculators
 
     def leave_year_start_end
       if self.leave_year_start_date
-        date_leave_year_start_date = leave_year_start_date
-
-        needs_offset = date_calc >= date_of_year(date_leave_year_start_date, date_calc.year)
+        needs_offset = date_calc >= date_of_year(leave_year_start_date, date_calc.year)
         number_years = date_calc.year - (needs_offset ? 0 : 1)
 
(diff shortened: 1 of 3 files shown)


chrisroos:
Is it worth adding another unit test around `HolidayEntitlement#fraction_of_year` to cover the case where the `start_date` is later than the `leave_year_start_date`? This would complement the test you've added where `start_date` is earlier than the `leave_year_start_date`.

chrisroos:
I also wonder whether it's worth making the `date_of_year` method public so that we can add unit tests to describe the behaviour when the `date` passed is the 29 Feb. What do you think?

Aside from my comments, I think this looks OK.

erik-eide:
@chrisroos I think that would be clearer, I'll give it a go and try cover start before, after and equal to the year value.

PR #2043, opened 2015-11-02
Remove Smartdown code

chrisroos:
We've now converted the final Smartdown flow (pay-leave-for-parents) to Ruby so we can remove a lot of the Smartdown related code.

cc @dsingleton


floehopper:
I've found a couple of other bits of code you might want to consider removing:

* https://github.com/alphagov/smart-answers/blob/master/bin/smartdown
* https://github.com/alphagov/smart-answers/blob/master/app/assets/stylesheets/smart_answers.scss#L63-L119
* https://github.com/alphagov/smart-answers/blob/master/lib/smart_answer/flow.rb#L42-L45 (and related code)

floehopper:
I assume you've intentionally not removed documentation relating to Smartdown. I think we should update the docs to make it clear that Smartdown flows are deprecated, but as long as it's on our list of things to do, it doesn't necessarily need to be part of this PR.

Otherwise this looks good to me. Good work! :smile:

dsingleton:
Taking a look at this now.

dsingleton:
The [`transition` status](https://github.com/alphagov/smart-answers/blob/187e682e2ec2d52bbf3f042574d1eedaf29cfda8/lib/smart_answer/flow.rb#L42-L45) was exclusively for Smart Answer -> Smartdown conversion, and can be removed. Up to you if it's in this PR, or as a follow up.

dsingleton:
Other bits i've spotted:

- There's some CSS that might be smartdown related, [smartdown/multiple-questions CSS](https://github.com/alphagov/smart-answers/blob/187e682e2ec2d52bbf3f042574d1eedaf29cfda8/app/assets/stylesheets/smart_answers.scss#L63-L119) it would be worth investigating (and this might be better for custom to do rather than you)
- There's still [bin/smartdown](https://github.com/alphagov/smart-answers/blob/master/bin/smartdown)

But broadly this looks :+1: I'd be happy merging as is and the other bits in follow up PRs.

dsingleton:
OK, so finished looking. Still happy for this to be merged as is, but it would be good to address the bits @floehopper and I raised - in particular updating/removing documentation.

We should also update the README on the `smartdown` repo itself, to avoid confusion there. Let us know if you don't have access to that repo

chrisroos:
> I've found a couple of other bits of code you might want to consider removing:
> https://github.com/alphagov/smart-answers/blob/master/bin/smartdown

Removed in 0cf3da159e33c091fe0eaedff9ef370f8dddecc3

chrisroos:
> I've found a couple of other bits of code you might want to consider removing:
> https://github.com/alphagov/smart-answers/blob/master/lib/smart_answer/flow.rb#L42-L45 (and related code)

Removed in 793f09128b56e4c110d1de390e893b81c9f66e6c.

chrisroos:
> The transition status was exclusively for Smart Answer -> Smartdown conversion, and can be removed. Up to you if it's in this PR, or as a follow up.

Removed in 793f09128b56e4c110d1de390e893b81c9f66e6c.

chrisroos:
The CSS you've both mentioned as being Smartdown specific actually appears to affect Ruby Smart Answers too. See the screenshots below for before/after removing this CSS:

## Before

before

## After

after

I think I'm going to remove the comment that talks about this being Smartdown specific instead of trying to understand which bits of this CSS I can remove.

chrisroos:
> We should also update the README on the smartdown repo itself, to avoid confusion there. Let us know if you don't have access to that repo.

@dsingleton: Are there any examples of how we've updated similar READMEs in the past? Do we have a standard way of saying a project is no longer under active development/not being used?

chrisroos:
I've chosen not to remove the documentation as part of this PR. I've created a [Trello card to capture that task](https://trello.com/c/jwB5AVmJ/118-remove-documentation-about-smartdown).

chrisroos:
OK, I'm going to rebase this on master and get it merged.

chrisroos:
I'm going to close this pull request and open a new one with the rewritten commits. I wonder whether this will preserve more history than if I force push over this branch.

PR #2047, opened 2015-11-02
Consistently use I18nTestHelper

chrisroos:
This should mean that we're now consistently modifying the `I18n.load_path`.

### DIFF
--- a/test/functional/smart_answers_controller_test.rb
+++ b/test/functional/smart_answers_controller_test.rb
@@ -106,7 +106,7 @@ class SmartAnswersControllerTest < ActionController::TestCase
     end
 
     should "display first question after starting" do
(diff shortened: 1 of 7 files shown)


floehopper:
Other than my minor comments, this looks good to me.

chrisroos:
> Since your new methods #using_additional_translation_file and #use_additional_translation_file no longer have a ! suffix, I'd suggest removing the one on #reset_translation_files!. I'm not sure it's serving any purpose and I find it confusing that this has it, but the others don't.

Good spot. I've made this change in 21e9c415059f29a1ffd4f9fab7a7094e69a8e18c.

chrisroos:
I've going to rebase on master and force push.

PR #2048, opened 2015-11-03
Move smart answer test flows and locales to single directory

chrisroos:
This is based on PR #2047 (i.e. I'm assuming that'll be merged first).

This moves all test Smart Answer flows and locale files to test/fixtures/smart_answer_flows, rather than having them spread across different directories in test/fixtures.

PR #2049, opened 2015-11-03
Fail fast if ERB template is missing

chrisroos:
This is based on PR #2048 (i.e. I'm assuming that'll be merged first).

In general we're trying to make more things fail fast rather than fall back to somewhat surprising defaults.

This change results in a `ActionView::MissingTemplate` exception being raised if we add a Smart Answer without the required coverpage or outcome templates.


floehopper:
Other than my minor comments, this look good to me. :+1:.

chrisroos:
I merged master into this branch in the hope that this PR view would only show me the commits from this branch, and not those that have already been merged to master. Unfortunately that hasn't quite worked out, presumably because the commits in this branch have different SHAs when they were merged to master.

chrisroos:
> It's not obvious to me why you've had to introduce the setup block i.e. the stubbing of SmartAnswer::FlowRegistry.instance. It's a bit of a shame the test has become so "stubby". If nothing else, I think it would be good to explain why it's needed.

The `FlowContentItem#payload` method (used in `ContentItemPublisher#publish`) returns a hash containing the `FlowRegistrationPresenter#title` which is looked up in the cover page ERB template. If we don't stub `SmartAnswer::FlowRegistry.instance` then the `FlowRegistrationPresenter#title` method results in an exception because it tries to find the cover page ERB template for bridge-of-death in lib/smart_answer_flows.

chrisroos:
I've addressed the feedback (thanks, @floehopper), rebased on master and force-pushed to this branch in preparation for merging.

chrisroos:
Actually, I can't merge from this pull request because the base branch isn't master. Opening a new PR instead...

chrisroos:
This has been superseded by PR #2055. I'm going to merge that to master once all the tests pass.

PR #2052, opened 2015-11-04
Remove MultipleChoice#option shortcut

chrisroos:
This removes the ability to call `MultipleChoice#option` with a hash. Passing a hash allowed you to specify the option key and target next node. While occasionally convenient, our main focus at the moment is on reducing the number of ways to do the same thing, and on simplifying the code. Removing this functionality reduces the number of ways in which the next node can be specified.


floehopper:
I understand why you've done it (to keep the changes as small as possible), but I think it's a bit confusing that the keys passed to the `option` method are mostly/all either symbols or double-quoted strings; whereas the values in the `when` statements are single-quoted strings.

floehopper:
Seeing all these case statements with the `permitted_next_nodes` defined separately makes me think that it would be good to introduce a new form of `next_node` which accepted some kind of data structure representing the mapping (e.g. Hash of option keys vs outcome keys) so that we could DRY things up a bit. But we can do that separately. I'm going to add a Trello card for that now.

floehopper:
Other than my minor comments, this LGTM.

chrisroos:
> I understand why you've done it (to keep the changes as small as possible), but I think it's a bit confusing that the keys passed to the option method are mostly/all either symbols or double-quoted strings; whereas the values in the when statements are single-quoted strings.

I agree that it's confusing. I've added the ["Consistently pass the option key to `MultipleChoice#option` Trello card](https://trello.com/c/qFSuTC0m/123-consistently-pass-the-option-key-to-multiplechoice-option) to capture this.

chrisroos:
> Seeing all these case statements with the permitted_next_nodes defined separately makes me think that it would be good to introduce a new form of next_node which accepted some kind of data structure representing the mapping (e.g. Hash of option keys vs outcome keys) so that we could DRY things up a bit. But we can do that separately. I'm going to add a Trello card for that now.

Trello card for reference - https://trello.com/c/tSI0JX4b/122-dry-up-next-node-blocks-with-case-statements

chrisroos:
I've addressed the feedback (thanks for reviewing, @floehopper) and am now going to close this, rebase the branch on master and created a new PR.

chrisroos:
I've opened PR #2056 with the tidied and rebased version of this PR.

PR #2057, opened 2015-11-05
Use Publishing API v2

tijmenb:
The new V2 endpoint:

- works by content_id, which means we can more easily change slugs in the future
- allows us to only send the content, not the links. The links will contain the taggings for this content item, and tagging them separately will enable us to tag from more than one tool. Without this "content-only", this app would overwrite taggings done by other apps.

cc @rboulton

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -49,7 +49,7 @@ end
 if ENV['API_DEV']
   gem 'gds-api-adapters', path: '../gds-api-adapters'
 else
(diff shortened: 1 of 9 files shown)


danielroseman:
v2 `put_content` sends the item as draft only. You need to also call `publish(content_id, update_type)` to actually mark the content as live and sent it to the live content-store,.

tijmenb:
@danielroseman updated PR.

PR #2058, opened 2015-11-05
Add missing calculated value in paternity calculator

erik-eide:
Adds tests to the adoption, maternity and paternity smart answers to ensure that questions with interpolated values are rendered correctly.

Adds a test helper method #question_title reads the value of a question as a String, forcing the interpolation to occur, so it can be asserted with a has_content? matcher in integration tests.

Updates the paternity calculator, adding the missing calculated value that is interpolated into the question title.

## Verify step

[Example URL](http://localhost:3000/maternity-paternity-calculator/y/paternity/no/2015-10-29/2015-10-29/yes/yes/yes/yes/yes/2015-10-29/two_weeks/2015-07-16)
 * Shows a question "Was the employee (or will they be) on your payroll on 12 July 2015?" within the previous answers without breaking the page.

screen shot 2015-11-05 at 4 16 09 pm

### DIFF
--- a/lib/smart_answer_flows/shared_logic/paternity-calculator.rb
+++ b/lib/smart_answer_flows/shared_logic/paternity-calculator.rb
@@ -102,6 +102,10 @@ multiple_choice :employee_responsible_for_upbringing? do
     due_date
   end
 
+  calculate :qualifying_week_start do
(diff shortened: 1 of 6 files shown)


chrisroos:
Looks good to me @erik-eide!

PR #2060, opened 2015-11-09
Inline #presenter_for method

chrisroos:
Since becb787a1247c3192fed391184d93956b047bc7d we've only been dealing with
question nodes in #indexable_content so we can always instantiate a
QuestionPresenter.


floehopper:
LGTM :+1:

PR #2061, opened 2015-11-09
Use next_node blocks in calculate-state-pension

chrisroos:

### DIFF
--- a/lib/smart_answer_flows/calculate-state-pension.rb
+++ b/lib/smart_answer_flows/calculate-state-pension.rb
@@ -40,8 +40,17 @@ module SmartAnswer
         option :male
         option :female
 
-        next_node_if(:dob_age?, variable_matches(:relevant_calculation, "age"))
-        next_node :dob_amount?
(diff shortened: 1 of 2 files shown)


floehopper:
There are quite a few places where you've replaced `define_predicate` blocks with `next_node_calculation` blocks where the code inside the block is calling one (or a combination of) "predicate" method(s) on the calculator. Did you consider inlining the calls to these methods into the `next_node` block? I guess my slight concern is that if we don't get round to doing that, then people might wonder why the code is how it's been left after this PR. But I don't feel strongly about it.

floehopper:
LGTM :+1:

chrisroos:
> Did you consider inlining the calls to these methods into the next_node block?

I did but decided it was outside of the scope of the changes in this PR. I wanted to get rid of the use of `next_node_if` before going on to make any further improvements.

chrisroos:
I've rebased on master and force pushed ready to merge.

PR #2062, opened 2015-11-09
Use next_node blocks in calculate-your-child-maintenance

chrisroos:


floehopper:
I think it would at least be good to explain your choice of `case` or `if/elsif` constructions in the PR description or commit messages.

floehopper:
Other than my very minor comments, this LGTM :+1:

chrisroos:
I've changed the `next_node` blocks for `:gross_income_of_payee?` and `:how_many_nights_children_stay_with_payee?` so that they use a `case` statement instead of `if .. else .. end`. This is more consistent with what I've done elsewhere, as highlighted by @floehopper.

PR #2068, opened 2015-11-11
Refactor calculate-statutory-sick-pay

floehopper:
This PR is an attempt at separating the concerns for `calculate-statutory-sick-pay` as we did in `part-year-profit-tax-credits` as follows:

* The calculator is converted into an `ActiveModel::Model` and instantiated in a `next_node_calculation` in the first question.
* Next node logic is all converted to use `next_node` with a block
* Responses are set as attributes on the calculator in these `next_node` blocks
* All policy & validation logic is moved out of the flow into the calculator
* Only routing logic remains in the flow
* Presentation logic is moved out of the calculator and the flow and into the outcome templates

All these changes should not affect the outwardly observable behaviour as attested by the fact that no changes to the regression test artefacts were needed.

This PR leaves the calculator in a less than ideal state, but given the number of changes, I thought it would be good to get this merged as soon as possible.

### DIFF
--- a/lib/smart_answer/calculators/statutory_sick_pay_calculator.rb
+++ b/lib/smart_answer/calculators/statutory_sick_pay_calculator.rb
@@ -150,6 +150,10 @@ module SmartAnswer
         self.class.lower_earning_limit_on(sick_start_date)
       end
 
+      def pattern_days_total
+        pattern_days * 28
+      end
+
       # define as static so we don't have to instantiate the calculator too early in the flow
       def self.lower_earning_limit_on(date)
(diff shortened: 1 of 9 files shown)


chrisroos:
It looks like there's one remaining use of `save_input_as` in `:has_linked_sickness?` that can be removed.

chrisroos:
This looks great to me. I think the resulting flow is much clearer and the commits should help anyone else wanting to make similar changes to another Smart Answer.

floehopper:
I'll remove the unnecessary `save_input_as` statement in `: has_linked_sickness?` as you suggest.

floehopper:
I've made the changes suggested by @chrisroos, rebased against master, re-run the SSP regression tests and force-pushed in preparation for merging.

PR #2070, opened 2015-11-11
Use next_node blocks in overseas-passports

chrisroos:


floehopper:
Other than my comment about the regression test checksums, this looks good to me.

chrisroos:
I've amended the "Use next_node block in :child_or_adult_passport?" and "Use next_node block in :country_of_birth?" commits based on @floehopper's feedback.

PR #2071, opened 2015-11-11
Spike: Fail fast on missing methods in SmartAnswer::State object

chrisroos:
__This has been superseded by PR #2099__

This is for discussion only and currently contains 4 failing tests that I've not investigated fully.

The `SmartAnswer::State` object is an `OpenStruct` which means that it responds to undefined methods with `nil`. This can lead to hard to find bugs in `next_node` blocks (particularly larger ones, e.g. [this one in pay-leave-for-parents](https://github.com/alphagov/smart-answers/blob/master/lib/smart_answer_flows/pay-leave-for-parents.rb#L385)) as a simple typo will mean that the program runs but doesn't return the desired result.

The main change in this PR overrides `SmartAnswer::State#method_missing` so that it raises a `NoMethodError` exception if we try to call a method that's not previously been defined.

```
# Behaviour before this change
>> state = SmartAnswer::State.new(start_node = nil)
>> state.foo
=> nil
>> state.foo = 'bar'
>> state.foo
=> 'bar'

# Behaviour after this change
>> state = SmartAnswer::State.new(start_node = nil)
>> state.foo
=> NoMethodError: undefined method 'foo' for SmartAnswer::State
>> state.foo = 'bar'
>> state.foo
=> 'bar'
```

I've also updated all Smart Answer flows where the code was relying on the old behaviour of the `State` object, by explicitly setting the variables to nil that are used later in the flow.

I'm not 100% convinced about this but would like to hear what other people think.

### DIFF
--- a/lib/smart_answer/state.rb
+++ b/lib/smart_answer/state.rb
@@ -6,6 +6,14 @@ module SmartAnswer
       super(current_node: start_node, path: [], responses: [], response: nil, error: nil)
     end
 
+    def method_missing(method_name, *args)
+      if method_name =~ /=$/
+        super
+      else
+        raise NoMethodError.new("undefined method '#{method_name}' for #{self.class}")
+      end
+    end
+
     def transition_to(new_node, input, &blk)
       dup.tap { |new_state|
         new_state.path << self.current_node
--- a/lib/smart_answer/state.rb
+++ b/lib/smart_answer/state.rb
@@ -3,7 +3,7 @@ require 'ostruct'
 module SmartAnswer
   class State < OpenStruct
     def initialize(start_node)
-      super(current_node: start_node, path: [], responses: [])
+      super(current_node: start_node, path: [], responses: [], response: nil, error: nil)
     end
 
     def transition_to(new_node, input, &blk)
--- a/lib/smart_answer_flows/additional-commodity-code.rb
+++ b/lib/smart_answer_flows/additional-commodity-code.rb
@@ -17,6 +17,10 @@ module SmartAnswer
 
         save_input_as :starch_glucose_weight
 
+        calculate :milk_protein_weight do
+          nil
+        end
+
         permitted_next_nodes = [
           :how_much_sucrose_1?,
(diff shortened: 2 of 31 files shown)


floehopper:
I think this is a worthwhile change. It's not ideal that we have to introduce so many missing variables, but then that's kind of the point! Is that what you are unconvinced about?

It would be good to implement `respond_to?` or `respond_to_missing?` to match `method_missing`.

Ideally, I think each flow would automatically instantiate a flow-specific "model" class with the relevant methods defined on it. Then we'd get this kind of behaviour for free. But I think this is a good first step.

chrisroos:
> It would be good to implement respond_to? or respond_to_missing? to match method_missing.

Do you think this is necessary given that we've not had it before now?

Interestingly, it would appear that `OpenStruct` doesn't define `respond_to?` or `respond_to_missing?` (this is against the `State` object before making the changes in this branch):

```
>> state = SmartAnswer::State.new(:start_node)
=> #<SmartAnswer::State current_node=:start_node, path=[], responses=[]>
>> state.respond_to?(:foo)
=> false
>> state.respond_to?(:foo=)
=> false
>> state.foo = 'bob'
=> "bob"
>> state.respond_to?(:foo)
=> true
>> state.respond_to?(:foo=)
=> true
```

chrisroos:
I've rebased this on master, fixed the failing tests and force-pushed.

I've also moved the commit that overrides `State#method_missing` to the end, so that the tests should be passing for each commit in the branch.

chrisroos:
@floehopper: Are you still happy for us to make this change?

floehopper:
I guess it's not critical to define `respond_to?` - I tend to try to always do it whenever I define `method_missing`. However, I can see here that `OpenStruct` is already behaving weirdly.

I'm still happy for this change to be merged.

LGTM :+1:

chrisroos:
I've tidied this branch up and opened a new pull request (PR #2099). I'm closing this and will merge from the new PR.

PR #2075, opened 2015-11-12
Require explicit question titles

floehopper:
Supersedes #2072.

Fail fast if no explicit question title is specified.

We think it's dangerous/surprising that the code was falling back to a human-friendly version of the question key for the question title.

I implemented the change to fail fast first and then ran all the regression tests to identify the small number of questions which did not have an explicitly defined title. I then moved the change to fail fast to be the last commit.

There are no expected changes in behaviour - hence no change to the regression test artefacts.

### DIFF
--- a/app/presenters/question_presenter.rb
+++ b/app/presenters/question_presenter.rb
@@ -18,7 +18,7 @@ class QuestionPresenter < NodePresenter
   end
 
   def title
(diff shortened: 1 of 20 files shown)


chrisroos:
Looks good to me.

floehopper:
I've rebase this against master and force-pushed in preparation for merging.

PR #2076, opened 2015-11-12
Reduce dependencies on I18nRenderer

floehopper:
I want to be able to change the `SmartAnswer::QuestionPresenter` to use the `SmartAnswer::ErbRenderer` instead of the `SmartAnswer::I18nRenderer` and so anything I can do to reduce the dependencies on the `I18nRenderer` will make this easier.


chrisroos:
I've left a couple of small comments but in general this looks good to me.

floehopper:
I've addressed @chrisroos' suggestions, rebased against master and force-pushed in preparation for merging.

PR #2077, opened 2015-11-12
Make QuestionPresenterTest more realistic

floehopper:
Since phrase lists are now *only* interpolated for question nodes, these tests
didn't make much sense. Especially as they were using the `QuestionPresenter`
with an `Outcome` node.

Also I'm not sure why the log statement was only being executed for outcome
nodes. If anything the opposite is now true and in any case the code in
`SmartAnswer::I18nRenderer#value_for_interpolation` will only ever be executed
for question nodes.


chrisroos:
Looks good to me.

PR #2078, opened 2015-11-12
Remove flow-level options from i18n fixtures

floehopper:
Since #2005, flow-level options are no longer used as a fallback, so we can
safely remove these.


chrisroos:
Looks good to me.

floehopper:

PR #2085, opened 2015-11-13
Remove title fallback for start nodes

floehopper:
Previously if not title was specified for a start node, we fell back to using
a human-friendly version of the node key. This was a carry over from code
originally in the `NodePresenter` and needlessly complicates the code.

Ideally we'd fail fast if no title was found (as in #2075), but given that the
regression tests render the title of the landing page for each flow, I think
this step is a safe one and it will make something else I'm working on easier.


chrisroos:
Looks good to me.

floehopper:

PR #2086, opened 2015-11-13
Improve unit tests for QuestionPresenter#error

floehopper:
My main objective here was to cover the behaviour when an
`SmartAnswer::InvalidResponse` is raised with a text message and the default
error message in the i18n YAML file interpolates this error. Currently this
behaviour is only tested in the `InputValidationTest` integration test, but
there's so much going on in there it's hard to see the wood for the trees.

I'm not convinced that this behaviour is very sensible nor do I think it is used
anywhere in the published smart answer flows. However, it does appear as if
there are quite a few places where a custom `String` is set as the exception
message rather than an error message key. Since I believe the mechanism
described above is the only way such messages could currently be displayed to
the user, I'm loathe to remove it without providing a replacement.

I've also clarified that the final fallback error message is system-wide and not
just for the question's flow.

### DIFF
--- a/test/fixtures/smart_answer_flows/locales/en/question-presenter-sample.yml
+++ b/test/fixtures/smart_answer_flows/locales/en/question-presenter-sample.yml
@@ -17,9 +17,11 @@ en-GB:
         title: "Is today a %{day}?"
         body: "Today is %{day}"
       question_with_custom_error_message:
-        custom_error_message: custom error message
+        custom_error_message_key: custom error message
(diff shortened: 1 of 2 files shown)


chrisroos:
Looks good to me.

floehopper:

floehopper:
I believe the build failed due to [a race condition in one of the JS-enabled integration tests](https://github.com/alphagov/smart-answers/issues/2092). I've kicked off another build manually.

PR #2087, opened 2015-11-16
Remove unused data attribute from content partial

floehopper:
This was added in a *very* early [commit][1] and as far as I can see has never
been used. The value of the data attribute also sounds suspiciously
flow-specific even though it appears in a generic template. On that basis, I
think it's safe to remove it.

[1]: b5df0c1bfb43e9840b4251cb84214e66eb1820a9

### DIFF
--- a/app/views/smart_answers/_content.html.erb
+++ b/app/views/smart_answers/_content.html.erb
@@ -20,7 +20,7 @@
     <% if @presenter.page_title %>
       <h2 class="page-title"><%= @presenter.page_title %></h2>
     <% end %>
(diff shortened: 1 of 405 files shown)


chrisroos:
Looks good to me, although I think you're going to regenerate the question page artefacts as they all seem to contain this `data-partial` attribute.

floehopper:
Well spotted. That's a shame :confounded:

chrisroos:
I'm running all the regression tests locally in preparation for getting this merged.

chrisroos:
I've rebased and force-pushed in preparation for merging.

PR #2088, opened 2015-11-16
Remove deprecated Flow#section & #subsection_slug

floehopper:
Most of the code relating to these methods was removed in 2012 in the
following two commits:

* 8bee67ea43dc4de5e024342931c8028785d10f4d
* d397d13352b3832d263867ba742f256955272374

The deprecation warnings were left in place in case any un-merged flows used
either of the methods. I think it's safe to say that this danger has passed and
it's now safe to removed these methods.

### DIFF
--- a/lib/smart_answer/flow.rb
+++ b/lib/smart_answer/flow.rb
@@ -48,16 +48,6 @@ module SmartAnswer
       @status
     end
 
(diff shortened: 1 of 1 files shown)


chrisroos:
Looks good to me.

floehopper:

PR #2089, opened 2015-11-16
Improve test coverage for FormattingHelper#format_money

floehopper:
I believe that the origin of the code in `FormattingHelper#format_money` was
from the [code which is now in `SmartAnswer::I18nRenderer#value_for_interpolation`][1].
Thus at least in some scenarios I would expect a `SmartAnswer::Money` object to be
passed in rather than a `String`. This commit adds assertions for the former.

[1]: https://github.com/alphagov/smart-answers/blob/7e26c554be106b2888f44bf92746e5fc7f6d819f/lib/smart_answer/i18n_renderer.rb#L33-L34


chrisroos:
Looks good to me.

floehopper:

PR #2090, opened 2015-11-16
Remove unnecessary guard conditions before calls to String#chomp in node presenters

floehopper:
`SmartAnswer::ErbRenderer#content_for` never returns `nil` - it always returns
a `String` even if it's an empty one. Thus the guard condition before each
call to `String#chomp` in the `OutcomePresenter` & `StartNodePresenter` is
unnecessary and can be removed.


chrisroos:
Looks good to me.

floehopper:

PR #2091, opened 2015-11-16
Avoid relying on old fallback behaviour for question title

floehopper:
Since [this commit][1] we no longer fallback to the humanized version of the question name,
so it doesn't make sense to rely on that behaviour.

Also by asserting against a string literal we can avoid delving into the
internals of the flow.

[1]: aad10d5b0b0dbd1826c68523e9c69d8f70e38687


chrisroos:
Looks good to me.

floehopper:

PR #2093, opened 2015-11-16
Refactor SmartAnswersControllerTest

floehopper:
* This controller test was huge and thus difficult to follow, so I've split it up into a bunch of different test cases.

* Many of the tests were sharing bits of flow fixtures which meant they were brittle/coupled, so I've introduced explicit flow classes for each test.

* In the cases where a different translation file was used for a subset of tests, I've added extra questions to the relevant flow, rather than creating a whole new explicit flow class. Although this isn't great, it feels like a reasonable compromise.

* Previously `SmartAnswer::FlowRegistry.find` was being stubbed. I thought this was a bit clunky and it wasn't much harder to use the real registry, although it seemed sensible to disable flow pre-loading to prevent a big slow-down in the tests.

All in all, these controller tests are still testing too much, but hopefully these changes make things a bit easier to follow. More to the point, I'm hoping they're going to make it easier to switch the question presenter to ERB rendering vs I18n rendering.


chrisroos:
Looks good to me.

PR #2094, opened 2015-11-16
Use FormattingHelper#format_money in entitled_to_sick_pay outcome in SSP

floehopper:
We're trying to standardize on using this method to format money objects.

The effect of this is to leave off the decimal point and the pence when the
value is a whole number of pounds.

@lutgendorff confirmed that this was probably just an accidental difference and
so I'm fixing it here.


chrisroos:
Looks good to me.

PR #2098, opened 2015-11-16
ErbRenderer should always return an HTML-safe String

floehopper:
Content that has already been rendered through `ErbRenderer` doesn't need to be HTML-escaped, because `ActionView::Base` will have already taken care of that. So we can mark content returned from `ErbRenderer` as being `html_safe`.


chrisroos:
Looks good to me.

floehopper:

PR #2103, opened 2015-11-17
Allow questions to use ERB templates

floehopper:
This PR makes it possible to optionally switch a question or flow to use `SmartAnswer::ErbRenderer` instead of the default `SmartAnswer::I18nRenderer`.

To do this for an individual question, you need to pass in `use_erb_template: true` when declaring the question in the flow. If you call `Flow#use_erb_templates_for_questions` in the flow definition, then all subsequent questions will use an ERB template.

I've implemented a hacky script (`script/convert-question-i18-yaml-to-erb-templates.rb`) to do most of the conversion automatically. It warns you if there are any top-level phrases in the i18n YAML file, because you'll probably need to make some manual changes to implement the same functionality in the ERB templates. It also warns you if an individual ERB template contains interpolation, because you will need to manually convert these to use ERB tags. Note that this script hasn't been exhaustively tested, so it may need tweaking to run against the published flows.

I have also switched all the fixture flows over to use ERB templates as a way of trying out the new functionality. The only one I haven't changed over is `QuestionPresenterTest`. This is because I've added a "parallel" `QuestionPresenterWithErbRendererTest` which exercises the same presenter but with the `ErbRenderer`. Once we've switched all the published flows over, we will be able to remove the `I18nRenderer` and replace `QuestionPresenterTest` with `QuestionPresenterWithErbRendererTest`.


chrisroos:
This all looks good to me.

floehopper:

PR #2112, opened 2015-11-18
ErbRenderer should always return HTML-safe string for option labels

floehopper:
The option label text is all provided by developers; none of it is supplied by external users. This means we can legitimately mark all option label text as HTML-safe. c.f. #2098

There's no need to update any regression test artefacts, because this change only affects the options in question pages and these are all currently rendered by the `I18nRenderer`.


chrisroos:
Looks good to me.

PR #2116, opened 2015-11-18
Fix indentation for multi-line strings in script to convert flows to use ERB templates for questions

floehopper:
Previously only the first line was indented.


chrisroos:
Looks good to me.

PR #2119, opened 2015-11-18
Remove Smartdown documentation

chrisroos:
This is no longer required now that Smartdown has been removed from the application (see PR #2050).


floehopper:
Apart from my one minor comment, this LGTM :+1:

PR #2125, opened 2015-11-18
Make FormattingHelper methods available to question ERB templates

floehopper:
I need this in order to be able to convert `calculate-statutory-sick-pay` (and probably others) to use ERB templates for questions.


chrisroos:
Looks good to me.

PR #2133, opened 2015-11-18
Convert estimate-self-assessment-penalties to use ERB templates for questions

floehopper:


chrisroos:
I think we might want to fix the ampersand escaping before merging this one.

floehopper:
@chrisroos: I've addressed the [ampersand-escaping issue](https://github.com/alphagov/smart-answers/commit/6aae53ecdc8e872253eb1eb5e35f187d66dc02a3#commitcomment-14497029) using an interactive rebase in this [new commit](1383b67f8ac8f515f145c2edb32994e5b01bb186). I've also updated the regression test checksums and force-pushed.

chrisroos:
Looks good to me.

chrisroos:

PR #2138, opened 2015-11-19
Fix FlowRegistrationPresenter to work with both i18n & ERB

floehopper:
In #2103 I forgot that we still need to support flows/questions with their
content in i18n YAML files vs ERB templates.

In this commit I've renamed the `FlowRegistrationPresenterTest` to
`FlowRegistrationPresenterWithErbRendererTest` and reinstated the earlier
version of `FlowRegistrationPresenterTest` to give suitable test coverage.

I've also reinstated the two i18n YAML files and use an explicit call to
`Flow#use_erb_templates_for_questions` in
`FlowRegistrationPresenterWithErbRendererTest` to switch over to ERB rendering.

Finally I've also reinstated the handling of
`I18n::MissingInterpolationArgument` so that the tests all pass.

Note that we saw [an exception in preview](https://errbit.preview.alphagov.co.uk/apps/533c2ee40da115303f0129a5/problems/564d91d365786306b4860500) which alerted me to this problem.


chrisroos:
Looks good to me.

floehopper:

PR #2139, opened 2015-11-19
Fix GraphPresenter to work with both i18n & ERB

floehopper:
In #2103 I forgot that we still need to support flows/questions with their
content in i18n YAML files vs ERB templates.

In this commit I've renamed the `GraphPresenterTest` to
`GraphPresenterWithErbRendererTest` and reinstated the earlier
version of `GraphPresenterTest` to give suitable test coverage.

I've also reinstated the i18n YAML file and used an explicit call to
`Flow#use_erb_templates_for_questions` in `GraphPresenterWithErbRendererTest` to
switch over to ERB rendering.

I've avoided needing to handle `I18n::MissingInterpolationArgument` by passing
in an empty `Hash` as the `state` when instantiating the `QuestionPresenter`.

### DIFF
--- a/lib/graph_presenter.rb
+++ b/lib/graph_presenter.rb
@@ -1,6 +1,7 @@
 class GraphPresenter
   def initialize(flow)
     @flow = flow
+    @i18n_prefix = "flow.#{@flow.name}"
   end
 
   def labels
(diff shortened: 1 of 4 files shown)


chrisroos:
Looks good to me.

floehopper:

PR #2143, opened 2015-11-19
Convert part-year-profit-tax-credits to use ERB templates for questions

floehopper:
Note that this flow has no integration tests, but all the other unit & integration tests are passing, so I think all is well.


chrisroos:
Looks good to me.

chrisroos:

PR #2159, opened 2015-11-19
Combine data partials and shared partials

chrisroos:
We had shared partials in both the smart_answer_flows/data_partials/ and smart_answer_flows/shared/ directories, which was confusing.

The smart_answer_flows/data_partials/ directory is a hangover from the days before ERB templates and all partials should now live in the smart_answer_flows/shared/ directory.


floehopper:
LGTM :+1:

chrisroos:

PR #2162, opened 2015-11-24
Update marriage abroad with fact check amends from August

erik-eide:
This PR introduces a country specific outcome for marriage abroad in Poland, as step of moving away from the generic outcome template with a lot of branching logic that is hard to maintain.

Included are copy changes from the fact check and updates to download links for marriage in Spain that now go to splash pages rather than the download files directly.

## Expected changes
 * Under the "What you need to do" section, there should be two new sub-sections "Get your documents witnessed in front of a notary public" and "Get your documents witnessed at the embassy".
  [Example URL](https://www.gov.uk/marriage-abroad/y/poland/ceremony_country/partner_british/opposite_sex)

screen shot 2015-11-24 at 11 52 11 am


chrisroos:
Looks good to me.

erik-eide:
Now rebased with master

PR #2163, opened 2015-11-24
Remove predicate functionality

chrisroos:
We deprecated the use of the predicate logic a while back. This branch removes all the unused predicate logic code.


chrisroos:
I've run all the regression tests in this branch and am happy that I've not broken anything. There may be other things that can also be removed but I think this is good enough to get merged to master.

chrisroos:

PR #2167, opened 2015-11-26
Remove i18n code

chrisroos:
I also tried to remove the "rails-i18n" Gem as part of this pull request. Removing it caused a number of regression tests to fail because, for example, `number_to_currency` started returning `$` instead of `£`.


chrisroos:
I'm currently running all regression tests in this branch to ensure that I haven't inadvertently broken something.

chrisroos:
All regression tests passed in this branch on my local machine.

chrisroos:
I've tweaked some of the commit messages and force pushed in preparation for merging.

PR #2168, opened 2015-11-27
Add common fee table for marriage abroad

erikse:
This change pulls all the fee table rendering for the marriage abroad outcomes into one partial, and allows fee items as well as the price to charge for them, to be set in the calling outcome.

Prior to this, the same markdown for the fee table appeared in many places through-out the smart answer, possibly as a result of many outcomes having slightly different sets of fees.


erik-eide:
Now rebased with master

chrisroos:
Is there a reason you're using `local_assigns` in _fee_table.govspeak.erb? We're not using it anywhere else so I think it'd be better to access the local variables directly, e.g. `receiving_notice_of_marriage` instead of `local_assigns[:receiving_notice_of_marriage]`.

chrisroos:
I've just realised that you're using `local_assigns` as you need to know whether the key exists to decide whether to render that line. I think I would've done this using `if defined?(local)` but that's probably only because I haven't come across `local_assigns` before.

chrisroos:
I _think_ I'd be tempted to hardcode the prices that never change in the "_fee_table.govspeak.erb" partial. My worry with passing them in is that it'll be easy for someone to change a price in one place and not necessarily realise that it should also be changed elsewhere.

It looks like there are only 3 fees that vary (and 21 that don't):

* `:affidavit_for_marriage` - £55 / £65
* `:affirmation_for_marriage` - £55 / £65
* `:affirmation_for_freedom_to_marry_in_english` - £45 / £55

I'd also be tempted to check with someone whether or not these should actually be different? Maybe they've inadvertently fallen out of sync.

What do you think, @erik-eide?

chrisroos:
I should mention that in general I think this is an improvement - it's just the passing in of prices that I'm slightly wary about.

erik-eide:
Ok, I'll check with content regarding the differences, agree it would be simpler if they were the same/consistent.

erik-eide:
Content has forwarded this query to FCO, waiting clarification

chrisroos:
@erikse: Let me know if you'd like me to rebase this now that my changes to marriage-abroad have been merged to master.

erikse:
FCO have just come back with an answer on this, that the fees for each service should be the same in each country - I'm not sure if it's worth updating this PR as it was based on providing flexibility of each outcome specifying what a fee for each service should be.

chrisroos:
> FCO have just come back with an answer on this, that the fees for each service should be the same in each country

@erikse: Does this mean that the fees for the following services should all be the same?

```
:affidavit_for_marriage - £55 / £65
:affirmation_for_marriage - £55 / £65
:affirmation_for_freedom_to_marry_in_english - £45 / £55
```

If so, do you know what the fee should be in each case?

I still think the introduction of the single fees partial is an improvement. Rather than passing a hash of services and prices, we could just pass an array of services and have the prices in the fees partial itself.

For example:

```
# In a template
<%= render partial: 'fee_table.govspeak.erb',
    locals: { services: [:receiving_notice_of_registration] } %>
```

```
# In _fee_table.govspeak.erb
<% if services.include?(:receiving_notice_of_registration) %>
  Receiving a notice of registration | £nnn %>
<% end %>
```

erikse:
@chrisroos - my understanding is they should all be the same, I guess we'll need to double check which fees are correct in the 3 cases where there are variances. I would guess it's the default one, but can check. A single partial receiving an array to turn on the fees to show seems good :+1:

erikse:
I think given the discussion seems to agree on taking a different direction with the implementation of this PR, I'll close this one for now.

chrisroos:
@erikse:

Did you check with the department about the fees that currently vary? I think we should at least capture the variable fees in a PT story or Trello card.

Are you planning to do any further work on this, i.e. toward the introduction of a single fees table?

erikse:
@chrisroos - we're still waiting to hear regarding the fee variances, the indication seems to be fees will be reviewed in the future, and that's all we know at this stage

chrisroos:
Having heard from back FCO, @erikse has now made all the fees consistent In PR #2350.

Issue #1681, opened 2015-06-08
Upgrade the Airbrake Gem


chrisroos:
We're using version 3.1.7 of the Airbrake Gem. This version doesn't report on `ActionController::BadRequest` exceptions correctly. They're [reported in Errbit][errbit-notification] as `Notification`s, which isn't very helpful.

[Pull request 271 in the Airbrake project][pr-271] appears to have fixed this problem.

The pull request was updated on Aug 26 2014 with a comment saying that the fix would be released the following week. Airbrake version 4.1.0 was released on Sep 4 2014 so I suspect we'll want at least this version.

[errbit-notification]: https://errbit.production.alphagov.co.uk/apps/533c35ae0da1159384044f5f/problems/553c03c66578631e12190300
[pr-271]: https://github.com/airbrake/airbrake/pull/271

chrisroos:
I've moved this to Trello in an attempt to end up with a single place containing Smart Answers stories/tasks/bugs.

See the [Trello card](https://trello.com/c/wAWWo86p/139-upgrade-the-airbrake-gem) (not public).

Closing this issue.

PR #2170, opened 2015-12-01
Upgrade airbrake gem to version 4.1.0 (as used in alphagov/whitehall).

erik-eide:
Addresses issue #1681 by upgrading to a newer version of the Airbrake gem to ensure Errbit reports ActionController::BadRequest correctly as exceptions, and not notifications.

Whilst not the latest version, its the same version as used in [alphagov/whitehall](https://github.com/alphagov/whitehall/blob/master/Gemfile#L40).

Further details in Trello [card](https://trello.com/c/wAWWo86p/139-upgrade-the-airbrake-gem) (private board).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,7 +4,7 @@ ruby '2.2.3'
 
 gem 'rails', '4.2.5'
 
-gem 'airbrake', '~> 3.1.17'
+gem 'airbrake', '4.1.0'
 gem 'extlib', '0.9.16'
 gem 'govuk-content-schema-test-helpers', '~> 1.3.0'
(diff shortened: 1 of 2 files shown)


jennyd:
The [changelog](https://github.com/airbrake/airbrake/blob/34fcbe8e6aae20aad0e4f5174df16945dd6f1356/CHANGELOG#L219-L297) doesn't mention anything which sounds like it would require other changes for this bump, and Whitehall [didn't have any non-Gemfile changes](https://github.com/alphagov/whitehall/pull/1938) for a similar bump :+1:

PR #2172, opened 2015-12-02
Memoize node presenters

floehopper:
We were creating far more instances of `NodePresenter` and its subclasses than we needed. This in turn meant we were creating far more instances of `SmartAnswer::ErbRenderer` and thence `ActionView::Base` than necessary. It may well also have meant that we were unnecessarily re-rendering the same template multiple times.

In a recent release, all flows were converted to use ERB rendering for questions instead of YAML I18n rendering. This release coincided with the app using a lot more memory ([private link](https://trello.com/c/qhSd3VPD/162-investigate-memory-leak)). We have some evidence to suggest that making this change will (at the very least) reduce memory usage of the app somewhat. However, we haven't been able to unambiguously reproduce the memory usage problem we're seeing in production, so it's difficult to know whether this will actually resolve the problem.

In working on this change, it's highlighted how tangled an inefficient this area of the codebase is (e.g. `FlowPresenter#collapsed_questions` indirectly calls `Flow#process` multiple times). I'm convinced that the code could be considerably simplified, but that's going to have to wait for another day.

### DIFF
--- a/app/presenters/flow_presenter.rb
+++ b/app/presenters/flow_presenter.rb
@@ -95,7 +95,7 @@ class FlowPresenter
 
   def start_node
     node = SmartAnswer::Node.new(@flow, @flow.name.underscore.to_sym)
-    StartNodePresenter.new(node)
+    @start_node ||= StartNodePresenter.new(node)
   end
 
   def change_collapsed_question_link(question_number)
--- a/app/presenters/flow_presenter.rb
+++ b/app/presenters/flow_presenter.rb
@@ -14,6 +14,7 @@ class FlowPresenter
     @request = request
     @params = request.params
     @flow = flow
+    @node_presenters = {}
   end
 
   def artefact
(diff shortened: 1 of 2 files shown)


floehopper:
Based on the suspicion that `ActionView::Base` is the root of the problem, I think some sensible follow-up steps to this PR would be to:

1. Re-use the same instance of `ActionView::Base` for all question node presenters for a given flow and another for all outcome node presenters.
2. Re-use the same instance of `ActionView::Base` for all node presenters for a given flow.

floehopper:
@chrisroos, @erkede: Is there any reason why I shouldn't merge this? I've addressed @erkede's earlier comment to memo-ize node presenters keyed on name in 1a79b4f3d66ac8c0abd18f978b2925da73261cfa.

chrisroos:
Looks good to me.

PR #2176, opened 2015-12-04
WIP: Spike into using single ActionView::Base for all node presenters

floehopper:
This turned out to be quite awkward with the current structure of the code.

With the changes as they stand, all the *regression* tests pass, but I haven't
fixed up any of the other tests. Also there are two other places where we use
node presenters which will also need attention:

* `GraphPresenter`
* `FlowRegistrationPresenter`

The biggest problem I ran into was the content in `content_for` blocks for
different node presenters, but the same key were appended to each other. I got
round this by doing the rendering in the node presenter constructor and using
a different instance of `ActionView::OutputFlow` for each presenter.

This means the node presenters don't need to hang on to a reference to the
(now shared) instance of `ActionView::Base`.

Other downsides of this approach are:

* I had to include all helper modules in the one instance of `ActionView::Base`.
  However, we could still feasibly restrict this to the generic helper modules
  and *one* flow-specific helper module relating to the "current" flow.

* I ended up adding the flow directory and both the `questions` & `outcomes`
  sub-directories to the view path. I assume this might cause problems if we
  had a question and an outcome template with the same name for the same flow.
  Thankfully this doesn't seem to be a problem at the moment, but it would be
  good to understand what would happen if this were the case. There might well
  be a better way of setting up the view paths, but I couldn't find it.

Note: It wouldn't be a big step to use Rails' instance of `ActionView::Base`
instead of our own special instance. However, I decided not to do this, to avoid
running into more problems with sharing helpers and view paths with the main
Rails app.

### DIFF
--- a/app/controllers/smart_answers_controller.rb
+++ b/app/controllers/smart_answers_controller.rb
@@ -69,7 +69,7 @@ private
   def find_smart_answer
     @name = params[:id].to_sym
     @smart_answer = flow_registry.find(@name.to_s)
-    @presenter = SmartAnswerPresenter.new(request, @smart_answer)
+    @presenter = FlowPresenter.new(request, @smart_answer)
   end
 
   def flow_registry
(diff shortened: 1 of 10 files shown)


floehopper:
I'm a bit less convinced about the approach I've taken in this spike and in any case the code needs a lot of work to make it merge-able. So I'm going to close this PR for now.

PR #2177, opened 2015-12-04
Update the accommodation rate for minimum wage

erkede:
As the minimum wage data is shared between the
am-i-getting-minimum-wage and the
minimum-wage-calculator-employers smart answers,
the file artefact checksums in change in both.

The rest artefacts remain the same, as the
regression testing suite freezes time to before
the current wage data that applies from October
2015 onwards.


erik-eide:
Approved for release once code reviewed

marksheldonGDS:
Reviewed with @jennyd. Looks good to us.

PR #2179, opened 2015-12-08
Rename SmartAnswerPresenter to FlowPresenter

floehopper:
Since we have a Flow with many Nodes, it makes more sense to me that we have
a FlowPresenter with many NodePresenters.

This is also part of #2172, but I reckon it's much less controversial, so thought it was worth a separate PR.


chrisroos:
Looks good to me.

PR #2181, opened 2015-12-08
Remove multiple question per page code

floehopper:
Multiple questions per page were never supported by Ruby Smart Answers. All this code was only there to support Smartdown's multiple questions per page functionality. Now that all flows have been converted to Ruby Smart Answers (i.e. single question per page), it's safe to remove this code.

Even if we decide we want to support multiple questions per page in the future, I think it's still worth doing this. Having made the changes in this PR, I'm sure we can simplify the code a lot further which would then make it easier to re-introduce multiple questions per page if it becomes desirable.

These changes should indirectly make it a bit easier to move towards directly rendering the landing page, question & outcome templates from the controller action.

I think there was already some unused CSS and this PR probably makes more CSS redundant. However, think that would be better tackled separately.

PS: Hopefully this PR isn't as bad as it looks for reviewing - I've tried to keep all the commits very small and focussed which is why there are so many of them.

### DIFF
--- a/app/helpers/current_question_helper.rb
+++ b/app/helpers/current_question_helper.rb
@@ -13,7 +13,8 @@ module CurrentQuestionHelper
     end
   end
 
-  def prefill_value_is?(value, question_number = nil)
+  def prefill_value_is?(value)
+    question_number = nil
     if params[previous_response_key(question_number)]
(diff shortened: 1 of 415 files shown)


chrisroos:
Looks good to me.

floehopper:

PR #2186, opened 2015-12-10
Remove duplicate payment content

erikse:
PT story: https://www.pivotaltracker.com/story/show/108740582

Picked up during the above story, this PR replaces duplicated content in two outcomes in the marriage-abroad smart answer, with content that already exists in a partial.

This PR can be merged either before or after the above story is completed.

### Expected changes:

Regression tests show no change in output.


tadast:
LGTM

erikse:
I've just rebased this PR with master and forced pushed

PR #2188, opened 2015-12-11
Test against v2 schemas

tijmenb:
This app was still sending old-style payload to the publishing-api.

https://trello.com/c/catWOsWC

### DIFF
--- a/app/presenters/flow_content_item.rb
+++ b/app/presenters/flow_content_item.rb
@@ -9,11 +9,9 @@ class FlowContentItem
     {
       base_path: base_path,
       title: flow_presenter.title,
-      content_id: flow_presenter.content_id,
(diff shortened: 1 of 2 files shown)


e2ff:
We have a lot of references to [content_id](https://github.com/alphagov/smart-answers/search?utf8=%E2%9C%93&q=content_id) still left over... do we still need these?

tijmenb:
Yes definitely. Previously we sent `content_id` in both the payload and in the URL params (`PUT /v2/content/:content_id`). This commit removes the content_id from the payload, but not from the URL:

https://github.com/alphagov/smart-answers/blob/master/app/services/content_item_publisher.rb#L5-L6

e2ff:
Ok cool, let's give it a try

PR #2192, opened 2015-12-15
State pension calculator amends

erikse:
PT story: https://www.pivotaltracker.com/story/show/108748030

DWP are launching their own State Pension calculator.

This PR removes the 'estimate State Pension amount' option and associated logic. The start page is amended to reflect the reduced functionality of the updated flow and many of the questions and outcomes are removed.

## Fact check

https://pt-108748030.herokuapp.com/state-pension-age

## Expected changes
 * [URL on gov.uk](https://www.gov.uk/calculate-state-pension)
  * The start page title shows 'Check your State Pension age'
  * Removes sections "If you're under 55" and "If you're over 55" 

### Before
screen shot 2015-12-15 at 08 12 35

### After
screen shot 2015-12-15 at 08 12 38

 * [URL on gov.uk](https://www.gov.uk/calculate-state-pension/y)
  * First question no longer has option "Amount - estimate of your basic State Pension amount"

### Before
screen shot 2015-12-15 at 08 20 49

### After
screen shot 2015-12-15 at 08 20 53


floehopper:
As far as I can see there is no explanation as to why this functionality has been removed in the commit notes, the PR description or the Pivotal Tracker story.

erikse:
There is a router that can do the redirect, but I guess we need to deploy the new smart answer before we can redirect away from the old one. So this would just cover that window.

erikse:

erikse:

erikse:
Changed this to don't merge until we have a known release date

floehopper:
Minor: There are quite a few commits with exactly the same first line which makes it tricky to work out which one is which. I'd be inclined to include some kind of differentiation in these first lines (e.g. method name) to make things clearer.

floehopper:
I think it would be useful to include a "before" and "after" visualisation of the flow in the PR description.

floehopper:
In [this line](https://github.com/alphagov/smart-answers/blob/master/lib/smart_answer_flows/calculate-state-pension/questions/which_calculation.govspeak.erb#L6), the lac of capitalisation of "pension credit" seems to be inconsistent with other copy. However, it looks as if this was the case before this PR anyway.

floehopper:
The code in `StatePensionAgeCalculator` which uses `FriendlyTimeDiff` should eventually be moved out of the calculator and into view helper methods. However, this change is probably out-of-scope for this PR.

floehopper:
I've now finished reviewing this PR. The majority of my comments were about better ways to structure the commits . If time is pressing, I don't think there is anything preventing this PR from being merged. However, if there is time, it would be much better to address as many of the comments as possible before merging.

floehopper:
With the caveats mentioned in my most recent comment, I'm going to remove the "Don't merge" label and add the "LGTM" label.

erikse:
Have just rebased with master

floehopper:
The 5 new commits look good to me.

PR #2197, opened 2015-12-16
Italy, Philippines and South Africa marriage abroad amends

erikse:
PT story: https://www.pivotaltracker.com/story/show/108740582

This PR is a collection of changes requested by FCO content design for the marriage abroad outcomes for Italy, Philippines and South Africa.

## Fact check

https://fierce-woodland-9779.herokuapp.com/marriage-abroad/y

## Expected changes

 * [URL on gov.uk](https://www.gov.uk/marriage-abroad/y/italy/ceremony_country/partner_british/opposite_sex)
  * Changes sentence "To get a Nulla Osta, you’ll first need to give notice" to "To get a Nulla Osta, you’ll first need to post notice"
   * Moves appointment link below above sentence
   * Moves 3 day residency statement to before "Applying for a Nulla Osta section"

### Before
screen shot 2015-12-16 at 5 02 09 pm

### After
screen shot 2015-12-16 at 5 02 16 pm


floehopper:
Typo: "FTO content design" -> "FCO content design" in the PR description.

erikse:
Have updated

floehopper:
I realise that these changes probably came in on the same ticket/story, but, unless they are actually closely related in some way, I think it would be a lot clearer to split them into one PR per country. At the very least it would be good if all the Italy-related commits in this PR were contiguous.

floehopper:
A number of the early commits appear to be making changes related to a "Nulla Osta" in Italy. I think it might be better to group these related changes into a single commit rather than having one commit per file.

floehopper:
Very minor: I've noticed that the line length in many of the commit messages varies quite a bit. For example, in [this one](91596836741d8c943a442edad5bcb90913ab8227) and a few others, the subject line is nearly 100 characters. Also in [this one](a74bd5393b36b81c3cedae1cdfcaf4e1d0374a3b) and a number of others, the non-subject lines appear to wrap at ~45 characters.

The former means the subject lines are truncated on GitHub and in other git clients. The latter makes the text a bit hard to read.

I think the standard recommendation is that the subject line should be <= 50 chars and the other lines should be <= 72 chars.

floehopper:
I think this PR was rebased while I was reviewing it, so I may have missed something.

As I think is usual with Marriage Abroad changes, there doesn't seem to be much information about *why* various changes are being made.

Most (if not all) my comments are minor and none of them should prevent this PR from being merged.

erikse:
This was one of the last tickets from August 2015 in which we were receiving with changes to sevearl countries grouped together, so this kind of PR shouldn't happen any longer. 

The initial work was applied per country, but after fact check, revisions were made to some outcomes which is why the commits are not ordered per country across the entire PR. I'm inclined to leave them as they are to reflect this.

PR #2201, opened 2015-12-18
Remove unnecessary dependency on extlib

floehopper:
The dependency on `extlib` was originally just an implicit dependency via
`google-api-client`, but in [this commit][1] it was converted into an explicit
dependency in order to fix a [security vulnerability][2].

The dependency on `google-api-client` was removed in [this commit][3]
and so now, as far as I can see, we no longer need `extlib` at all.

[1]: https://github.com/alphagov/smart-answers/commit/d0709ebaf107286ac35229829a08feb741d002cf
[2]: http://osvdb.org/show/osvdb/90740
[3]: https://github.com/alphagov/smart-answers/commit/ac0c081d09553d3ac12d9e608044fb591fb5b3ef

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -5,7 +5,6 @@ ruby '2.2.3'
 gem 'rails', '4.2.5'
 
 gem 'airbrake', '4.1.0'
-gem 'extlib', '0.9.16'
 gem 'govuk-content-schema-test-helpers', '~> 1.3.0'
 gem 'govuk_frontend_toolkit', '3.1.0'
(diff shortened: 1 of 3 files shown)


erikse:
I got 404s on those commit links (seems to work when I change pull to commit in the path), but otherwise - looks good to me

floehopper:
Thanks. That's weird. I'm sure that form of link to commit used to work. Presumably something changed in GitHub-flavoured markdown. Anyway, I've fixed the links now.

PR #2205, opened 2015-12-21
Remove debugger gem from Gemfile

erikse:
This gem doesn't seem to support the Ruby version [2.2.3](https://github.com/alphagov/smart-answers/blob/master/Gemfile#L3) we specify in the Gemfile.

From the debugger gem [README.md](https://github.com/cldwalker/debugger/blob/master/README.md#description):

"A fork of ruby-debug(19) that works on 1.9.2 and 1.9.3 and installs easily for rvm/rbenv rubies.
ruby >= 2.0 are *not* supported"

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -57,10 +57,6 @@ else
   gem 'govspeak', '~> 3.3.0'
 end
 
-if ENV['RUBY_DEBUG']
-  gem 'debugger', require: "ruby-debug"
-end
-
 if ENV['SLIMMER_DEV']
   gem 'slimmer', path: '../slimmer'
 else


floehopper:
Looks good to me. I've never been keen on having gems like this in the `Gemfile`, especially with the environment variable conditional.

I've never used it, but I assume that to use it you'd have to set the env var, run `bundle install`, do your debugging, and then when you've finished remember to unset the env var, revert the changes to `Gemfile.lock` and run `bundle install` again.

If this is true, I reckon it's not much to ask a developer to *manually* add the relevant gem to the `Gemfile` before peforming the other steps i.e. I don't see that much is gained by having the conditional gem dependency in the version of the `Gemfile` in the repo.

erikse:
Agreed, be nice to remove the last few.

I PR'ed this one separately as the debugger gem doesn't even build towards the Ruby version we use.

PR #2207, opened 2015-12-21
Remove conditional selection of gem version/path

erikse:
Developers can easily modify the Gemfile to include a ```path: <path>``` option or ```git: <uri>``` option to a particular version of a gem if they want to work towards, rather than setting ```ENV``` variables which are very easy to forget to unset.


floehopper:
Looks good to me :+1:

PR #2211, opened 2015-12-22
Landlord immigration check amends

erikse:
This PR supersedes #2210, #2212, #2213, #2214 and #2224

The landlord immigration check pilot now rolls out to England wide, requiring landlords to perform checks on tenancies start on or after Feb 1, 2016, as well as the existing requirement for checks to be performed inside the pilot Boroughs and Councils since 2014.

## Fact check
https://afternoon-citadel-5721.herokuapp.com/landlord-immigration-check/y

## Expected changes
 * [URL on gov.uk](https://www.gov.uk/landlord-immigration-check)
  * New content allowing tenancy checks after Feb 1, 2016 for all of England, e.g. "You can only do this check for tenancies which start on or after 1 February 2016."

### Before
screen shot 2015-12-22 at 11 32 42 am

### After
screen shot 2015-12-22 at 11 32 52 am

PR #2214, opened 2015-12-22
Updated for rest of the the UK

keithiopia:
Changed this so it is relevant for residents of NI, Wales, Scotland (outside of England).
From: https://govuk.zendesk.com/agent/tickets/1209106

This will require a preview please.


erikse:
I've pulled these two commits into PR #2211

PR #2224, opened 2015-12-23
Policy update

keithiopia:
Updated this for 3rd Jan policy change - this rolls out across England (not just Birmingham etc).

A preview would be very useful please.

From: https://govuk.zendesk.com/agent/tickets/1209106
url: https://www.gov.uk/landlord-immigration-check


erikse:
These three commits have been cherry-picked into PR #2211

PR #2231, opened 2016-01-04
Split list of acceptable docs by user's location

keithiopia:
Trying to simplify the view of documents that a prospective tenant has to show a landlord. 
I'm not sure if I've done the conditional renders correctly, but I'm hoping it will show the user a slightly different bullet list of documents depending on the location of their property. 
If they're in included_borough, they should see _pilot_documents.govspeak.erb.
If they're in included_country, they should see _england_documents.govspeak.erb.

Currently appears at: /landlord-immigration-check/y/B5%204RU/yes/yes/no/no/no/somewhere_else

This will need to go to the department for fact check.

From: https://www.pivotaltracker.com/story/show/110973118


erikse:
This PR has been superseded by #2232

PR #2234, opened 2016-01-11
Register a birth abroad in Turkey

erikse:
PT story: https://www.pivotaltracker.com/story/show/109527156

Updates the outcome for Register a birth abroad in Turkey to include the name of the local birth certificate.

## Fact check

https://pt-109527156.herokuapp.com/register-a-birth

## Expected changes
 * [URL on gov.uk](https://www.gov.uk/register-a-birth/y/turkey/mother_and_father/yes/same_country)
  * Adds the name of the local birth certificate ‘Formül A’

### Before
screen shot 2016-01-13 at 1 00 41 pm

### After
screen shot 2016-01-13 at 1 01 35 pm

### DIFF
--- a/lib/smart_answer_flows/register-a-birth/outcomes/oru_result.govspeak.erb
+++ b/lib/smart_answer_flows/register-a-birth/outcomes/oru_result.govspeak.erb
@@ -113,7 +113,7 @@
         - the child's full local birth certificate
         - the hospital declaration of live birth (if the father isn't named on the birth certificate)
       <% when 'turkey' %>
-        - the child’s full local birth certificate - it must have both parents’ names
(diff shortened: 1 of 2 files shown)


erikse:
I've just rebased this with the master branch and forced back to this PR's branch

tadast:
I guess Turkey is not on the set of inputs for regression tests?
If so then it LGTM, otherwise artefacts should have changed :) :+1:

erikse:
It's not in the [sample answers[(https://github.com/alphagov/smart-answers/blob/master/test/data/register-a-birth-questions-and-responses.yml#L2), but given it wasn't a logic change - I didn't add it

PR #2235, opened 2016-01-11
Update Heroku settings and documentation from preview to integration

erikse:
Now that preview has been replaced in favour of integration, updating the Heroku variables set by this script for setting up new staging servers

@dsingleton @jennyd

### DIFF
--- a/doc/common-errors.md
+++ b/doc/common-errors.md
@@ -56,7 +56,7 @@ Unable to fetch: 'wrapper' from 'http://static.dev.gov.uk/templates/wrapper.html
 
 There are a couple of solutions to this problem:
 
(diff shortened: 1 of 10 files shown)


jennyd:
The Errbit URL on line 45 also needs updating - otherwise :+1:

erikse:
Thanks, updated that. But I found a bunch of other references to preview in the docs folder, just looking at those

erikse:
Think I've got them all now, there are two historical references to the preview environment about how smart answers was previously developed, which doesn't seem correct to change.

daibach:
:+1:

erikse:
Just rebased this with master & force pushed back to this PR's branch

PR #2238, opened 2016-01-12
Increase the location cache max_size from 100 (default) to 250

erikse:
Smart Answers uses an LRU cache for storing world locations returned from Whitehall, but the [default max_size of 100](https://github.com/kindkid/lrucache/blob/0f6193e9bce0b88d571ad23ed7b1c330a72dedde/lib/lrucache.rb#L10) kicks out 116 of the 226 countries we currently offer in our dropdown:

For example:

 https://www.gov.uk/overseas-passports/y

 ```javascript
   $('#response > option').length;
   226
 ```

@dsingleton @alexmuller @jennyd

### DIFF
--- a/test/unit/world_location_test.rb
+++ b/test/unit/world_location_test.rb
@@ -260,4 +260,10 @@ class WorldLocationTest < ActiveSupport::TestCase
       end
     end
   end
+
+  context "cache configuration" do
+    should "have capacity for 250 cache entries" do
+      assert_equal WorldLocation.cache.max_size, 250
+    end
+  end
 end
--- a/app/models/world_location.rb
+++ b/app/models/world_location.rb
@@ -4,7 +4,7 @@ class WorldLocation
   extend Forwardable
 
   def self.cache
-    @cache ||= LRUCache.new(soft_ttl: 24.hours, ttl: 1.week)
+    @cache ||= LRUCache.new(max_size: 250, soft_ttl: 24.hours, ttl: 1.week)
   end
 
   def self.reset_cache


erikse:
I've just rebased this with master and forced pushed back to this PR's branch

test/unit/world_location_test.rb
```diff
@@ -260,4 +260,10 @@ class WorldLocationTest < ActiveSupport::TestCase
       end
     end
   end
+
+  context "cache configuration" do
+    should " have capacity for 250 cache entries" do
```
tadast: minor: extra space

tadast:
Was this discovered as a part of the memory leak investigations?

I guess we're not going to have an extra 26 countries in the world any time soon so this should do the trick. LGTM :+1:

erikse:
Not really, if anything this will use a bit more memory (maybe a few KB).

But it should cut down the number of remote calls across to Whitehall to fetch countries, as we have 226 countries now, and only 100 of those can be kept in memory at any one time.

floehopper:
Am I right in thinking that increasing the maximum size of the cache will mean that it will never reach capacity and thus the LRU aspect of the cache will no longer be used (at least with the current set of countries)?

I'm guessing we're still getting some benefit from the `LRUCache` class' TTL / soft-TTL configuration, although it's not immediately obvious to me what this is doing. Otherwise I might suggest we ditch the `LRUCache` class and use a simple `Hash` instead.

erikse:
At least until there more than 249 countries (we also have the 'all' key), there will be no LRU triggered eviction.

I guess the [soft_ttl functionality](https://github.com/kindkid/lrucache#soft_ttl) let's smart answers update its data once a day without a restart, so we'd lose that with a ```Hash``` but could probably achieve something similar with ```Rails.cache.fetch``` and the ```expires_in``` option.

floehopper:
Cool. I see that there are actually some unit tests which kind of explain the TTL and soft-TTL behaviour, so I reckon it's probably sensible to leave it be for now.

PR #2241, opened 2016-01-13
Use bearer token for publishing API authentication

danielroseman:
https://trello.com/c/sTGL6WGn/472-move-update-type-validation-to-content-item-and-check-change-note


erikse:
What happens if you're missing the ENV variable and it passes 'example' as the token?

danielroseman:
In production, it will fail with a 401. But we have already configured Puppet to ensure that the env variable is indeed there.

PR #2242, opened 2016-01-14
Add benefit calculator rates for London

erikse:
PT story: https://www.pivotaltracker.com/story/show/106198568

Introduces London specific benefit cap rates for the calculator, as well as updating the existing rates for all other non-London regions. Adds a postcode question to the end of the flow, to determine if the user is in London or not, and selects the appropriate rates for calculating the benefit cap.

## Fact check

https://pt-106198568.herokuapp.com/benefit-cap-calculator

## Expected changes
 * [URL on gov.uk](https://www.gov.uk/benefit-cap-calculator/y/yes/no/no/bereavement,carers,child_benefit/230.0/2.0/4.0/5.0/parent)
  * Introduces a new question at the end of the flow to lookup the postcode to determine if London or non-London rates should be applied by the calculator.

### Before
screen shot 2016-01-14 at 3 03 05 pm

### After
screen shot 2016-01-14 at 3 03 19 pm


lib/smart_answer/calculators/benefit_calculator.rb
```diff
@@ -0,0 +1,62 @@
...
+      @benefits[benefit]
+    end
+
+    def total_benefits
+      @benefits.values.inject(0) {|sum, value| sum + value }
+    end
+
+    def benefit_cap
+      single_couple_lone_parent == 'single' ? benefit_cap_rate(:single) : benefit_cap_rate(:couple)
+    end
```
floehopper: This method seems to implicitly assume that the `couple` and `parent` rates are the same. While this is true at the moment the `benefit_caps.yml` file suggests it may not always be true.

floehopper:
This PR definitely moves the `benefit-cap-calculator` in the direction of the new approach e.g. by extracting a calculator object from the flow and moving presentational logic out of the flow and into the templates. Yay! :smile:

I don't think any of my comments should prevent this from being merged.

erikse:
This PR is superseded by #2372

PR #2247, opened 2016-01-19
Add logging config for when deployed to Heroku

erikse:
As we use Heroku for our all out staging servers for fact checking processes, we need to point Rails logging to STDOUT for the application log to be available when on Heroku:

https://devcenter.heroku.com/articles/rails4

@dsingleton @jennyd

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -71,5 +71,9 @@ SmartAnswers::Application.configure do
     config.logstasher.enabled = true
     config.logstasher.logger = Logger.new("#{Rails.root}/log/#{Rails.env}.json.log")
     config.logstasher.suppress_app_logs = true
+  else
+    # enable STDOUT logging for Heroku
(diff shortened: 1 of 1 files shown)


erikse:
Just rebased with master, and force pushed back

config/environments/production.rb
```diff
@@ -71,5 +71,12 @@
     config.logstasher.enabled = true
     config.logstasher.logger = Logger.new("#{Rails.root}/log/#{Rails.env}.json.log")
     config.logstasher.suppress_app_logs = true
+  else
```
tadast: Unless-else seems to be considered as a something that reads badly. It would be better to invert the conditional and have it as if - else.

erikse: Ok, just inverted it for readability

daibach:
I think this is ok.

PR #2166, opened 2015-11-25
Child maintenance calculator update for November 2015

erikse:
PT story: https://www.pivotaltracker.com/story/show/103405692

Updates the calculator start page and two of the outcomes to include a note that the calculator gives an estimate rather than exact figure.

## Factcheck

https://fierce-sands-8457.herokuapp.com/calculate-your-child-maintenance

## Expected changes
 * [URL on gov.uk](https://www.gov.uk/calculate-your-child-maintenance)
  * Updates content on start page
  * Adds a callout with the text "The calculator only gives an estimate - it may not give the right result for complicated circumstances, eg if the paying parent has children from a different relationship"

### Before
screen shot 2015-12-15 at 5 26 00 pm

### After
screen shot 2015-12-15 at 5 26 05 pm

This PR supersedes #2165 #2193 and #2249

PR #2249, opened 2016-01-25
Child maintenance content nov2015 

stwalsh:
Minor amends: 2 small commits for the Child Maintenance Calculator from 25 Jan. 

They deal with some comments from the Child Maintenance agency, and a point I'd missed from 2i. 

If this looks good, it can be deployed (at last!).


erikse:
Thanks for this, I've cherry-picked the last two updates into the existing PR #2166

PR #2253, opened 2016-01-25
Upgrade Rails to version 4.2.5.1

erikse:
Updates smart answers to the use the latest patch release of Ruby on Rails, addressing a list of known CVEs.
```
[CVE-2015-7576] Timing attack vulnerability in basic authentication in Action Controller
[CVE-2016-0751] Possible Object Leak and Denial of Service attack in Action Pack
[CVE-2015-7577] Nested attributes rejection proc bypass in Active Record
[CVE-2015-7579] XSS vulnerability in rails-html-sanitizer
[CVE-2016-0752] Possible Information Leak Vulnerability in Action View
[CVE-2016-0753] Possible Input Validation Circumvention in Active Model
[CVE-2015-7578] Possible XSS vulnerability in rails-html-sanitizer
[CVE-2015-7581] Object leak vulnerability for wildcard controller routes in Action Pack
```
Changelog links:
 * [ActionMailer](https://github.com/rails/rails/blob/v4.2.5.1/actionmailer/CHANGELOG.md)
 * [ActionPack](https://github.com/rails/rails/blob/v4.2.5.1/actionpack/CHANGELOG.md)
 * [ActionView](https://github.com/rails/rails/blob/v4.2.5.1/actionview/CHANGELOG.md)
 * [ActiveModel](https://github.com/rails/rails/blob/v4.2.5.1/activemodel/CHANGELOG.md)
 * [ActiveRecord](https://github.com/rails/rails/blob/v4.2.5.1/activemodel/CHANGELOG.md)
 * [ActiveSupport](https://github.com/rails/rails/blob/v4.2.5.1/activesupport/CHANGELOG.md)
 * [ActiveJob](https://github.com/rails/rails/blob/v4.2.5.1/activejob/CHANGELOG.md)


erikse:
Pending fix in slimmer's resolver sub-class https://github.com/alphagov/slimmer/pull/138

erikse:
Seems like the render partial calls with prefixes of "../.." are getting rejected by the new [inside_path?(path, filename)](https://github.com/rails/rails/blob/4-2-stable/actionview/lib/action_view/template/resolver.rb#L231) method.
```ruby
From: /Users/erik/Coding/erikse/rails/actionview/lib/action_view/template/resolver.rb @ line 234 ActionView::PathResolver#inside_path?:

    231: def inside_path?(path, filename)
    232:   filename = File.expand_path(filename)
    233:   path = File.join(path, '')
 => 234:   binding.pry
    235:   filename.start_with?(path)
    236: end

[1] pry(#<ActionView::OptimizedFileSystemResolver>)> filename
=> "/Users/erik/Coding/alphagov/smart-answers/test/fixtures/smart_answer_flows/data_partials/_array_partial.erb"
[2] pry(#<ActionView::OptimizedFileSystemResolver>)> path
=> "/Users/erik/Coding/alphagov/smart-answers/test/fixtures/smart_answer_flows/data-partial-sample/outcomes/"
[3] pry(#<ActionView::OptimizedFileSystemResolver>)> filename.start_with?(path)
=> false
```

issyl0:
Might https://github.com/alphagov/publisher/commit/e7b4495bc32646d0b234a60c56cdd29468021ca8 be related? I don't know, just saw it and thought it might be useful.

erikse:
Yep, that looks like the same issue

erikse:
Regression is green
```
Finished in 3718.666119s, 5.9078 runs/s, 0.0307 assertions/s.

21969 runs, 114 assertions, 0 failures, 0 errors, 0 skips
```

tadast:
Looks good to me. Nicely dealt with the paths issue  :+1:

erikse:
Thanks @tadas !

floehopper:
I think this is definitely an improvement and it's great that it allowed the upgrade to Rails v4.2.5.1.

However, it would be nice if we could avoid proliferating calls to `FlowRegistry.instance`. Ideally ultimately it would be good to get rid of this singleton and replace it with explicit dependencies being passed to the relevant objects.

After this change there is now a call to `FlowRegistry.instance.load_path` in both `SmartAnswer::Node#template_directory` and `SmartAnswer::ErbRenderer#initialize`. Ideally we'd only be calling this in one place. I'm not sure where the right place is, but I'm not convinced that place is in the `ErbRenderer`. Does that make sense?

Also it would be nice to have a unit test which demonstrates the use of the shared path and/or explains why it's needed.

PR #2258, opened 2016-01-30
Marriage abroad amends for Kuwait

erikse:
PT story: https://www.pivotaltracker.com/story/show/109437846

Update content for any users resident in Kuwait getting married abroad.

## Fact check

https://pt-109437846.herokuapp.com/marriage-abroad/y

## Expected changes
 [URL on gov.uk](https://www.gov.uk/marriage-abroad/y/kuwait/ceremony_country/partner_british/opposite_sex)
  * Under supporting documents, proof of residency should be Civil ID

### Before
screen shot 2016-01-30 at 2 57 58 pm

### After
screen shot 2016-02-01 at 2 17 32 pm


floehopper:
Minor: I have mixed feelings about having two pairs of commits updating the regression test artefacts and checksums in one PR like this. In one sense it's nice to be able to see the smaller diff for the second pair, but conversely why then do we not update them after each content change commit, although I know the latter is probably because it takes a long time to run the regression tests. In this case, I think I'd be tempted to combine the two pairs of commits into one pair at the end of the PR, but it's not a big deal.

floehopper:
Other than [my comment about the inconsistent name of the partial filename](https://github.com/alphagov/smart-answers/pull/2258#commitcomment-15931876), this should be ok to merge.

erikse:
I've removed the first pair of artefact/checksum updates, I think they were left over after the work was first sent for fact check, and we had some additional changes requested afterwards.

erikse:
Have now rebased with master

PR #2259, opened 2016-02-01
Update therubyracer gem from v0.12.1 to v0.12.2

erikse:
Updating to version 0.12.2 which was released with PR [FIX: severe memory leak in WeakValueMap](https://github.com/cowboyd/therubyracer/pull/336)


Gemfile.lock
```diff
@@ -295,4 +295,4 @@ DEPENDENCIES
   webmock (= 1.20.4)
 
 BUNDLED WITH
-   1.10.6
+   1.11.2
```
benlovell: This will probably cause an issue given the current ruby version is deployed with bundler `1.10.6`

erikse: Yeah, we don't need this, can revert

benlovell:
I'd argue we should drop this dependency altogether. `ExecJS` will find a JavaScript runtime and I'd wager most (if not all) developers have a recent node.js and our app servers have node.js out-of-the-box.

There's prior art for dropping this dependency elsewhere also.

erikse:
I'm not sure, I think in the short term, this might help with the memory issues we are seeing on smart answers, and brings therubyracer version inline with what's running in frontend etc.

erikse:
Maybe relevant? https://github.com/reactjs/react-rails/pull/290

benlovell:
Dropping the dependency altogether will remove it from memory entirely :confused: 

This has been done before (in examples below and several others including alphagov/whitehall) and we're not seeing any issues. 

alphagov/calendars#104
alphagov/publisher#438

benlovell:
@erikse If this is causing issues currently and needs resolving quickly, feel free to merge and bring up the discussion after.

erikse:
I'm not sure it'll make much of a difference, I can't find any explicit dependency on therubyracer in the smart answers code, so I guess it's just used for compiling the sass/assets

erikse:
This PR has been superseded by #2260

PR #2260, opened 2016-02-01
Remove therubyracer gem

erikse:
This PR supersedes #2259 and removes the therubyracer gem from smart answers in favour of having node.js running on machines precompiling assets during deployment, similar to:

alphagov/calendars#104
alphagov/publisher#438

The side effect of this is it introduces an implicit dependency that is no longer resolved by bundler - so being able to deploy is now dependent the user having node.js installed or reading the README


benlovell:
:heart_eyes:

PR #2263, opened 2016-02-09
Fix typos in the regression test docs

issyl0:
- "changed" => "changes".
- Change a plural "regression test are" to "regression test is" to fit
  in with the rest of the paragraph.


issyl0:
:sunny:

PR #2269, opened 2016-02-11
Update to Smart Answers to use Ruby version 2.3 

erikse:
This is a working branch, there seems to be a change in OpenStruct which we sub-class in ```state.rb```. 

In Ruby 2.2.4
```
Eriks-MacBook-Pro:smart-answers erik$ git checkout master
Switched to branch 'master'
Eriks-MacBook-Pro:smart-answers erik$ rvm use 2.2.4
Using /Users/erik/.rvm/gems/ruby-2.2.4
Eriks-MacBook-Pro:smart-answers erik$ rails c
Loading development environment (Rails 4.2.5.1)
```
```ruby
2.2.4 :001 > state = SmartAnswer::State.new('test')
 => #<SmartAnswer::State current_node="test", path=[], responses=[], response=nil, error=nil> 
2.2.4 :002 > state.methods.include?(:error)
 => true 
```
In Ruby 2.3.0
```
Eriks-MacBook-Pro:smart-answers erik$ git checkout update_to_ruby_2_3
Switched to branch 'update_to_ruby_2_3'
Eriks-MacBook-Pro:smart-answers erik$ rvm use 2.3.0
Using /Users/erik/.rvm/gems/ruby-2.3.0
Eriks-MacBook-Pro:smart-answers erik$ rails c
Loading development environment (Rails 4.2.5.1)
```
```ruby
2.3.0 :001 > state = SmartAnswer::State.new('test')
 => #<SmartAnswer::State current_node="test", path=[], responses=[], response=nil, error=nil> 
2.3.0 :002 > state.methods.include?(:error)
 => false
```
In ```flow#process``` the call to ```state.error```
```ruby
def process(responses)
  responses.inject(start_state) do |state, response|
    return state if state.error
    ...
  end
end
```
probably triggers ```state#method_missing``` when running in Ruby 2.3, which will raise NoMethodError as it only passes methods ending in ```=``` to ```OpenStruct```

```ruby
def method_missing(method_name, *args)
  if method_name =~ /=$/
    super
  else
    raise NoMethodError.new("undefined method '#{method_name}' for #{self.class}")
  end
end
```

Note: A user report of speed and memory improvements (the later maybe more interesting to smart answers) running Rails with Ruby 2.3 http://ruby-performance-book.com/blog/2016/02/is-ruby-2-3-faster-rails-erb-template-rendering-performance.html


floehopper:
Ultimately I'd like to move away from using `OpenStruct` for the `State` class and the extraction of calculator/policy objects (as described [here](https://github.com/alphagov/smart-answers/blob/master/doc/refactoring.md)) is gradually achieving this.

In the meantime, I wonder whether the right solution here is to explicitly define `current_node`, `path`, `responses`, `response`, and `error` using `attr_accessor` on the `State` class. Although I have to confess, I don't know how doing that will interact with the `OpenStruct` behaviour.

floehopper:
The `CGI.escapeHTML` optimisation (https://github.com/ruby/ruby/pull/1164) examined in that article is really interesting. Well spotted!

floehopper:
I'm going to have a look at this to see whether I can fix the problem with `SmartAnswer::State` that @erikse ran into.

floehopper:
This has been superseded by #2345.

PR #2280, opened 2016-02-17
SPIKE: Introduce Period of Incapacity for Work (PIW) into calculate-statutory-sick-pay

floehopper:
When I was fixing problems with `calculate-statutory-sick-pay` at the end of last year, I realised that there was at least one missing concept in the code - the [Period of Incapacity for Work][1] (PIW). The smart answer deals with a current PIW and an (optional) linked PIW. I felt that it ought to be possible to share some of the logic connected with these two PIWs. See this [Trello card][2] for details.

This pull request contains my initial efforts at this. Unfortunately it turned out that the code was even more tangled and incomprehensible than I remembered and it's proved hard to get as far as I would've liked. Also I think I've discovered a bug in the validation logic for a linked PIW. I've contacted the Content Team about this and they are asking HMRC for clarification.

For these reasons and the fact that this smart answer doesn't tend to get updated very often means that I've decided to park this work for now.

I'm also conscious that my changes would mean the code diverging further from the [logic document][3]. I'm not convinced that the latter is a reason not to make the changes - in fact quite the opposite - it makes me more convinced that the idea of the logic document process is flawed - but I thought it was worth noting.

I'm fairly happy that most of the changes in this PR will be worth making once we hear back from HMRC. The only thing I've become a bit less convinced about is the two commits titled:

* "Replace 2 current PIW instance vars with 1 in SSP calculator"
* "Replace 2 linked PIW instance vars with 1 in SSP calculator"

On other thing I've noticed is that at the moment we pass the value of the last response into validation methods as a parameter, because the value hasn't been set on the calculator. It would make life a lot easier if we adopted a more `ActiveModel` style validation approach whereby the value was set on the calculator and then a method like `valid?` was called.

[1]: http://www.hmrc.gov.uk/manuals/spmmanual/spm110500.htm
[2]: https://trello.com/c/uu0HeZee
[3]: https://drive.google.com/open?id=1qHAJUhXYR1ophsUkewRmkiSeY39StsElMvJzKMlhMTc

### DIFF
--- a/lib/smart_answer/calculators/statutory_sick_pay_calculator.rb
+++ b/lib/smart_answer/calculators/statutory_sick_pay_calculator.rb
@@ -117,7 +117,9 @@ module SmartAnswer
       end
 
       def at_least_1_day_before_first_sick_day?(value)
-        value < sick_start_date - 1
+        prospective_linked_piw = linked_piw.ending_on(value)
+        gap = prospective_linked_piw.gap_between(current_piw)
+        !gap.empty?
       end
 
       def valid_period_of_incapacity_for_work?
--- a/lib/smart_answer/calculators/statutory_sick_pay_calculator.rb
+++ b/lib/smart_answer/calculators/statutory_sick_pay_calculator.rb
@@ -1,9 +1,9 @@
 module SmartAnswer
   module Calculators
     class StatutorySickPayCalculator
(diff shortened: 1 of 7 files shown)


floehopper:
As I explain in the description, I want to wait until I hear back from HMRC about a potential bug before trying to merge this. So I'm going to close it for now, but at least it will still be available for reference.

PR #2283, opened 2016-02-17
Docker Development environment

ikennaokpala:
Here, I share some scripts that assist in setting up a docker environment for development purposes

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -70,4 +70,4 @@ EXPOSE 5000
 
 # Setup the entrypoint
 ENTRYPOINT ["/bin/bash", "-l", "-c"]
-CMD ["/smart-answers/startup.sh"]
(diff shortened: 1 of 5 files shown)


floehopper:
I don't use it, but I imagine there is some overlap between this and the [development VM][1] in use across GOV.UK. I'll leave it to someone who knows about the development VM to comment. If there is significant duplication, I'm not sure it would be such a good idea to merge this.

[1]: https://github.com/alphagov/govuk-puppet/tree/master/development

floehopper:
@daibach: Can you recommend someone who knows about the development VM to review this? (see my comment above).

floehopper:
@ikennaokpala: Unless you can find someone suitable to review this, I think we should probably close the PR for now. The work won't be lost, but it'll stop it cluttering up the list of open PRs.

PR #2288, opened 2016-02-19
Move logic from country_of_ceremony question to MarriageAbroadCalculator

chrisroos:
I've moved the logic from the `calculate` blocks in the `:country_of_ceremony?` question to methods on the `MarriageAbroadCalculator`. I've added unit tests for these methods to describe the existing behaviour.

The next step will be to remove these `calculate` blocks and access the methods on the calculator directly in the flow, questions and outcomes.

I've used `stub.quacks_like(<instance-of-object>)` in a number of tests. This is instead of using a real instance of the object, or of using a plain stub. It gives me confidence that I'm not accidentally relying on behaviour in the real object, and that I'm not stubbing/calling a method that doesn't exist on the real object.

### DIFF
--- a/lib/smart_answer/calculators/marriage_abroad_calculator.rb
+++ b/lib/smart_answer/calculators/marriage_abroad_calculator.rb
@@ -108,5 +108,19 @@ module SmartAnswer::Calculators
     def country_name_uppercase_prefix
       @country_name_formatter.definitive_article(ceremony_country, true)
     end
+
+    def country_name_partner_residence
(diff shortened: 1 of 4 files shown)


floehopper:
Apart from my all minor comments, this looks good to me.

chrisroos:
Thanks for reviewing this, @floehopper. I'm going to rebase on master and open a new PR to preserve these comments.

chrisroos:
This has been superseded by PR #2297.

PR #2281, opened 2016-02-17
Update outcomes for countries in Electronic Visa Waiver program

erikse:
Pivotal ticket: https://www.pivotaltracker.com/story/show/113366241

Updates an existing outcome with corrections for countries included EVW program and adds three new outcomes specific to the purpose of visit in cases of work (for less than 6 months), study or to visit a child in school.

## Fact check

https://pt-113366241.herokuapp.com/check-uk-visa

## Expected changes
 [URL on gov.uk](https://www.gov.uk/check-uk-visa/y/kuwait/work/six_months_or_less)
  * H2 heading Business visits becomes Business and academic visits
  * adds the option of applying for an Electronic Visa Waiver (EVW) in the first option under business and academic visits

### Before
screen shot 2016-02-17 at 2 09 56 pm

### After
screen shot 2016-02-17 at 2 10 16 pm

Note: this PR supersedes #2274, #2275, #2276 and #2277

PR #2275, opened 2016-02-16
New outcome for EVW countries (work< 6mth)

Lilboyce:
Outcome based on outcome_work_m.govspeak.erb

Changes based on:
https://govuk.zendesk.com/agent/tickets/1239005
https://www.pivotaltracker.com/story/show/112416855

2i-ed by Al

PR #2289, opened 2016-02-19
Update outcomes for countries in Electronic Visa Waiver program

erikse:
Pivotal ticket: https://www.pivotaltracker.com/story/show/113366241

Updates an existing outcome with corrections for countries included EVW program and adds three new outcomes specific to the purpose of visit in cases of work (for less than 6 months), study or to visit a child in school.

## Fact check

https://pt-113366241.herokuapp.com/check-uk-visa

## Expected changes
 [URL on gov.uk](https://www.gov.uk/check-uk-visa/y/kuwait/work/six_months_or_less)
  * H2 heading Business visits becomes Business and academic visits
  * adds the option of applying for an Electronic Visa Waiver (EVW) in the first option under business and academic visits

### Before
screen shot 2016-02-17 at 2 09 56 pm

### After
screen shot 2016-02-17 at 2 10 16 pm

Note: this PR supersedes #2274, #2275, #2276, #2277 and #2281


erikse:

floehopper:
Observations:

1. It would've been nicer if the addition of each of the new outcomes was done in a single atomic commit i.e. if the outcome was added to the flow in the same commit as the template was added. However, I assume it wasn't done this way, because the template was added by a member of the Content Team and the flow was changed by a developer. Either way, I'd be interested to understand how it came about.

2. It looks as if quite a bit of duplication has been introduced in at least one of the new outcomes - `outcome_work_waiver` is almost exactly the same as `outcome_work_waiver `. I'm guessing this might be because the member of the Content Team didn't know they could use partial templates to reduce the duplication? If so, it would be good to extract any relevant partials.

3. I think there's a balance to be struck between having more outcomes and having more "routing" conditionals in the flow versus having fewer outcomes and having conditionals in the templates. In this case, I might've been inclined to use fewer outcomes and more conditional logic in the templates, but perhaps that's also a result of the Content Team not knowing how to use conditionals in templates? I'd be interested to know if this was the case.

4. It looks as if you've chosen to cover the new outcomes in integration tests i.e. there don't appear to be any new responses for the regression test. If that's true, it might be worth explaining why you chose to do that in the relevant commit note(s).

However, all of the above are minor & subjective, and otherwise this looks good to me. I apologise if I've repeated anything @chrisroos said - I haven't yet had a chance to read his comments on the original PR.

erikse:
Thanks for these comments, just briefly:

1. The content team uses the github web UI for submitting PRs so they are restricted to only being able to edit or add one file per commit. In some cases, the original commits are followed with subsequent commits with changes from 2i and fact check. So in these cases, I think it's clearer to retain the original contributions and update the code/tests around them so it's easier for the content editor to add additional contributions later on.

2. I'm not sure which outcome template you mean, you mentioned the same one. I think these new ones don't seem to be any worse in terms of duplication than the ones already in place, for example: ```outcome_work_n``` and ```outcome_work_m``` repeats content, so I think this is following the existing pattern.

3. This seems to follow the existing pattern also - in all the outcomes of the check-uk-visa smart answers, there is no conditional branching controlling what content is shown in an outcome template.

4. I will add Kuwait to the regression test!

floehopper:
The responses & expected results file wasn't fully updated in this PR. I have fixed this in #2524.

PR #2290, opened 2016-02-19
Remove calculate blocks from country_of_ceremony question in marriage-abroad

chrisroos:
This builds on the work in PR #2288.

I've updated the flow, questions and outcomes to use the methods on `MarriageAbroadCalculator` directly, and then removed the unused `calculate` blocks from the flow. This is a step toward moving the policy logic from the flow to the calculator as described in the [refactoring doc](https://github.com/alphagov/smart-answers/blob/master/doc/refactoring.md).

Removing some of the `calculate` blocks caused some assertions to fail in marriage_abroad_integration_test.rb. In the past I think we've deleted these failing assertions but in this branch I updated them so that they asserted against the method on the calculator object (see 25030e60587c0bc0b649df7a938212df82954094 for an example). I'm not really sure about this. I _think_ it might be better to remove the failing assertions because they're probably duplicating the tests that I now have in marriage_abroad_calculator_test.rb. What do you think, @erikse / @floehopper?


chrisroos:
I realise that this can't be automatically merged to master but I don't think that should stop someone reviewing it.

chrisroos:
I've rebased this on master and force pushed now that PR #2297 has been merged. There were no comments on the commits in this branch so I've not lost anything by rebasing/force pushing.

floehopper:
I think the PR description is now out-of-date since you've rebased - I don't think this sentence is relevant any more: "Ignore the first 9 commits (all starting with "Add MarriageAbroadCalculator#...") as they're in the other PR."

floehopper:
Minor: Quite a few commits have commit note subject lines of more than 50 characters. I suspect you've done this intentionally, because you think the long method names are important to distinguish the (small, focussed) commits. I guess it would be useful to have some kind of consistent criteria for deciding when we can ignore the 50 character guideline.

floehopper:
Very minor: As I mentioned in [one of my comments][1], it would be nice at some point to tidy up all the `locals` Hashes passed into partial templates now that many of them are much shorter. However, that doesn't need to be done as part of this PR.

[1]: https://github.com/alphagov/smart-answers/commit/121b0a160a5436521dff9f6ffef46958e7d738f8#commitcomment-16293597

floehopper:
This all looks good to me :+1:

chrisroos:
> I think the PR description is now out-of-date since you've rebased - I don't think this sentence is relevant any more: "Ignore the first 9 commits (all starting with "Add MarriageAbroadCalculator#...") as they're in the other PR."

Good spot. I've updated the PR description.

chrisroos:
> Very minor: As I mentioned in one of my comments, it would be nice at some point to tidy up all the locals Hashes passed into partial templates now that many of them are much shorter. However, that doesn't need to be done as part of this PR.

I've made this change in my local branch because if I don't do it now I can imagine not getting round to it.

chrisroos:
> Minor: Quite a few commits have commit note subject lines of more than 50 characters. I suspect you've done this intentionally, because you think the long method names are important to distinguish the (small, focussed) commits. I guess it would be useful to have some kind of consistent criteria for deciding when we can ignore the 50 character guideline.

It is indeed intentional. I tried a few variations but struggled to come up with anything shorter that still summarised the change.

I'm not sure I'd be able to come up with a criteria for when to ignore the 50 character guideline.

chrisroos:
I've rebased my local branch on master. I'm going to close this PR and delete this branch so that I can open a new PR with my changes.

chrisroos:
This has been superseded by PR #2307.

PR #2294, opened 2016-02-22
Marriage abroad spain replace download links

ikennaokpala:
PT story: https://www.pivotaltracker.com/projects/1270592/stories/110175582

I have replaced the PDF download links on all Spanish marriage outcome pages with links to certificate of no impediment and marital status respectively.

* https://www.gov.uk/government/publications/example-certificate-of-no-impediment-cni (Certificate of no Impediment)
* https://www.gov.uk/government/publications/marital-status-example (Marital Status)

## FactCheck
[Marriage abroad spain replace download links](https://smart-answers-pr-110175582.herokuapp.com/marriage-abroad/y/spain/uk/partner_british/opposite_sex)

## Expected Changes
* [URL on gov.uk](https://www.gov.uk/marriage-abroad/y/spain/uk/partner_british/opposite_sex)
     * Remove PDF download links on Spanish outcome pages.
     * Change links to point to the certificate of no impediment and marital status on Spanish outcome pages respectively.

### Before
screen shot 2016-02-22 at 15 29 03

### After

screen shot 2016-02-25 at 16 12 13

### DIFF
--- a/lib/smart_answer_flows/marriage-abroad/outcomes/outcome_spain.govspeak.erb
+++ b/lib/smart_answer_flows/marriage-abroad/outcomes/outcome_spain.govspeak.erb
@@ -35,7 +35,7 @@
   <%= render partial: 'what_you_need_to_do.govspeak.erb',
       locals: { calculator: calculator } %>
 
-  You'll be asked to provide a [certificate of no impediment (CNI)](//gov.uk/government/publications/example-certificate-of-no-impediment-cni), [marital status](//gov.uk/government/publications/marital-status-example) certificate or both to prove you’re allowed to marry.
+  You'll be asked to provide a [certificate of no impediment (CNI)](//www.gov.uk/government/publications/example-certificate-of-no-impediment-cni), [marital status](//www.gov.uk/government/publications/marital-status-example) certificate or both to prove you’re allowed to marry.
 
   <% if calculator.resident_of_third_country? %>
     Check with the civil registry  where you’re getting married to find out what certificate you’ll need.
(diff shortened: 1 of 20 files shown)


floehopper:
Apart from the relative URLs this looks good to me.

erikse:
We'll get a `rel='external'` attribute on the anchor tag if we specify a domain, including an icon to show it's not an gov.uk page. Maybe we should raise a ticket in govspeak - so that it allows the gov.uk domain without generating the attribute.

floehopper:
@erikse:

Are you suggesting that these link URLs need a domain? i.e. that they *cannot* be relative links? Surely if they are relative links they will *not* be marked as external which is what we want (isn't it?). I'm a bit confused!

In case it's useful, I seem to remember that I originally added the external link behaviour to Govspeak (!) and I think there's a `document_domains` option in which you can list domains that should be considered as internal.

erikse:
@floehopper - just that it's weird that gov.uk isn't recognised as an internal link by govspeak

floehopper:
@erikse: `gov.uk` is redirected to `www.gov.uk`, so it makes sense to me.

floehopper:
I'm still confused why these need to be absolute URLs. The vast majority of GOV.UK URLs referenced from within Smart Answers are relative URLs e.g. search for `"/government/"` to see all the links to content served by the "Whitehall" app. I think ideally all such URLs should be relative.

erikse:
@floehopper - I think we're all in agreement on this, and it looks like the PR has been updated to reflect that.

floehopper:
@erikse: Sorry, I must've missed the later commit. Looks good.

PR #2295, opened 2016-02-22
Update the govuk-lint gem to version 0.5.3

erikse:
Moves smart answers from version `0.3.0` to the latest  version of the `govuk-lint` gem to ensure we are up to date with the latest cops/fixes.

In turn, `govuk-lint` updates its dependency on `Rubucop` from version `0.32` to `0.35.0`.


floehopper:
Looks good to me :+1:

PR #2298, opened 2016-02-24
WIP: Delegate to MarriageAbroadDataQuery from MarriageAbroadCalculator

chrisroos:
My main motivation is to have a single object that encapsulates all the logic required by the marriage-abroad flow and views.

This branch isn't finished but I'd appreciate a second pair of eyes on it to ensure that it's a sensible direction.

Ignore the first commit, which simply merges all the changes from PR #2297.


floehopper:
Overall I think this is a good approach. If nothing else, I think it highlights for me that there are a bunch of methods relating to the ceremony country. I can imagine in the future that it might be worth extracting some kind of `Country` class and moving this group of methods onto that.

As I think I've mentioned in the past, it feels to me as if the client code would be simpler if you could call something like `ceremony_country.french_overseas_territory?`. At the moment it feels as if everything is based on loads of lists of countries. I think this is useful for optimising bulk updates, but not so good for the "client" code.

floehopper:
I have to confess I'm a bit nervous about some of the renaming of methods you've done as part of this PR. I think your motivation is good, but I think we should only rename where we're *very confident* that the new name is correct. Otherwise I think it might be better to stick to the original names, albeit expanding any abbreviations.

floehopper:
Other than my mostly minor comments, this looks good to me so far. Please continue!

chrisroos:
> I have to confess I'm a bit nervous about some of the renaming of methods you've done as part of this PR. I think your motivation is good, but I think we should only rename where we're very confident that the new name is correct. Otherwise I think it might be better to stick to the original names, albeit expanding any abbreviations.

I became more nervous, the more I started changing the names. I'm going to revise the commits and stick to the original names as much as possible. Renaming can come later.

chrisroos:
I've started reworking this branch and sticking to the original method names as closely as possible. I'll close this for now and re-open a new PR with that branch soon.

PR #2301, opened 2016-02-24
Update the sass and sass-rails gems to latest versions

erikse:
To be able to update the `govuk-lint` gem to the latest version 0.6.2, we need to update our version of the `sass` gem to a newer version.

`govuk-lint` gets the `sass` gem via the `scss_lint` gem, which specifies `~> 3.4.15` in the [gemspec](https://github.com/brigade/scss-lint/blob/master/scss_lint.gemspec#L29)

Smart Answers gets the `sass` gem via the `sass-rails` gem.

In the `sass-rails` gem [release notes](https://github.com/rails/sass-rails/releases), affecting changes seem to be only really in [version 5.0.0](https://github.com/rails/sass-rails/releases/tag/v5.0.0):

- allow support for sass ~> 3.1
- deprecates css.scss and css.sass extensions, neither in use within smart answers
- limit require to css files, and @import to .scss files, which smart answers already does

In the [sass gem changelog](https://github.com/sass/sass/blob/stable/doc-src/SASS_CHANGELOG.md) there are really an overwhelming number of changes between 3.2.19 and 3.4.21!

## Preview

https://smart-answers-pr-sass-update.herokuapp.com/marriage-abroad


floehopper:
It might be worth explaining in the commit note and/or PR description that this change is [necessary for govuk-link >= v0.6.0][1].

[1]: https://github.com/alphagov/govuk-lint/blob/master/CHANGELOG.md#060

floehopper:
This looks good to me. I'd hope that the fact that we're using sass-rails should insulate us a bit from the changes to sass itself. Besides, it doesn't look as if we're doing anything complicated in our SASS files. We could conceivably do the upgrades in smaller steps, but I don't think I would bother.

:+1:

erikse:
Have update the commit message and PR description to specify the version of the govuk-lint gem that this is needed for.

PR #2292, opened 2016-02-19
Update next node in check UK visa

erikse:
This PR reorganises the conditional logic in a next_node block for the check UK visa flow, moving the branching logic related to the current question to the top level, and nesting any sub-branching within each top level branch.

Related to #2281 and #2289

PR #2306, opened 2016-02-24
Update next node in check UK visa smart answer

erikse:
This PR reorganises the conditional logic in a `next_node` block for the check UK visa flow, moving the branching logic related to the current question to the top level, and nesting any sub-branching within each top level branch.

Related to #2281, #2289 and #2292

### DIFF
--- a/lib/smart_answer_flows/check-uk-visa.rb
+++ b/lib/smart_answer_flows/check-uk-visa.rb
@@ -82,9 +82,13 @@ module SmartAnswer
 
           if calculator.study_visit? || calculator.work_visit?
             next :staying_for_how_long?
-          elsif calculator.diplomatic_visit?
+          end
+
+          if calculator.diplomatic_visit?
             next :outcome_diplomatic_business
-          elsif calculator.school_visit?
+          end
+
+          if calculator.school_visit?
             if calculator.passport_country_in_electronic_visa_waiver_list?
               next :outcome_school_waiver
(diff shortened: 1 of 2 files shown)


erikse:
Alternatively to the second commit, we could just add a `.rubocop.yaml` to the project and turn off this cop for the whole repo.

I'm not convinced banning single line if/unless statements is necessarily always a good restriction for readability such as when we have a sequence of if statements such as the next_node block of this PR, or just long method names that could leave the if condition off screen when folded down to one line, e.g.:
```ruby
if some_condition
  do_a_really_long_method_name_that_might_take_up_most_of_the_screen_width_with_no_room_left
end
```
verses:
```ruby
do_a_really_long_method_name_that_might_take_up_most_of_the_screen_width_with_no_room_left if some_condition
```

floehopper:
I agree with your comment about the `Style/IfUnlessModifier` cop not always giving the most readable code and I'd certainly be happy for it to be disabled for the whole project.

However, given that the whole point of the `govuk-lint` gem is presumably to enforce a common approach across all GDS projects, if we do disable it in Smart Answers, I think we ought to put the arguments for doing so in a PR against `govuk-lint` to disable the cop by default. I think this would be the best way to communicate what we're doing. What do you think?

floehopper:
Having said that, I think this PR looks good as it is and I'd be happy for it to be merged.

erikse:
This PR is now superseded by #2310

PR #2308, opened 2016-02-25
Update regression tests artefacts for marriage-abroad

chrisroos:
These differences were caused by
2583d8ec2fd113f5e9a06766d7b9f0cc531d20fd. I should've updated these
artefacts as part of PR #2307 but completely failed to do so. Sorry!


chrisroos:
I've merged this despite not having anyone look at it because the risk is minimal and I want to get the regression tests passing again.

PR #2309, opened 2016-02-25
Update the `govuk-lint` gem to version 0.7.0

erikse:
Version 0.7.0 was released to disabled the `IfUnlessModifier` cop dis-allowing single line blocks inside `if` and `unless` statements.

For smart answers, the change will allow us to write sequences of `if` statements in a consistent form, regardless of how many lines are inside the block.

If accepted, we can remove the second commit adding Rubocop `enable` and `disable` directives to the `check-uk-visa` smart answer in PR #2306


floehopper:
Great! :+1:

PR #2310, opened 2016-02-25
Update next node for check UK visa

erikse:
This PR reorganises the conditional logic in a `next_node` block for the check UK visa flow, moving the branching logic related to the question `purpose_of_visit` to the top level within the block, and nesting any sub-branching within each if statement for the particular purpose of visit.

Related to #2281, #2289, #2292 and #2306


floehopper:
Looks good to me. :+1:

PR #2311, opened 2016-02-25
WIP: Automatically detect possible next nodes

floehopper:
This is an investigation following on from [this Trello card][1].

This introduces an alternative style of `next_node` block. Instead of returning the key for the next node from the block, we now call the new `goto_node` method. We can then reliably detect all the possible next nodes by parsing the source code of the `next_node` block and finding all the calls to the `goto_node` method.

The advantage of doing this is that we reduce duplication and make the flow definitions terser, hopefully improving their readability.

The way I've implemented this means that we could gradually migrate flows across to the new style. Once all flows were migrated, we could simplify the code further.

See the individual commit notes for more details.

I'm submitting this as a PR in order to solicit feedback - please let me know what you think.

## Before

```ruby
multiple_choice :question? do
  option :option_one
  option :option_two

  permitted_next_nodes = [
    :outcome_one,
    :outcome_two
  ]

  next_node(permitted: permitted_next_nodes) do |response|
    case response
    when :option_one
      :outcome_one
    when :option_two
      :outcome_two
    end
  end
end
```

## After

```ruby
multiple_choice :question? do
  option :option_one
  option :option_two

  next_node do |response|
    case response
    when :option_one
      goto_node :outcome_one
    when :option_two
      goto_node :outcome_two
    end
  end
end
```

[1]: https://trello.com/c/K66eybKM


chrisroos:
I think this is a great improvement. Good work!

I'm interested in whether this will have a detrimental effect on flow start-up/processing but I can see you've already got that on your todo list.

I've only scanned the code so let me know if you want me to look at it in more detail.

erikse:
I'm seeing about 7 times slower:
```ruby
require 'benchmark'

Benchmark.bmbm do |x|

  x.report('Next') do
    10_000_000.times do
      next :benchmarking
    end
  end

  x.report('Catch/Throw') do
    10_000_000.times do
      catch(:benchmarking) do
        throw(:benchmarking)
      end
    end
  end
end
```

```bash
$ ruby benchmark.rb

Rehearsal -----------------------------------------------
Next          0.490000   0.000000   0.490000 (  0.488633)
Catch/Throw   3.350000   0.010000   3.360000 (  3.355013)
-------------------------------------- total: 3.850000sec
```
Maybe there is an impact on memory/GC frequency if it has to save the stack frame each time

erikse:
Just as context to the last comment, I guess that's 3.5 seconds / 10,000,000 so 350 nanoseconds. Given there is only 1 throw/catch per request (to go to the next node), I don't think this an issue.

erikse:
Well, maybe that's wrong - and it's not 1 per request, but 1 per node in the flow to find out the next node?

floehopper:
@chrisroos, @erikse: Thanks for your comments.

I've done some more thinking about this and I'm not entirely happy that this is the right direction. I don't think it's worth thinking about the performance and/or memory usage until we're convinced it's the right direction. I've tried to step back a bit from the problem and here are some of my notes.

* Both the use of `throw` & `catch` and the use of method-source/ruby-parsing seem like quite "hacky" solutions.

* We only need to use `throw` & `catch` if we want to support multiple exits from the `next_node` block e.g. the use of `next`. I'm pretty sure that any combinatorial conditional with multiple exits can be transformed into a combinatorial conditional with a single exit, albeit possibly at the cost of readability.

* The method-source/ruby-parsing approach (i.e. 2nd commit) could be decoupled from the introduction of `goto_node` and use of `throw` & `catch` (i.e. 1st commit), although it might be harder to come up with a robust implementation of the Ruby parsing.

* In fact it should be possible to use the the method-source/ruby-parsing approach without *any* changes to the `next_node` blocks. However, the Ruby parsing would definitely be significantly more complicated.

* I'm pretty sure we only need to be able to work out the permitted next nodes in order to generate a visualisation of all the paths through the flow. If we're using something like the method-source/ruby-parsing approach (as per the 2nd commit), then I think we could move the code somewhere so that it is only executed when we're trying to visualise the flow and not during normal user interactions.

* I'd probably be a lot happier with the method-source/ruby-parsing approach if we were only ever doing it in the visualisation code, because then it wouldn't feel central to the app.

* If we move as much of the policy logic out of flow `next_node` blocks and into "policy" objects, we might be able to re-introduce a simplified version of the predicate DSL while retaining readability/familiarity. This would give us a way to introspect the conditional logic in order to generate the visualisation.

* In any case, I think moving as much of the policy logic out of the `next_node` blocks would a good thing to do. I suspect it might make it easier to avoid the need for multiple exits i.e. the use of `next`.

* I think we need to decide how valuable/sensible it is to allow arbitrary Ruby code in `next_node` blocks. I think the main advantages are the flexibility and familiarity i.e. no need to re-invent Ruby `if`, `else`, `case`, etc. A disadvantage is that not all the flexibility is desirable e.g. it might be sensible (at some point) to prevent any modification of state within the `next_node` block. Another disadvantage is that it's harder to analyse arbitrary Ruby code for the visualisation.

* Perhaps the pragmatic approach is to handle the multiple exit scenario with `throw` & `catch` for the moment, but to aim to remove this code in the future. If we also move the method-source/ruby-parsing code so it's only executed for visualisation, I think that might be a combination that I can live with.

Let me know what you think. Is there anything I've missed?

chrisroos:
I found this really useful, @floehopper. Particularly pointing out that we only need to specify `permitted_next_nodes` to support Smart Answer visualisation.

At the risk of repeating what you’ve written, and assuming I understand correctly, I think there are two problems we’re trying to solve with this spike:

1. Remove duplication. We want to avoid having to specify `permitted_next_nodes` while retaining the ability to visualise the Smart Answers.
2. Avoid unusual Ruby syntax. I don’t think the use of `next :<symbol>` to return a value from a block is very common and might cause confusion for developers new to the project.

Using method-source/ruby-parsing in the visualisation code sounds like a good approach to solving number 1.

Converting “combinatorial conditional with multiple exits […] into a combinatorial conditional with a single exit”, along with moving as much logic out of the flow as possible, sounds like the right approach to number 2.

At the moment, I’m inclined to think that number 1 is the higher priority of those two points.

floehopper:
@chrisroos: Thanks. I agree with your conclusion. I'm going to close this PR and focus on addressing point 1 in a new PR.

PR #2315, opened 2016-02-27
Use `instance_eval` instead of `eval` for `use_shared_logic`

floehopper:
This change reduces the scope of the context in which the code in the shared
logic files is executed which feels like a good thing to do.

This change also supplies a file path and a line number to the call to
`instance_eval` which means we get much better error messages when there is a
`SyntaxError` in a shared logic file.

### DIFF
--- a/lib/smart_answer/flow.rb
+++ b/lib/smart_answer/flow.rb
@@ -23,7 +23,8 @@ module SmartAnswer
     end
 
     def use_shared_logic(filename)
-      eval File.read(Rails.root.join('lib', 'smart_answer_flows', 'shared_logic', "#{filename}.rb")), binding
+      path = Rails.root.join('lib', 'smart_answer_flows', 'shared_logic', "#{filename}.rb")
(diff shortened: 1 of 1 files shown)


chrisroos:
Can you explain how the change "reduces the scope of the context in which the code in the shared
logic files is executed"? I'm afraid I don't know about the difference between `eval` and `instance_eval` without looking it up.

Would it be possible/useful to add a test that demonstrates the behaviour of the stack trace given the additional filename and line number? That might be a silly idea so apologies in advance if it is.

floehopper:
Superseded by #2318. Closing.

floehopper:
@chrisroos: I realised that the sentence you asked about was wrong or at least confusing. I've now opened a [new PR][1] which focusses on my real motivation which was to add the file and line number to the stack trace. I've still switched from using `eval` to `instance_eval`, but this is really just a readability/simplification change i.e. there is no difference in behaviour due to this switch.

[1]: #2318

PR #2317, opened 2016-02-29
Add tags to allow for govspeak rendering for questions.

ikennaokpala:
PT Story: https://www.pivotaltracker.com/projects/1270592/stories/114770219

This PR started with the intention of rendering call outs within questions, just like outcomes and landing pages do.

The initial attempt made, wrapped the question div with the article container div (with the article tag following it), to allow the question share the same styling that is used by the outcome and landing pages.

This appeared to work initially, but broke a couple of other questions with form inputs like checkboxes etc.

A further wrapping of the question div in an article tag.

## Factcheck
https://smart-answers-pr-113993127.herokuapp.com/check-uk-visa/y/
https://smart-answers-pr-113993127.herokuapp.com/check-uk-visa/y/afghanistan/transit
## Expected Changes
[First URL on Gov.UK](https://www.gov.uk/check-uk-visa/y/)
[Second URL on Gov.UK](https://www.gov.uk/check-uk-visa/y/afghanistan/transit)
     * Changes made to the wording of a question 1 and question 3 in the check uk visa smart answer. These are only content changes and doesn't affect any logic or smart flow.

### Before

screen shot 2016-03-02 at 09 23 52

screen shot 2016-03-02 at 09 23 39

### After

screen shot 2016-03-02 at 09 21 19

screen shot 2016-03-02 at 09 22 44

### DIFF
--- a/app/views/smart_answers/_current_question.html.erb
+++ b/app/views/smart_answers/_current_question.html.erb
@@ -3,7 +3,9 @@
     <%= question.title %>
   </h2>
   <div class="question-body">
-    <%= question.body %>
+    <article role="article">
+      <%= question.body %>
+    </article>
 
     <% if question.hint.present? %>
       <p class="hint"><%= question.hint %></p>
--- a/test/artefacts/additional-commodity-code/0.html
+++ b/test/artefacts/additional-commodity-code/0.html
@@ -35,8 +35,10 @@
     How much sucrose, invert sugar or isoglucose does the product contain?
   </h2>
   <div class="question-body">
-    
+    <article role="article">
+      
 
+    </article>
 
       <p class="hint">The values represent % by weight</p>
 
--- a/test/artefacts/additional-commodity-code/0/0.html
+++ b/test/artefacts/additional-commodity-code/0/0.html
@@ -35,8 +35,10 @@
     How much milk fat does the product contain?
   </h2>
   <div class="question-body">
-    
+    <article role="article">
+      
 
+    </article>
 
       <p class="hint">The values represent % by weight</p>
 
--- a/test/artefacts/additional-commodity-code/0/0/0.html
+++ b/test/artefacts/additional-commodity-code/0/0/0.html
@@ -35,8 +35,10 @@
     How much milk proteins does the product contain?
   </h2>
   <div class="question-body">
-    
+    <article role="article">
+      
 
+    </article>
 
       <p class="hint">The values represent % by weight</p>
 
--- a/test/artefacts/additional-commodity-code/0/0/18.html
+++ b/test/artefacts/additional-commodity-code/0/0/18.html
@@ -35,8 +35,10 @@
     How much milk proteins does the product contain?
   </h2>
   <div class="question-body">
-    
+    <article role="article">
+      
 
+    </article>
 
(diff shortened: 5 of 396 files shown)


floehopper:
This PR seems to have been changing while I've been reviewing it. I think it might be worth closing a PR if you realise it's not finished.

floehopper:
It looks as if this might've broken some styling. If I visit this path: `/state-pension-age/y/`, then I see:

Before:

2016-02-29 at 11 10

After:

2016-02-29 at 11 13

Is that change intentional?

floehopper:
Is this change intended to change the appearance of question pages? If so, I think it would be good to include screenshots demonstrating the difference.

ikennaokpala:
Thanks for sharing the screenshots.. I did realise how long the entire tests take to run.

ikennaokpala:
@floehopper 

screen shot 2016-03-01 at 10 14 04

floehopper:
I'm still don't have a clear understanding of what this PR is intended to achieve. Can you expand on what you mean by "more uniform rendering in the questions" in the PR description? And can you make it clear whether this PR is intended to change the rendering of question pages? If it does, can you add screenshots to explain how they are expected to change?

ikennaokpala:
I have updated the PR description with more information.

floehopper:
I've unassigned myself, because I'm not going to have a chance to look at this for a while. Anyone else should feel free to review it now, because it has changed significantly since I first reviewed it.

ikennaokpala:
@dsingleton Here is the PR with commit to enable call outs with questions

erikse:
@ikennaokpala - could you rebase this was master, and regenerate the artefacts again - all the JS and CSS file should not longer have the checksum digest included in the file name since #2323 was merged in

ikennaokpala:
Ok will do

dsingleton:
For anyone following, @ikennaokpala and I had a chat offline and we agree'd a slightly different approach, that avoids the complexity of applying `article` styling to question UI.

Rather than wrapping the content of `_current_question.html.erb`, the `article` tag can go around `<div class="question-body">...</div>`, which will give the govspeak markup in `question.body` the right styles, but not require any resetting CSS on the rest of the page.

In the long-term we'd want to use the `govspeak-component` style the govspeak content within smart-answers, but thats a much larger piece of work, that would involve converting the app to a new layout approach. It's something we'll likely look at as part of the format consolidation work in the coming months.

ikennaokpala:
@dsingleton I have made the changes you recommended

dsingleton:
LGTM. Could someone from @alphagov/team-custom-formats take a look as it's a change to the app/presentation? :+1:

davidbasalla:
:+1:

erikse:
I wonder if we should render the article tag if there is no question body? Does article introduce any margins/padding to the page?

floehopper:
I know this has been merged and then reverted, but I find the title of the PR and the name of the branch confusing, since I think of "rendering as govspeak" or "govspeak rendering" to be the conversion of govspeak text to HTML. I haven't followed in detail, but I believe this PR is more about the *styling of the HTML* generated from that rendering process. I think it would be good to make that clearer in any follow-on PR.

PR #2318, opened 2016-02-29
Include file & line number in the messages of exceptions raised during eval of shared logic

floehopper:
Supersedes #2315.

This adds a file path and a line number to exceptions raised when evaluating shared logic files.

Before:

```
  1 + 1 = 2 # => SyntaxError
         ^
```

After:

```
  1 + 1 = 2 # => SyntaxError
         ^
```

### DIFF
--- a/lib/smart_answer/flow.rb
+++ b/lib/smart_answer/flow.rb
@@ -24,7 +24,7 @@ module SmartAnswer
 
     def use_shared_logic(filename)
       path = Rails.root.join('lib', 'smart_answer_flows', 'shared_logic', "#{filename}.rb")
-      eval File.read(path), binding
+      eval File.read(path), binding, path.to_s, 0
     end
 
     def name(name = nil)
--- a/lib/smart_answer/flow.rb
+++ b/lib/smart_answer/flow.rb
@@ -23,7 +23,8 @@ module SmartAnswer
     end
 
     def use_shared_logic(filename)
-      eval File.read(Rails.root.join('lib', 'smart_answer_flows', 'shared_logic', "#{filename}.rb")), binding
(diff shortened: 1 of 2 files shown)


erikse:
Looking at this, I'm pretty curious as to how the stack trace looks currently without this change, verses how it looks with it - maybe there is some value in adding an example of that to the PR.

Is changing from `eval` to `instance_val` needed for the change in the stack trace, or does that purely come from adding the file name and line number to the `eval` call? If not, maybe it's okay to change that in the previous PR #2315 and have this change focused on the stack trace improvement

floehopper:
Adding an example of the stack trace before and after the change is a good idea. I'll do that shortly.

floehopper:
The change from `eval` to `instance_eval` is *not* necessary to improve the stack trace. I tried to explain my rationale for changing from `eval` to `instance_eval` in the [relevant commit note][1]. Did you read that? If so, did it not make sense?

Note that my explanation in the [previous closed PR][2] was wrong or at least misleading.

If my latest commit note doesn't make sense or is confusing then I'd probably prefer just to drop that commit rather than have two PRs.

[1]: 589f3b22bf41fdc306e7001d004a6bd4c16fb997
[2]: #2315

erikse:
I read the commit note, and agree with the rationale for moving from `eval` to `instance_eval`, I just wondered how it was related to improving the stack trace, and if not related, would maybe expect it to see in a separate PR.

floehopper:
I've now added an example of how the exception message has changed in the PR description.

floehopper:
I'm going to remove the commit which changes from `eval` to `instance_eval` based on @erikse pointing out that it is unrelated to this PR. I plan to apply the above fixup commits and open a new PR. Closing.

PR #2321, opened 2016-03-01
Improve message of exceptions raised in eval of shared logic

floehopper:
This means any exception raised during the evaluation will report a sensible file path and line number rather than the file path and line number of the `Flow#use_shared_logic` method.

Before:

```
  1 + 1 = 2 # => SyntaxError
         ^
```

After:

```
  1 + 1 = 2 # => SyntaxError
         ^
```

Note that I had to change `eval` to `instance_eval` to fix a Rubocop violation.

### DIFF
--- a/lib/smart_answer/flow.rb
+++ b/lib/smart_answer/flow.rb
@@ -24,7 +24,7 @@ module SmartAnswer
 
     def use_shared_logic(filename)
       path = Rails.root.join('lib', 'smart_answer_flows', 'shared_logic', "#{filename}.rb")
-      eval File.read(path), binding
+      eval File.read(path), binding, path.to_s
     end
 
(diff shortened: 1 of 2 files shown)


erikse:
Great, let's ship it :+1:

floehopper:
Thanks. I've just rebased against master and force-pushed in preparation for merging.

floehopper:
This broke the [CI build on master][1] after it was merged, so I [reverted the merge][2]. I need to investigate why the tests were failing.

[1]: https://ci.dev.publishing.service.gov.uk/job/govuk_smart_answers/3080/console
[2]: https://github.com/alphagov/smart-answers/commit/75f7f39a2fde8cf24abf71c29c09a6c3140b7411

PR #2320, opened 2016-02-29
Update Rails to security patched version 4.2.5.2

erikse:
Updates Rails to the latest version to address `[CVE-2016-2098]` for Possible remote code execution vulnerability in Action Pack, and `[CVE-2016-2097]` for Possible Information Leak Vulnerability in Action View.

Both vulnerabilities are exploited by code that allows a carefully set value in `params[:id]` when passed to the render method, which smart answer’s doesn’t do, instead `params[:id]` being used to find the flow and sanitised with `to_sym` in `smart_answers_controller#find_smart_answer`:

```ruby
  def find_smart_answer
    @name = params[:id].to_sym
    @smart_answer = flow_registry.find(@name.to_s)
    @presenter = FlowPresenter.new(request, @smart_answer)
  end
```

ChangeLog details for affected dependencies in Rails 4.2.5.2:
 * [Action Pack](https://github.com/rails/rails/blob/4-2-stable/actionpack/CHANGELOG.md#rails-4252-february-26-2016)
 * [Action View](https://github.com/rails/rails/blob/4-2-stable/actionview/CHANGELOG.md#rails-4252-february-26-2016)

PR #2322, opened 2016-03-01
Revert "Update Rails to security patched version 4.2.5.2"

erikse:
Reverts alphagov/smart-answers#2320

This commit was created using `bundle update rails` but has bumped the major version of sprockets from 2.x to 3.x, which seems to results in a checksum being appended to each compiled JS and CSS asset when we run the regression tests and save the output. 

I guess we can either regenerate the artefacts, or turn this behaviour off with a config setting that has been flipped since [sprockets version 3.0.2](https://github.com/rails/sprockets-rails/commit/3cb84ea29aa321c97364dca86e5685f1ed447df2) was released.


erikse:
Just tried adding the following to `config/environments/test.rb`

```ruby
  # disable asset digest for test environment
  config.assets.digest = false
```

and it seems to prevent the checksum being added to the JS and CSS assets when running the regression tests.

floehopper:
This revert looks fine to me. :+1: 

I'd suggest a new PR to re-do the upgrade to `rails` without the upgrade to `sprockets`. Then we can think about upgrading `sprockets` separately.

erikse:
There is a line in Rails's Gemfile that they require sprockets ~> 3.0.0.rc.1, so not sure how it's possible to not bump the version https://github.com/rails/rails/blob/4-2-stable/Gemfile#L18

floehopper:
I think it's the constraint(s) in `rails.gemspec` (and possibly the sub-gems) which are used by Bundler. As far as I can see there is no constraint on `sprocket-rails`: https://github.com/rails/rails/blob/v4.2.5.2/rails.gemspec#L31. So it ought to be possible to upgrade either by editing the `Gemfile.lock` as I did above or alternatively adding an explicit dependency on `sprocket-rails` in the `Gemfile` with a `~> 2.0` constraint.

erikse:
Maybe it's worth updating sprockets before Rails, in the 4.2-stable branch, the [Gemfile](https://github.com/rails/rails/blob/4-2-stable/Gemfile#L18) is specifying `sprockets ~> 3.0.0.rc.1`

PR #2323, opened 2016-03-02
Update sprockets-rails gem to version 3.0.3

erikse:
Using: `bundle update sprockets-rails`

Upgrades the sprockets-rails gem in smart answers from version 2.3.3 to version 3.0.3. The major release from 2.x to 3.x introduces breaking changes, including enabling asset digests by default (JS and CSS files will receive a checksum value in the in filename).

[Sprockets-rails release note](https://github.com/rails/sprockets-rails/releases/tag/v3.0.0)
 * Enable digest by default rails/sprockets-rails#221

With the smart answers regression tests saving the generated HTML in the repository, change  `config.assets.digest` from the new default value of `true` to `false` to prevent all the artefacts changing as a result.

The upgrade to sprockets-rails in turn updates the sprockets gem from version 2.12.4 to 3.5.2. This major version brings a pretty significant list of changes, and notably brings smart answers inline with the major version 3.x which the `4-2-stable` branch of Rails itself is developed towards. 

See: https://github.com/rails/rails/blob/4-2-stable/Gemfile#L18

Running smart answers locally with this change, no assets appear to be failing and the browser console is free of errors. The [guide to upgrading sprockets from version 2.x to 3.x](https://github.com/rails/sprockets/blob/c49f42475335c2fbde2d5ad4346cc7a659d6890b/UPGRADING.md) suggests:

> For the most part, the application facing APIs have remained the same with the majority of the changes at the extension layer.

Smart Answers also seems unaffected by the following changes:

[Prefer just foo.coffee and foo.scss](https://github.com/rails/sprockets/blob/c49f42475335c2fbde2d5ad4346cc7a659d6890b/UPGRADING.md#prefer-just-foocoffee-and-fooscss)
 * The smart answers [SCSS files](https://github.com/alphagov/smart-answers/tree/master/app/assets/stylesheets) are already following this convention 

[Removed //= include directive](https://github.com/rails/sprockets/blob/c49f42475335c2fbde2d5ad4346cc7a659d6890b/UPGRADING.md#removed--include-directive)
 * Seems unused within smart answers

### Preview on Heroku

https://smart-answers-pr-2323.herokuapp.com/marriage-abroad

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -62,6 +62,7 @@ GEM
       builder (>= 2.1.2)
     cliver (0.3.2)
     coderay (1.1.0)
+    concurrent-ruby (1.0.1)
     crack (0.4.2)
       safe_yaml (~> 1.0.0)
     debug_inspector (0.0.2)
@@ -92,7 +93,6 @@ GEM
     govuk_frontend_toolkit (4.8.0)
       rails (>= 3.1.0)
       sass (>= 3.2.0)
-    hike (1.2.3)
     htmlentities (4.3.3)
     http-cookie (1.0.2)
       domain_name (~> 0.5)
@@ -220,15 +220,13 @@ GEM
       rack (>= 1.3.5)
       rest-client
     slop (3.6.0)
-    sprockets (2.12.4)
-      hike (~> 1.2)
-      multi_json (~> 1.0)
-      rack (~> 1.0)
-      tilt (~> 1.1, != 1.3.0)
-    sprockets-rails (2.3.3)
-      actionpack (>= 3.0)
-      activesupport (>= 3.0)
-      sprockets (>= 2.8, < 4.0)
+    sprockets (3.5.2)
+      concurrent-ruby (~> 1.0)
+      rack (> 1, < 3)
+    sprockets-rails (3.0.3)
+      actionpack (>= 4.0)
+      activesupport (>= 4.0)
+      sprockets (>= 3.0.0)
     thor (0.19.1)
     thread_safe (0.3.5)
     tilt (1.4.1)
@@ -296,4 +294,4 @@ DEPENDENCIES
   webmock (= 1.20.4)
 
 BUNDLED WITH
-   1.10.6
+   1.11.2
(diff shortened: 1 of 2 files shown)


floehopper:
I think it's worth explaining that we want to do this so we can do a simple `bundle update rails` to upgrade `rails` to v4.2.5.2. Give that this was the original intention, I might've been inclined just to upgrade `rails` and not to upgrade `sprockets-rails` at this point. It might be good to explain why you've chosen not to do that.

It's worth noting that this PR also upgrades `sprockets` from v2.12.4 to v3.5.2, a fairly significant change. It's hard to be sure that none of the changes listed in the [changelog][1] affect Smart Answers. However, I'd hope that `sprockets-rails` would shield us from most (if not all) of these changes.

If you haven't done so already, I think it would be worth precompiling the assets and running the app in production locally to check that everything looks OK - I'm not sure we have much test coverage of that side of things. Alternatively this could be checked in the environment that used to be called "preview" (is that now "integration"?).

Ideally this change would be deployed separately from any other upgrades, e.g. the upgrade to `rails`, in case there's a problem, so that it's easier to work out why the problem happened.

Otherwise, I can't see a reason not to merge this.

[1]: https://github.com/rails/sprockets/blob/v3.5.2/CHANGELOG.md

erikse:
@floehopper I've updated the PR description, and will push the branch to integration to verify. 

I think it's worth doing this upgrade before the Rails upgrade, as we're behind the version of sprockets that Rails is developed towards, and attempting to update Rails now to receive the latest security patches, forces sprockets-rails and sprockets to be to upgraded as well, as we saw in #2320

floehopper:
@erikse:

The PR description looks great.

Regarding your last comment, strictly speaking I don't think you're right in saying that updating `rails` *forces* `sprockets-rails` to be upgraded - I was able to upgrade `rails` without upgrading `sprockets-rails` by reverting the relevant parts of `Gemfile.lock` as per [this comment](https://github.com/alphagov/smart-answers/pull/2320#commitcomment-16424968) i.e. there are no dependency constraints which *force* the upgrade of `sprockets-rails`.

However, your explanation in the PR description talks about wanting to come into line with *future* v4.2.x versions, which I think is a perfectly reasonable motivation.

:+1:

ikennaokpala:
@erikse LGTM

erikse:
Ok, that's true - we can undo the upgrade to sprockets-rails and sprockets that occur as a result running `bundle update rails`, perhaps forces isn't the correct wording.

PR #2200, opened 2015-12-18
Marriage abroad amends Bulgaria

erikse:
PT story: https://www.pivotaltracker.com/story/show/109437586

This PR changes the outcome content for opposite sex marriages for Bulgarian and non-UK residents in Bulgaria.

## Factcheck

https://powerful-dusk-4063.herokuapp.com/marriage-abroad/y/bulgaria/

## Expected changes
 * [URL on gov.uk](https://www.gov.uk/marriage-abroad/y/bulgaria/ceremony_country/partner_british/opposite_sex)
  * Adds the name/location of the local authorities to contact for getting married

### Before
screen shot 2015-12-18 at 2 28 38 pm

### After
screen shot 2015-12-18 at 2 28 43 pm

 * [URL on gov.uk](https://www.gov.uk/marriage-abroad/y/bulgaria/ceremony_country/partner_british/opposite_sex)
  * Adds a link to making an appointment at embassy
  * Adds a link of required documents

### Before
screen shot 2015-12-18 at 2 30 19 pm

### After
screen shot 2015-12-18 at 2 30 22 pm

 * [URL on gov.uk](https://www.gov.uk/marriage-abroad/y/bulgaria/ceremony_country/partner_british/opposite_sex)
  * Updates the fee payment methods

### Before
screen shot 2015-12-18 at 2 33 47 pm

### After
screen shot 2015-12-18 at 2 34 04 pm

PR #2325, opened 2016-03-02
Replace appointment booking link (Marriage abroad SUDAN)

ikennaokpala:
PT Story: https://www.pivotaltracker.com/projects/1270592/stories/109439186
Trello Story: https://trello.com/c/YaDRvSFu/32-marriage-abroad-sudan

* Replace the appointment booking link with Embassy address details

## Factcheck
https://smart-answers-pr-109439186.herokuapp.com/marriage-abroad
https://smart-answers-pr-109439186.herokuapp.com/marriage-abroad/y/sudan/ceremony_country/partner_local/opposite_sex

### Expected changes
*  [URL on GOV.UK](//gov.uk/marriage-abroad/y/sudan/ceremony_country/partner_local/opposite_sex)

### Before

screen shot 2016-03-02 at 11 53 33

### After
 
screen shot 2016-03-02 at 11 53 44

### DIFF
--- a/lib/smart_answer_flows/marriage-abroad/outcomes/outcome_os_consular_cni.govspeak.erb
+++ b/lib/smart_answer_flows/marriage-abroad/outcomes/outcome_os_consular_cni.govspeak.erb
@@ -200,7 +200,7 @@
 
       Email: <information.khartoum@fco.gov.uk>
 
-      [British Embassy Khartoum - opening hours](https://www.gov.uk/government/world/organisations/british-embassy-khartoum/office/british-embassy-khartoum-main-contact)
+      [British Embassy Khartoum - opening hours](/government/world/organisations/british-embassy-khartoum/office/british-embassy-khartoum-main-contact)
       $C
     <% else %>
       <%= render partial: 'contact_method.govspeak.erb',
--- a/lib/smart_answer_flows/marriage-abroad/outcomes/outcome_os_consular_cni.govspeak.erb
+++ b/lib/smart_answer_flows/marriage-abroad/outcomes/outcome_os_consular_cni.govspeak.erb
@@ -183,6 +183,25 @@
 
       %You can’t book an appointment to get advice at the embassy in person.%
 
+    <% elsif calculator.ceremony_country == 'sudan' %>
+      Contact the embassy or consulate to make an appointment.
+
+      $A
+      British Embassy Khartoum
+      off Sharia Al Baladia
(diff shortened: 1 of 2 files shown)


pmanrubia:
If we have fully migrated to Trello, it might be useful to add a link to the Trello card, so we have both links (PT and Trello)?

ikennaokpala:
@pmanrubia  I have added the trello link for this PR

pmanrubia:
@ikennaokpala Any reason why there are no update `artefacts`?

ikennaokpala:
@pmanrubia that is becuase Sudan is not included in this test data file.. test/data/marriage-abroad-questions-and-responses.yml

pmanrubia:
@ikennaokpala I guess I might be missing something.
I will hand over the PR to @erikse as I would still expect to see the artefact update as the outcome is being modified. My understanding is that the `questions-and-responses` file would need to be amended manually in case Sudan is not there.

erikse:
@pmanrubia please see the following https://github.com/alphagov/smart-answers/pull/2200#issuecomment-199122150 for discussion on this, at the moment, it's kind of undefined when we add a country, and looking at the responses file, there are only 104 entries covering about half of the countries available in question one

floehopper:
@pmanrubia: When @chrisroos & I originally added the regression tests, we added responses to the `questions-and-responses` file until we had the minimum set that exercised all the paths through the code using a coverage tool (see [adding-new-regression-tests docs][1]).

Since we converted from using phrase lists and YAML files (don't ask!) to using ERB templates, I think using a coverage tool might be hard, because I don't think they support ERB templates.

If new "logic paths" have been added then in the short term it would probably be sensible to make the relevant additions to `questions-and-responses`. However, I'd be wary of adding them just to see content changes.

In the long run we want to replace the regression tests with unit tests and integration tests. `part-year-profit-tax-credits` was a new flow that was written with no regression tests. However, that flow isn't particularly content-heavy and it's not entirely clear to me what we'd ideally want to do with a content-heavy flow like `marriage-abroad`. However, I doubt that having brittle tests around the exact content are going to be the answer.

Sorry - that was a bit of a ramble, but hopefully some of it helps explain things.

[1]: https://github.com/alphagov/smart-answers/blob/master/doc/adding-new-regression-tests.md

pmanrubia:
Sure it does. Thank you @ikennaokpala, @erikse and @floehopper for the clarifications.

LGTM @ikennaokpala

PR #2328, opened 2016-03-02
Improve message of exceptions raised in eval of shared logic (v2)

floehopper:
This means any exception raised during the evaluation will report a sensible file path and line number rather than the file path and line number of the `Flow#use_shared_logic` method.

Before:

```
  1 + 1 = 2 # => SyntaxError
         ^
```

After:

```
  1 + 1 = 2 # => SyntaxError
         ^
```

Note that I've temporarily suppressed a Rubocop rule in order to make this change. See the 3rd commit message for details.

Note also previous failed attempts at doing this in #2321 and #2315.


erikse:
Other than minor comment - this looks good to me.

floehopper:
Thanks for reviewing. I'm going to rebase against master and force-push in preparation for merging.

PR #2333, opened 2016-03-04
Update round_up_weekly.govspeak.erb

reverend-al:
CHANGED FROM

Your weekly childcare costs are the total amount you pay each week rounded up to the nearest pound. Use this figure on your claim form.

REASON

Users were reading "Your weekly childcare costs are ..." and "use this figure ..." and expecting to see a figure on the page. By changing the first part to an action ("Round up"), users will hopefully understand that there's something else they need to do to get the figure they need - rather than expecting the tool to do the calculation for them.


erikse:
Thanks @reverend-al this looks good to cherry-pick and add to a PR with updated artefacts/checksums

erikse:
This PR has been superseded by #2333

PR #2339, opened 2016-03-07
Remove apply-tier-4-visa smart answer

boffbowsh:
This is now archived and should redirect to `/browse/visas-immigration/student-visas`
The redirect is stored separately in router-data.

Zendesk: https://govuk.zendesk.com/agent/tickets/1272124


erikse:
Other than the minor comment about the left over data query class, this looks good to me. I guess some coordination is required getting the redirect in router-data deployed, and the archiving of the artefact in panopticon?

boffbowsh:
Done. I've archived the artefact in Panopticon. This doesn't yet need deploying, and the router-data redirect is already deployed.

floehopper:
If there's an corresponding PR in another repo (`router-data`?), it would be nice to reference it from here.

PR #2319, opened 2016-02-29
Refactor SmartAnswer::Question::Base

floehopper:
These changes were prompted by me preparing the ground for a new version of #2311.

There are still lots of improvements that could be made to this code, but I think the changes in this PR should make the work I want to do on automatic detection of permitted next nodes easier.

PR #2340, opened 2016-03-07
Refactor unit test for SmartAnswer::Question::Base

floehopper:
This follows on from #2319 and is further work in preparation for a proper version of #2311.

I was finding it increasingly hard to read and edit the unit test for `SmartAnswer::Question::Base`, so I decided to to some preparatory refactoring.

The main change in this PR is the grouping together of related tests into Shoulda contexts. I hope the commits are small enough that you can see what's going on in the diffs - you might need to ignore whitespace changes to make things clearer. I've also extracted a method to DRY up the test code a bit.

See the individual commit notes for details.

Given that I've only made changes to a single unit test and no "production" code, these changes should be pretty low risk, but it would be good if someone could give them the once over.

### DIFF
--- a/test/unit/question_base_test.rb
+++ b/test/unit/question_base_test.rb
@@ -1,15 +1,36 @@
 require_relative '../test_helper'
 
 class QuestionBaseTest < ActiveSupport::TestCase
-  should 'permitted next nodes must be supplied if next_node is called with block' do
-    e = assert_raises(ArgumentError) do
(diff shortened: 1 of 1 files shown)


erikse:
Other than minor comments, looks good to me

floehopper:
I'm going to do the interactive rebase against master now and force push in preparation for merging.

PR #2343, opened 2016-03-08
Lazy automatic detection of possible next nodes

floehopper:
Previously the possible node keys returned by a `next_node` block had to be specified explicitly using the `permitted` keyword argument. This PR introduces a mechanism for automatically detecting the possible node keys by parsing the source code in the `next_node` block.

In order to make use of this mechanism, you should set the `permitted` keyword argument to `:auto` and wrap the node keys returned by the `next_node` block with calls to `#question` or `#outcome` as appropriate. See the example below for details.

Note that if you fail to wrap the node key, you will see an exception when that node is chosen much as you would've done if a node not in the list of permitted next nodes was chosen. Note also that at the moment the `#question` and `#outcome` methods are identical i.e. it doesn't actually matter which you use. However, I think using them improves the readability of the code and if we thought it was useful, we could later introduce a check that the chosen node is of the correct type.

## Before

```ruby
multiple_choice :question? do
  option :option_one
  option :option_two

  permitted_next_nodes = [
    :solution_one,
    :solution_two
  ]

  next_node(permitted: permitted_next_nodes) do |response|
    case response
    when :option_one
      :solution_one
    when :option_two
      :solution_two
    end
  end
end
```

## After

```ruby
multiple_choice :question? do
  option :option_one
  option :option_two

  next_node(permitted: :auto) do |response|
    case response
    when :option_one
      outcome :solution_one
    when :option_two
      outcome :solution_two
    end
  end
end
```

This work follows on from #2311 and makes use of the changes in #2319 & #2340.

However, in contrast to #2311, the automatic detection of possible next nodes is now only triggered when a flow is being visualised via the `GraphPresenter`. Also this implementation does not use `throw` & `catch`. Hopefully this allays the concerns about a possible deterioration in performance.

Unfortunately the logic in `SmartAnswer::Question::Base` is more complicated than I would like, but much of that complexity is due to having to support both the old and new styles of `next_node` blocks. I'm confident that it'll be possible to simplify the logic considerably once we've converted all the flows over to the new style. We should also be able to remove the `permitted: :auto` option once all the flows have been converted.

I did contemplate adding a new engine/integration test for the new behaviour. However, in the end I decided that updating all the fixtures for the existing engine/integration tests gave enough coverage and was more in keeping with the current state of affairs. Also I was pleased to see that the `GraphPresenterTest` uses one of these fixture flows - now that the fixture has been updated, the fact that the test still passes gives me confidence that the automatic detection of possible next nodes is working OK.

I have not updated any published flows to use the new style `next_node` blocks in this PR. I'm imagining doing that in one or more subsequent PRs. It should be easy to do flow-by-flow or even question-by-question, because of the support for the old style syntax.

For further details, please see the individual commit notes.

[Trello card](https://trello.com/c/ZbbZ0ruT)


chrisroos:
I think we might have problems using this approach if we also want to use `next` to return early from the `next_node` block. Fortunately, it looks as though the only Smart Answer using `next` is check-uk-visa, and the work that @erikse did in PR #2310 should mean that we can easily remove that use of `next`.

chrisroos:
This all looks good to me. It took me quite a while to understand what was going on but I think I've got it now! As you've already mentioned, the code should be easier to understand once we've removed support for supplying the permitted next nodes via the `permitted` argument.

Great work, @floehopper!

floehopper:
@chrisroos: Can you explain why we might have problems using this approach in flows which use `next` to exit early from a `next_node` block? I just changed the first such occurrence in `check-uk-visa` to:

```ruby
if calculator.study_visit? || calculator.work_visit?
  next question(:staying_for_how_long?)
end
```

And ran its regression test and they all passed. Am I missing something?

floehopper:
@chrisroos: Thanks for reviewing. I think I've addressed all your comments. It'd be great if you could review the latest state of play so I can try to get this merged tomorrow.

chrisroos:
> Can you explain why we might have problems using this approach in flows which use `next` to exit early from a `next_node` block? 

@floehopper: Sorry - my mistake! When I was trying this branch locally I ran into problems visualising the check-uk-visa flow after updating it to use the `question` and `outcome` methods. I obviously didn't investigate properly because I've just done the same thing again and the visualisation works as expected. Sorry for the confusion.

chrisroos:
Your changes look good to me, @floehopper.

floehopper:
@chrisroos: Thanks for reviewing again.

I'm going to rebase this against master, apply the fixup & squash commits, and force-push.

PR #2345, opened 2016-03-08
Upgrade from Ruby v2.2.4 to v2.3.0

floehopper:
### Description

This upgrades Ruby from v2.2.4 to v2.3.0 and addresses a couple of issues which this caused.

I've addressed the problem with `SmartAnswer::State` and the changes to `OpenStruct`. I've also worked around a warning that was being generated from the `WorldLocationTest`. All the tests now pass. Also the full regression test suite passes. See individual commit notes for details.

### Motivation

My main motivations for working on this were (a) the PR had been open for a long time and I kept seeing it in the list; and (b) the `CGI.escapeHTML` optimisation (ruby/ruby#1164) examined in the [Is Ruby 2.3 Faster? Rails ERB Template Rendering Performance](http://ruby-performance-book.com/blog/2016/02/is-ruby-2-3-faster-rails-erb-template-rendering-performance.html) article which @erikse linked to in #2269.

### Deployment

The unicorn processes should be restarted after this PR is deployed. @erikse says that when we upgraded Ruby last time, we saw that the child unicorn processes didn't pick up the new Ruby version until their parent process was restarted.

This supersedes #2269.


chrisroos:
Can we update the PR description with the motivation for upgrading the version of Ruby?

chrisroos:
Is Ruby 2.3.0 already available on CI and all staging/production machines?

erikse:
@chrisroos there are a handful of apps running 2.3.0 in production now

erikse:
In the description, it might be worth adding that the unicorns need to be restarted when this is deployed. I When we upgraded Ruby last time, we saw that the child unicorn processes didn't pick up the new Ruby version without restarting the parent unicorn process.

floehopper:
I've just addressed some of the issues raised by @chrisroos & @erikse which involved re-writing the history on the branch i.e. I've just force-pushed.

floehopper:
I've just updated the PR description to explain my motivations for the PR as requested by @chrisroos.

floehopper:
I've just added a note to the PR description about restarting unicorn processes on deployment as suggested by @erikse.

floehopper:
After @chrisroos highlighted the risk of merging this PR with the `rest-client` related warning happening, I outlined our possible options in [this comment](https://github.com/alphagov/smart-answers/commit/049e207f9b92e02a5a8f1642bdb20b96c13afee0#commitcomment-16590958).

What do people think we should do?

chrisroos:
Given that it's only a warning, and we don't expect it to happen in production[1], I think it's probably fine to merge this branch.

[1]: I _think_ we'll see the warning in production if someone manually changes the country slug in a URL to something that doesn't exist in the API (e.g. https://www.gov.uk/marriage-abroad/y/foobar).

floehopper:
@chrisroos: Thanks.

floehopper:
I'm about to rebase against master, apply the fixup commit, and force-push in preparation for merging.

PR #2347, opened 2016-03-09
Update Rails from version 4.2.5.2 to 4.2.6

erikse:
Updates smart answers to the latest version `4.2.6` of Rails.

Updated with `bundle update rails`, all non-Rails gems in `Gemfile.lock` remain at previous versions with the exception of `sprockets-rails` which moves from `3.0.3` to `3.0.4`. The changelogs for the relevant dependancies are:

[actionmailer](https://github.com/rails/rails/blob/4-2-stable/actionmailer/CHANGELOG.md) and [actionpack](https://github.com/rails/rails/blob/4-2-stable/actionpack/CHANGELOG.md)
- no changes

[actionview](https://github.com/rails/rails/blob/4-2-stable/actionview/CHANGELOG.md)
- a fix in the generation of the img alt tag when using sprockets
- a fix for `content_for` to flush content rails/rails#19890
- a fix to allow the value `false` to be passed to partials rails/rails#22260
- a change in the index used for generating the `week_field` form helper

[activejob](https://github.com/rails/rails/blob/4-2-stable/activejob/CHANGELOG.md), [activemodel](https://github.com/rails/rails/blob/4-2-stable/activemodel/CHANGELOG.md), [activesupport](https://github.com/rails/rails/blob/4-2-stable/activesupport/CHANGELOG.md) and [railties](https://github.com/rails/rails/blob/4-2-stable/railties/CHANGELOG.md)
- no changes

[sprockets-rails](https://github.com/rails/sprockets-rails/releases/tag/v3.0.4)
 - fix that adds Rails environment to be part of cache key

Note: these tests may require an update to older version of `phantomjs` to suppress a warning during execution of the test suite, version `1.8.2` outputs the following:
```
phantomjs[10028:237125] *** WARNING: Method
userSpaceScaleFactor in class NSView is deprecated on 10.7 and later. It should not be
used in new applications. Use convertRectToBacking: instead.
```
versions `1.9.8` seems to have resolved it.


erikse:
The test suite passes, will kick off the regression test to verify
```bash
$ rake test
  ...
 
Finished in 122.643899s, 26.3690 runs/s, 71.9318 assertions/s.
3234 runs, 8822 assertions, 0 failures, 0 errors, 0 skips
```

erikse:
Seems like the regression suite passes as well:
```bash
$ RUN_REGRESSION_TESTS=true ruby test/regression/smart_answers_regression_test.rb
  ...

Finished in 4036.451689s, 5.4097 runs/s, 0.0275 assertions/s.
21836 runs, 111 assertions, 0 failures, 0 errors, 0 skips
```

ikennaokpala:
@erikse  I got the following whilst running  `bundle exec rake test` on my machine:

```
➜  smart-answers git:(update_rails_version_to_4_2_6) bundle exec rake test
Run options: --seed 8821

Running:

.....................................................................................................
...............2016-03-10 12:06:27.202 phantomjs[10028:237125] *** WARNING: Method
userSpaceScaleFactor in class NSView is deprecated on 10.7 and later. It should not be
used in new applications. Use convertRectToBacking: instead.
```

erikse:
@ikennaokpala - can you let me know what version of phantomjs you have installed, eg:
```bash
$ phantomjs -v
1.9.8
```

ikennaokpala:
@erirkse My PhantomJS version

```
➜  ~ phantomjs -v
1.8.2
```

erikse:
Seems like `1.8.2` is just over 3 years old, I wonder if it is worth try to upgrade to a new phantomjs release as I'm not seeing that warning?

ikennaokpala:
Done that waiting for the regression test to finish before re-running `bundle exec rake test`

ikennaokpala:
After upgrading my phantomjs and a re-run of `bundle exec rake test` the warning message (below) disappeared.

```
WARNING: Method
userSpaceScaleFactor in class NSView is deprecated on 10.7 and later. It should not be
used in new applications. Use convertRectToBacking: instead
```

erikse:
thanks @ikennaokpala - which version of `phantomjs` did you upgrade to?

floehopper:
I'd find it helpful if the PR description mentioned the version we are upgrading *from*.

floehopper:
I know this is a pain, but I wonder whether we might be better deploying this separately from the upgrade to Ruby so that if any problems occur in production, we have better idea of what the source of the problem might be.

erikse:
I've updated the description.

floehopper:
:+1:

floehopper:
I know it's a bit late given that this has been merged, but I've just realised that I don't think any of our recent Rails upgrades have included updating the Rails-generated files. Perhaps there haven't been any changes, but when I run `rake rails:update` as per [these instructions][1] at first glance it looks to me as if there might be some.

[1]: http://guides.rubyonrails.org/upgrading_ruby_on_rails.html#the-rake-task

PR #2348, opened 2016-03-09
Remove action mailer dependency

erikse:
It doesn't appear smart answers uses any of the Rails mailer functionality, so this PR removes the requiring of the `action_mailer` gem that is in the project Gemfile.lock via the `rails` dependency, and removes the mailer related config and directory in the application.

For further detail regarding why `action_mailer` may exist in smart_answers (along with other gov.uk projects) see alphagov/frontend#930

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -1,9 +1,8 @@
 require File.expand_path('../boot', __FILE__)
 
-# Don't include all of rails, we don't need activerecord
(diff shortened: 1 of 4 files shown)


floehopper:
Other than my two minor suggestions, this looks good to me. :+1:

erikse:

PR #2353, opened 2016-03-11
Fix marriage-abroad test data

chrisroos:
I noticed that the responses-and-expected-results.yml file was no longer in sync with the questions-and-responses.yml.

This branch updates them so that they're back in sync.

The marriage-abroad regression tests continue to pass after these changes.


floehopper:
Assuming the test artefacts are still correct, this looks good to me :+1:

chrisroos:
Thanks @floehopper.

I've updated the PR description to state that the marriage-abroad regression tests are continuing to pass.

PR #2354, opened 2016-03-11
SPIKE: Country specific outcomes for marriage-abroad

chrisroos:
I've been spiking on country specific outcomes for marriage-abroad. I arbitrarily chose Germany to start with.

I created an outcome template for each of the 18 possible sets of responses that can lead to an outcome for marrying in Germany. I then started to remove some duplication from the templates - at the moment I've only removed _some_ duplication from the opposite sex outcome templates.

I'm interested in taking this spike further - by removing more duplication from these Germany outcome templates and by adding one or two more countries. Are there any obvious reasons why I shouldn't continue?

I've half wondered whether it might even be an improvement to produce response-specific outcomes for all 225 countries (225 countries * 18 templates per country = 4,050 templates) and live with the duplication - at least initially. I don't think it would be too much work to produce all these templates as we can use the complete set of countries in our regression test and then copy the content from the artefacts back to the templates. I can imagine that the duplication would become problematic, but I can also imagine that in the short term it would become much easier to respond to marriage-abroad content change requests.

I'd be interested in any thoughts you have about this spike, @floehopper, @erikse and @ikennaokpala.


floehopper:
In general, I'm really happy that you're having a go at this. And I don't see any reason for you not to take it further - in fact I'd encourage you to do so!

I think the advantage of generating the new outcome templates directly from the regression test artefacts is that we can avoid any preconceptions about how the duplication should best be removed. However, the disadvantage is that we immediately lose all the ways duplication is currently removed, some of which might be good ones!

I don't feel strongly about this, but perhaps there might be some middle ground where we can recognise some existing sensible ways to remove duplication and avoid losing them at the the point where we do the expansion. I was going to say that an example of this might be the outcome title, but looking at the original Germany outcome, I see that this was already hard-coded anyway! So perhaps your approach is the best one!

I think the amount of duplication being introduced in this just this PR is pretty scary. Even merging just this PR, we'd have to make sure that everyone fully understood the implications i.e. being super careful when making a change, that the same change is made to all other copies of that content. It would just be so easy for stuff to get out of kilter and then we'd be in an even worse situation.

So at the moment I'm not convinced that producing response-specific outcomes for *all* 225 countries. I think I'd like to see this PR for Germany taken a bit further and perhaps a few more for other countries, before we commit to this approach across the board. However, I'm willing to be convinced!

I'd also love to hear what @erikse & @ikennaokpala think, given that they have the most experience of maintaining `marriage-abroad`.

chrisroos:
Thanks for your feedback, @floehopper. I'm going to close this while I concentrate on making some smaller refactoring changes/improvements to marriage-abroad, although I hope to come back to it in the near future.

PR #2361, opened 2016-03-14
Remove incorrect permitted next node from check-uk-visa

floehopper:
The `outcome_visit_waiver` is not a valid return value for the `next_node` block for this question.

I came across this while double-checking that the mechanism for automatic detection of permitted next nodes was working correctly. The fact that this had managed to get out of sync highlights the duplication inherent in the current implementation and encourages me that the automatic detection mechanism will be an improvement.

### Expected changes

None

### DIFF
--- a/lib/smart_answer_flows/check-uk-visa.rb
+++ b/lib/smart_answer_flows/check-uk-visa.rb
@@ -163,7 +163,6 @@ module SmartAnswer
           :outcome_transit_taiwan,
           :outcome_transit_taiwan_through_border_control,
           :outcome_transit_venezuela,
-          :outcome_visit_waiver
         ]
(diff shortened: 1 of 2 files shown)


erikse:
Good catch @floehopper. I think our Rubocop config wants trailing commas, which would reduce the diff by a line, otherwise looks good

floehopper:
@erikse: Thanks. Rubocop wasn't failing, but I've added the trailing comma anyway in this [fixup commit][1].

[1]: https://github.com/alphagov/smart-answers/commit/d200a2b4415939536885c4015ec39b0f7c280882

floehopper:
I'm about to rebase this against master and force-push in preparation for merging.

PR #2358, opened 2016-03-14
Remove regression test checksum file for part-year-profit-tax-credits flow

floehopper:
Remove unnecessary regression test checksum file.

This flow uses a newer style of testing and so doesn't have a regression test associated with it.

A flow [only has a regression test][1] if it has a `*-responses-and-expected-results.yml` file associated with it and this flow does not have such a file.

I believe this checksum file was accidentally added in [this commit][2].

### Expected changes

None

[1]: https://github.com/alphagov/smart-answers/blob/aabf580f3fbf53d6cdd4475380cb197d6dcacaea/lib/smart_answer_test_helper.rb#L10-L12
[2]: https://github.com/alphagov/smart-answers/commit/e757b611f3c1bae61ab26d9fdaf642ad0c0bb3c4

PR #2362, opened 2016-03-14
Use automatic next node detection in existing flows

floehopper:
### Dependencies

This PR is based on the relatively small changes in the following three other PRs:

* [Remove regression test checksum file for part-year-profit-tax-credits flow](#2358)
* [Prevent duplicate permitted next nodes](#2360)
* [Remove incorrect permitted next node from check-uk-visa](#2361)

I've included a single squashed merge commit for each of these PRs at the start of the branch for this PR. These will be removed once those PRs have been merged into master and this branch is then rebased against master.

### Changes

This PR changes all the existing flows (and shared logic) to use the mechanism to automatically detect permitted next nodes which was introduced in #2343. This was all relatively straightforward, if a little tedious. The only things worthy of note are:

* The `benefit-cap-calculator` flow uses "dynamic" return values in many of its `next_node` blocks. I couldn't see a quick way to use the auto-detect mechanism, so I've left them be for the moment. I'll think about addressing this in a separate PR.

* The `plan-adoption-leave` flow doesn't use `next_node` with a block for *any* of its questions and so I didn't need to make any changes to it.

### Testing

I added a script to generate a YAML representation of the possible next nodes for all the flows. I ran this script before and after the changes to double-check that using the automatic detection hadn't changed anything. However, the fact that the existing tests (including the regression tests) all pass also give me a lot of confidence.

### Note

If I can find a way to use the automatic detection mechanism in `benefit-cap-calculator`, I plan to remove the code which allows the manual specification of permitted next nodes, but this will be in a separate PR.

### Expected changes

None. However it might be worth checking that a couple of flow visualisation URLs work OK (not forgetting that the visualisation is [currently broken in Chrome][1] for other reasons):

* https://www.gov.uk/part-year-profit-tax-credits/visualise
* https://www.gov.uk/overseas-passports/visualise

[1]: https://trello.com/c/9qXvwUXM


chrisroos:
As a complete aside, mixing GitHub's auto linking of PRs with Markdown's manual linking doesn't appear to work (i.e. `[description-of-the-pr](#pr-number)`). The links to the three related PRs in the description are currently broken because they use this mixed style. For what it's worth, I tend to use `PR #pr-number - description-of-the-pr` when linking to other PRs.

chrisroos:
Is there a plan to differentiate between the `question` and `outcome` methods, @floehopper? If not, I wonder whether it might be simpler to have a single `node` (or similarly named) method. I'm asking because I've come across a couple of instances of `question` being used to return the name of an `outcome` node, and `outcome` being used to return the name of a `question` node.

chrisroos:
I've added comments for a few instances I spotted of `question` and `outcome` being used in place of each other. Aside from those, this all looks great to me, @floehopper. Removing those arrays of permitted next nodes leaves the code looking much better.

floehopper:
@chrisroos: Thanks for pointing out the broken PR links in the description. I've now fixed them.

floehopper:
@chrisroos:

> Is there a plan to differentiate between the question and outcome methods?

Yes, I would like to add a check that the node is of the specified type. However, I don't think it'll be completely straightforward to implement and so I'm planning on doing it later. I did start by using a single `node` method, but I felt that it made the code less readable, so that's why I added the `question` & `outcome` methods in #2343. I'll add a note to the description of that PR to explain this.

floehopper:
@chrisroos: Actually I've just noticed that I did explain this in the 3rd paragraph of the description of #2343.

floehopper:
@chrisroos: Thanks for reviewing and spotting all those `question` & `outcome` mistakes. I thought I'd been so careful! I'm going to create a new branch from this branch, rebase it against master, remove the 3 initial squash commits, and apply all the other fixup & squash commits. I'll then open a new PR.

floehopper:
Superseded by #2368.

PR #2364, opened 2016-03-14
Split integration test for two redundancy pay flows into two tests

floehopper:
### Explanation

This integration test was testing two flows, albeit ones with a significant amount of shared logic. It seems less surprising to have one test per flow as we do elsewhere.

There are a bunch of indentation changes in the first commit, so it's better to view it using the following command:

    $ git show -p --ignore-space-change <sha>

Alternatively you can add `w=1` to the query string to the commit URL on GitHub.

I fixed the filename of the original test in a separate commit to keep the first commit's diff as simple as possible.

### Expected changes

None


chrisroos:
This looks good to me.

floehopper:

PR #2342, opened 2016-03-08
Refactor overseas passports flow

erikse:
This PR aims to tidy up the smart answer flow for overseas passports, extracting logic from the flow definition into a calculator class where it can be unit tested.

With the exception of the `overseas_passports_embassies` value calculated in question 1 (which are not currently available for the location options of question 1a), the calculator class provides a complete encapsulation of the logic required by the flow, outcomes and partials in overseas passports.

This PR supersedes #2331, and is based on #2341, which should be merged before this PR.

PR #2367, opened 2016-03-15
Refactor overseas passports flow updates

erikse:
This PR aims to tidy up the smart answer flow for overseas passports, extracting logic from the flow definition into a calculator class where it can be unit tested.

This PR supersedes #2342.


erikse:
I think I've covered all the non-minor comments in #2342 as well as most of the minor ones, so now marking as ready for review.

floehopper:
I'm probably missing something obvious, but I don't understand what the following bit of the PR description means:

> With the exception of the overseas_passports_embassies value calculated in question 1 (which are not currently available for the location options of question 1a)

floehopper:
I've just reviewed this by going through the comments on #2342 and trying to find the relevant changes made in this PR. Ideally the responses to comments in the older PR would've had links to the new commits.

I have not reviewed all the commits again i.e. I have assumed that the only changes between the two PRs are the ones identified in the comments on the old PR.

On this basis, this PR looks good to me, although I note that this branch currently has conflicts with master.

erikse:
I've updated the description, removing that comment which isn't really related to the changes in this PR but rather one of the things the PR doesn't change.

floehopper:
@erikse: It looks to me as if the regression test artefacts have changed and no longer match those in the repo. Are you aware of this?

PR #2368, opened 2016-03-15
Use automatic next node detection in existing flows (with feedback addressed)

floehopper:
This PR incorporates the feedback on #2362.

### Changes

This PR changes all the existing flows (and shared logic) to use the mechanism to automatically detect permitted next nodes which was introduced in #2343. This was all relatively straightforward, if a little tedious. The only things worthy of note are:

* The `benefit-cap-calculator` flow uses "dynamic" return values in many of its `next_node` blocks. I couldn't see a quick way to use the auto-detect mechanism, so I've left them be for the moment. I'll think about addressing this in a separate PR.

* The `plan-adoption-leave` flow doesn't use `next_node` with a block for *any* of its questions and so I didn't need to make any changes to it.

* Since there's currently no automatic check that `question` or `outcome` methods are called with the correct node types, there's a danger that some of them are called with the wrong node types. However, since I plan to add such a check in the future and calling with the wrong node type won't do any harm, I think it's more readable to use `question` & `outcome` rather than some kind of generic `node` method.

### Testing

I added a script to generate a YAML representation of the possible next nodes for all the flows. I ran this script before and after the changes to double-check that using the automatic detection hadn't changed anything. However, the fact that the existing tests (including the regression tests) all pass also give me a lot of confidence.

### Note

If I can find a way to use the automatic detection mechanism in `benefit-cap-calculator`, I plan to remove the code which allows the manual specification of permitted next nodes, but this will be in a separate PR.

### Expected changes

None. However it might be worth checking that a couple of flow visualisation URLs work OK (not forgetting that the visualisation is [currently broken in Chrome][1] for other reasons):

* https://www.gov.uk/part-year-profit-tax-credits/visualise
* https://www.gov.uk/overseas-passports/visualise

[1]: https://trello.com/c/9qXvwUXM


chrisroos:
Does this need to be reviewed again, @floehopper? Or have you opened this ready to merge it to master given that PR #2362 is good to merge?

floehopper:
I think it's basically ready to merge as long as you're happy that I've addressed all your feedback.

floehopper:

PR #2369, opened 2016-03-15
Add callout to green deal in the energy grants calculator.

pmanrubia:
Paired with @erikse 

PT story: https://www.pivotaltracker.com/story/show/107991402
Trello card: https://trello.com/c/toqXU8S6/37-energy-grant-smart-answer-remove-green-deal-from-outcomes

On all the outcomes that fall on the under green deal we are showing a callout that informs the user that the Green Deal Home Improvement Fund is currently closed to new applications, but they are not affected if they have already applied.

We have also added the missing value to the Warm Home Discount scheme.

## Factcheck

https://smart-answers-pr-2369.herokuapp.com/energy-grants-calculator/y/all_help/benefits/1954-01-01/pension_credit/before-1940/house/mains_gas,electric_heating,modern_double_glazing,loft_attic_conversion,loft_insulation,solid_wall_insulation,modern_boiler,draught_proofing

## Expected changes

* [URL on gov.uk](https://www.gov.uk/energy-grants-calculator/y/all_help/benefits/1954-01-01/pension_credit/before-1940/house/mains_gas,electric_heating,modern_double_glazing,loft_attic_conversion,loft_insulation,solid_wall_insulation,modern_boiler,draught_proofing)
  * Render content "The government has stopped funding the Green Deal Finance Company, which was set up to lend money to Green Deal providers. Find out how this affects you."

### Before
screen shot 2016-03-29 at 14 40 00

### After with Colon
screen shot 2016-03-29 at 14 39 56

### After without colon (Latest)

screen shot 2016-03-31 at 16 26 47


lib/smart_answer_flows/energy-grants-calculator.rb
```diff
@@ -470,6 +470,9 @@ def define
         precalculate :flat_type do
           flat_type
         end
+        precalculate :under_green_deal do
```
ikennaokpala: I this `under_green_deal` will return a boolean.. is there a case against not using a question mark at the end of this definition `under_green_deal?` For me this reads a lot more nicer.

floehopper: @ikennaokpala: I agree about being more readable. Unfortunately this isn't possible when using `precalculate` or other similar blocks. This is because the result is assigned to the state using an attribute writer derived from the block's key. In this case that would mean `state.under_green_deal? = true` which gives a `SyntaxError`. We could look at fixing this, but I think the better solution is to extract a calculator/policy object as per the [refactoring docs][1].

floehopper: @ikennaokpala: It's probably worth adding a sentence or two to explain this in the [storing data docs][1]. I've added a [Trello card][2] to capture this.

[1]: https://github.com/alphagov/smart-answers/blob/master/doc/storing-data.md
[2]: https://trello.com/c/Ldn6KJeE

ikennaokpala: @floehopper I don't have access to /that/[this](https://trello.com/c/Ldn6KJeE) Trello board

pmanrubia: @floehopper @erikse @ikennaokpala Happy to give a try to the calculator/policy object. It feels like a better approach. Not sure, though, if including it in this PR as we would have two different styles for precalculated values?

floehopper: Yes, I think a separate PR would be sensible. :+1:

ikennaokpala:
Factcheck has been carried out by Paul Heron

PR #2370, opened 2016-03-15
Use automatic next node detection in benefit-cap-calculator

floehopper:
### Description

In #2368 I didn't manage to convert all of the questions in the `benefit-cap-calculator` flow to use automatic detection of permitted next nodes. This PR converts the remaining questions in that flow to do so.

This PR is based on #2368 and starts with a squashed merge commit of its changes. That initial squash commit will be removed before this PR is merged.

By using case statements to convert from benefit type to the relevant benefit amount question, we can make the `next_node` logic less "dynamic" i.e. we can explicitly enumerate all the possible results from the `next_node` blocks.

I'm not really very happy with this solution, but I don't think it's any worse than the previous implementation and doing this means that we should be able to remove the logic which deals with permitted next nodes that are explicitly passed into a `next_node` block.

The main commit is the one entitled: "Auto-detect permitted next nodes in benefit-cap-calculator". See individual commit notes for further details.

### Notes

I'm not convinced that the visualisation for this flow is actually that useful, particularly regarding the optional benefit amount questions. We could probably get something that was almost as useful by extracting the "full" case statement from the first of these questions into a method and re-using it in each question. This would imply that you could always get to any other benefit amount question from any benefit amount question. Obviously this isn't actually accurate, but the resulting visualisation might be good enough.

Before attempting any better solutions for this flow, I think it would also be worth considering whether the user experience for this flow is actually that good e.g. it might be better to group all the benefit amount questions onto a single page.

### Testing

I added a script (borrowed from #2368) to generate a YAML representation of the possible next nodes for all the flows. I ran this script before and after the changes to double-check that using the automatic detection hadn't changed anything. However, the fact that the existing tests (including the regression tests) all pass also give me a lot of confidence.

### Expected changes

None. However it might be worth checking that the [flow visualisation URL][1] works OK (not forgetting that the visualisation is [currently broken in Chrome][2] for other reasons).

[1]: https://www.gov.uk/benefit-cap-calculator/visualise
[2]: https://trello.com/c/9qXvwUXM

### DIFF
--- a/lib/smart_answer_flows/benefit-cap-calculator.rb
+++ b/lib/smart_answer_flows/benefit-cap-calculator.rb
@@ -68,11 +68,8 @@ module SmartAnswer
         option :widow_pension
         option :widows_aged
 
-        next_node_calculation :benefit_related_questions do |response|
-          questions = response.split(",").map { |r| :"#{r}_amount?" }
-          questions << :housing_benefit_amount? if housing_benefit == 'yes'
-          questions << :single_couple_lone_parent?
-          questions
+        next_node_calculation :benefit_types do |response|
+          response.split(",").map(&:to_sym)
         end
 
         calculate :total_benefits do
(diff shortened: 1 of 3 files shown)


lib/smart_answer_flows/benefit-cap-calculator.rb
```diff
@@ -68,11 +68,8 @@ def define
         option :widow_pension
         option :widows_aged
 
-        next_node_calculation :benefit_related_questions do |response|
-          questions = response.split(",").map { |r| :"#{r}_amount?" }
-          questions << :housing_benefit_amount? if housing_benefit == 'yes'
-          questions << :single_couple_lone_parent?
-          questions
+        next_node_calculation :benefit_types do |response|
+          response.split(",").map(&:to_sym)
```
chrisroos: Did you consider adding `:housing_benefit` to this `benefit_types` array? I _think_ that would mean you could simplify the current `else` in the `case` statements:

```ruby
# Old
case benefit_types.shift
...
else
  if housing_benefit == 'yes'
    question :housing_benefit_amount?
  else
    question :single_couple_lone_parent?
  end
end

# New
case benefit_types.shift
...
when :housing_benefit then :housing_benefit_amount?
else :single_couple_lone_parent?
end
```

Does that make sense, or am I missing something?

chrisroos:
This looks good to me. I agree that it's not completely ideal but I think the benefit of being able to simplify the code dealing with permitted next nodes makes it worth while.

I see what you mean by the current visualisation not being very useful - it's a complete mess!

lib/smart_answer_flows/benefit-cap-calculator.rb
floehopper: I did consider it, but I thought that having the `if housing_benefit == 'yes'` condition in that `next_node_calculation` block, rather than in each of the `next_node` blocks was harder to follow. Although it results in more duplication, I quite like the explicitness of having it in the `next_node` blocks.

floehopper:
Removing squash merge commit, rebasing against master and force-pushing in preparation for merging.

floehopper:
Rebasing against master has removed the commits which I borrwed from #2368, but I think that's OK as they were just double-checking that everything was working OK.

PR #2374, opened 2016-03-17
Make auto-detecting permitted next nodes the default behaviour

floehopper:
In #2368 & #2370, all occurrences of `next_node` with a block were converted to use the automatic detection of permitted next nodes instead of explicitly supplying them. So we can now make the code less verbose by making `permitted: :auto` the default arguments for `next_node` with a block.

Note that the old mechanism is still available, it's just no longer the default behaviour.

I've updated the documentation to reflect this change, noting that the old approach is now deprecated.

### Expected changes

This is just a refactoring - there should be no changes in externally observable behaviour.


floehopper:
I'm going to rebase this against master to resolve the conflicts that have been introduced by recent changes in master and then force-push it.

floehopper:
Rebasing & force-pushing done. Conflicts resolved. Ready for review again.

test/unit/question_base_test.rb
```diff
@@ -8,7 +8,7 @@ class QuestionBaseTest < ActiveSupport::TestCase
   context '#next_node' do
     should 'raise exception if called with block but no permitted next nodes' do
       e = assert_raises(ArgumentError) do
-        @question.next_node do
+        @question.next_node(permitted: []) do
           :done
```
erikse: should this line be?
```ruby
outcome :done
```

erikse: should this line be?
```ruby
outcome :done
```

erikse:
Even though this code doesn't change the outward appearance of any page, it might be useful to add a before and after of what a next_node block looked(will look) like with this change.

test/unit/question_base_test.rb
floehopper: No, although I see why you ask. This is still testing the "old" mechanism which doesn't require the wrapping of the result with `question` or `outcome`. I hope to remove this functionality soon.

floehopper:
I've edited the PR description to explain the internal changes to the flow DSL.

test/unit/question_base_test.rb
```diff
@@ -90,6 +90,19 @@ class QuestionBaseTest < ActiveSupport::TestCase
...
+          if response == 'yes'
+            outcome :done
+          else
+            question :another_question
+          end
+        end
+        assert_equal [:done, :another_question], @question.permitted_next_nodes
+      end
+    end
   end
```
erikse: In the context description, when I read "not permitted option specified" I'm looking for the 'not permitted' option, but I guess it means without a permitted option.

In the should description, I'm not sure I'm aware of what syntactic sugar methods relates to. I think a more obvious expectation would be that it expects to return a list of nodes from the outcomes and questions defined within the block.

floehopper: Regarding your first comment - that was a typo - well spotted! Fixup commit: 30d92cccbbbfcb8d3952414a26b76a85d8d3a3c6.

floehopper: Regarding your second comment - I've tried to improve the test name. Fixup commit: 7c9ea426dd2e1dadda4607db5e74cb7908ab27af.

floehopper:
@erikse: Thanks for reviewing. I've addressed your comments. Are you happy for me to merge this?

erikse:
Yep, this looks really good

floehopper:
Thanks. I'll get it rebased, force-pushed & merged.

PR #2375, opened 2016-03-17
Revert "Refactor overseas passports flow updates"

erikse:
Since merging this branch, the regression test are showing diffs in the artefacts. As there appears to be a number of issues, we should probably revert so we can resolve without effecting other work.

Reverts alphagov/smart-answers#2367


erikse:
I think given the size of this revert, I'll close this and I'll prepare a PR with a fix to flow, the cause seems to a poor rebase with master prior to the merge with master that included a small change to smart answer DSL.

chrisroos:
In general, I think we should prefer fixing problems by improving the code rather than by reverting changes, so I think closing this and fixing the problem in #2376 was the right thing to do.

Out of interest, was the problem caused by GitHub's automatic merging, or was it simply due to a complicated rebase/merge?

floehopper:
@chrisroos: I believe the problems were introduced when the branch was manually rebased against master in preparation for merging.

PR #2378, opened 2016-03-18
Remove `permitted` option from `next_node` with block

floehopper:
In #2374 auto-detecting of permitted next nodes was made the default behaviour and most of the flows were changed to not specify the `permitted` option on calls to `next_node`.

This PR does the following:

* Removes the few remaining occurrences of the deprecated `permitted` option
* Removes the ability to specify the deprecated `permitted` option
* Removes some code which becomes unreachable because of the latter change
* Updates the documentation to match the new behaviour

### Motivation

This is another small step towards simplifying the flow DSL and in particular the code in `SmartAnswer::Question::Base`.

Also it will prevent any accidental "regression" to the deprecated syntax.

#### Before

Deprecated usage:

```ruby
next_node(permitted: [:green?, :red]) do |response|
  if response == 'green'
    :green?
  else
    :red
  end
end
```

#### After

The above is no longer allowed and *must* be written as:

```ruby
next_node do |response|
  if response == 'green'
    question :green?
  else
    outcome :red
  end
end
```

### External changes

None. This is just a refactoring and should not cause any externally visible changes.

### DIFF
--- a/doc/next-node-rules.md
+++ b/doc/next-node-rules.md
@@ -12,24 +12,6 @@ next_node do |response|
 end
 ```
 
-## Explicitly setting permitted next nodes
-
-This approach is deprecated.
-
-```ruby
-permitted_next_nodes = [
-  :green,
-  :red
-]
-next_node(permitted: permitted_next_nodes) do |response|
-  if response == 'green'
-    :green # Go to the :green node
-  else
-    :red   # Go to the :red node
-  end
-end
-```
-
 ## Shortcut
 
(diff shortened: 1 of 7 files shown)


erikse:
This looks good, maybe it's worth a run through the regression test suite before merging?

floehopper:
@erikse: Thanks. I'll run all the regression tests locally before rebasing this and getting it merged.

floehopper:
The regression tests all passed. Merging.

PR #2380, opened 2016-03-18
Convert calls to `next_node` without a block into calls with a block

floehopper:
### Summary

* Changes all the places which call `next_node` *without* a block and with a single node key to call `next_node` *with* a block.
* Removes the ability to call `next_node` *without* a block.
* Simplifies logic in `SmartAnswer::Question::Base`.
* Updates documentation.

### Design choice

I considered a couple of other alternatives:

* Introduce a `next_question` & `next_outcome` method.
* Change signature of `next_node` to be something like `next_node question: :q1` & `next_node outcome: :o1`.

However, I didn't think either of these was as readable and it was still awkward to simplify the code in `SmartAnswer::Question::Base`.

### Motivation

* Allows us to simplify code in `SmartAnswer::Question::Base`
* Simplifies the flow DSL - there's now only a single way to call `next_node`
* Makes the type (question/outcome) of the next node explicit
  * I think this is more readable & consistent
  * Will allow us to add checks in the future that the node referred to by the key passed into `#question` or `#outcome` is actually a question node or an outcome node respectively

#### Before

```ruby
  multiple_choice :how_much_sucrose_1? do
    # ...
    next_node :how_much_milk_fat?
  end

  multiple_choice :how_much_milk_protein_ab? do
    # ...
    next_node :commodity_code_result
  end
```

#### After

```ruby
  multiple_choice :how_much_sucrose_1? do
    # ...
    next_node do
      question :how_much_milk_fat?
    end
  end

  multiple_choice :how_much_milk_protein_ab? do
    # ...
    next_node do
      outcome :commodity_code_result
    end
  end
```

### External changes

None. This is just a refactoring and should not cause any externally visible changes.


erikse:
What does that last bullet point under motivation "Will allow us to add checks in the future" mean?

test/unit/question_base_test.rb
```diff
@@ -15,8 +15,8 @@ class QuestionBaseTest < ActiveSupport::TestCase
 
     should 'raise exception if next_node is invoked multiple times' do
       e = assert_raises do
-        @question.next_node :outcome_one
-        @question.next_node :outcome_two
+        @question.next_node { outcome :one }
+        @question.next_node { outcome :two }
       end
```
erikse: Minor. I've add a comment to the Trello card about moving away from outcome names prefixed with "outcome", and why having it can be useful.

test/unit/question_base_test.rb
```diff
@@ -6,11 +6,11 @@ class QuestionBaseTest < ActiveSupport::TestCase
   end
 
   context '#next_node' do
-    should 'ensure single next node key is supplied if next_node called without block' do
+    should 'raise exception if next_node is called without a block' do
```
erikse: Minor. More of a question, are defining or calling, maybe the term calling a block means something specific in Ruby

erikse:
I'm out of laptop battery, will continue reviewing at this [commit](https://github.com/alphagov/smart-answers/pull/2380/commits/96b4ef2dc4f40b04352c88bca6113eeb51e518d6)

doc/testing.md
```diff
@@ -24,21 +24,27 @@ multiple_choice :question_1 do
   option :A
   option :B
 
-  next_node :question_2
+  next_node do
+    question :question_2
+  end
```
erikse: Should the question symbol end with a `?` as we do elsewhere

doc/testing.md
```diff
@@ -24,21 +24,27 @@ multiple_choice :question_1 do
...
 end
 
 multiple_choice :question_2 do
   option :C
   option :D
 
-  next_node :question_3
+  next_node do
+    question :question_3
+  end
```
erikse: Should the question symbol end with a `?` as we do elsewhere

doc/storing-data.md
```diff
@@ -31,7 +31,9 @@ The flow below illustrates the data available to the different Question node met
 multiple_choice :question_1 do
   option :q1_option
 
-  next_node :question_2
+  next_node do
+    question :question_2
+  end:
```
erikse: Should the question symbol end with a `?` as we do elsewhere

test/unit/question_base_test.rb
floehopper: I'm not quite sure what you're getting at here. I don't mention "calling a block" in the test names.

However, I do mention "calling next_node *with* a block" and "calling next_node *without* a block". Is there something about this that doesn't make sense? If so, what?

Here's some example code which might help clarify what I mean when I say a method is called with/without a block:

```ruby
def foo
  if block_given?
    puts "foo method called *with* block"
  else
    puts "foo method called *without* block"
  end
end

foo() # => foo method called *without* block

foo do
  # some code
end
# => foo method called *with* block 
```

erikse: Argh, I think I totally misread the test name! I guess I was more questioning if we calling next_node, or defining a next_node in the flow.

floehopper:
> What does that last bullet point under motivation "Will allow us to add checks in the future" mean?

In #2368 & #2370 we started using `question(:key)` & `outcome(:key)` inside `next_node` blocks. This PR means that we now use them everywhere.

Although currently there is no difference between using `#question` or `#outcome`, I've had it in mind that we could introduce a check that the node referred to by the key is of the relevant type i.e. if you tried to call `question(:outcome_key)` or `outcome(:question_key?)` then we could fail-fast.

I agree this was not well explained in the PR description - I've now updated it.

doc/testing.md
floehopper: Although I don't think there's currently anything forcing us to do this, I agree that it's a standard that has been adopted in the code and the documentation should reflect it. However, since it was already missing, I think it's out of scope for this PR.

I've opened a new PR to address this: #2392

doc/testing.md
floehopper: See my [previous comment][1].

[1]: https://github.com/alphagov/smart-answers/pull/2380#discussion_r56827078

doc/storing-data.md
floehopper: See my [previous comment][1].

[1]: https://github.com/alphagov/smart-answers/pull/2380#discussion_r56827078

floehopper:
@erikse: Thanks for reviewing this. Had you finished reviewing it? I saw that you ran out of laptop battery at one point. Also are you happy that I've now addressed all your comments?

erikse:
Yep, this all looks good to merge

floehopper:
Rebasing to apply fixup commit(s). Then rebasing against master and force-pushing in preparation for merging.

PR #2381, opened 2016-03-18
Remove marriage-abroad outcome path precalculate blocks

chrisroos:
I'm continuing to refactor marriage-abroad. This branch removes the `precalculate` blocks from outcome_os_kuwait, outcome_spain and outcome_consular_cni_os_residing_in_third_country by moving logic to the `MarriageAbroadCalculator.`

### DIFF
--- a/lib/smart_answer/calculators/marriage_abroad_calculator.rb
+++ b/lib/smart_answer/calculators/marriage_abroad_calculator.rb
@@ -231,5 +231,13 @@ module SmartAnswer::Calculators
         @partner_nationality, @sex_of_your_partner
       ].join('/')
     end
+
+    def outcome_path_when_resident_in_ceremony_country
+      [
+        '', 'marriage-abroad', 'y',
+        @ceremony_country, 'ceremony_country',
+        @partner_nationality, @sex_of_your_partner
+      ].join('/')
+    end
   end
 end
--- a/lib/smart_answer/calculators/marriage_abroad_calculator.rb
+++ b/lib/smart_answer/calculators/marriage_abroad_calculator.rb
@@ -223,5 +223,13 @@ module SmartAnswer::Calculators
     def same_sex_alt_fees_table_country?
       @data_query.ss_alt_fees_table_country?(ceremony_country, self)
     end
+
+    def outcome_path_when_resident_in_uk
+      [
+        '', 'marriage-abroad', 'y',
+        @ceremony_country, 'uk',
+        @partner_nationality, @sex_of_your_partner
+      ].join('/')
+    end
   end
 end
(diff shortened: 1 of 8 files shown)


lib/smart_answer/calculators/marriage_abroad_calculator.rb
```diff
@@ -223,5 +223,13 @@ def requires_7_day_notice?
...
       @data_query.ss_alt_fees_table_country?(ceremony_country, self)
     end
+
+    def outcome_path_when_resident_in_uk
+      [
+        '', 'marriage-abroad', 'y',
+        @ceremony_country, 'uk',
+        @partner_nationality, @sex_of_your_partner
+      ].join('/')
+    end
```
floehopper: Can you explain why you chose not to use `State#current_path` here? I think it probably does make sense, but I think at the very least you should explain the reasoning in the commit note.

floehopper: Oh. I've just noticed that `current_path` isn't a method on `State`.

I suppose I'm wary of the idea that the path to a question or outcome is built up of a set of responses separated by slashes being spread further around the codebase. But perhaps that was already a problem in the `current_path` precalculate block.

floehopper:
I'm not sure why my comments are showing up as being on an outdated diff - as far as I can tell I added my comments to the latest version of the commit.

floehopper:
Minor: In the PR description, I would say you are *moving* the logic to the calculator and not just *adding* it.

floehopper:
I think it might be worth trying to encapsulate the concept of a path to a question or outcome being made up of slash-separated responses somewhere. Perhaps a `path_to_outcome(responses)` helper method similar to `url_for` might be worth considering. However, it doesn't necessarily need to be done in this PR.

chrisroos:
> I'm not sure why my comments are showing up as being on an outdated diff - as far as I can tell I added my comments to the latest version of the commit.

I noticed this happen to some of the comments I left on another PR too. Presumably related to GitHub's recent code review updates...

chrisroos:
> Minor: In the PR description, I would say you are moving the logic to the calculator and not just adding it.

Done.

chrisroos:
I can see that something like `path_to_outcome(responses)` might be useful but I'm not going to worry about it for now. As far as I can see, marriage-abroad is the only Smart Answer that links to other outcomes so I don't think it's too terrible that this logic is contained within the `MarriageAbroadCalculator` at the moment. I hope that makes sense.

chrisroos:
Thanks for reviewing, @floehopper. I'm going to close this and open a new PR now that I've addressed feedback and rebased the branch on master.

chrisroos:
This has been superseded by PR #2387.

PR #2383, opened 2016-03-19
Add OverseasPassportsCalculator to regression test checksums

floehopper:
### Description

This should've been added as part of #2367, so that changes to the calculator cause the regression tests to run.

I did this by running the following command:

```
$ rails r script/generate-checksums-for-smart-answer.rb overseas-passports \ 
  lib/smart_answer/calculators/overseas_passports_calculator.rb
```

### External changes

None.


erikse:
Well spotted, thanks for adding this @floehopper

floehopper:

PR #2385, opened 2016-03-19
Remove unnecessary instance variable assignment in SmartAnswer::Question::Base

floehopper:
### Description

Instance variables are set to `nil` by default, so this assignment is unnecessary.

### External changes

None.


erikse:
This looks good to merge

floehopper:

PR #2386, opened 2016-03-19
Remove unused `state` attribute from flow

floehopper:
### Description

As far as I can tell this is not used and I'm pretty sure it's *never* been used. The state for a request seems to be handled entirely locally within the `Flow#process` method.

### External changes

None.


erikse:
That's weird - nice to remove the, lgtm

floehopper:

PR #2392, opened 2016-03-21
Consistently use question mark suffix for question node keys in docs

floehopper:
Documentation change only.

### DIFF
--- a/doc/next-node-rules.md
+++ b/doc/next-node-rules.md
@@ -5,7 +5,7 @@ The `next_node` method takes a block that should return the next node (question 
 ```ruby
 next_node do |response|
   if response == 'green'
-    question :green? # Go to the :green question node
+    question :green? # Go to the :green? question node
   else
     outcome :red # Go to the :red outcome node
   end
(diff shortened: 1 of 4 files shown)


doc/next-node-rules.md
```diff
@@ -73,5 +73,5 @@ Occurs if the `next_node` blocks returns a value that isn't defined as a questio
 next_node do
   outcome :red
 end
-outcome :green
+outcome :green?
```
chrisroos: I don't think this one should've been changed.

chrisroos:
Aside from one change that doesn't look right, this all looks good to me.

doc/next-node-rules.md
floehopper: Good catch. I've changed it to `outcome :blue` so that it's less confusing. Fix-up commit: 87af37eab369ce184941969f39f3911b95083957.

floehopper:
Rebasing to incorporate fix-up commit, then rebasing against master, resolving conflicts and force-pushing in preparation for merge.

PR #2390, opened 2016-03-21
[WIP] National minimum wage over 25 above

pmanrubia:
paired with @erikse 

Trello ticket: https://trello.com/c/kZ0I4a4G/58-national-minimum-wage-april-2016-update

## Fact check

https://smart-answers-pr-2390.herokuapp.com/national-minimum-wage

PR #2393, opened 2016-03-21
Documentation for the new-style testing approach

floehopper:
I've had a first stab at this, because @pmanrubia asked about it in the context of the new version of the minimum wage flow - see #2390.

I don't think we've written this down anywhere else - please correct me if I'm wrong.

@chrisroos: I'd definitely like your feedback on this.

### DIFF
--- a/doc/new-style-testing.md
+++ b/doc/new-style-testing.md
@@ -1,0 +1,80 @@
+# New-style Testing
+
+## Introduction
+
+If you're writing a new smart answer or are [refactoring an existing one](refactoring.md) then you should be using the new approach to testing which we're gradually adopting. The idea is that with the various concerns better separated, we should be able to provide more test coverage at the unit-test level i.e. using the idea of a [Test Pyramid][].
+
+### Advantages
+
+* Avoids combinatorial explosion
+* Tests are less brittle
+* Easier to diagnose failing tests
+* Faster test suite
+
+## New style
+
+When we built the [part-year-profit-tax-credits][3], we adopted the following strategy:
+
+### Unit tests
+
+* [PartYearProfitTaxCreditsFlowTest][4]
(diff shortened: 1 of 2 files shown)


doc/new-style-testing.md
```diff
@@ -0,0 +1,80 @@
...
+* Faster test suite
+
+## Old style
+
+* Previously most of the test coverage for a flow was provided by a single integration test often with [lots of nested contexts](testing.md#deprecated-a-test-using-nested-contexts).
+* Flows that have a calculator also tend to have a unit test for that calculator.
+* When we started doing significant refactoring of the app, we weren't happy that we had sufficient test coverage to support this work.
+* Notably none of the tests actually *rendered* the question or outcome pages.
+* So at this point we introduced the [regression tests](regression-tests.md) as a relatively quick way to improve the test coverage.
+* However, the intention has always been that these regression tests are a a only temporary measure.
```
chrisroos: Typo: "...are a a only temporary..."

chrisroos:
I'd be inclined to swap the order of the "New style" and "Old style" sections of the document. I found it confusing that the introduction talks about a new style of testing but then tells me about the old style.

Aside from that, this looks great to me!

doc/new-style-testing.md
floehopper: Fix-up commit: 7b0be62f41e5f505ba70b488a6cea79d3e1557bc

floehopper:
I've swapped the old-style and new-style sections as suggested by @chrisroos. Fix-up commit: dbd7f898875cfd47ebb39b9dbaef5118b92ec129

floehopper:
Rebasing to include fixup commits, then rebase against master and force-push in preparation for merging.

PR #2395, opened 2016-03-22
Add regression tests for all marriage-abroad countries

chrisroos:
Given the recent discussion/confusion around when to add new countries to the marriage-abroad-questions-and-responses.yml I thought it'd be interesting to see how long the regression tests take to run if we add _all_ countries to marriage-abroad-questions-and-responses.yml. This branch does just that.

Anecdotally, it would appear that the time to run the marriage-abroad regression tests doubles after the addition of the 121 countries in this branch:

* In master on my laptop: 4m46.487s
* In this branch on my laptop: 9m33.561s

I used the following command to get the time to run the tests: `time RUN_REGRESSION_TESTS=marriage-abroad ruby test/regression/smart_answers_regression_test.rb `.

What do you think @erikse, @pmanrubia, @ikennaokpala and @floehopper? Do the potential benefits of having regression tests for all countries outweigh the additional time it takes to run the tests?

## Other changes in this PR

I updated the worldwide organisation data for all countries as part of this PR, which also affected the regression test artefacts for overseas-passports, register-a-birth, register-a-death and report-a-lost-or-stolen-passport. It's also left me with some failing tests in worldwide_organisation_test.rb. I've not fixed those as I'm not confident that this is a good approach.


pmanrubia:
Nice PR @chrisroos: real data is always helpful.

If we are able to exercise all our codebase without the need to check the content of every page, I’d say we would have a right balance between code coverage and fast tests.

A 4 minutes increase in regression tests would not be very painful for a flow that is not under active development, but following the same approach in other flows might impact the speed of our tests.

We are currently validating the content per PR, release (in staging and in production) and also by the content team with fact-checks; so I would be good with selecting the answers that exercise all our branches, and relying on human review for content changes.

So, If I would need to pick up an approach for test coverage in our tests, I would do what @erikse and @chrisroos discussed in https://github.com/alphagov/smart-answers/pull/2200#issuecomment-199122150: update the answers when a new branch is found.

chrisroos:
Thanks for your feedback, @pmanrubia. Adding input data when required sounds like a good approach to me. I've [added a card to Trello to capture this idea](https://trello.com/c/ltA9b90r) and will now close this PR.

PR #2399, opened 2016-03-22
Fix RatesQuery#rates method to respect relevant_date argument

erikse:
There seems to be a subtle bug in this code, in that the query class saves the results of invoking #rates the first time in a @rates instance variable.

But on subsequent invocations, it returns the previously calculated value, regardless if it is invoked with a different `relevant_date` argument.

### DIFF
--- a/lib/smart_answer/calculators/rates_query.rb
+++ b/lib/smart_answer/calculators/rates_query.rb
@@ -5,13 +5,12 @@ module SmartAnswer::Calculators
     end
 
     def rates(relevant_date = Date.today)
-      return @rates if @rates
       relevant_rates = data.find do |rates_hash|
         rates_hash[:start_date] <= relevant_date && rates_hash[:end_date] >= relevant_date
       end
       relevant_rates ||= data.last
 
-      @rates = OpenStruct.new(relevant_rates)
+      OpenStruct.new(relevant_rates)
     end
 
   private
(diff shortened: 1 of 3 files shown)


chrisroos:
Was commit ed1e6fedd904f0f641f2ae59845b3b52342cfb65 (titled "Rename rates local shadowing enclosing method name") necessary? Or was it more of a good practice change?

Aside from my question, this all looks good to me.

erikse:
@chrisroos - I don't think it is necessary, just my preference given the enclosing method signature is the same as the local variable, when called without an argument (using the default value). I just checked the [shadowing lint in Rubocop](http://www.rubydoc.info/github/bbatsov/rubocop/Rubocop/Cop/Lint/ShadowingOuterLocalVariable) and it seems like it doesn't apply, so can probably be safely removed from the PR.

erikse:
Btw, I'm can no longer merge my PR's, not even sure about pushing a rebase of my branches to `alphagov`. Let me know if you want me to help with my remaining open ones via a fork.

chrisroos:
Thanks for clarifying, @erikse. I'm happy for the first commit to remain, I was simply interested in the motivation.

I'll get this rebased and merged shortly.

chrisroos:
I've rebased this on master and force pushed in preparation for merging. I was happy to force push as there were no comments on individual commits that would be lost.

erikse:
Cool, thanks for the help merging

On Wednesday, 30 March 2016, Chris Roos <notifications@github.com> wrote:

> I've rebased this on master and force pushed in preparation for merging. I
> was happy to force push as there were no comments on individual commits
> that would be lost.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/alphagov/smart-answers/pull/2399#issuecomment-203272650>
>


-- 
Sent from Gmail Mobile

PR #2400, opened 2016-03-22
Avoid double render error if exception raised after render

floehopper:
The `if performed?` conditional in `ApplicationController#error` was introduced in 168b69edf25c2a816be2b947c93c4d6630b3f6a6 in order to avoid a `DoubleRenderError` if the relevant exception was raised after rendering had completed. However, the change was made along with a test that didn't test what it was meant to test - I think the test is triggering an exception *in the middle of* rendering and not *after* rendering.

To remedy this, I replaced the originally added controller test and added a new Rails integration test, `ErrorHandlingTest`. This gives coverage of both branches of the `if/else` statement and a failure of one of these tests surfaced the fact that the line, `response_body = error_message` is not doing anything useful - it's just assigning the value in `error_message` to a local variable, `response_body` which is not then used for anything.

In fact I first came across this because this method was causing a `Lint/UselessAssignment` Rubocop violation where a `response_body` local variable was being assigned, but not used. I _think_ the intention was probably to use the `ActionController::Metal#response_body=` writer method.

I fixed the failing test by changing `response_body = error_message` to `self.response_body = error_message`.

I also decided to change `response.status = status_code` so that it too uses a writer method on `ActionController::Metal` rather than accessing the `response`. The latter is only made available by `ActionController::RackDelegation` i.e. it may not be available in all requests. Doing this felt more consistent and more robust.

I had to use `set_slimmer_headers(skip: true)` in the test controller action, because Slimmer ignores the body content of an error page and so we'd never see what it's being set to. The body content is ignored because `Slimmer::Skin#error` does not include `Slimmer::Processors::BodyInserter` in the list of processors for the request.

Thus with Slimmer enabled the relevant ERB error template (e.g. [404.html.erb][1] and its [related partial [2]) from the static app is the main source of the error page, albeit with a few things changed e.g. the page title.

[1]: https://github.com/alphagov/static/blob/master/app/views/root/404.html.erb
[2]: https://github.com/alphagov/static/blob/master/app/views/root/_error_page.html.erb

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -26,8 +26,8 @@ protected
 
     # handle cases where exception occured during render
     if performed?
-      response.status = status_code
-      response_body = error_message
+      self.status = status_code
+      self.response_body = error_message
     else
       render status: status_code, text: error_message
     end
--- a/test/functional/smart_answers_controller_test.rb
+++ b/test/functional/smart_answers_controller_test.rb
@@ -103,13 +103,6 @@ class SmartAnswersControllerTest < ActionController::TestCase
       assert_equal 503, response.status
     end
 
-    should 'cope with 503 exception during rendering' do
-      @controller.view_context_class.any_instance.stubs(:content_for).raises(GdsApi::TimedOutException)
-
-      get :show, id: 'smart-answers-controller-sample'
-      assert_equal 503, response.status
-    end
-
     should "404 Not Found if request is for an unknown format" do
       @controller.stubs(:respond_to).raises(ActionController::UnknownFormat)
 
--- a/test/integration/error_handling_test.rb
+++ b/test/integration/error_handling_test.rb
@@ -1,0 +1,97 @@
+require_relative '../integration_test_helper'
+
+class ErrorHandlingTest < ActionDispatch::IntegrationTest
+  class ExampleController < ApplicationController
+    class << self
+      attr_accessor :exception_to_raise_before_render
+      attr_accessor :exception_to_raise_after_render
+    end
+
+    def test
+      set_slimmer_headers(skip: true)
+      if self.class.exception_to_raise_before_render
+        raise self.class.exception_to_raise_before_render
+      end
+      render text: 'rendered-from-test-action'
+      if self.class.exception_to_raise_after_render
+        raise self.class.exception_to_raise_after_render
+      end
+    end
+  end
+
+  setup do
+    Rails.application.routes.draw do
+      get '/test' => 'error_handling_test/example#test'
+    end
+  end
+
+  teardown do
+    Rails.application.reload_routes!
+    ExampleController.exception_to_raise_before_render = nil
+    ExampleController.exception_to_raise_after_render = nil
+  end
+
+  context 'when GdsApi::TimedOutException raised before render' do
+    setup do
+      ExampleController.exception_to_raise_before_render = GdsApi::TimedOutException
+    end
+
+    should 'set response status code to 503' do
+      get '/test'
+      assert_response 503
+    end
+
+    should 'render error message as response body' do
+      get '/test'
+      assert_equal '503 error', response.body
+    end
+  end
+
+  context 'when ActionController::UnknownFormat raised before render' do
+    setup do
+      ExampleController.exception_to_raise_before_render = ActionController::UnknownFormat
+    end
+
+    should 'set response status code to 404' do
+      get '/test'
+      assert_response 404
+    end
+
+    should 'render error message as response body' do
+      get '/test'
+      assert_equal '404 error', response.body
+    end
+  end
+
+  context 'when GdsApi::TimedOutException raised after render' do
+    setup do
+      ExampleController.exception_to_raise_after_render = GdsApi::TimedOutException
+    end
+
+    should 'set response status code to 503' do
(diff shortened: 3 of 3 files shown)


chrisroos:
What are the symptoms of a `DoubleRenderError`?

Having read the comments in PR #2255 and here, I'm afraid I don't really understand what problem this area of code is solving.

erikse:
I had look in Errbit last week and couldn't find any of the old `DoubleRenderError` entries, so not sure what happened to them, it would be useful to see a stack trace.

At the time of the fix - I thought the cause was when rendering question pages with a country list, the call to GDS API to refresh the cached location data was timing out, raising an exception. The exception handling code in `ApplicationController` would catch the GDS API exception, and call render a second time with an error message triggering `DoubleRenderError`. 

Maybe my understanding at that time was wrong, if I understand this PR's description, a `DoubleRenderError` is triggered only after the first call to render complete? If that's the case, I'm curious what code after render is complete would trigger a GDS API exception to invoke the exception handling in `ApplicationController`.
```ruby
  def show
    set_slimmer_artefact(@presenter.artefact)
    respond_to do |format|
      format.html { render }
      format.json {
        html_fragment = with_format('html') {
          render_to_string(partial: "content")
        }
        render json: {
          url: smart_answer_path(params[:id], 'y', @presenter.current_state.responses),
          html_fragment: html_fragment,
          title: @presenter.current_node.title
        }
      }
      if render_text?(@presenter)
        format.text {
          render
        }
      end
    end

    set_expiry
  end
```
Either way, the fix adding `self.` to `response_body` to retain the error message correctly, and improving the tests to understand the problem better, look great to me.

erikse:
Looking at the [rails 4.2 implementation of render](https://github.com/rails/rails/blob/4-2-stable/actionpack/lib/action_controller/metal/rendering.rb#L14-L17), seems having any previous content in  `self.response_body` will raise a `DoubleRenderError`.
```ruby
module ActionController
  module Rendering
    extend ActiveSupport::Concern

    def render(*args) #:nodoc:
      raise ::AbstractController::DoubleRenderError if self.response_body
      super
    end
```

floehopper:
I think Errbit might delete reported exceptions on every deployment, although I thought it just marked them as resolved. Anyway, that might explain why @erikse couldn't find the double render exceptions.

floehopper:
@chrisroos:

I wasn't involved with the original PR which added this functionality, so I can't add anything to @erikse's explanation of why it got added.

I'm not quite sure what you mean when you say "What are the symptoms of a `DoubleRenderError`?". However, maybe it will help if I tell you that if I remove the `if performed?` branch of the code in `ApplicationController#error` then I see test failures like this one:

```
ErrorHandlingTest#test_: when ActionController::UnknownFormat raised after render should replace response body with error message. :
AbstractController::DoubleRenderError: Render and/or redirect were called multiple times in this action. Please note that you may only call render OR redirect, and at most once per action. Also note that neither redirect nor render terminate execution of the action, so if you want to exit an action after redirecting, you need to do something like "redirect_to(...) and return".
    app/controllers/application_controller.rb:32:in `error'
    app/controllers/application_controller.rb:16:in `error_404'
    test/integration/error_handling_test.rb:93:in `block (2 levels) in <class:ErrorHandlingTest>'
```

Since there is at least a theoretical possibility of an exception being raised after rendering, then I think this functionality is useful in avoiding the masking of the underlying exception.

floehopper:
@chrisroos: Do you have any objections to this being merged?

chrisroos:
Apologies for taking so long to get back to this PR, @floehopper. The short version is that this looks good to me. The longer version is below.

When I asked about the symptoms of a `DoubleRenderError`, I was trying to understand what we'd see in the logs/errbit and what a user would see in the browser. I guess I was ultimately trying to understand whether we even needed this code. I've now found some errbit emails with `DoubleRenderError` exceptions that appear to stop after PR #2255 was deployed at the end of January, so I presume the original fix is preventing them. It'd be great to really understand where these `DoubleRenderError`s were coming from (something in Slimmer or other middleware perhaps?) but I appreciate that that's out of scope of this PR.

The other question I had was about the setting of `self.status` and `self.response_body`. I wondered whether we could just remove this code given that it hasn't been working for the past few months. After a little investigation (see below), I can see that the behaviour in this branch is correct.

In order to see the effect of setting `self.status` and `self.response_body` I added `raise GdsApi::TimedOutException` on the last line of `SmartAnswersController#show`. I then observed the behaviour in a few different situations:

* Remove the `if performed?` block and just use `render status: status_code, text: error_message`
  * This raises the `DoubleRenderError` exception when viewing /marriage-abroad/y

* Keep the `if performed?` block but remove `self.status = status_code` and `self.response_body = error_message`.
  * This renders the page at /marriage-abroad/y but with slightly wonky styling

* Keep the behaviour in this branch
  * This renders the "Sorry, we are experiencing technical difficulties" page if slimmer is enabled (default)
  * This renders "503 error" if slimmer is disabled (by setting `set_slimmer_headers(skip: true)`).

floehopper:
@chrisroos: Thanks for reviewing. I did some similar tests to you and that's the origin of the `ErrorHandlingTest` which I added in this PR.

floehopper:
I'm going to rebase this against master and (since there are no commit comments) force-push in preparation for merging.

PR #2401, opened 2016-03-23
Use auto-generated Rubocop config

floehopper:
### Motivation

Previously we were using the `--diff --cached` options with `govuk-lint-ruby` to only run the Rubocop checks against code that had been *changed* in the given branch i.e. it was a way to avoid having to fix all the violations up-front.

However, I'm not convinced this is a very sensible approach, because it means we fix the violations in rather a piecemeal way. Also I think we've run into a catch-22 situation a couple of times where we wanted to revert a merge commit to put the code back *exactly* as it was previously. The problem is that the revert commit can draw Rubocop's attention to "old" lines of code that may not pass the Rubocop checks - thus preventing us from doing the straightforward revert.

Trello card: https://trello.com/c/k4VNXtA2

### Description

This PR takes a different approach as follows:

* Use `govuk-lint-ruby --auto-gen-config` to generate `.rubocop_todo.yml` file to [disable cops for existing violations][1]
* Include `.rubocop_todo.yml` into configuration via `inherit_from` key in `.rubocop.yml` - note that this does not override the configuration supplied by the `govuk-lint` gem, but just disables some of the cops for some files.
* Use Rubocop auto-correct to fix violations for 7 cops that were completely disabled in `.rubocop_todo.yml`
* Change `jenkins.sh` CI script to stop using `--diff --cached` options
* Update documentation about Rubocop usage

The reason that I fixed violations of the 7 cops in this PR was because I wanted to end up with a `rubocop_todo.yml` file with *no* cops *completely* disabled. I was concerned that this would make it all to easy to accidentally introduce further violations of the disabled cops.

Even so, this isn't a perfect solution, because it is still possible to accidentally introduce violations of the other cops in files which are already excluded. However, I still think this is a better approach.

Hopefully we can fix the majority of the violations in the not too distant future - I notice that 35 of the 47 cops remaining in the todo list allow auto-correction.

### Internal changes

This PR should change the options for the `govuk-lint-ruby` command executed on the CI server.

### External changes

None.

[1]: https://github.com/bbatsov/rubocop/blob/master/README.md#automatically-generated-configuration


doc/rubocop.md
```diff
@@ -4,11 +4,17 @@ We're using the govuk-lint Gem to include Rubocop and the GOV.UK styleguide rule
 
 ### Jenkins
 
-We run Rubocop as part of the test suite executed on Jenkins (see jenkins.sh). Rubocop only tests for violations in files introduced in the branch being tested. This should prevent us from introducing new violations without us having to first fix all existing violations.
+We run Rubocop as part of the test suite executed on Jenkins (see jenkins.sh). We've generated a project-specific `.rubocop_todo.yml` file using `govuk-lint-ruby --auto-gen-config --exclude-limit 99999` which [excludes existing violations on a per-cop basis][1]. The limit was set to `99999` to minimise the number of cops that are disabled entirely - most are only disabled for a specific set of files.
+
+The generated `.rubocop_todo.yml` file is included into the configuration via a local `.rubocop.yml` file using an `inherit_from` key. This means that the exclusions in `.rubocop_todo.yml` are used for all `govuk-link-ruby` commands by default.
```
chrisroos: Typo: govuk-link-ruby

doc/rubocop.md
```diff
@@ -4,11 +4,17 @@ We're using the govuk-lint Gem to include Rubocop and the GOV.UK styleguide rule
...
+
+Using this approach should prevent us from introducing new violations into the project without needing to fix all the existing violations first. However, in the long run, the idea is for us to gradually remove these exclusions to bring the project fully in line with the default `govuk-link-ruby` configuration.
```
chrisroos: Typo: govuk-link-ruby

chrisroos:
Very minor. Commit note for b9697f2de8e4fcdf0755832f6e767ea97b94390a contains typo: govuk-link-ruby instead of govuk-lint-ruby.

chrisroos:
It wasn't immediately obvious to me whether we were still using the Rubocop rules from govuk-lint. I think it'd be good to update the rubocop.md docs and/or this PR description to explain that we are still using them.

chrisroos:
Do you any thoughts as to how we go about fixing existing violations after this change?

chrisroos:
I'm slightly wary that we might be moving Smart Answers away from the standard use of `govuk-lint` (assuming there is one) but I think the advantages of using Rubocop's built-in functionality outweigh that potential problem.

This looks good to me.

doc/rubocop.md
floehopper: Fixup commit: 016ec222a50cad7c7235db190589bdcb30fef7ad

doc/rubocop.md
floehopper: Fixup commit: 016ec222a50cad7c7235db190589bdcb30fef7ad

floehopper:
> Commit note for b9697f2 contains typo: govuk-link-ruby instead of govuk-lint-ruby.

I've added an empty squash commit to remind me to fix this: d4f1cdd695711c801291aab6092e1b750ba7d2d7

floehopper:
> It wasn't immediately obvious to me whether we were still using the Rubocop rules from govuk-lint. I think it'd be good to update the rubocop.md docs and/or this PR description to explain that we are still using them.

* I've tried to clarify this in the docs in this fixup commit: e455a137c5c51a6830f1241bdf3bcd2df473ba69
* I've also tried to clarify this by updating the PR description
* I've added an empty squash commit to remind me to improve the relevant commit note along similar lines: f479508d01b5f35e7b0b171fa668391040a72ec0

floehopper:
> Do you any thoughts as to how we go about fixing existing violations after this change?

I was thinking of trying to use the Rubocop auto-correct mechanism for those cops which support it. I think that could be done in a single PR with the violations for one cop auto-corrected per commit and that cop enabled. That should cover most of the violations. We can then look at what's left.

floehopper:
> I'm slightly wary that we might be moving Smart Answers away from the standard use of govuk-lint (assuming there is one) but I think the advantages of using Rubocop's built-in functionality outweigh that potential problem.

I'm not sure I understand what you're saying here. This PR is still using `govuk-lint` - it's just that the latter delegates almost all of its Ruby lint-ing to Rubocop. Also we are still using all the configuration from the `govuk-lint` gem - it's just that we've disabled some of its cops for some files.

floehopper:
@chrisroos: Thanks for all the feedback. I think I've addressed it all - I'm planning on doing an interactive rebase to incorporate the fixup & squash commits and then I'll rebase against master, force-push and get this merged.

floehopper:
Superseded by #2401 - I chose to open a new PR to preserve the comments on this one.

chrisroos:
>> I'm slightly wary that we might be moving Smart Answers away from the standard use of govuk-lint (assuming there is one) but I think the advantages of using Rubocop's built-in functionality outweigh that potential problem.

> I'm not sure I understand what you're saying here. This PR is still using `govuk-lint` - it's just that the latter delegates almost all of its Ruby lint-ing to Rubocop. Also we are still using all the configuration from the `govuk-lint` gem - it's just that we've disabled some of its cops for some files.

Apologies for not being clearer.

I guess I'm a bit concerned that most/all other projects will be using `govuk-lint-ruby --diff --cached` and we're now doing something different to essentially achieve the same thing (ensuring we don't introduce new violations without having to fix all existing violations first). I half wonder whether it's worth suggesting this approach over the use of `--diff --cached` in the `govuk-lint` project. What do you think, @floehopper?

PR #2402, opened 2016-03-23
Remove use_legacy_data option for country_select question

floehopper:
This option isn't used in any production code and so it seems safe to remove it.


chrisroos:
Looks good to me.

floehopper:

PR #2403, opened 2016-03-23
Remove unused FCOEmbassyScraper

floehopper:
### Description

* This class was [added 4 years ago in 2012][1].
* There are no references to it in the code or the documentation.
* I'm not aware of any developers using it in the last year or so.
* The www.fco.gov.uk URL which it scrapes now redirects to [a page on www.gov.uk][2], which I think is the source for the Worldwide API that the app is currently using to obtain embassy data anyway.

Note that this PR includes a squash commit of the branch from #2402 which I plan to remove before merging.

### External changes

None.

[1]: 0c19b1596706691c943951cc05d1b8b3c273282e
[2]: https://www.gov.uk/government/world/organisations


chrisroos:
Looks good to me.

floehopper:
Rebasing against master, removing temporary squash commit and force-pushing in preparation for merging.

PR #2405, opened 2016-03-24
Introduce on_response blocks to allow setting of attributes on calculator in new-style flows

floehopper:
This is a spike to explore the ideas in this [Trello card][1].

The idea is that we'll use these `on_response` blocks in a question block to set attributes on the calculator/policy object instead of doing that in `next_node` blocks which is the current "new-style" approach.

One advantage of doing this is to separate out the attribute-setting code which should make it easier to do automatically in future.

Another advantage is that we can avoid having to pass the response into any validation-related methods on the calculator which should make them more ActiveModel-like.

The new `on_response` blocks are run after `next_node_calculation` blocks, but before `validate` blocks. Ideally I would've preferred to have them run *before* `next_node_calculation` blocks too, but since the `calculator` is currently usually instantiated in a `next_node_calculation` bloc, this seemed like the simplest solution.

An alternative would be to use a `precalculate` block to instantiate the `calculator`, however, currently these don't get executed for the first question.

Although I have some reservations about this approach in the longer term (see Trello card comments), I think this is a step forward.

I'd love feedback, especially from @erikse & @chrisroos who were involved in the [original comment thread][2]. If this makes sense, I would open a new PR to do it properly and to use `on_response` blocks in all new-style flows.

[1]: https://trello.com/c/V6nqIWcJ
[2]: https://github.com/alphagov/smart-answers/commit/211734a881a56edeb0d73ecd88c57e7d9a040f67#commitcomment-16645397

### DIFF
--- a/lib/smart_answer/block.rb
+++ b/lib/smart_answer/block.rb
@@ -1,0 +1,14 @@
+module SmartAnswer
+  class Block
+    def initialize(&block)
+      @block = block
+    end
+
+    def evaluate(previous_state, response = nil)
+      args = [response].compact
+      previous_state.instance_exec(*args, &@block)
+      new_state = previous_state.dup
+      new_state.freeze
+    end
+  end
+end
--- a/lib/smart_answer/calculators/statutory_sick_pay_calculator.rb
+++ b/lib/smart_answer/calculators/statutory_sick_pay_calculator.rb
@@ -55,39 +55,39 @@ module SmartAnswer
         period.number_of_days
       end
 
-      def valid_last_sick_day?(value)
-        period = DateRange.new(begins_on: sick_start_date, ends_on: value)
+      def valid_last_sick_day?
+        period = DateRange.new(begins_on: sick_start_date, ends_on: sick_end_date)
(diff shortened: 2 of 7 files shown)


chrisroos:
This looks good to me, @floehopper. I understand the reservations you've mentioned in the Trello card but still think this is an improvement over what we have right now.

Did you consider instantiating the calculators in the `on_response` block of the first question? I think that would allow us to evaluate these blocks before `next_node_calculation`.

Can you explain why you went with the rather generic `Block` class name (not that I have any alternative suggestions)?

floehopper:
Superseded by #2408.

floehopper:
@chrisroos:

Thanks for your feedback.

* I didn't consider instantiating the calculator in the `on_response` blocks. In the [new PR][1], I've now done so and I've made the `on_response` blocks execute *before* `next_node_calculation` blocks as you suggested.

* I've attempted to explain why I used the name `SmartAnswer::Block` in aaa6694003e680f1eb8800aabc80892cb93b5064.

[1]: https://github.com/alphagov/smart-answers/pull/2408

PR #2408, opened 2016-03-25
Introduce on_response blocks to set attributes on calculator

floehopper:
## Description

Supersedes #2405.

Trello card: https://trello.com/c/V6nqIWcJ

The idea is that we'll use these `on_response` blocks in a question block to set attributes on the calculator/policy object instead of doing that in `next_node` blocks which is the current "new-style" approach.

I've updated the following flows to use the new `on_response` blocks:

* check-uk-visa
* marriage-abroad
* overseas-passports
* part-year-profit-tax-credits

These are the only flows which use the new-style pattern of instantiating a calculator object in the first question, setting each response as an attribute on the calculator, and encapsulating all the policy logic in the calculator. I don't think it makes sense to apply the same approach to any of the other flows until they have been converted closer to the new-style.

## Motivation

One advantage of doing this is to separate out the attribute-setting code which should make it easier to do automatically in future. Theoretically the `next_node` block will only need access to the calculator and not to the response.

Another advantage is that we can avoid having to pass the response into any validation-related methods on the calculator which should make them more ActiveModel-like.

The new `on_response` blocks are run before both `next_node_calculation` and `validate` blocks.

## External changes

None. This is just a refactoring.

### DIFF
--- a/test/unit/smart_answer_flows/calculate_statutory_sick_pay_flow_test.rb
+++ b/test/unit/smart_answer_flows/calculate_statutory_sick_pay_flow_test.rb
@@ -46,58 +46,67 @@ module SmartAnswer
 
     context 'when answering linked_sickness_end_date? question' do
       context 'and linked sickness ends more than 8 weeks before sickness starts' do
+        setup do
+          @calculator.linked_sickness_start_date = Date.parse('2015-01-01')
+          @calculator.days_of_the_week_worked = %w(1 2 3 4 5)
+          @calculator.stubs(
+            within_eight_weeks_of_current_sickness_period?: false,
+            at_least_1_day_before_first_sick_day?: true,
+            valid_linked_period_of_incapacity_for_work?: true
+          )
+        end
+
         should 'raise an exception' do
           exception = assert_raise(SmartAnswer::InvalidResponse) do
             setup_states_for_question(:linked_sickness_end_date?,
               responding_with: '2015-01-07',
-              initial_state: {
-                calculator: stub('calculator', {
-                  linked_sickness_start_date: Date.parse('2015-01-01'),
-                  days_of_the_week_worked: %w(1 2 3 4 5),
-                  within_eight_weeks_of_current_sickness_period?: false,
-                  at_least_1_day_before_first_sick_day?: true,
-                  valid_linked_period_of_incapacity_for_work?: true
-                }),
-              })
+              initial_state: { calculator: @calculator }
+            )
           end
           assert_equal 'must_be_within_eight_weeks', exception.message
         end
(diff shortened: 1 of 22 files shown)


floehopper:
Since no comments have been added to the commits in this PR, I'm going to rebase it against master and force-push to fix the conflicts reported by GitHub above.

floehopper:
Rebase & force-push done - GitHub is now showing no conflicts. This is ready for review.

test/unit/smart_answer_flows/calculate_statutory_sick_pay_flow_test.rb
```diff
@@ -46,58 +46,67 @@ class CalculateStatutorySickPayFlowTest < ActiveSupport::TestCase
```
chrisroos: Would you be able to update the commit note to explain why you've used a real instance instead of a stub?

chrisroos:
Would you be able to update the commit note for 4a26b6c831b3551cafeee6de3a6441554c994a7a ("Make CalculateStatutorySickPayFlowTest more realistic") to explain why you've used a real instead instead of a stub?

chrisroos:
> Fix Style/ClosingParenthesisIndentation Rubocop violations (7bf40ba6d8f462b046e95e57d8e0befbbaf53488)
> 
> Annoyingly I don't think these are real violations - I think they are caused by
> a problem with the `-diff` option on `govuk-lint-ruby`. However, this seems to
> fix the problem and I don't mind the style of the resulting code.

You might be able to change this back now that we've stopped using the `--diff` option of `govuk-lint-ruby`.

chrisroos:
This looks great to me, @floehopper. Good work!

floehopper:
> Would you be able to update the commit note for 4a26b6c831b3551cafeee6de3a6441554c994a7a ("Make CalculateStatutorySickPayFlowTest more realistic") to explain why you've used a real instead instead of a stub?

I've added an empty squash-commit with some notes to remind me to improve the commit note:  21c04eeaa030e34f46bde01efc9166d63168a959.

floehopper:
> You might be able to change this back now that we've stopped using the `--diff` option of `govuk-lint-ruby`.

I tried reverting that commit, but the violations still happened, so I think I'll leave it in place. I've added an empty squash-commit to remind me to update the commit note accordingly: dacb3c93db522b4646b2ad7f92ac5dc8e3efa8ea.

floehopper:
I'm going to do an interactive rebase to incorporate the squash-commits, then rebase against master to resolve the conflicts reported by GitHub, before force-pushing (since there are no commit comments) in preparation for merging.

PR #2337, opened 2016-03-07
Stop abc test on start now buttons

h-lame:
Now that the abc test on start now buttons (see: #2282) has been running for a while we are ready to stop it.

This reverts commit c950be53933ca6a2a4600507fd304131cc1137b2, reversing
changes made to 5a1581d6edbba1a5d20b25fb92e15b988cc48ba6.

Note this isn't a clean revert as we've left in the Gemfile and Gemfile.lock changes to keep the govuk_frontend_toolkit upgrade.

This is for: https://trello.com/c/NaRemNDC/270-end-the-start-now-button-a-b-c-test

Do not merge/deploy until 8th March 2016.

PR #2411, opened 2016-03-29
Re-start A/B/C test on 'Start now' buttons

davidbasalla:
Reverts alphagov/smart-answers#2337

For https://trello.com/c/KbYQzkNV/351-run-start-now-button-test-again-1

We ran a multivariate test for 'start now' buttons on 3 start pages (https://trello.com/c/GYqg2EEj) which was merged March 8 2016.

We want to run the test again to increase our confidence that the improvement trends shown in the original test are valid, with the same options winning and with a similar scale if difference. The original test was our first experiment and some of the data seems odd, so a chance to cross check the results is valuable.

We want to run the test for at least 14 days so that we can easily compare the results with those from the original test.

### DIFF
--- a/app/assets/javascripts/smart-answers.js
+++ b/app/assets/javascripts/smart-answers.js
@@ -1,3 +1,5 @@
+//= require start-now-button-abc-tests
+
 function browserSupportsHtml5HistoryApi() {
   return !! (history && history.replaceState && history.pushState);
 }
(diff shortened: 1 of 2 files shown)


floehopper:
While I understand that you've used the default `git revert` commit note, I think it might be more helpful to use a custom commit note more like this:

> Re-start abc test on start now buttons

Or even this:

> Re-start A/B/C test on 'Start now' buttons

It would also be nice to update the PR title to match.

floehopper:

davidbasalla:
Thank you for your comments @floehopper, have addressed and rebased.

PR #2419, opened 2016-03-30
[wip] National minimum living wage

pmanrubia:


floehopper:
In general this looks pretty good to me, but as you requested, I'll add some comments with suggestions about how it might be good to organise the commits.

floehopper:
Regarding the regression tests, I think we've used two main approaches:

1. This is the preferred approach, but it's not always realistic e.g. if the regression tests for the flow take too long to run. Update *all* the relevant regression test files after *every* commit, so that the regression tests all pass after every commit. If your commits are already small enough, then the changes to the regression test files shouldn't be too big and it should be easy to see what's going on.

2. This is a variation of approach 1. If one or more of your commits is quite big and you can't see a way to make it smaller while still keeping it atomic, then I think it's reasonable to split the related changes to the regression test files into one or more separate commits immediately following the commit where the actual change was made. This should make it easier for someone looking at the diffs to see what's going on.

3. If the tests take too long to run to make either approach 1 or 2 feasible, I think it's reasonable to group all the changes to the regression test files into one or more commits which appear *after* the commits containing all your actual changes. The problem with this is when there are a lot of changes it's hard to know whether all the changes were intentional.

Note that we haven't used approach 1 or 2 much recently, because most of the changes have been to flows whose regression tests take a long time to run e.g. marriage-abroad.

floehopper:
I hope that helps. Let me know if you have any questions. I should be online, at least for a while, first thing tomorrow (Thursday).

pmanrubia:
Hi @floehopper, 

1. Thanks for the review of this branch, and for being so responsive when I pinged you out of normal working hours.
2. Unifying commits `national living wage` into `National Living Wage`: Good call. :+1:
3. Agree on option 1 as the preferred for regression tests and commits; not sure if I will be able to do it for this PR considering the deadline of 1st April; I guess I will try to do 2), and follow 3) if I see any risks with the deadline.
3. Commits granularity: I am happy to break them up. I’m used to squash them into bigger commits with detailed descriptions, but seems like a good idea to follow the same approach you are following in other PRs. Thanks for sharing your suggestion about the way to break them up, by the way.
4. I will try to address the other comments are part of the remaining work.

Again, thank you for the early review

floehopper:
@pmanrubia:

My suggestions are all inevitably somewhat subjective, so please feel free to do things differently if you're short of time or you can see a better way.

I'd suggest closing this PR now that it's served it's purpose.

pmanrubia:
Thanks for your suggestions @floehopper.
They have been very valuable. I am trying to address all of them.

Closing this PR as it was intended for getting feedback.

PR #2420, opened 2016-03-31
Remove space stripping ERB tags from marriage-abroad templates

chrisroos:
This replaces all uses of `<%- -%>` and `<%= -%>` with `<% %>` and `<%= %>` respectively.

We started using `<%= -%>` when converting from i18n YAML templates to ERB templates. Using `-%>` to strip trailing spaces/newlines allowed us to exactly replicate the outcomes that were previously generated by the i18n YAML system. This gave us confidence that the conversion was accurate. These are no longer necessary and so I've removed them in favour of consistently using `<%= %>`.

Using `<%- -%>` doesn't have any effect and can safely be replaced with `<% %>`. From the [`ActionView::Base` docs](http://api.rubyonrails.org/classes/ActionView/Base.html):

> When on a line that only contains whitespaces except for the tag, <% %> suppress leading and trailing whitespace, including the trailing newline. <% %> and <%- -%> are the same. Note however that <%= %> and <%= -%> are different: only the latter removes trailing whitespaces.

  I ran all regression tests after making these changes. All changes to the artefacts are due to the addition of non-significant whitespace so I'm happy to accept the changes.


floehopper:
Minor: I think it might be clearer to group the 3 commits for each flow together, i.e. remove space-stripping ERB tags, update artefacts, update checksums. But I don't feel strongly about this.

floehopper:
Very minor: I find "Update artefacts" and "Update checksums" in the commit note subjects overly generic - I'd prefer "Update regression test artefacts" and "Update regression test checksums". However, I realise this would mean very long subject lines and perhaps we all know what you mean.

floehopper:
Very minor: It's pretty obvious, but I think it would be good to say that the regression tests were all passing in the commit notes for updating checksums.

floehopper:
Do you think it would be worth trying to add some kind of check that we don't re-introduce any of these? I wonder whether `govuk-lint` would allow this?

floehopper:
I think it might be worth making explicit the motivation behind this PR, which I assume is to simplify the code i.e. the non-space-stripping ERB tags are conceptually simpler and terser.

floehopper:
Apart from my minor comments, this looks good to me.

chrisroos:
> I think it might be worth making explicit the motivation behind this PR, which I assume is to simplify the code i.e. the non-space-stripping ERB tags are conceptually simpler and terser.

I've added the following sentence to the PR description: "We don't need the space/newline stripping behaviour and consistently using the same ERB tags leaves us with a slightly simpler system." Hopefully that makes sense!

chrisroos:
> Do you think it would be worth trying to add some kind of check that we don't re-introduce any of these? I wonder whether govuk-lint would allow this?

A very, very quick search suggests that Rubocop (and therefore `govuk-lint-ruby`) [doesn't support parsing ERB templates](https://github.com/bbatsov/rubocop/issues/1113).

I'm not sure it's worth the effort of trying to ensure these aren't introduced in future. I'm happy to leave it to see if we notice any more being added before trying to do something about it.

chrisroos:
> Minor: I think it might be clearer to group the 3 commits for each flow together, i.e. remove space-stripping ERB tags, update artefacts, update checksums. But I don't feel strongly about this.

This was a side effect of how I made these changes:

1. I used find/replace to replace the space-removing ERB tags.
2. I used `git add <dir>` and `git commit <dir>` to add the changes for a specific set of files (e.g. a single Smart Answer)
3. I ran _all_ the regression tests
4. I used `git add <dir>` and `git commit <dir>` to add the changes for each Smart Answer
5. I updated the checksum data for each Smart Answer and committed those changes

I'm going to leave the commits as they are given the time I think it might take to combine the commits and ensure I've not missed anything.

chrisroos:
> Very minor: I find "Update artefacts" and "Update checksums" in the commit note subjects overly generic - I'd prefer "Update regression test artefacts" and "Update regression test checksums". However, I realise this would mean very long subject lines and perhaps we all know what you mean.

I'm happy to leave these as they are. Any potential confusion will hopefully be resolved when looking at the files in the commit (i.e. it contains changes to the regression test checksum file or regression test artefacts).

PR #2441, opened 2016-04-07
Fix minimum wage rates for people aged 24

chrisroos:
Trello card: https://trello.com/c/n72PekVB

A problem in the minimum_wage_data.yml file meant that responding to the am-i-getting-minimum-wage "age" question with "24" resulted in exceptions being logged in errbit.

I've fixed the problem in minimum_wage_data.yml and have also updated the data so that the National Living Wage only applies from 1 Apr 2016 (instead of 1 Oct 2015).


pmanrubia:
:+1: Thanks for fixing @chrisroos

PR #2438, opened 2016-04-05
Update marriage-abroad services

chrisroos:
Trello cards:

* Standardise service names - https://trello.com/c/qMeYVdfh
* Update country/service groups - https://trello.com/c/D0ZMM9v5

I've used the data from the spreadsheet attached to the Trello cards to update the marriage-abroad services data.

I created the update-marriage-abroad-services.rb script to make this process easier. The script takes a yaml file of data necessary to update the services for a combination of country, partner sex, residency and partner nationality. I used it to update each of the groups of countries that share a set of services. I updated the regression test artefacts in each of these commits so that it's easier to see the differences that have been introduced.

With all the services updated I was able to remove a number of no longer used services, before finally renaming two services whose names had changes.

It's important to note that not all of the changes to the marriage_abroad_services.yml data results in services being displayed. Notably, the changes to France and French overseas territories ("certificate of custom and law") are not currently being (and never have been) displayed. I plan to investigate these separately.

## Fact check

There are lots of changes in this branch. I've chosen a single example to illustrate one of them:

* Opposite sex ceremony in Spain with British partner, when resident in the UK:
  * [GOV.UK](https://www.gov.uk/marriage-abroad/y/spain/uk/partner_british/opposite_sex#fees) - Observe the Fees section is present
  * [Heroku](https://smart-answers-pr-2438.herokuapp.com/marriage-abroad/y/spain/uk/partner_british/opposite_sex#fees) - Observe the Fees section has been removed, , which matches the data in the spreadsheet

* Opposite sex ceremony in Spain with British partner, when resident in Spain:
  * [GOV.UK](https://www.gov.uk/marriage-abroad/y/spain/ceremony_country/partner_british/opposite_sex#fees) - Observe the Fees table contains 5 items
  * [Heroku](https://smart-answers-pr-2438.herokuapp.com/marriage-abroad/y/spain/ceremony_country/partner_british/opposite_sex#fees) - Observe the Fees table contains 2 items, which matches the data in the spreadsheet

PR #2443, opened 2016-04-07
Remove marriage-abroad unused service data

chrisroos:
Trello cards:

* [Opposite sex services in Norway for third_country residents](https://trello.com/c/xQGZ5uuL)
* [Opposite sex services in Slovenia for third_country residents](https://trello.com/c/Vy3XJJVl)
* [Opposite sex services in Ukraine for Ukrainian residents](https://trello.com/c/2jG2LmZi)
* [Same sex services in Romania](https://trello.com/c/NwWDH1m4)

PR #2438 updated the marriage-abroad service data based on a spreadsheet supplied by FCO. The changes to the service data implied that services/fees should be visible in some outcomes that weren't currently showing them. I've now had confirmation that we _shouldn't_ be showing them in four of those cases.

This branch updates the marriage_abroad_services.yml so that it matches our current understanding of when the services/fees should be shown.

This change isn't strictly necessary as logic in the flow/templates meant that these removed services/fees weren't being shown anyway. I think it's good to make this change as it means the data matches our expectations of what's being shown.

## Expected changes

None.

### DIFF
--- a/lib/data/marriage_abroad_services.yml
+++ b/lib/data/marriage_abroad_services.yml
@@ -515,16 +515,6 @@ romania:
       default:
       - :receiving_notice_of_marriage
       - :issuing_cni_or_nulla_osta
-  same_sex:
-    default:
-      partner_british:
-      - :receiving_notice_of_registration
-      - :registering_marriage
-      - :issuing_marriage_certificate
-      partner_other:
-      - :receiving_notice_of_registration
-      - :registering_marriage
-      - :issuing_marriage_certificate
 russia:
   opposite_sex:
     ceremony_country:
--- a/lib/data/marriage_abroad_services.yml
+++ b/lib/data/marriage_abroad_services.yml
@@ -70,7 +70,10 @@ morocco:
       - :affidavit_or_affirmation_for_marriage
 norway:
   opposite_sex:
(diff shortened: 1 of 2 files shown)


pmanrubia:
The commit notes in 433073493bf23951df9f57af769f34ed53d66779: 
> This data implied that the services would be shown for all same ceremonies
where the partner isn't from `Ukraine`.

might need to be updated to
> This data implied that the services would be shown for all same ceremonies
where the partner isn't from `Romania`.

pmanrubia:
@chrisroos minor comment. Good to merge.

chrisroos:
> The commit notes in 433073493bf23951df9f57af769f34ed53d66779: 
>> This data implied that the services would be shown for all same ceremonies
where the partner isn't from `Ukraine`.
>
> might need to be updated to
>> This data implied that the services would be shown for all same ceremonies
where the partner isn't from `Romania`.

Good spot! I'll get it updated.

chrisroos:
Thanks for reviewing, @pmanrubia.

chrisroos:
I rewrote the commit comment to replace 'Ukraine" with "Romania". I rebased on master, force pushed and then merged.

PR #2444, opened 2016-04-07
Refactor marriage-abroad PACS outcomes

chrisroos:
I've added a couple of intention revealing methods to the `MarriageAbroadCalculator`, and split the single `outcome_ceremonies_in_monaco` outcome into two.

## Expected changes

None.


lib/smart_answer/calculators/marriage_abroad_data_query.rb
```diff
@@ -8,6 +8,10 @@ class MarriageAbroadDataQuery
 
     FRENCH_OVERSEAS_TERRITORIES = %w(french-guiana french-polynesia guadeloupe martinique mayotte new-caledonia reunion st-pierre-and-miquelon wallis-and-futuna)
 
+    CEREMONY_COUNTRIES_OFFERING_PACS = %w(
+      france monaco new-caledonia wallis-and-futuna
+    )
+
```
floehopper: Minor: The formatting of this array is inconsistent compared to the ones immediately above and below.

lib/smart_answer_flows/marriage-abroad.rb
```diff
@@ -66,16 +66,18 @@ def define
         end
 
         next_node do
-          if calculator.ceremony_country == 'monaco'
-            if calculator.want_to_get_married?
+          if calculator.want_to_get_married?
+            if calculator.ceremony_country == 'monaco'
               outcome :outcome_marriage_in_monaco
             else
```
floehopper: Out of interest, how do we know that the `else` here means the outcome is one in France or french overseas territory? It seems as if some logic before this must have restricted the possibilities so this `else` works. If so that means the logic here is coupled to that other logic which seems like a bad thing. Obviously this is out-of-scope for this PR, but I thought I'd mention it.

lib/smart_answer/calculators/marriage_abroad_data_query.rb
```diff
@@ -12,6 +12,10 @@ class MarriageAbroadDataQuery
       france monaco new-caledonia wallis-and-futuna
     )
 
+    FRENCH_OVERSEAS_TERRITORIES_OFFERING_PACS = %w(
+      new-caledonia wallis-and-futuna
+    )
+
```
floehopper: Minor: As I mentioned earlier the formatting of arrays in this file seems inconsistent.

test/unit/calculators/marriage_abroad_calculator_test.rb
```diff
@@ -938,6 +938,22 @@ class MarriageAbroadCalculatorTest < ActiveSupport::TestCase
...
+        should 'return true if a PACS is available in the ceremony country and the country is a French overseas territory' do
+          @calculator.ceremony_country = 'new-caledonia'
+          assert @calculator.french_overseas_territory_offering_pacs?
+        end
+
+        should "return false if PACS is available in the ceremony country but it's not a French overseas territory" do
+          @calculator.ceremony_country = 'france'
+          refute @calculator.french_overseas_territory_offering_pacs?
+        end
+      end
```
floehopper: I don't understand the naming of these tests. The implementation only has a single condition, but the two test names mention two conditions. The test names might indeed by correct, but it would be good to explain how you derived them in the commit note.

floehopper:
Other than my minor comments, this looks good to me. 👍

chrisroos:
I'm closing this to rebase on master and address @floehopper's feedback.

lib/smart_answer_flows/marriage-abroad.rb
chrisroos: This question (`marriage_or_pacs?`) can only be reached if the ceremony country offers [PACS](https://en.wikipedia.org/wiki/Civil_solidarity_pact). According to our system, only France, Monaco, New Caledonia and Wallis and Futuna offer PACS. So, if the ceremony country isn't Monaco (which is a special case for reasons I don't know) then it must be France or one of the French overseas territories.

lib/smart_answer/calculators/marriage_abroad_data_query.rb
chrisroos: This was because I was trying to avoid the long lines that the existing style leads to. However, consistency is more important so I've changed these in https://github.com/alphagov/smart-answers/pull/2470/commits/80e8c02950f22a5da52da69b1697c588ca2aa1fa.

lib/smart_answer/calculators/marriage_abroad_data_query.rb
chrisroos: This was because I was trying to avoid the long lines that the existing style leads to. However, consistency is more important so I've changed these in https://github.com/alphagov/smart-answers/pull/2470/commits/1a316a1b8d8f6b4c2b8733750c588559937263bc.

test/unit/calculators/marriage_abroad_calculator_test.rb
chrisroos: I've amended this commit in https://github.com/alphagov/smart-answers/pull/2470/commits/1a316a1b8d8f6b4c2b8733750c588559937263bc - hopefully that makes the naming of these tests clearer.

chrisroos:
Thanks for the review, @floehopper. I've addressed your comments and opened PR #2470 ready to merge to master.

PR #2452, opened 2016-04-08
Refactor Benefit cap calculator (Making benefit rates and type configurable)

ikennaokpala:
Trello Story: https://trello.com/c/LvffB57l/100-refactor-benefit-cap-calculator

## Motivation 

A final consideration is that when the new cap go live this autumn, the existing flow will need to be removed, and the new flow will become the only flow.

See Trello Story: https://trello.com/c/IVrsJIfM/34-22-april-requires-fix-but-hold-deployment-benefit-cap-calculator-2016-2017-changes

### Options

When reviewing the code for the benefit-cap-calculator, a few things became apparent:

1. The benefit caps for the different family types (single, couple, parent) are hard-coded.
2. The question and outcome templates for exempt benefits contains a hard-coded list of benefits
3. The template for Question 4, that asks you to select the benefits you are receiving is a hard-coded list.
4. Question 5, that asks you to enter the amount of each benefit you receive, is comprised of 15 case statements. Each case statement has a decreasing list of the next valid benefit amount question that could be asked. This was done to make sure that the questions are asked in the order that they are displayed in the checkbox list.

**Option 1: Duplicate the entire flow**

Have a routing question at the beginning that directs to one of two flow sub classes. The two sub classes would be virtually identical except for the points raised in 1-4 above. i.e. the case statements would need to be edited to change the lists of valid next questions.

All of the flow tests would also have to be duplicated.

We decided that this was inefficient, and could lead to things being missed in the interim if the exempt benefits/benefits that affect the cap change before the new caps are applied.

**Option 2: Assign a flag**

Have a routing question at the beginning that assigns a flag value. This flag holds the value of which flow the user is on. 

Using the flag, something like question 5 would have an if/else added to each case statement to determine whether or not the benefit amount question is valid for that flow.

This option appears messy to implement and even harder to unpick come the autumn.

**Option 3: Read config from a YAML file**

To be inline with existing calculators, this option is to move cap values (rates) into a YAML file. This would make it easier to update the benefit caps in the future when they inevitably change again.

Then we looked at the two lists of benefits (exempt and not exempt) and determined that we could also read those from a YAML file as they are also subject to change by policy.

We decided to go with this option as going forward (after the new caps are applied) it will be easier to make changes to exempt benefits, benefits that affect the cap and cap limits in a YAML file rather than throughout the code base.

## Changes

* Introduce YAML to hold benefit cap data:
   * Rates or Weekly benefits cap (For different family types)
   * Exempt Benefits
   *  Benefits that are affected by the cap

*  Update benefit cap calculator flow to read data from YAML for benefit cap calculator.

*  Refactored Question 5 a to o 
     * Replace hard coded case/switch with dynamic loop driven by data from the YAML.
 
* Updated all affected tests.

## Factcheck 

 Link: https://smart-answers-pr-2452.herokuapp.com/benefit-cap-calculator/
 Link on GOV.UK: https://gov.uk/benefit-cap-calculator/

The [factcheck](https://smart-answers-pr-2452.herokuapp.com/benefit-cap-calculator/) should work exactly the same as [GOV.UK](https://gov.uk/benefit-cap-calculator)

### DIFF
--- a/lib/data/benefit_cap_data.yml
+++ b/lib/data/benefit_cap_data.yml
@@ -1,4 +1,21 @@
 ---
-single: 350
-couple: 500
-parent: 500
+rates:
+  single: 350
+  couple: 500
+  parent: 500
+benefits:
+  - bereavement
+  - carers
+  - child_benefit
+  - child_tax
+  - esa
+  - guardian
+  - incapacity
+  - income_support
+  - jsa
+  - maternity
+  - sda
+  - widowed_mother
+  - widowed_parent
+  - widow_pension
+  - widows_aged
--- a/lib/data/benefit_cap_data.yml
+++ b/lib/data/benefit_cap_data.yml
@@ -1,5 +1,5 @@
 ---
-rates:
+weekly_benefit_cap:
   single: 350
   couple: 500
   parent: 500
--- a/lib/data/benefit_cap_data.yml
+++ b/lib/data/benefit_cap_data.yml
@@ -4,18 +4,18 @@ rates:
   couple: 500
   parent: 500
 benefits:
-  - bereavement
-  - carers
-  - child_benefit
-  - child_tax
-  - esa
-  - guardian
-  - incapacity
-  - income_support
-  - jsa
-  - maternity
-  - sda
-  - widowed_mother
-  - widowed_parent
-  - widow_pension
-  - widows_aged
+  bereavement: :bereavement_amount?
+  carers: :carers_amount?
+  child_benefit: :child_benefit_amount?
+  child_tax: :child_tax_amount?
+  esa: :esa_amount?
+  guardian: :guardian_amount?
+  incapacity: :incapacity_amount?
+  income_support: :income_support_amount?
+  jsa: :jsa_amount?
+  maternity: :maternity_amount?
+  sda: :sda_amount?
+  widowed_mother: :widowed_mother_amount?
+  widowed_parent: :widowed_parent_amount?
+  widow_pension: :widow_pension_amount?
+  widows_aged: :widows_aged_amount?
--- a/lib/data/benefit_cap_data.yml
+++ b/lib/data/benefit_cap_data.yml
@@ -1,8 +1,14 @@
 ---
-weekly_benefit_cap:
-  single: 350
-  couple: 500
-  parent: 500
+weekly_benefit_caps:
+  single:
+    amount: 350
+    description: "Single"
+  couple:
+    amount: 500
+    description: "Living as a couple (with or without children)"
+  parent:
+    amount: 500
+    description: "A lone parent with 1 or more dependent children"
 exempt_benefits:
   - Attendance Allowance
   - Disability Living Allowance
--- a/lib/data/benefit_cap_data.yml
+++ b/lib/data/benefit_cap_data.yml
@@ -4,18 +4,48 @@ rates:
   couple: 500
   parent: 500
 benefits:
-  bereavement: :bereavement_amount?
-  carers: :carers_amount?
-  child_benefit: :child_benefit_amount?
-  child_tax: :child_tax_amount?
-  esa: :esa_amount?
-  guardian: :guardian_amount?
-  incapacity: :incapacity_amount?
-  income_support: :income_support_amount?
-  jsa: :jsa_amount?
-  maternity: :maternity_amount?
-  sda: :sda_amount?
-  widowed_mother: :widowed_mother_amount?
-  widowed_parent: :widowed_parent_amount?
-  widow_pension: :widow_pension_amount?
-  widows_aged: :widows_aged_amount?
+  bereavement:
+    question: :bereavement_amount?
+    description: "Bereavement Allowance"
+  carers:
+    question: :carers_amount?
+    description: "Carer's Allowance"
+  child_benefit:
+    question: :child_benefit_amount?
+    description: "Child Benefit"
+  child_tax:
+    question: :child_tax_amount?
+    description: "Child Tax Credit"
+  esa:
+    question: :esa_amount?
+    description: "Employment and Support Allowance"
+  guardian:
+    question: :guardian_amount?
+    description: "Guardian's Allowance"
+  incapacity:
+    question: :incapacity_amount?
+    description: "Incapacity Benefit"
+  income_support:
+    question: :income_support_amount?
+    description: "Income Support"
+  jsa:
+    question: :jsa_amount?
+    description: "Jobseeker’s Allowance"
+  maternity:
+    question: :maternity_amount?
+    description: "Maternity Allowance"
+  sda:
+    question: :sda_amount?
+    description: "Severe Disablement Allowance"
+  widowed_mother:
+    question: :widowed_mother_amount?
+    description: "Widowed Mother's Allowance"
+  widowed_parent:
+    question: :widowed_parent_amount?
+    description: "Widowed Parent's Allowance"
+  widow_pension:
+    question: :widow_pension_amount?
+    description: "Widow's Pension"
+  widows_aged:
+    question: :widows_aged_amount?
+    description: "Widow's Pension (age related)"
--- a/lib/data/benefit_cap_data.yml
+++ b/lib/data/benefit_cap_data.yml
@@ -1,0 +1,4 @@
+---
+single: 350
+couple: 500
+parent: 500
--- a/lib/data/benefit_cap_data.yml
+++ b/lib/data/benefit_cap_data.yml
@@ -3,6 +3,15 @@ rates:
   single: 350
   couple: 500
   parent: 500
+exempt_benefits:
+  - Attendance Allowance
+  - Disability Living Allowance
+  - Industrial Injuries Benefit
+  - Personal Independence Payment
+  - Employment and Support Allowance (support component)
+  - War Widow’s or War Widower’s Pension
+  - Armed Forces Compensation Scheme
+  - Armed Forces Independence Payment
 benefits:
   bereavement:
     question: :bereavement_amount?
--- a/lib/smart_answer/calculators/benefit_cap_calculator_data_query.rb
+++ b/lib/smart_answer/calculators/benefit_cap_calculator_data_query.rb
@@ -1,8 +1,28 @@
 module SmartAnswer::Calculators
   class BenefitCapCalculatorDataQuery
-    attr_reader :data
+    attr_reader :data, :benefits, :questions
     def data
       @data ||= YAML.load_file(Rails.root.join('lib', 'data', 'benefit_cap_data.yml'))
+    end
+
+    def benefits
+      @benefits ||=  @data.fetch("benefits").with_indifferent_access
+    end
+
+    def questions
+      @questions ||=
+        benefits.inject(HashWithIndifferentAccess.new) do |benefits_and_questions, (key, value)|
+          benefits_and_questions[key] = value.fetch("question")
+          benefits_and_questions
+        end
+    end
+
+    def descriptions
+      @descriptions ||=
+        benefits.inject(HashWithIndifferentAccess.new) do |benefits_and_descriptions, (key, value)|
+          benefits_and_descriptions[key] = value.fetch("description")
+          benefits_and_descriptions
+        end
     end
   end
 end
--- a/lib/smart_answer/calculators/benefit_cap_calculator_data_query.rb
+++ b/lib/smart_answer/calculators/benefit_cap_calculator_data_query.rb
@@ -1,12 +1,15 @@
 module SmartAnswer::Calculators
   class BenefitCapCalculatorDataQuery
-    attr_reader :data, :benefits, :questions
     def data
       @data ||= YAML.load_file(Rails.root.join('lib', 'data', 'benefit_cap_data.yml'))
     end
 
+    def rates
+      @rates ||= data.fetch("rates")
+    end
+
     def benefits
-      @benefits ||=  @data.fetch("benefits").with_indifferent_access
+      @benefits ||= data.fetch("benefits").with_indifferent_access
     end
 
     def questions
--- a/lib/smart_answer/calculators/benefit_cap_calculator_data_query.rb
+++ b/lib/smart_answer/calculators/benefit_cap_calculator_data_query.rb
@@ -5,7 +5,7 @@ module SmartAnswer::Calculators
     end
 
     def rates
-      @rates ||= data.fetch("rates")
+      @rates ||= data.fetch("rates").with_indifferent_access
     end
 
     def benefits
--- a/lib/smart_answer/calculators/benefit_cap_calculator_data_query.rb
+++ b/lib/smart_answer/calculators/benefit_cap_calculator_data_query.rb
@@ -4,8 +4,8 @@ module SmartAnswer::Calculators
       @data ||= YAML.load_file(Rails.root.join('lib', 'data', 'benefit_cap_data.yml'))
     end
 
-    def rates
-      @rates ||= data.fetch("rates").with_indifferent_access
+    def weekly_benefit_cap
+      @weekly_benefit_cap ||= data.fetch("weekly_benefit_cap").with_indifferent_access
     end
 
     def benefits
--- a/lib/smart_answer/calculators/benefit_cap_calculator_data_query.rb
+++ b/lib/smart_answer/calculators/benefit_cap_calculator_data_query.rb
@@ -1,0 +1,8 @@
+module SmartAnswer::Calculators
+  class BenefitCapCalculatorDataQuery
+    attr_reader :data
+    def data
+      @data ||= YAML.load_file(Rails.root.join('lib', 'data', 'benefit_cap_data.yml'))
+    end
+  end
+end
--- a/lib/smart_answer_flows/benefit-cap-calculator.rb
+++ b/lib/smart_answer_flows/benefit-cap-calculator.rb
@@ -109,9 +109,9 @@ module SmartAnswer
 
       #Q6
       multiple_choice :single_couple_lone_parent? do
-        option :single
-        option :couple
-        option :parent
+        query.rates.keys.each do |rate|
+          option rate
+        end
 
         calculate :benefit_cap do |response|
           sprintf("%.2f", query.rates.fetch(response))
--- a/lib/smart_answer_flows/benefit-cap-calculator.rb
+++ b/lib/smart_answer_flows/benefit-cap-calculator.rb
@@ -7,7 +7,7 @@ module SmartAnswer
       satisfies_need "100696"
 
       benefit_cap_query = SmartAnswer::Calculators::BenefitCapCalculatorDataQuery.new
-      benefit_cap_data = benefit_cap_query.data
+      benefit_cap_rates = benefit_cap_query.data.fetch('rates')
 
       # Q1
       multiple_choice :receive_housing_benefit? do
@@ -522,11 +522,11 @@ module SmartAnswer
         option :parent
 
         calculate :benefit_cap do |response|
-          sprintf("%.2f", benefit_cap_data.fetch(response))
+          sprintf("%.2f", benefit_cap_rates.fetch(response))
         end
 
         next_node do |response|
-          if total_benefits > benefit_cap_data.fetch(response)
+          if total_benefits > benefit_cap_rates.fetch(response)
             outcome :outcome_affected_greater_than_cap
           else
             outcome :outcome_not_affected_less_than_cap
--- a/lib/smart_answer_flows/benefit-cap-calculator.rb
+++ b/lib/smart_answer_flows/benefit-cap-calculator.rb
@@ -43,6 +43,13 @@ module SmartAnswer
         option :yes
         option :no
 
+        precalculate :exempt_benefits do
+          exempt_benefits = query.data.fetch("exempt_benefits").map do |exempt_benefit|
+            "- #{exempt_benefit}"
+          end
+          exempt_benefits.join("\n")
+        end
+
         next_node do |response|
           if response == 'yes'
             outcome :outcome_not_affected_exemptions
--- a/lib/smart_answer_flows/benefit-cap-calculator.rb
+++ b/lib/smart_answer_flows/benefit-cap-calculator.rb
@@ -8,7 +8,6 @@ module SmartAnswer
 
       query = Calculators::BenefitCapCalculatorDataQuery.new
       rates = query.data.fetch('rates')
-      benefits = query.data.fetch('benefits').with_indifferent_access
 
       # Q1
       multiple_choice :receive_housing_benefit? do
@@ -56,8 +55,8 @@ module SmartAnswer
 
       #Q4
       checkbox_question :receiving_non_exemption_benefits? do
-        benefits.keys.each do |benefit|
-          option benefit.to_sym
+        query.benefits.keys.each do |benefit|
+          option benefit
         end
 
         next_node_calculation :benefit_types do |response|
@@ -76,13 +75,13 @@ module SmartAnswer
           if response == "none"
             outcome :outcome_not_affected
           else
-            question BenefitCapCalculatorFlow.next_benefit_amount_question(benefits, benefit_types)
+            question BenefitCapCalculatorFlow.next_benefit_amount_question(query.questions, benefit_types)
           end
         end
       end
 
       #Q5a-o
-      benefits.each do |(benefit,method)|
+      query.questions.each do |(_benefit, method)|
         money_question method do
 
           calculate :total_benefits do |response|
@@ -90,7 +89,7 @@ module SmartAnswer
           end
 
           next_node do
-            question BenefitCapCalculatorFlow.next_benefit_amount_question(benefits, benefit_types)
+            question BenefitCapCalculatorFlow.next_benefit_amount_question(query.questions, benefit_types)
           end
         end
       end
@@ -175,6 +174,7 @@ module SmartAnswer
       ## Outcome 5
       outcome :outcome_not_affected
     end
+
     def self.next_benefit_amount_question(benefits, selected_benefits)
       benefits.fetch(selected_benefits.shift, :housing_benefit_amount?)
     end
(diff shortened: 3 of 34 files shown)


lib/smart_answer/calculators/benefit_cap_calculator_data_query.rb
```diff
@@ -0,0 +1,8 @@
+module SmartAnswer::Calculators
+  class BenefitCapCalculatorDataQuery
+    attr_reader :data
```
chrisroos: This is unnecessary as it's overridden by the `data` method added below.

lib/smart_answer/calculators/benefit_cap_calculator_data_query.rb
```diff
@@ -0,0 +1,8 @@
...
+    def data
```
chrisroos: I would've been inclined to name this `weekly_benefit_cap` or similar.

test/unit/calculators/benefit_cap_calculator_data_query_test.rb
```diff
@@ -0,0 +1,25 @@
...
+module SmartAnswer::Calculators
+  class BenefitCapCalculatorDataQueryTest < ActiveSupport::TestCase
+    context BenefitCapCalculatorDataQuery do
+      setup do
+        @query = BenefitCapCalculatorDataQuery.new
+      end
+
+      context "benefit_cap_data" do
+        context "family types" do
+          should "contain value for single person" do
```
chrisroos: Assuming the rename of `#data` to `#weekly_benefit_cap` (mentioned above) then these tests could be renamed from `"family types" should "contain value for single person"` to `"#weekly_benefit_cap" should "be £350 for a single person"`.

lib/data/benefit_cap_data.yml
```diff
@@ -0,0 +1,4 @@
+---
```
chrisroos: I would've preferred to see this commit combined with the next one ("Implement getting benefit cap data from YAML in the flow") so that it's clear to see where these weekly caps come from.

lib/data/benefit_cap_data.yml
```diff
@@ -1,4 +1,21 @@
 ---
-single: 350
-couple: 500
-parent: 500
+rates:
```
chrisroos: As suggested in the original commit, I'd name this `weekly_benefit_cap` or similar.

lib/data/benefit_cap_data.yml
```diff
@@ -1,4 +1,21 @@
...
+  single: 350
+  couple: 500
+  parent: 500
+benefits:
```
chrisroos: I don't think it makes sense to add these benefits in this commit as they're not being used for anything.

test/unit/calculators/benefit_cap_calculator_data_query_test.rb
```diff
@@ -4,8 +4,8 @@ module SmartAnswer::Calculators
   class BenefitCapCalculatorDataQueryTest < ActiveSupport::TestCase
     context BenefitCapCalculatorDataQuery do
       setup do
-        query = BenefitCapCalculatorDataQuery.new
-        @rates = query.data.fetch("rates")
+        @query = BenefitCapCalculatorDataQuery.new
+        @rates = @query.data.fetch("rates")
```
chrisroos: This commit is essentially empty. Are there other changes that should be included?

lib/smart_answer/calculators/benefit_cap_calculator_data_query.rb
ikennaokpala: @chrisroos this has already been taken off.

lib/smart_answer_flows/benefit-cap-calculator.rb
```diff
@@ -6,8 +6,9 @@ def define
       status :published
       satisfies_need "100696"
 
-      benefit_cap_query = SmartAnswer::Calculators::BenefitCapCalculatorDataQuery.new
-      benefit_cap_rates = benefit_cap_query.data.fetch('rates')
+      query = Calculators::BenefitCapCalculatorDataQuery.new
+      rates = query.data.fetch('rates')
+      benefits = query.data.fetch('benefits')
```
chrisroos: I think the benefits added to benefit_cap_data.yml in "Update data structure for benefit cap rates and add list of benefits" should've been added in this commit (i.e. at the point they're extracted from the flow).

lib/smart_answer_flows/benefit-cap-calculator.rb
```diff
@@ -6,8 +6,9 @@ def define
       status :published
       satisfies_need "100696"
 
-      benefit_cap_query = SmartAnswer::Calculators::BenefitCapCalculatorDataQuery.new
-      benefit_cap_rates = benefit_cap_query.data.fetch('rates')
+      query = Calculators::BenefitCapCalculatorDataQuery.new
+      rates = query.data.fetch('rates')
```
chrisroos: As mentioned previously, I'd name this `weekly_benefit_cap`.

lib/data/benefit_cap_data.yml
```diff
@@ -4,18 +4,18 @@ rates:
...
+  guardian: :guardian_amount?
+  incapacity: :incapacity_amount?
+  income_support: :income_support_amount?
+  jsa: :jsa_amount?
+  maternity: :maternity_amount?
+  sda: :sda_amount?
+  widowed_mother: :widowed_mother_amount?
+  widowed_parent: :widowed_parent_amount?
+  widow_pension: :widow_pension_amount?
+  widows_aged: :widows_aged_amount?
```
chrisroos: I think the addition of these question names should be combined with the next commit: "Use the YAML to fetch relevant next question to be called." (d0e240c42d95e92fd4a0a224d23602721fd7b5b8) to make it clearer where they've come from, and why they've been extracted.

lib/smart_answer_flows/benefit-cap-calculator.rb
```diff
@@ -8,7 +8,7 @@ def define
 
       query = Calculators::BenefitCapCalculatorDataQuery.new
       rates = query.data.fetch('rates')
-      benefits = query.data.fetch('benefits')
+      benefits = query.data.fetch('benefits').with_indifferent_access
```
chrisroos: Is the call to `#with_indifferent_access` important here? It'd be good to mention it in the commit note if it is.

lib/smart_answer_flows/benefit-cap-calculator.rb
```diff
@@ -76,29 +76,7 @@ def define
...
-            when :sda then question :sda_amount?
-            when :widowed_mother then question :widowed_mother_amount?
-            when :widowed_parent then question :widowed_parent_amount?
-            when :widow_pension then question :widow_pension_amount?
-            when :widows_aged then question :widows_aged_amount?
-            else
-              if housing_benefit == 'yes'
-                question :housing_benefit_amount?
-              else
-                question :single_couple_lone_parent?
```
chrisroos: I would've preferred to see this change (i.e. "Remove housing benefit conditional in Q4 and Q5") in a separate commit.

lib/smart_answer_flows/benefit-cap-calculator.rb
```diff
@@ -76,188 +76,22 @@ def define
           if response == "none"
             outcome :outcome_not_affected
           else
-           question benefits.fetch(benefit_types.shift, :housing_benefit_amount?)
+            question benefits.fetch(benefit_types.shift, :housing_benefit_amount?)
```
chrisroos: I prefer to fix these sort of minor problems in the commit that introduced them, i.e. amend the original commit so that the problem was never there in the first place.

lib/smart_answer_flows/benefit-cap-calculator.rb
```diff
@@ -76,188 +76,22 @@ def define
...
           end
         end
       end
 
       #Q5a
-      money_question :bereavement_amount? do
+      benefits.each do |(benefit,method)|
```
chrisroos: It's possible Rubocop will complain about the unused `benefit` parameter and the lack of space between `benefit` and `method`.

chrisroos: I think I'd prefer `question_name` over `method` here.

lib/smart_answer_flows/benefit-cap-calculator.rb
```diff
@@ -56,7 +56,7 @@ def define
 
       #Q4
       checkbox_question :receiving_non_exemption_benefits? do
-        benefits.each do |benefit|
+        benefits.keys.each do |benefit|
```
chrisroos: Can we guarantee that these `benefits.keys` will always come back in the same order? Am I right in thinking that if the order changes then the order of the benefit options will change for the user?

lib/smart_answer_flows/benefit-cap-calculator.rb
```diff
@@ -175,5 +175,8 @@ def define
       ## Outcome 5
       outcome :outcome_not_affected
     end
+    def self.next_benefit_amount_question(benefits, selected_benefits)
+      benefits.fetch(selected_benefits.shift, :housing_benefit_amount?)
+    end
```
chrisroos: I find it confusing that `.next_benefit_amount_question` modifies the `benefit_types` array and returns the name of the next question. I actually wonder whether it might be better to live with the duplication of `benefits.fetch(benefit_types.shift...` as at least that makes it clearer that `benefit_types` is being modified.

test/unit/smart_answer_flows/benefit_cap_calculator_flow_test.rb
```diff
@@ -0,0 +1,27 @@
...
+          first_benefit: :first_benefit_amount?,
+          second_benefit: :second_benefit_amount?,
+          third_benefit: :third_benefit_amount?,
+          fourth_benefit: :fourth_benefit_amount?
+        }
+
+        @selected_benefits = [:second_benefit, :fourth_benefit]
+      end
+      context "next question" do
+        should "go the the next selected benefit question" do
```
chrisroos: If we stick with the implementation of `.next_benefit_amount_question` then I think it'd be good to add an assertion to this test to demonstrate that `@selected_benefits` is also modified.

lib/data/benefit_cap_data.yml
```diff
@@ -4,18 +4,48 @@ rates:
...
-  jsa: :jsa_amount?
-  maternity: :maternity_amount?
-  sda: :sda_amount?
-  widowed_mother: :widowed_mother_amount?
-  widowed_parent: :widowed_parent_amount?
-  widow_pension: :widow_pension_amount?
-  widows_aged: :widows_aged_amount?
+  bereavement:
+    question: :bereavement_amount?
+    description: "Bereavement Allowance"
```
chrisroos: I wonder whether `question_name` and `question_text` might be better/clearer key names?

lib/smart_answer/calculators/benefit_cap_calculator_data_query.rb
```diff
@@ -1,8 +1,28 @@
 module SmartAnswer::Calculators
   class BenefitCapCalculatorDataQuery
-    attr_reader :data
+    attr_reader :data, :benefits, :questions
```
chrisroos: This is unnecessary as you've also defined `benefits` and `questions` methods below.

lib/smart_answer/calculators/benefit_cap_calculator_data_query.rb
```diff
@@ -1,8 +1,28 @@
...
     def data
       @data ||= YAML.load_file(Rails.root.join('lib', 'data', 'benefit_cap_data.yml'))
     end
+
+    def benefits
```
chrisroos: I'd prefer to see the addition of this method combined with the next commit ("Update the benefit cap flow to use the data convenience methods" - 54b51dfbd0d560cf0be68a265733fee12c57d758) to make it clearer how it's being used.

lib/smart_answer/calculators/benefit_cap_calculator_data_query.rb
```diff
@@ -1,8 +1,28 @@
...
+      @benefits ||=  @data.fetch("benefits").with_indifferent_access
+    end
+
+    def questions
```
chrisroos: I'd prefer to see the addition of this method combined with the next commit ("Update the benefit cap flow to use the data convenience methods" - 54b51dfbd0d560cf0be68a265733fee12c57d758) to make it clearer how it's being used.

lib/smart_answer/calculators/benefit_cap_calculator_data_query.rb
```diff
@@ -1,8 +1,28 @@
...
+      @questions ||=
+        benefits.inject(HashWithIndifferentAccess.new) do |benefits_and_questions, (key, value)|
+          benefits_and_questions[key] = value.fetch("question")
+          benefits_and_questions
+        end
+    end
+
+    def descriptions
```
chrisroos: It looks as though this method isn't used until commit bf2f88fbdf995fcfc261de62f73ff9ef179a3634 ("Populate checkbox list of non-exempt benefits from YAML"). Ideally it'd be combined with that commit.

lib/smart_answer/calculators/benefit_cap_calculator_data_query.rb
```diff
@@ -1,8 +1,28 @@
 module SmartAnswer::Calculators
   class BenefitCapCalculatorDataQuery
-    attr_reader :data
+    attr_reader :data, :benefits, :questions
     def data
       @data ||= YAML.load_file(Rails.root.join('lib', 'data', 'benefit_cap_data.yml'))
     end
+
+    def benefits
+      @benefits ||=  @data.fetch("benefits").with_indifferent_access
```
chrisroos: Does this need to be memoized?

lib/smart_answer/calculators/benefit_cap_calculator_data_query.rb
```diff
@@ -1,8 +1,28 @@
...
+    end
+
+    def questions
+      @questions ||=
```
chrisroos: Does this need to be memoized?

lib/smart_answer/calculators/benefit_cap_calculator_data_query.rb
```diff
@@ -1,8 +1,28 @@
...
+        benefits.inject(HashWithIndifferentAccess.new) do |benefits_and_questions, (key, value)|
+          benefits_and_questions[key] = value.fetch("question")
+          benefits_and_questions
+        end
+    end
+
+    def descriptions
+      @descriptions ||=
```
chrisroos: Does this need to be memoized?

lib/smart_answer/calculators/benefit_cap_calculator_data_query.rb
chrisroos: Is the call to `#with_indifferent_access` important? It be good to test it if so. Actually, I'm not sure there's a test for the `#benefits` method at all, is there?

lib/smart_answer/calculators/benefit_cap_calculator_data_query.rb
```diff
@@ -1,12 +1,15 @@
 module SmartAnswer::Calculators
   class BenefitCapCalculatorDataQuery
-    attr_reader :data, :benefits, :questions
     def data
       @data ||= YAML.load_file(Rails.root.join('lib', 'data', 'benefit_cap_data.yml'))
     end
 
+    def rates
```
chrisroos: As I've mentioned elsewhere, I think this would be better named something like `weekly_benefit_cap`.

lib/smart_answer/calculators/benefit_cap_calculator_data_query.rb
```diff
@@ -1,12 +1,15 @@
...
+      @rates ||= data.fetch("rates")
+    end
+
     def benefits
-      @benefits ||=  @data.fetch("benefits").with_indifferent_access
```
chrisroos: Minor: I prefer to rewrite history to avoid these small problems appearing in the first place.

lib/smart_answer_flows/benefit-cap-calculator.rb
```diff
@@ -109,9 +109,9 @@ def define
 
       #Q6
       multiple_choice :single_couple_lone_parent? do
-        option :single
-        option :couple
-        option :parent
+        query.rates.keys.each do |rate|
+          option rate
+        end
```
chrisroos: I'm not completely sure about this change. It feels like we might be overloading the data in `rates` by using it to return the weekly benefit caps and to list the options to the `single_couple_lone_parent?`. How would it look if the question text for each of these options was also moved to the YAML file (as you've done with the benefits in 3ac69aaed833650a6b120de85d013fbd2a84b614)?

lib/smart_answer/calculators/benefit_cap_calculator_data_query.rb
```diff
@@ -5,7 +5,7 @@ def data
     end
 
     def rates
-      @rates ||= data.fetch("rates")
+      @rates ||= data.fetch("rates").with_indifferent_access
```
chrisroos: Is `#with_indifferent_access` important here? It should be tested if it is.

lib/smart_answer_flows/benefit-cap-calculator/questions/receiving_non_exemption_benefits.govspeak.erb
```diff
@@ -2,24 +2,7 @@
...
-  "income_support": "Income Support",
-  "jsa": "Jobseeker’s Allowance",
-  "maternity": "Maternity Allowance",
-  "sda": "Severe Disablement Allowance",
-  "widowed_mother": "Widowed Mother's Allowance",
-  "widowed_parent": "Widowed Parent's Allowance",
-  "widow_pension": "Widow's Pension",
-  "widows_aged": "Widow's Pension (age related)",
-  "none_above": "None of the above"
-) %>
```
chrisroos: I'd prefer to see this merged with commit 3ac69aaed833650a6b120de85d013fbd2a84b614 ("Refactor Benefit cap calculator (Making benefit rates and type configurable)") as mentioned in a comment on that commit.

lib/smart_answer_flows/benefit-cap-calculator.rb
```diff
@@ -43,6 +43,13 @@ def define
         option :yes
         option :no
 
+        precalculate :exempt_benefits do
+          exempt_benefits = query.data.fetch("exempt_benefits").map do |exempt_benefit|
+            "- #{exempt_benefit}"
+          end
+          exempt_benefits.join("\n")
+        end
```
chrisroos: This is a view concern and so should really be in the receiving_exemption_benefits.govspeak.erb template.

chrisroos:
I've finished going through this pull request now.

I've added quite a few comments to individual commits. Unfortunately, this PR interface is now showing quite a few of them as being on "outdated diffs". I don't understand why that is.

I think most of my comments are about the order/content of commits. I think this is somewhat subjective so you should feel free to ignore them as you see fit.

Some of my comments are about what I believe to be missing tests/unexplained behaviour. I think it'd be good to add these.

## Other questions/comments

* I'm fairly certain this branch breaks the visualisation of the benefit-cap-calculator flow. This may or may not be important but I think it should be acknowledged in the PR description either way. There's more discussion about the visualisation in @floehopper's PR #2370. Actually, it'd probably be good to mention that PR in the description of this one as the work in this branch undoes some of the changes in that other branch.

* Are you confident that this change will make it easier to implement the changes required by https://trello.com/c/IVrsJIfM? I'm not 100% convinced by the approach in this PR (not that I have any alternative suggestions) so it'd be great if we're really confident that it's going to make future work easier.

chrisroos:
One final thought. _If_ we're not sure about this making future changes easier, then an alternative approach might be to implement the new changes required by https://trello.com/c/IVrsJIfM and then to refactor out the duplication after. That might reveal some alternative approaches that aren't apparent in isolation.

chrisroos:
> Unfortunately, this PR interface is now showing quite a few of them as being on "outdated diffs". I don't understand why that is.

I think I understand this now. They're appearing as outdated because the lines I commented on were updated in a later commit. Presumably because the GitHub way is to review the combined diff (i.e. the result of all the commits) rather than the individual commits. Apologies if this has made is hard to respond to my feedback. Please let me know if you have any questions about any of the points I've mentioned.

chrisroos:
I wonder whether we could simplify this flow by asking a single question for the combined amount of benefits someone receives, rather than asking a separate question for the amount of each benefit they receive.

This might well lead to a terrible user experience (i.e. if obtaining the combined amount requires more work from the user) but it might be worth asking the question of someone that knows this Smart Answer. It would certainly make the code much easier to understand, and would make the future changes easier to implement.

leenagupte:
@chrisroos I think it would be better to add a card to the backlog to fix visualise to handle flows with repeating questions rather than adding a test that could delay the progress of the benefit-cap-calculator update that needs to be live on April 22nd and that restrains our ability to write code in succinct and configurable way.

floehopper:
@pmanrubia, @leenagupte, @ikennaokpala:

I thought I'd just try to write up some thoughts on this - mainly to draw attention to the bigger picture:

Currently there aren't many/any restrictions on the Ruby we can write within the flow DSL and this gives us a huge amount of flexibility/power. This power is obviously very useful in cases like this one where we feel the need to reduce duplication. However, we need to be careful how we use this power.

Our longer term goal is to make Smart Answers "editable" (in some sense) by the Content Team. If we have flows that use the DSL in lots of different ways, that's going to make the longer term goal harder to achieve. To this end many of the changes @chrisroos and I have been making over the last year have been to *reduce* the number of ways in which the DSL is used.

Thus I was happy to introduce some duplication in `benefit-cap-calculator` in #2370, because it removed another "special" way the DSL was being used. Although I now realise I didn't explain that very well in the PR description.

The fact that the visualisation is broken by this PR is less important to me than the fact that it introduces a new "special" way to use the DSL (effectively reverting #2370).

I hope that makes some kind of sense. However, don't worry if it doesn't - I realise that this is the kind of thing that is better explained in person. I'm planning to come into Aviation House in the not too distant future, so hopefully we can talk it through properly then.

Ideally I'd prefer if this PR was not merged and the changes to `benefit-cap-calculator` were made to the version in `master`. However, I realise there may be deadline pressures, etc which mean that it's more expedient to merge it.

chrisroos:
Thanks for your comments, @floehopper. I think the background you've added is really useful. I'm sorry that I wasn't able to articulate it when I was reviewing this PR.

pmanrubia:
Hi @floehopper,

We definitely need to spend some time understanding the bigger picture, as we all need to work aligned wit it; I am really looking forward to it, and of course meeting you and @chrisroos in person.

As far as I understand, this is what happened here: 
1. We were tasked with an update to the flow
2. This update would need to duplicated 7 questions (with their big case statement) making it quite difficult to follow and maintain, as we need to live with both flows for a few weeks / months.
3. As a prior step to introduce this second set of questions, and following good software design principles, @ikenna and @leenagupte extracted some logic out, so adding the second path is easier (they have been tackling the long list of comments from @chrisroos, as far as I know)

So, the process they followed makes sense to me. I guess @ikenna and @leenagupte didn’t realise you have updated it recently in #2370.

Having said this, we are sorry if this is not following the bigger picture principles, because I am an advocate of consistency within a project, so we are open to suggestions here. 

Our main requirement is the deadline of April 22nd for having the second flow working within the cap-calculator.

I hope this makes sense for you.

ikenna:
@pmanrubia: I believe you mean @ikennaokpala rather than me. (Secretly I like gloating over the fact that I beat @ikennaokpala to getting the @ikenna github handle :sunglasses:  ).

ikennaokpala:
@ikenna I know.. I know..

lib/smart_answer/calculators/benefit_cap_calculator_data_query.rb
leenagupte: I agree, this has been implemented in a later commit.

test/unit/calculators/benefit_cap_calculator_data_query_test.rb
leenagupte: I agree, this test has been updated in a later commit.

lib/smart_answer_flows/benefit-cap-calculator.rb
leenagupte: benefits.fetch(selected_benefits.shift, :housing_benefit_amount?) was moved to a function to make the benefit cap flow readable.

test/unit/smart_answer_flows/benefit_cap_calculator_flow_test.rb
leenagupte: I agree, the tests have been updated in a later commit.

lib/smart_answer_flows/benefit-cap-calculator.rb
leenagupte: This was missed. The description text for single, couple, parent has been moved from the govspeak template and into the YAML file in a later commit.

leenagupte:
@chrisroos @floehopper @ikennaokpala @pmanrubia 

I've expanded the PR description to give a more details about what motivation behind this refactoring was and the options we considered.

If you have any questions, or can suggest another option or improvement that we haven't considered, please let us know.

lib/data/benefit_cap_data.yml
ikennaokpala: @chrisroos These were added as part of the modeling phase and later used.. [here](https://github.com/alphagov/smart-answers/blob/benefit-cap-calculator-refactoring/test/unit/calculators/benefit_cap_calculator_configuration_test.rb#L11)

floehopper:
@pmanrubia: Thanks for explaining more about the evolution of the changes in this PR.

@leenagupte: Thanks for expanding the PR description to explain what other options that were considered. That's really useful.

Now that I better understand the changes you are trying to make (especially the need to have two very similar flows operating in parallel for a period of time), I can see why you decided to reduce the duplication in the benefit type questions in the way you did.

I can't immediately see any significantly better ways to do this, so I suggest you go ahead and merge the changes in this PR in order to meet the deadline. In the meantime, I've added a [Trello card][1] to remind us to investigate whether we can address the bigger picture concerns.

[1]: https://trello.com/c/c7lOkqgw

PR #2455, opened 2016-04-11
Add a developer checklist to the docs

leenagupte:
This checklist is something that new developers on the team can refer when
they are first becoming familiar with the code base and structure of the tests.

### DIFF
--- a/doc/regression-tests.md
+++ b/doc/regression-tests.md
@@ -68,6 +68,25 @@ The regression test for a given flow is triggered automatically when you run the
 
 If you've added new classes, modules or data which is used by a flow, you should add the relevant files to the checksums file.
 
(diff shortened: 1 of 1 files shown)


pmanrubia:
I would [update this document](https://github.com/alphagov/smart-answers/blob/master/doc/regression-tests.md) with your checklist. It feels to me like a better place as the checklist is focused in regression tests. 
When it comes to the title, could you think on something more concrete than `checklist`? Something like: `Steps to update the tests after a change`?

leenagupte:
@pmanrubia Thanks for the suggestion, I'll do that.

leenagupte:
@pmanrubia I've updated the regression tests doc as suggested.

pmanrubia:
LGTM 👍

PR #2457, opened 2016-04-13
Add test to ensure all flows are visualisable

chrisroos:
We've gone to quite some effort to ensure that the graph visualisation
continues to work while we've been refactoring Smart Answers. I'm adding this
test to ensure that we don't undo that work in future.

This was highlighted by a proposed refactoring of benefit-cap-calculator (see
PR #2452) that results in the flow not being visualisable.

Assuming this commit, and the proposed benefit-cap-calculator refactoring make
it into master, we'll have to work out how to handle the broken visualisation.

### DIFF
--- a/test/integration/flow_visualisation_test.rb
+++ b/test/integration/flow_visualisation_test.rb
@@ -1,0 +1,12 @@
+require_relative '../integration_test_helper'
+
+class FlowVisualisationTest < ActionDispatch::IntegrationTest
+  SmartAnswer::FlowRegistry.instance.flows.each do |flow|
+    should "be able to visualise #{flow.name}" do
+      presenter = GraphPresenter.new(flow)
+      assertion_message = "The #{flow.name} Smart Answers isn't visualisable"
+
+      assert presenter.visualisable?, assertion_message
+    end
+  end
+end


leenagupte:
I think it would be better to add a card to the backlog to fix visualise to handle flows with repeating questions rather than adding a test that could delay the progress of the benefit-cap-calculator update that needs to be live on April 22nd and that restrains our ability to write code in succinct and configurable way.

pmanrubia:
Hi @chrisroos, thank you for work on this PR. I think it is a good idea to detect automatically when the graph visualization has been broken.

But, as a note, I would wait before merging this PR, and this is the reason:
We are discussing with the content team about which flows visualisation are useful for us. Some of them, like the Benefit cap calculator, is not being very helpful because it's too complex.

So what about adding a list of exceptions to the test? What are your thoughts?

chrisroos:
> I think it would be better to add a card to the backlog to fix visualise to handle flows with repeating questions rather than adding a test that could delay the progress of the benefit-cap-calculator update that needs to be live on April 22nd and that restrains our ability to write code in succinct and configurable way.

I don't think this test prevents the changes in PR #2452 being merged. We can easily update this test to add an explicit exclusion for benefit-cap-calculator so that flow is skipped. The benefit of having the test for all other flows is that it'll alert us if we accidentally break the visualisation of other flows in future.

chrisroos:
> We are discussing with the content team about which flows visualisation are useful for us. Some of them, like the Benefit cap calculator, is not being very helpful because it's too complex.

This discussion sounds good. I'm sure there are other flows where the visualisation is too complicated for it to be useful.

> So what about adding a list of exceptions to the test? What are your thoughts?

Adding exceptions to the test sounds good to me. Although I'd only add them at the point the visualisation is broken (i.e. when this test starts failing). That way we'll be prompted to question whether or not the visualisation was useful on a flow by flow basis.

chrisroos:
My vote would be to get this test merged and then to add exceptions if/when they're required.

pmanrubia:
@chrisroos agree with the exception on the `flow by flow` basis. 
LGTM 👍

chrisroos:
Thanks, @pmanrubia!

PR #2469, opened 2016-04-18
Fix auto-correctable Rubocop violations

floehopper:
## Description

Trello card: https://trello.com/c/BqZ8i1b6

This PR fixes all the violations of cops which are supposedly auto-correctable. There are still quite a few disabled cops, but hopefully this is a good step forward.

All the commits up to and including the one fixing the `Style/AndOr` violations were generated using only Rubocop's auto-correct functionality by running `govuk-lint-ruby --auto-correct`.

The next batch of commits up to and including the one fixing the `Lint/UnusedBlockArgument` violations were done manually. In the middle of this batch, the commit fixing the `Style/TrivialAccessors` violation was done manually, because the auto-correct option didn't seem to do anything.

The commit fixing the `Lint/UnusedMethodArgument` violations was generated using only Rubocop's auto-correct functionality, but then the subsequent 7 commits attempt to make improvements to those auto-corrections.

Finally I've updated all the affected regression checksums in a single commit - no regression test artefacts needed to be changed to make the regression tests pass.

## External changes


floehopper:
It'd be great to get this reviewed as soon as possible, because it's quite vulnerable to updates in `master`.

chrisroos:
A number of commit messages contain, "I chose to fix these violations manually, because I wasn't happy with the changes generated by Rubocop's auto-correct mechanism." Do you know whether it's possible to update Rubocop so that it would correct these violations in the way you prefer?

chrisroos:
It looks as though the subject line of f4616b8374fa5c50039559513928fabb3657f955 ("Fix Style/BracesAroundHashParameters violations") should be "Fix Style/ClosingParenthesisIndentation violations".

lib/smart_answer_flows/calculate-agricultural-holiday-entitlement.rb
```diff
@@ -35,11 +35,9 @@ def define
...
         option "1-day"
 
-        calculate :days_worked_per_week do |response|
-          # XXX: this is a bit nasty and takes advantage of the fact that
-          # to_i only looks for the very first integer
-          response.to_i
-        end
+        # XXX: this is a bit nasty and takes advantage of the fact that
+        # to_i only looks for the very first integer in the string response
+        calculate :days_worked_per_week, &:to_i
```
chrisroos: This looks odd to me but maybe it's not such a big problem as hopefully these `calculate` blocks won't be around for _too_ much longer.

lib/smart_answer_flows/uk-benefits-abroad.rb
```diff
@@ -41,7 +41,7 @@ def define
           going_or_already_abroad == 'already_abroad'
         end
 
-        calculate :already_abroad_text_two do |response|
+        calculate :already_abroad_text_two do |_response|
```
chrisroos: Why have you chosen to keep `_response` here?

chrisroos:
It'd be good to update the commit message that appears to be wrong but aside from that, this all looks good to me.

floehopper:
> A number of commit messages contain, "I chose to fix these violations manually, because I wasn't happy with the changes generated by Rubocop's auto-correct mechanism." Do you know whether it's possible to update Rubocop so that it would correct these violations in the way you prefer?

I don't believe so - at least not without making changes to the Rubocop code. There are configuration options for some cops and changing them _might_ make a difference to how the auto-correct works, but then we would've diverged from the GDS style-guide which I wanted to avoid.

There are usually multiple ways to change some code so it satisfies a cop. The auto-correct mechanism is inevitably somewhat naive, because it has to work in all situations; whereas a human developer can look at each change in context and make improvements, sometimes by making other changes not necessarily directly related to complying with the cop rules.

In hindsight, for each of these cops I should've blindly applied the auto-correct in one commit and then made my adjustments in another like I did for `Lint/UnusedMethodArgument` - then you could've seen what changes I made. However, since there were only a small number of changes in each case, I didn't think it was worth it.

floehopper:
Hopefully the commits related to the `Lint/UnusedMethodArgument` cop demonstrate how the auto-correct mechanism sometimes generates sub-optimal code e.g. the `_response` argument to the `:already_abroad_text_two` calculate block which you commented on. In this example the auto-correct mechanism adds an underscore to all unused arguments. However, we can see that it's possible to remove the argument altogether.

floehopper:
I'm going to create a new branch based on this on in order to rebase against `master`, resolve the conflicts highlighted by GitHub above, and address @chrisroos' feedback.

lib/smart_answer_flows/calculate-agricultural-holiday-entitlement.rb
floehopper: I agree this looks a bit odd. I added "in the string response" to try to make things clearer, but it sounds as if that wasn't enough. To address your concern I'm going to leave the code unchanged and disable the cop for these lines. Hopefully that will highlight how undesirable this code is and it will encourage us to remove the calculate block as you suggest.

Addressed in 4c23b45e9db97580079e5980b9e6d9f2200ce41f.

lib/smart_answer_flows/uk-benefits-abroad.rb
floehopper: I did mean to remove it, but forgot. Well spotted!

Addressed in 28d34a9494f26218033a9de5903c3585942ca98b.

floehopper:
> It looks as though the subject line of f4616b8374fa5c50039559513928fabb3657f955 ("Fix Style/BracesAroundHashParameters violations") should be "Fix Style/ClosingParenthesisIndentation violations".

Addressed in ab3f6b11b55a85bb008abbcf31e6f64e15bedd8c.

floehopper:
Superseded by #2474. Closing.

PR #2458, opened 2016-04-14
Update worldwide location organisation data

chrisroos:
I've occasionally found it useful to generate regression test artefacts for all countries in marriage-abroad (when I make a change that affects every outcome, for example). Each time I do that I have to add the worldwide organisation fixture data (stored in test/fixtures/worldwide) for each of the new countries. This branch updates the existing organisation fixture data and adds new fixtures for all countries stored in worldwide_locations.yml. This will make life easier when generating regression test artefacts for all countries in marriage-abroad.

This builds on PR #2448, which I hope/plan to merge first. Without the changes to worldwide_organisation_test.rb in that PR, updating this organisation data causes some of those tests to fail.

Updating these fixtures has resulted in changes to the regression test artefacts for marriage-abroad, overseas-passports, register-a-birth, register-a-death and report-a-lost-or-stolen-passport. For all but marriage-abroad I've also had to update the checksum data as it contains links to the organisation fixtures that those flows rely on. This seems somewhat inconsistent but I plan to address that separately.

PR #2473, opened 2016-04-19
Remove worldwide fixture data from regression test checksums

chrisroos:
We were inconsistently tracking worldwide fixture data in the regression test checksums. The regression test artefacts for Smart Answers that contain a `country_select` will be affected if the worldwide_locations.yml data changes, and several of those will be affected if the organisation data changes.

Rather than trying to keep the checksum data up to date for each of the Smart Answers that rely on the worldwide fixtures, I think it might be easier to rely on the [documentation in updating-worldwide-fixture-data](https://github.com/alphagov/smart-answers/blob/master/doc/updating-worldwide-fixture-data.md), which explains that you'll need to run all integration and regression tests after updating this data.

This came to light while I was working on PR #2458.

What do people think about this approach? Does it seem OK, or would you prefer that we try to keep the checksum data in sync with the worldwide fixture files required by the regression tests for each Smart Answer?


floehopper:
I don't feel very strongly about this either way. I think your suggested approach is probably fine. If all else fails, the full regression test build scheduled on Jenkins should catch any mistakes.

So I'm happy if this gets merged. 👍

chrisroos:
I'm closing this in preparation for rebasing on master, opening a new PR (to preserve commit comments on this one) and getting it merged.

PR #2475, opened 2016-04-19
Remove unnecessary explicit namespace qualifications

floehopper:
These classes were already within the scope of the relevant namespace.

Removing the explicit namespace qualifications makes the code terser and easier to read.


chrisroos:

floehopper:

PR #2476, opened 2016-04-19
Remove v2-related code and documentation

floehopper:
We haven't used the old versioning workflow for about a year and there are no longer any v2 flows in the app, so it feels safe to remove this code and documentation.


chrisroos:

PR #2477, opened 2016-04-19
Spike: Allow user to choose between 2 versions of benefit-cap-calculator

floehopper:
This is a really hack-y spike to demonstrate what I meant by option 1 in my email earlier today [1].

There's lots not to like about this approach. In particular, the duplication of Q4 and Q6 and their templates. Although this could almost certainly be improved, because I did a very naive copy and paste of these two questions.

Note that I had to duplicate these two questions, because they were the only two places which needed the value of `chosen_cap` at flow-definition time as opposed to at request-time. In my email I hadn't appreciated that Q6 needed it too, which is why I only suggested duplicating Q4.

Note that I've removed the memoisation in `BenefitCapCalculatorConfiguration`, because I suspected it would be problematic once I added a `version` argument to the `data` method. I'm not convinced the memoisation is really necessary anyway.

For demo purposes I moved the existing YAML data under a new `default` key and copied the same data under a `future` key. I then removed a bunch of the benefits and changed some of the other values.

Note that I have not updated any of the tests and I have only tested the solution very briefly - I may well have overlooked something.

I'm still not convinced that this is the best approach, but hopefully this is enough to explain what I meant.

[1]: https://groups.google.com/a/digital.cabinet-office.gov.uk/forum/#!topic/content-team-developers/2cCz7vvMXp8

### DIFF
--- a/lib/data/benefit_cap_data.yml
+++ b/lib/data/benefit_cap_data.yml
@@ -1,66 +1,88 @@
 ---
-weekly_benefit_caps:
-  single:
-    amount: 350
-    description: "Single"
-  couple:
-    amount: 500
-    description: "Living as a couple (with or without children)"
-  parent:
-    amount: 500
-    description: "A lone parent with 1 or more dependent children"
-exempt_benefits:
-  - Attendance Allowance
-  - Disability Living Allowance
-  - Industrial Injuries Benefit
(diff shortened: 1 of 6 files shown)


floehopper:

PR #2478, opened 2016-04-20
Spike: Duplicate benefit-cap-calculator flow for 2016 changes

chrisroos:
This is similar to @floehopper's PR #2477 in that it demonstrates another approach to introducing the benefit-cap-calculator changes required for 2016.

I hesitated to create this as I'm wary that it might add to the confusion/problem. I'm not particularly suggesting we go down this route but I thought it might be useful to demonstrate an alternative approach.

This approach is similar to that used in the maternity-paternity-calculator, where an initial question essentially routes between the adoption-calculator, maternity-calculator or paternity-calculator shared logic.

I've reverted PR #2452 so that I can work on the earlier, very verbose version of benefit-cap-calculator.

I've moved the benefit-cap-calculator logic to a "shared logic" file, made a copy for the post-2016 changes and added an initial question to route between the two shared logic flows.

I've then made some changes to the post-2016 flow logic to show that I can do it without having to worry about the current logic.

Once all the remaining changes necessary for post-2016 had been made (e.g. addition of post code check) I'd then investigate removing the duplication between these flows.


floehopper:
Despite the fact that this introduces a lot of duplication, I really like this approach. Given that the pre-2016 logic will soon be removed, I could happily live with this level of duplication. However, I think it might be possible to remove some of the duplication. I'm going to have a quick try at doing that now.

floehopper:

PR #2480, opened 2016-04-21
Spike: Duplicate benefit cap calculator flow for 2016 changes (with duplication then removed)

floehopper:
This builds on #2478 by @chrisroos, but uses a couple of tactics to remove duplication:

1. Introduces a mechanism to supply a set of local variables to a piece of shared flow logic.
2. Uses symbolic links to avoid duplicating question & outcome template files.

I've used all except the last commit of #2478 i.e. the one which introduces differences between pre-2016 and post-2016. I've then removed the duplication before then making the equivalent changes to remove the "Carers" and "Guardian" options & questions for post-2016.

Note that I've tried to break up the changes to the regression test artefacts into multiple commits in order to make sure all the changes were as I expected. I'm not sure what I've done is perfect, but hopefully it's more useful than a single giant commit.

There's lots in here that could be improved, but hopefully it's enough to demonstrate an alternative approach to the problem.


floehopper:

PR #2481, opened 2016-04-21
Benefit cap calculator pre autumn changes

ikennaokpala:
Paired with @leenagupte

Trello Story: https://trello.com/c/IVrsJIfM/34-22-april-requires-fix-but-hold-deployment-benefit-cap-calculator-2016-2017-changes

## Motivation

## Factcheck
[Benefit Cap Calculator](https://smart-answers-pr-2481.herokuapp.com/benefit-cap-calculator/y/)

## Expected Changes
[URL on GOV.UK](https://gov.uk/benefit-cap-calculator/y/)
* Add a first question to beginning of the flow to ask users which benefit cap they want to view.
* Add Carer's Allowance and Guardians Allowance to the list of exempt benefits in the autumn 2016 flow
* Remove Carer's Allowance and Guardians Allowance from the list of benefits to check in the autumn 2016 flow.

### Before

screen shot 2016-04-21 at 13 34 08

### After

screen shot 2016-05-06 at 16 30 12

## Removing the future flow (Autumn 2016)
Sometime in autumn 2016, the new benefit caps will be implemented. At this point, the future flow will become the default flow and the old default flow will need to be removed.

We have (as much as possible) tried to implement these changes to make removing a flow as easy as possible. As it currently stands, these are the changes that will need to be made in Autumn 2016:

1. Remove the first question.
2. Remove the configuration for "default" from benefit_cap_data.yml
3. Rename "future" to "default in benefit_cap_data.yml
4. Remove all references to chosen_cap in the benefit-cap-calculator flow
5. Questions that have a _future version (i.e. is duplicated):
    - Remove the original, and rename _future to the same name as the original. 
    - Consolidate the templates with the current versions.
    - Questions affected:
       * single_couple_lone_parent_future?
       * receiving_non_exemption_benefits_future?
6. Outcomes that have a _future version (i.e. is duplicated):
    - Consolidate the templates with the current versions. 
    - Outcomes affected:
        * outcome_not_affected_future
        * outcome_not_affected_no_housing_benefit_future
        * outcome_not_affected_exemptions_future
7. Outcomes that have National and London future versions:
    - Outcomes affected:
        * outcome_not_affected_less_than_cap_future_london
        * outcome_not_affected_less_than_cap_future_national
        * outcome_affected_greater_than_cap_future_london
        * outcome_affected_greater_than_cap_future_national
8. Remove the default flow from the integration tests.
9. Rename the default flow to future flow in the integration tests.
10. Remove the first question from the integration tests.
11. Regenerate all test artefacts for the regression and integration tests.
12. Remove the version parameter from the functions in benefit_cap_calculator_configuration. There will only be one version (future renamed to default) of the configuration, so it's no longer needed.
13. Update benefit_cap_calculator_configuration_test

Tests affected by the changes:
* test/regression/smart_answers_regression_test.rb
* test/integration/smart_answer_flows/benefit_cap_calculator_test.rb
* test/unit/smart_answer_flows/benefit_cap_calculator_flow_test.rb 
* test/unit/calculators/benefit_cap_calculator_configuration_test.rb


floehopper:
I'm slightly confused to see both the "Waiting on factcheck" **and** "Spike" labels on this. To me a "Spike" is something that I would expect to throw away - see [this definition][1].

[1]: http://www.extremeprogramming.org/rules/spike.html

lib/smart_answer/calculators/benefit_cap_calculator_configuration.rb
```diff
@@ -37,6 +37,10 @@ def descriptions(version)
       end
     end
 
+    def all_questions
+      dataset.keys.inject({}){|versions, version| versions.merge(questions(version))}
```
floehopper: Minor: I'd prefer some more spaces in here i.e. like this:

```ruby
dataset.keys.inject({}) { |versions, version| versions.merge(questions(version)) }
```

I suspect Rubocop will complain about this anyway.

floehopper: I find the name of the block argument, `versions` confusing. I think something like `questions` or `all_questions` would be less confusing.

test/unit/calculators/benefit_cap_calculator_configuration_test.rb
```diff
@@ -112,6 +120,15 @@ class BenefitCapCalculatorConfigurationTest < ActiveSupport::TestCase
             assert_equal "other_first_question", @config.benefits(:other).fetch("first_benefit")["question"]
           end
         end
+        context "merge questions" do
+          should "get all benefit questions from multiple configuration sets" do
+            questions = @config.all_questions
+            assert_includes questions.values, "other_first_question"
+            assert_includes questions.values, "other_second_question"
+            assert_includes questions.values, "third_question"
+            assert_includes questions.values, "other_third_question"
```
floehopper: I find it strange that we're not checking that `first_question` and `second_question` are included here, but we are checking all the others.

floehopper:
I think it might be better to use better names than `default` and `future`. I think something like @chrisroos's `pre-2016` and `post-2016` might be better. Or `pre-autumn-2016` and `post-autumn-2016` if those aren't specific enough. I just worry that when we get to autumn 2016, `future` might no longer be in the future!

lib/data/benefit_cap_data.yml
```diff
@@ -1,15 +1,16 @@
...
+    national:
+      single:
+        amount: 350
+        description: "Single"
+      couple:
+        amount: 500
+        description: "Living as a couple (with or without children)"
+      parent:
+        amount: 500
+        description: "A lone parent with 1 or more dependent children"
```
floehopper: I find it confusing that only the `default` version of `weekly_benefit_caps` is updated here. I would've expected the `future` version to be updated in this commit, but it seems to have been done in an earlier commit - before the change to `BenefitCapCalculatorConfiguration#weekly_benefit_caps` which uses it.

lib/smart_answer_flows/benefit-cap-calculator.rb
```diff
@@ -75,6 +75,7 @@ def define
 
       #Q4 default flow
       checkbox_question :receiving_non_exemption_benefits? do
+
```
floehopper: Minor: Delete this blank line. I suspect Rubocop will complain about it anyway.

lib/smart_answer/calculators/benefit_cap_calculator_configuration.rb
```diff
@@ -38,9 +38,19 @@ def descriptions(version)
     end
 
     def all_questions
-      dataset.keys.inject({}){|versions, version| versions.merge(questions(version))}
+      dataset.keys.inject({}) { |versions, version| versions.merge(questions(version)) }
```
floehopper: This change belongs in the commit where the code was first introduced.

lib/smart_answer_flows/benefit-cap-calculator.rb
```diff
@@ -26,6 +28,7 @@ def define
         save_input_as :housing_benefit
 
         next_node do |response|
+          chosen_cap = chosen_cap
```
floehopper: I'm not sure what this line is trying to do, but I'm pretty sure it's redundant and can be removed.

lib/smart_answer_flows/benefit-cap-calculator.rb
```diff
@@ -54,21 +57,25 @@ def define
         option :no
 
         precalculate :exempt_benefits do
-          config.exempt_benefits
+          config.exempt_benefits(chosen_cap.to_sym)
```
floehopper: It seems odd that in some places we're calling `String#to_sym` on `chosen_cap` and in others we are not. If this mixture of approaches is actually necessary, then the commit note should explain why. If it's not necessary, then I'd prefer all the method calls to pass `chosen_cap` consistently i.e. always pass as a string OR always pass as a symbol.

lib/smart_answer_flows/benefit-cap-calculator.rb
```diff
@@ -28,11 +28,14 @@ def define
         save_input_as :housing_benefit
 
         next_node do |response|
-          chosen_cap = chosen_cap
```
floehopper: It would be better if the git history was re-written so this line was never introduced in the first place.

lib/smart_answer/calculators/benefit_cap_calculator_configuration.rb
```diff
@@ -38,9 +38,19 @@ def descriptions(version)
...
     end
 
+    def is_london?(postcode)
+      area(postcode).any? { |result| result[:slug] == "london" }
```
floehopper: Minor: `#london?` vs `#is_london?` would be more idiomatic Ruby.

test/unit/calculators/benefit_cap_calculator_configuration_test.rb
```diff
@@ -134,6 +135,49 @@ class BenefitCapCalculatorConfigurationTest < ActiveSupport::TestCase
...
+                  }
+                ]
+              )
+            )
+          end
+          should "return true" do
+            assert_equal true, @config.is_london?("WC2B%206SE")
+          end
+        end
+      end
```
floehopper: Minor: I think it would be better to avoid `any_instance` and stub directly on `Services.imminence_api`.

test/unit/calculators/benefit_cap_calculator_configuration_test.rb
```diff
@@ -134,6 +135,49 @@ class BenefitCapCalculatorConfigurationTest < ActiveSupport::TestCase
...
+                    country_name: "England"
+                  }
+                ]
+              )
+            )
+          end
+          should "return false" do
+            assert_equal false, @config.is_london?("B1%201PW")
+          end
+        end
```
floehopper: Minor: Insert blank line between contexts.

test/unit/calculators/benefit_cap_calculator_configuration_test.rb
```diff
@@ -134,6 +135,49 @@ class BenefitCapCalculatorConfigurationTest < ActiveSupport::TestCase
...
+                    country_name: "England"
+                  },
+                  {
+                    slug: "london",
+                    country_name: "England"
+                  }
+                ]
+              )
+            )
+          end
```
floehopper: Minor: Insert blank line between `setup` and `should` blocks.

test/unit/calculators/benefit_cap_calculator_configuration_test.rb
```diff
@@ -134,6 +135,49 @@ class BenefitCapCalculatorConfigurationTest < ActiveSupport::TestCase
...
+          should "return true" do
+            assert_equal true, @config.is_london?("WC2B%206SE")
```
floehopper: Minor: Is it actually important that the method returns `true` or is it good enough for it to return something truth-y? I suspect the latter is true in which case, I'd prefer that we used `assert @config.is_london?("WC2B%206SE")` here. And use `refute @config.is_london?("B1%201PW")` for the related assertion above.

test/data/benefit-cap-calculator-questions-and-responses.yml
```diff
@@ -62,3 +62,4 @@
 - parent
 :enter_postcode?:
 - B1 1PW
+- IG6 2BA
```
floehopper: This belongs in the previous commit.

lib/smart_answer/calculators/benefit_cap_calculator_configuration.rb
```diff
@@ -38,9 +38,19 @@ def descriptions(version)
...
+    end
+
+    def area(postcode)
+      response = Services.imminence_api.areas_for_postcode(postcode)
+      (response&.results || {})
+    end
```
floehopper: Minor: I'd prefer it if we duplicated *exactly*  the code in [LandlordImmigrationCheckCalculator#areas_for_postcode][1] here to make it more obvious that this is ripe for refactoring. Ultimately I think we should extract some of this code into some kind of service adapter class which would wrap calls to `GdsApi::Imminence` in a friendlier interface.

[1]: https://github.com/alphagov/smart-answers/blob/0f9d51b3022f10a071b07a393a9802e701293d21/lib/smart_answer/calculators/landlord_immigration_check_calculator.rb#L21-L24

lib/smart_answer_flows/benefit-cap-calculator.rb
```diff
@@ -200,12 +201,23 @@ def define
...
+        next_node do
+          question :enter_postcode?
+        end
+      end
+
+      #Q7 Future flow - Enter a postcode
+      postcode_question :enter_postcode? do
         calculate :benefit_cap do |response|
-          sprintf("%.2f", config.weekly_benefit_cap_amount(chosen_cap, response))
+          region = config.is_london?(response) ? :london : :national
```
floehopper: Minor: It seems odd to have a method which returns a truth-y or false-y value which we then immediately convert into symbols, especially as we're doing that conversion in two places. I think it'd be better if we had a method called something like `#region` on `config` which returned the appropriate symbol.

lib/smart_answer_flows/benefit-cap-calculator.rb
```diff
@@ -92,21 +99,52 @@ def define
...
+          config.descriptions(chosen_cap).merge(none_above: "None of the above")
+        end
+
+        calculate :total_benefits do
+          0
+        end
+
+        calculate :benefit_cap do
+          0
+        end
```
floehopper: I think the `benefit_options`, `total_benefits` & `benefit_cap` blocks are common to both versions of Q4 and since we always arrive at these via Q3, I think these blocks could move there. Having said that I would do that in a separate commit i.e. introduce the duplication first and then remove it in a subsequent commit. Note also that `benefit_options` should become a `calculate` block if you do move them.

lib/smart_answer_flows/benefit-cap-calculator.rb
```diff
@@ -121,26 +159,53 @@ def define
...
         end
       end
 
-      #Q6
+      #Q6 current flow
       multiple_choice :single_couple_lone_parent? do
         precalculate :weekly_benefit_cap_descriptions do
-          config.weekly_benefit_cap_descriptions
+          config.weekly_benefit_cap_descriptions(chosen_cap)
+        end
```
floehopper: I think the `weekly_benefit_cap_descriptions` block is common to both versions of Q6. And since we always arrive at these via Q5p, we could move this block to be a `calculate` block in Q5p to avoid the duplication. As before, I would do this in a separate commit i.e. introduce duplication first and then remove it.

floehopper:
Overall the final solution looks good. However, the sequence of commits is pretty hard to follow, especially as the code and tests are broken in one of the early commits and are not fixed until one of the last commits.

I think it would be better to structure the commits more like the following:

1. Introduce the new initial question, but still send the user to the same next node.
2. Duplicate the `outcome_not_affected_no_housing_benefit` outcome and in Q1 send the user to it if `chosen_cap` is `future`.
3. Make the relevant changes to the new future version of the `outcome_not_affected_no_housing_benefit` outcome template.
4. Duplicate the `outcome_not_affected_exemptions` outcome and in Q2 & Q3 send the user to it if `chosen_cap` is `future`.
5. Make the relevant changes to the new future version of the `outcome_not_affected_exemptions` outcome template.
6. Duplicate the `receiving_non_exemption_benefits?` question and in Q3 send the user to it if `chosen_cap` is `future`.
7. Make the relevant changes to the new future version of the `receiving_non_exemption_benefits?` question template.
8. Introduce support for multiple configuration types
9. Duplicate the `default` configuration to a `future` configuration
10. Change the future version of the `receiving_non_exemption_benefits?` question to use the `future` configuration.

... and so on. I hope that makes sense.

I think it would be relatively straightforward to ensure the flow was working and the tests were passing after each commit.

The regression tests for this flow don't take very long to run, so I think you could even keep them passing after each commit by making the relevant changes to the inputs and artefacts. That way you wouldn't end up with a giant commit at the end in which lots of the artefacts change.

floehopper:
The order of your last 6 commits is a bit odd. First there are two which make changes to the regression tests, then there are two which make changes, then there are two which make more changes to the regression tests. I think it would be better to have the two which change the integration tests first and then the four which change the regression tests.

floehopper:
Lastly I just want to point out that GitHub is indicating that this branch has conflicts with `master` and so you'll need to resolve those before merging.

lib/smart_answer/calculators/benefit_cap_calculator_configuration.rb
leenagupte: @floehopper We found that "results" returned from `imminence_api.areas_for_postcode` does not return a value for "code", however, [LandlordImmigrationCheckCalculator#areas_for_postcode](https://github.com/alphagov/smart-answers/blob/0f9d51b3022f10a071b07a393a9802e701293d21/lib/smart_answer/calculators/landlord_immigration_check_calculator.rb#L21-L24) checks for this value. We found this to be inconsistent and chose to implement it a different way.

lib/smart_answer_flows/benefit-cap-calculator.rb
leenagupte: @floehopper I don't understand the difference between calculate and precalculate blocks. I know that variables defined by precalculate blocks can be passed to the templates. Is this true of calculate blocks?

lib/smart_answer_flows/benefit-cap-calculator.rb
leenagupte: @floehopper See previous note about confusion around calculate blocks.

lib/smart_answer/calculators/benefit_cap_calculator_configuration.rb
floehopper: @leenagupte:

I'm not quite sure I understand what you mean. I just tried the following in the rails console and it looks as if `#areas_for_postcode` returns an instance of `GdsApi::Response` which *does* have a `#code` method, even though this doesn't show up in the string generated by `#inspect`.

```
response = Services.imminence_api.areas_for_postcode('EC1A 1AA')
# => #<GdsApi::Response:0x007fa4d40a75f0 @http_response="{\"_response_info\":{\"status\":\"ok\",\"links\":[]},\"total\":14,\"start_index\":1,\"page_size\":14,\"current_page\":1,\"pages\":1,\"results\":[{\"slug\":\"house-of-commons\",\"name\":\"House of Commons\",\"country_name\":\"-\",\"type\":\"WMP\"},{\"slug\":\"islington-022\",\"name\":\"Islington 022\",\"country_name\":\"England\",\"type\":\"OMG\"},{\"slug\":\"london-assembly\",\"name\":\"London Assembly\",\"country_name\":\"England\",\"type\":\"LAE\"},{\"slug\":\"islington-022f\",\"name\":\"Islington 022F\",\"country_name\":\"England\",\"type\":\"OLF\"},{\"slug\":\"london-assembly\",\"name\":\"London Assembly\",\"country_name\":\"England\",\"type\":\"LAS\"},{\"slug\":\"greater-london-authority\",\"name\":\"Greater London Authority\",\"country_name\":\"England\",\"type\":\"GLA\"},{\"slug\":\"islington-borough-council\",\"name\":\"Islington Borough Council\",\"country_name\":\"England\",\"type\":\"LBO\"},{\"slug\":\"islington-022f\",\"name\":\"Islington 022F\",\"country_name\":\"England\",\"type\":\"OLG\"},{\"slug\":\"clerkenwell\",\"name\":\"Clerkenwell\",\"country_name\":\"England\",\"type\":\"LBW\"},{\"slug\":\"european-parliament\",\"name\":\"European Parliament\",\"country_name\":\"-\",\"type\":\"EUP\"},{\"slug\":\"islington-022\",\"name\":\"Islington 022\",\"country_name\":\"England\",\"type\":\"OMF\"},{\"slug\":\"north-east\",\"name\":\"North East\",\"country_name\":\"England\",\"type\":\"LAC\"},{\"slug\":\"islington-south-and-finsbury\",\"name\":\"Islington South and Finsbury\",\"country_name\":\"England\",\"type\":\"WMC\"},{\"slug\":\"london\",\"name\":\"London\",\"country_name\":\"England\",\"type\":\"EUR\"}]}">
response.code
# => 200
```

However, I'm not very familiar with `GdsApi::Imminence` and it's quite possible that the implementation in `LandlordImmigrationCheckCalculator` might be incorrect as I think you're suggesting.

Either way, I don't understand why the implementation in `BenefitCapCalculatorConfiguration` and `LandlordImmigrationCheckCalculator` should end up being very different from each other. If the implementation in the latter is wrong, then I think the solution is to fix it and bring it into line with that in `LandlordImmigrationCheckCalculator`.

lib/smart_answer_flows/benefit-cap-calculator.rb
floehopper: @leenagupte:

The [Storing data for later use][1] documentation & example flow explain the difference between the various types of calculation block. Essentially, a value stored in a `precalculate` block is available in other blocks within the same question and within the template for that question; whereas a value stored in a `calculate` block is only available in blocks in subsequent questions and their templates.

Since in this case you would be moving the `benefit_options` block into a previous question, there would no longer be any need for it to be a `precalculate` block - it can just be a `calculate` block. And I think the latter would be less confusing.

I know this flow DSL stuff is really confusing - it took @chrisroos & I ages to get our heads round it - hence that documentation page! One of our longer-term goals is to simplify/remove this DSL, but for now we're stuck with it.

[1]: https://github.com/alphagov/smart-answers/blob/master/doc/storing-data.md

floehopper: I've just realised that the [Storing data for later use][1] page doesn't say anything about at which point the template is rendered. I've added a [new Trello card][2] to suggest that we add that.

[1]: https://github.com/alphagov/smart-answers/blob/master/doc/storing-data.md
[2]: https://trello.com/c/WxANAohC

leenagupte: I've moved duplicated things into calculate blocks where there's a real benefit.
Things moved:
* `exempt_benefits`
* `total_over_cap`
* `housing_benefit_amount`
* `benefit_options`
* `total_benefits`
* `benefit_cap`

floehopper: :+1:

floehopper:
My original understanding was that these changes were required urgently by the department and hence why we decided it was expedient to merge the pre-emptive refactoring in #2452 despite it not necessarily being ideal. However, it looks as if this follow-up PR has now been in "fact check" for nearly two weeks. Has anyone heard any news about it?

leenagupte:
@floehopper We needed to have the changes ready for factcheck by April 22nd. We gave the department one week (April 29th) for factcheck. They have come back with a bug (that also exists in the current calculator) plus have hinted at other unnamed changes.

We are going to give them until stand up tomorrow to give us the outstanding changes, but in the mean time will work on the bug. 

As it stands, the changes needed to be deployed to Production on May 9th (Monday).

floehopper:
@ikennaokpala: Thanks for the update.

leenagupte:
I've added a Don't merge label, as the department got back to us this morning an pushed back the deployment date to May 24th. The [trello card](https://trello.com/c/IVrsJIfM/34-24-may-requires-fix-but-hold-deployment-benefit-cap-calculator-2016-2017-changes) has been updated with the details.

pmanrubia:
Thank you @leenagupte.
What about creating a Milestone for the 24th May, and linking the card?

lib/smart_answer_flows/benefit-cap-calculator.rb
```diff
@@ -8,6 +8,18 @@ def define
 
       config = Calculators::BenefitCapCalculatorConfiguration.new
 
+      # Routing question
+      multiple_choice :choose_cap_to_calculate? do
```
chrisroos: The text of this question is "Which period do you want to check for?". Did you consider giving the question a similar name?

lib/smart_answer_flows/benefit-cap-calculator.rb
```diff
@@ -8,6 +8,18 @@ def define
...
+        option :default
+        option :future
+
+        save_input_as :chosen_cap
```
chrisroos: Similar to my comment about the name of this question, I think this might be clearer if it was named `period_to_check`.

lib/smart_answer_flows/benefit-cap-calculator.rb
```diff
@@ -19,7 +31,11 @@ def define
           if response == 'yes'
             question :working_tax_credit?
           else
-            outcome :outcome_not_affected_no_housing_benefit
+            if chosen_cap == 'future'
```
chrisroos: I think this would be much easier to understand for people not familiar with this Smart Answer if it was written as `if period_to_check == 'post-autumn-2016'` (i.e. both `chosen_cap` and the `'future'` and `'current'` options are renamed).

lib/smart_answer/calculators/benefit_cap_calculator_configuration.rb
leenagupte: I disagree. Using questions (or all_questions) would not make it clear that we are only interested in the questions that apply that version (current or future) of the calculator and not every single question that is defined in benefit-cap-calculator.

test/unit/calculators/benefit_cap_calculator_configuration_test.rb
leenagupte: first_question and second_question are (and should be) overwritten by other_first_question and other_second_question as they are duplicates. The test is checking that we get all discreet questions from multiple configuration sets.

I have addressed this point by changing the test description in this commit: d374998165d392a7e1a6610b4d73e121453e34c8

leenagupte:
Regarding commit: https://github.com/alphagov/smart-answers/pull/2481#issuecomment-212973508

Future was picked as a name because it refers to something that hasn't happened yet. When the cap changes occur and the future becomes the present, Having two flows will no longer be valid. The "future" will become the default (current) and the future will no longer exist.

lib/data/benefit_cap_data.yml
leenagupte: I'm sorry, I don't understand.

lib/smart_answer_flows/benefit-cap-calculator.rb
leenagupte: I'll run the lint tests and fix all styling in one commit.

lib/smart_answer_flows/benefit-cap-calculator.rb
leenagupte: I think this has been fixed in a later commit.

lib/smart_answer/calculators/benefit_cap_calculator_configuration.rb
leenagupte: I'll run the lint tests and fix all styling in one commit.

lib/smart_answer_flows/benefit-cap-calculator.rb
leenagupte: This block was changed into calculate block in the later commit and re-written (including removing the to_sym).

lib/smart_answer_flows/benefit-cap-calculator.rb
leenagupte: I would agree if the git history was a bit simpler.

lib/smart_answer/calculators/benefit_cap_calculator_configuration.rb
leenagupte: I agree, though I think the "is" more readable. Changed in commit: fbe1f8029110df2f91a5cbf039b7a94ed1ebb8cc

test/unit/calculators/benefit_cap_calculator_configuration_test.rb
leenagupte: Why?

test/unit/calculators/benefit_cap_calculator_configuration_test.rb
leenagupte: I'll run the lint tests and fix any errors in one commit.

test/unit/calculators/benefit_cap_calculator_configuration_test.rb
leenagupte: I'll run the lint tests and fix any errors in one commit.

test/unit/calculators/benefit_cap_calculator_configuration_test.rb
leenagupte: The function returns a boolean, so testing for true is valid.

lib/smart_answer/calculators/benefit_cap_calculator_configuration.rb
leenagupte: I agree that `LandlordImmigrationCheckCalculator` should be updated to reflect what we have done here. However the change should be assigned to a new story.

lib/smart_answer_flows/benefit-cap-calculator.rb
leenagupte: This has been addressed in a later commit.

lib/smart_answer_flows/benefit-cap-calculator.rb
leenagupte: The name of the method reflects what the question is doing, i.e. that it's asking the user to select a flow even though the user is unaware that is what is happening.

lib/smart_answer_flows/benefit-cap-calculator.rb
leenagupte: The name of the method reflects what the question is doing, i.e. that it's asking the user to select a flow even though the user is unaware that is what is happening.

lib/smart_answer_flows/benefit-cap-calculator.rb
leenagupte: See comment: https://github.com/alphagov/smart-answers/pull/2481#issuecomment-220296815 for why default and future were chosen as names.

leenagupte:
Regarding the ordering of commits. They reflect the our intention of starting at the model layer and finishing at the presentation layer, and then implementing the new questions.

leenagupte:
I have moved to a new branch to resolve the merge conflicts with master to preserve the comment history: https://github.com/alphagov/smart-answers/pull/2542

chrisroos:
> Future was picked as a name because it refers to something that hasn't happened yet. When the cap changes occur and the future becomes the present, Having two flows will no longer be valid. The "future" will become the default (current) and the future will no longer exist.

@leenagupte: I understand the use of `current` and `future` but still think it'd be much clearer if it was something like `pre-autumn-2016` and `post-autumn-2016`. Having terms that match the language used by the department makes it much easier to go from talking and thinking about the Smart Answer to working on the code for the Smart Answer.

PR #2483, opened 2016-04-26
Remove space-stripping ERB tags from state-pension-through-partner

floehopper:
## Description

This replaces all uses of `<%= -%>` with `<%= %>`. We don't need the space/newline stripping behaviour and consistently using the same ERB tags leaves us with a slightly simpler system.

We started using `<%= -%>` when converting from i18n YAML templates to ERB templates. Using `-%>` to strip trailing spaces/newlines allowed us to exactly replicate the outcomes that were previously generated by the i18n YAML system. This gave us confidence that the conversion was accurate. These are no longer necessary and so I've removed them in favour of consistently using `<%= %>`.

I ran all regression tests after making these changes. There were no changes to the artefacts, so I'm happy to accept the changes and update the checksums accordingly.

c.f. #2420

## External changes

There should be no externally-visible changes - this is just an internal refactoring.


chrisroos:

floehopper:
I'm going to rebase this against `master` and (since there are no comments on individual commits) force-push in preparation for merging.

PR #2484, opened 2016-04-26
Refactor TranslatorLinks to handle alternate links

ikennaokpala:
Trello Story: https://trello.com/c/nIdy7kh6/75-register-a-birth-abroad-spain-translators-link

* Change Spanish translator link 

## Factcheck

[Preview link](https://smart-answers-pr-2484.herokuapp.com/register-a-birth/y/spain/mother/yes/same_country) 
[GOV.UK](https://gov.uk/register-a-birth/y/spain/mother/yes/same_country) 

### Expected changes
* Under the "Check your documents" header, the first para has a link to approved translators. The link should change from:
https://www.gov.uk/government/publications/spain-list-of-lawyers to http://www.exteriores.gob.es/Portal/en/ServiciosAlCiudadano/Paginas/Traductoresas---Int%C3%A9rpretes-Juradosas.aspx

### Before
screen shot 2016-04-26 at 14 58 42

### After
 
screen shot 2016-04-26 at 14 59 22

### DIFF
--- a/lib/data/alternative_translators_links.yml
+++ b/lib/data/alternative_translators_links.yml
@@ -1,0 +1,2 @@
+---
+spain: http://www.exteriores.gob.es/Portal/en/ServiciosAlCiudadano/Paginas/Traductoresas---Int%C3%A9rpretes-Juradosas.aspx
--- a/lib/smart_answer/calculators/translator_links.rb
+++ b/lib/smart_answer/calculators/translator_links.rb
@@ -1,13 +1,22 @@
 module SmartAnswer::Calculators
   class TranslatorLinks
-    attr_reader :links
+    attr_reader :links, :alternate_links
 
     def initialize
       @links = self.class.translator_links
+      @alternate_links = self.class.alternate_links
     end
 
     def self.translator_links
       YAML.load_file(Rails.root.join("lib", "data", "translators.yml"))
+    end
+
+    def self.alternate_links
+      YAML.load_file(Rails.root.join("lib", "data", "alternative_translators_links.yml"))
+    end
+
+    def alternate_link?(country)
+      @alternate_links.has_key?(country)
     end
   end
 end
--- a/lib/smart_answer_flows/register-a-birth.rb
+++ b/lib/smart_answer_flows/register-a-birth.rb
@@ -220,7 +220,11 @@ module SmartAnswer
         end
 
         precalculate :translator_link_url do
-          translator_query.links[country_of_birth]
+          if translator_query.alternate_link?(country_of_birth)
+            translator_query.alternate_links[country_of_birth]
+          else
+            translator_query.links[country_of_birth]
+          end
         end
       end
 
--- a/test/artefacts/register-a-birth/spain/father/no/2006-06-30.txt
+++ b/test/artefacts/register-a-birth/spain/father/no/2006-06-30.txt
@@ -1,0 +1,17 @@
+
+
+$!You must register the birth according to the regulations in Spain.$!
+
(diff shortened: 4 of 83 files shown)


gavancurley:
No need for fact check - good to go

floehopper:
Why could this URL not just have replaced the entry for `spain` in `lib/data/translators.yml`? i.e. why did we need to introduce a whole new mechanism for alternative links?

ikennaokpala:
The entry in the ```lib/data/translators.yml``` affect other parts of smart answers eg: register-a-death et al.

According to the [trello card](https://trello.com/c/nIdy7kh6/75-register-a-birth-abroad-spain-translators-link) 

This change relates to the "Register a birth abroad" tool on outcomes for Spain:

Under the "Check your documents" header, the first para has a link to approved translators. The link should change from:
https://www.gov.uk/government/publications/spain-list-of-lawyers
to:
http://www.exteriores.gob.es/Portal/en/ServiciosAlCiudadano/Paginas/Traductoresas---Int%C3%A9rpretes-Juradosas.aspx

So my implementation is meant to mitigate risk and ensure that this change only occurs for the country in question and in the paticular flow. Hence why I have altered the precalculate block as follows:

```
       precalculate :translator_link_url do
          if translator_query.alternate_link?(country_of_birth)
            translator_query.alternate_links[country_of_birth]
          else
            translator_query.links[country_of_birth]
          end
        end
```

floehopper:
Do we know for sure that this change is definitely **not** appropriate for `register-a-death`?

It seems odd to me that a translation service is OK for `register-a-birth`, but not for `register-a-death`. If this really is the case, I would expect to see an explanation in the commit note to say **why** this is so.

Alternatively, did the Content Team consider adding the new translation service to the [spain-list-of-lawyers][1] page?

The reason I'm concerned about this is that it's all too easy to make the code progressively more complicated and therefore less maintainable. I'd rather we didn't do this unless it's definitely necessary.

[1]: https://www.gov.uk/government/publications/spain-list-of-lawyers

gavancurley:
Thanks for pointing this out James. Yes I've checked with FCO and Register a death outcomes for Spain (eg https://www.gov.uk/register-a-death/y/overseas/spain/same_country) should also have the Spain translator link updated. It would be good to get a clearer understanding of how we link to these lists because there is a wider issue here, which is that a lot of these list links in fact take the user (as with Spain currently) to a list of lawyers. Historically the lawyer/translator lists were combined but they've been uncoupled by FCO. I think it's fair to say there's a lot of work to do at FCO though to get the 2 lists updated and available for all countries - it's very patchy at the moment. I'm drawing up a ticket to try to start tackling this issue. As I say, a lot is dependent on FCO getting their lists updated, but we should probably update wording to reset expectations. It's a bit odd clicking a link for translators and finding yourself at a lawyer list. Does this change affect other smart answers or does this mean we can make this change without overly complicating the code?

floehopper:
@gavancurley: Thanks for your response and for checking with FCO. I agree that it would be good to have a better understanding of how we link to these lists and from which contexts. My main concern was to avoid introducing (yet another) exceptional path in the code, so it's good to know that the same change should apply to `register-a-death` and that it sounds as if we can change the link for `spain` to the new URL rather than introducing a whole new mechanism.

I totally understand your point about translators vs lawyers. Once we understand the requirements for that a bit better we can think about how best to support it.

> Does this change affect other smart answers or does this mean we can make this change without overly complicating the code?

I'm not 100% sure what you mean here, but if we can just change the URL for `spain` and have it change  to the same value for both `register-a-birth` & `register-a-death`, I think that should be very straightforward. I don't believe these links are used in any other Smart Answers.

PR #2485, opened 2016-04-26
Fix YearRange#ends_on for year beginning on leap day (29 Feb)

floehopper:
## Description

In this [Trello card][1], @erikse pointed out that a year range beginning on 29th February should end on 28th February the following year and not 27th February as in the previous implementation. Also a year range should always include either 365 or 366 days depending on whether it includes a leap day.

In working on this I realised that there was a related problem with a year beginning on 1st March in the year before a leap year. Intuitively I think that this should end on 29th February in the leap year.

However, for both of the identified problems I can see reasons why you might not want to include the leap day (29th Feb). For now always including the leap day seems like the most unsurprising behaviour, so that what's I've gone with.

It's possible that this behaviour might need to be policy/legislation-specific, so we should bear that in mind when making use of the `YearRange` class elsewhere in the code.

This commit adds more test coverage to highlight the problems as well as fixing the implementation.

Although as far as we know this inconsistency/bug hasn't resulted in any user-facing problems, it's good to get it fixed before we roll out the use of `YearRange` to other flows.

## External changes

The `YearRange` class is only currently used within the `part-year-profit-tax-credits` flow. Although it's possible that the external behaviour of this flow has changed in some very specific circumstances, I couldn't immediately come up with such a scenario.

[1]: https://trello.com/c/M17mqlSN

### DIFF
--- a/test/unit/year_range_test.rb
+++ b/test/unit/year_range_test.rb
@@ -2,17 +2,17 @@ require_relative '../test_helper'
 
 module SmartAnswer
   class YearRangeTest < ActiveSupport::TestCase
-    context 'beginning on 7th June, 2001' do
+    context 'not including the 29th Feb of a leap year' do
       setup do
-        @year_range = YearRange.new(begins_on: Date.parse('2001-06-07'))
+        @year_range = YearRange.new(begins_on: Date.parse('2000-03-01'))
       end
 
       should 'begin on begins_on date' do
-        assert_equal Date.parse('2001-06-07'), @year_range.begins_on
+        assert_equal Date.parse('2000-03-01'), @year_range.begins_on
       end
 
       should 'end a day before a year after the begins_on date' do
-        assert_equal Date.parse('2002-06-06'), @year_range.ends_on
+        assert_equal Date.parse('2001-02-28'), @year_range.ends_on
       end
 
       should 'be 365 days long, because it does not include 29th Feb' do
@@ -20,7 +20,7 @@ module SmartAnswer
       end
     end
 
-    context 'beginning on 1st February, 2000' do
+    context 'including the 29th Feb of a leap year' do
       setup do
         @year_range = YearRange.new(begins_on: Date.parse('2000-02-01'))
       end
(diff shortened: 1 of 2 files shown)


erikse:
@floehopper There are some 29th of Feb triggered exceptions logged in Errbit that may be worth checking, I think they were for another smart answer, but if this resolves the issue, maybe that smart answer can use the YearRange class as well.

test/unit/year_range_test.rb
```diff
@@ -2,16 +2,50 @@
 
 module SmartAnswer
   class YearRangeTest < ActiveSupport::TestCase
-    setup do
-      @year_range = YearRange.new(begins_on: Date.parse('2000-02-01'))
+    context 'beginning on 7th June, 2001' do
```
chrisroos: I wonder whether it might be a bit clearer if the context was "beginning and ending in non leap years", or "beginning in the year following a leap year"?

test/unit/year_range_test.rb
```diff
@@ -2,16 +2,50 @@
...
+      end
+
+      should 'be 365 days long, because it does not include 29th Feb' do
+        assert_equal 365, @year_range.number_of_days
+      end
     end
 
-    should 'begin on begins_on date' do
-      assert_equal Date.parse('2000-02-01'), @year_range.begins_on
+    context 'beginning on 1st February, 2000' do
```
chrisroos: I wonder whether it might be a bit clearer if the context was "beginning on 1st February in a leap year"?

test/unit/year_range_test.rb
```diff
@@ -2,16 +2,50 @@
...
+      setup do
+        @year_range = YearRange.new(begins_on: Date.parse('2000-02-01'))
```
chrisroos: I wonder whether it'd be better to use 2012 (or 2000) for both the tests that deal with leap years?

chrisroos:

I've left a couple of comments about the test contexts that I think might help explain the significance of the dates being used. This is pretty minor so you should feel free to ignore it.

floehopper:
@chrisroos: Thanks for reviewing. I've added [a fixup commit][1] to address your feedback. I've added [another fixup commit][2] to extract `ends_on` into a local variable. In doing all this I realised that there's a similar issue with years _ending_ on a leap day, so I've added a [new commit][3] to address that. I've also updated the PR description accordingly. However, I'm not 100% convinced that this is the correct behaviour, so please could you review this again?

[1]: 9e042d618d50b9aa6f7d0d50c2046b9b3d6d49f8
[2]: 5e2217f6abf695a6d21bbe2229ca4163fed9172e
[3]: 9a779858b3462b9845b2edc47f074229fcc7cff4

chrisroos:
This still looks good to me, @floehopper. I agree that "always including the leap day seems like the most unsurprising behaviour".

floehopper:

floehopper:
Superseded by #2500. Closing.

PR #2487, opened 2016-04-26
Avoid raising InvalidResponse exception in calculators

floehopper:
## Description

Trello card: https://trello.com/c/aPkzlCZb

Avoid raising a `SmartAnswer::InvalidResponse` exception from within any calculators. It's better to have all of these exceptions raised from the same level of abstraction (i.e. within the flow itself) and ideally using a `validate` block. The following calculators for the following flows are affected:

* `marriage-abroad`
* `state-pension-topup`

## External changes


chrisroos:

floehopper:
I'm going to rebase this against `master` and (since there are no comments on individual commits) force-push in preparation for merging.

PR #2491, opened 2016-04-27
Spike: Avoid using `eval` for shared logic

floehopper:
Trello card: https://trello.com/c/II2iSzR2

Originally all the flows were defined like the shared logic (i.e. outside any class or method) and evaluated, but quite some time ago we moved them into `define` methods on sub-classes of `Flow` (e.g. see [this commit][1]). At the time, the main motivation behind this was so that we could use `rcov` to obtain coverage data to help us select suitable responses for the regression tests.

Using Ruby's `Kernel#eval` is generally considered to be a Bad Thing™ and we've had to disable the `Lint/Eval` cop in the implementation of `Flow#use_shared_logic` to satisfy Rubocop. So it feels like a step forward to move the shared logic in the same direction, if only to make things more consistent.

I'm not 100% happy with the approach I've taken in the PR which is why I've labelled it as a "spike", but I haven't been able to come up with anything that I consider better. So I thought I'd open this PR to solicit your feedback and see if you have any better ideas.

[1]: 09214a4378dac5ec408b5c47c36e74d1113b169a


chrisroos:
I haven't really thought this through, but could we somehow use `include` to make use of shared logic? So the class would end up something like:

```
class CalculateYourRedundancyPayFlow < Flow
  include SharedLogic::RedundancyPay
  ...
end
```

I'm not sure whether this is better or even possible but thought I'd ask nonetheless.

chrisroos:
I also wondered whether it'd be possible to remove the shared logic functionality altogether.

We could avoid using it in maternity-paternity-calculator because it uses shared logic as a way of splitting the flow, rather than to actually share logic. That would leave am-i-getting-minimum-wage/minimum-wage-calculator-employers and calculate-employee-redundancy-pay/calculate-your-redundancy-pay which do share logic. _If_ we refactored/simplified the minimum-wage/redundancy-pay shared logic then it _might_ be OK to duplicate the questions in each flow, but I'm really not very sure about this.

Having written all that, I think that it is probably useful to be able to split flow logic even if it's not being shared between flows. This would help organise flows where the first question essentially leads to two independent branches (e.g. minimum-wage has two branches for checking current or past payments, childcare-costs-for-tax-credits has two branches depending on whether you're currently claiming).

chrisroos:
It doesn't feel quite right to me that the shared logic uses the `define` method - it makes them appear/feel as though they could potentially be standalone flows, rather than a way of grouping/sharing logic.

chrisroos:
Despite my reservation about the use of `define` in the shared logic, I still think this is an improvement over the use of `eval`. I don't have any bright ideas about how else to implement this so I'd be happy to see this tidied up and merged.

pmanrubia:
A few thoughts from a newbie in Smart Answers.

1. Agree that removing `eval` is a good idea, and the work around it looks in a good direction.
2. Using `include SharedLogic::RedundancyPay` feels good to me, as we would be mixing things of the same nature; but it might be confusing as we would need to rely in some magic  (`Module` hooks) to evaluate the flow definition? Being explicit with the `add` method makes it easier to read, IMO. We could also use `include_flow` or `add_flow`.
3. If we aim to have no business logic nor presentation in the flow, and the DSL get simpler, then I agree with some of the points from @chrisroos about duplication: it might not hurt sometimes.
4. This could be out of the scope of this PR, but I am not sure about the name `SharedLogic`:

- I don’t see it as only logic, as we might also be including new presentation, flow navigation, and new business logic. 
- I wouldn’t say either that its main purpose is to be shared: we might want to extract / split a big flow into smaller pieces without sharing it, in order to make it more maintainable / readable (as @chrisroos explains)
- It is more natural to me to talk about flows including other flows, or flows including subflows… not sure, just brainstorming here :)
- This is why I like the idea of wrapping the `SharedLogic` [with an object][1].

So, as @chrisross suggests, I would also merge this PR once it is finished as it’s better approach than using `eval`.

Great work! 👍 

[1]: https://github.com/alphagov/smart-answers/blob/spike-avoid-using-eval-for-shared-logic/lib/smart_answer_flows/shared_logic/adoption-calculator.rb#L3-3

floehopper:
@chrisroos: Thanks for your feedback.

> I haven't really thought this through, but could we somehow use include to make use of shared logic? 

I did contemplate a mechanism like the one you describe. However, I don't think there are any very easy ways of getting such a mechanism to work. Essentially the problem is that the `Flow#define` method is operating on an instance of the flow whereas `include` would operate on the flow's class.

Also I think that using such a mechanism would be more like using inheritance (i.e. the ancestor modules), whereas my approach is more like composition. In this case, I feel as if composition is more appropriate and easier to understand.

> I also wondered whether it'd be possible to remove the shared logic functionality altogether.

I agree with your conclusions about this, but still think this PR is useful.

> It doesn't feel quite right to me that the shared logic uses the define method - it makes them appear/feel as though they could potentially be standalone flows, rather than a way of grouping/sharing logic.

Hmm. It feels right to me that the shared logic uses the `define` method, however, I agree that it would be good to make a more explicit distinction between standalone flows and shared logic.

floehopper:
@pmanrubia: Thanks for your feedback.

> Agree that removing `eval` is a good idea, and the work around it looks in a good direction.

👍 

> Using `include SharedLogic::RedundancyPay` feels good to me, as we would be mixing things of the same nature; but it might be confusing as we would need to rely in some magic  (`Module` hooks) to evaluate the flow definition? Being explicit with the `add` method makes it easier to read, IMO. We could also use `include_flow` or `add_flow`.

I agree with your conclusion about using module inclusion - see my response to @chrisroos (above). I'll have a think about renaming the `Flow#add` method as you suggest.

> If we aim to have no business logic nor presentation in the flow, and the DSL get simpler, then I agree with some of the points from @chrisroos about duplication: it might not hurt sometimes.

👍 

> This could be out of the scope of this PR, but I am not sure about the name `SharedLogic`:
>
> - I don’t see it as only logic, as we might also be including new presentation, flow navigation, and new business logic. 
> - I wouldn’t say either that its main purpose is to be shared: we might want to extract / split a big flow into smaller pieces without sharing it, in order to make it more maintainable / readable (as @chrisroos explains)
> - It is more natural to me to talk about flows including other flows, or flows including subflows… not sure, just brainstorming here :)
> - This is why I like the idea of wrapping the `SharedLogic` [with an object][1].

I agree with everything you've said. I'll investigate the idea of better naming and better distinguishing sub-flows from flows.

[1]: https://github.com/alphagov/smart-answers/blob/spike-avoid-using-eval-for-shared-logic/lib/smart_answer_flows/shared_logic/adoption-calculator.rb#L3-3

floehopper:
Superseded by #2509.

PR #2494, opened 2016-04-29
Fix no method error undefined method transition on outcome

floehopper:
## Description

This reverts commit 0929036ffa0c92dbea372c2d570fe6f25a087e29.

It was a mistake to remove the `Outcome#transition` method which raised a `SmartAnswer::InvalidNode` exception. I hadn't noticed that [this exception is rescued and the app responds with a 404 Not Found][1]. The latter seems like sensible behaviour and so I'm restoring the previous implementation in this PR. This should prevent [exceptions like this one][2] from appearing in Errbit.

This PR also adds test coverage to hopefully avoid the same mistake happening again.

## External changes

Adding extra responses to the path of an outcome page should result in a `404 No Found` error instead of a `5XX` error and no exception should be recorded in Errbit.

### Before

https://www.gov.uk/overseas-passports/y/australia/renewing_new/adult/born-in-uk-pre-1983 -> "Sorry, we are experiencing technical difficulties"`

### After

https://www.gov.uk/overseas-passports/y/australia/renewing_new/adult/born-in-uk-pre-1983 -> "Page not found"

[1]: https://github.com/alphagov/smart-answers/blob/bf77a584737d8dbfee5694b64d5ad8870ebe0445/app/controllers/smart_answers_controller.rb#L7
[2]: https://errbit.publishing.service.gov.uk/apps/533c35ae0da1159384044f5f/problems/572087f96578631e0c660b00

### DIFF
--- a/lib/smart_answer/outcome.rb
+++ b/lib/smart_answer/outcome.rb
@@ -3,5 +3,9 @@ module SmartAnswer
     def outcome?
       true
     end
+
+    def transition(*_args)
+      raise InvalidNode
+    end
   end
 end
--- a/test/integration/smart_answers_controller_error_handling_test.rb
+++ b/test/integration/smart_answers_controller_error_handling_test.rb
@@ -1,0 +1,37 @@
+require_relative '../integration_test_helper'
+require 'gds_api/test_helpers/content_api'
+
+class SmartAnswersControllerErrorHandlingTest < ActionDispatch::IntegrationTest
+  include GdsApi::TestHelpers::ContentApi
+
+  setup do
+    @flow_registry = stub('flow-registry')
+    SmartAnswer::FlowRegistry.stubs(:instance).returns(@flow_registry)
+    stub_content_api_default_artefact
+  end
+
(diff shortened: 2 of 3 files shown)


pmanrubia:
Good catch @floehopper; LGTM

floehopper:
@pmanrubia: Thanks. Sorry for making the mistake in the first place!

floehopper:
For some reason it doesn't look as if a CI build ran against this build. And it looks as if the CI build of master has just failed. I'm running the build locally to see whether I can see what's going on.

floehopper:
Oh - it's a Rubocop violation. I'll submit a PR to fix it.

pmanrubia:
Thank you for the follow-up @floehopper, and for adding the [missing tests covering](https://github.com/alphagov/smart-answers/pull/2494/commits/07a6d33c699184206c3f333c628aacb105179897) ```Outcome#transition``` 👍

floehopper:
See https://github.com/alphagov/smart-answers/pull/2495

PR #2495, opened 2016-04-29
Fix Lint/UselessAssignment Rubocop violation

floehopper:
This was accidentally introduced in #2494.


floehopper:
I've labelled this as "Don't merge" until the CI build runs against it.

floehopper:
It turns out that there's a problem with the Jenkins instance that runs the branch CI builds. 2nd line are working on fixing it. However, I have run the build for this branch locally and it all passes (including Rubocop checks), so I think we could safely merge this PR in order to fix the master branch CI build. This I'm going to remove the "Don't merge" label.

pmanrubia:
Thank you for fixing it 👍

PR #2497, opened 2016-04-30
Add content_api to Services module and use it from FlowPresenter

floehopper:
## Description

Trello card: https://trello.com/c/YAhS8i8Y

@tijmenb introduced a `Service` module in [this commit][1] to make the `publishing_api` available. Since then we've gradually been moving `GdsApi` services into this module for consistency.

As well as making the code more consistent, this change also makes it more explicit that we're calling an external service and should make it easier to stub that service at a higher level.

The implementation of the new `Services.content_api` method was copied from `GdsApi::Helper#content_api`, but modified slightly to be consistent with the existing methods in `Services` i.e. using `Plek.new` vs `Plek.current` (the former being the preferred style according to the comment on the method definition in `Plek`).

## External changes

[1]: https://github.com/alphagov/smart-answers/commit/34f8f7f4b169c8f10471edc630c7c1088cc205aa

### DIFF
--- a/app/presenters/flow_presenter.rb
+++ b/app/presenters/flow_presenter.rb
@@ -1,8 +1,6 @@
 require 'node_presenter'
-require 'gds_api/helpers'
 
 class FlowPresenter
-  include GdsApi::Helpers
   extend Forwardable
   include Rails.application.routes.url_helpers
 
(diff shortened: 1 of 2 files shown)


chrisroos:

floehopper:

floehopper:
Rebasing against master and force-pushing (since no commit comments) in preparation for merging.

PR #2499, opened 2016-05-01
Move register_a_death context out of married_couples_allowance context

floehopper:
## Description

This just fixes a mistake made in 919071b5910c4c54557952ced8ec9868005e502b in which the `register_a_death` test context was accidentally added *inside* the `married_couples_allowance` test context. I don't think the mistake actually had any bad effects, but I found the test quite confusing, so I think it's worth fixing.

The diff is best viewed using the `--ignore-all-space` option.

## External changes


chrisroos:
Looks good to me, @floehopper. Apologies for making the mistake in the first place!

floehopper:

floehopper:
Rebasing against master and force-pushing (since no commit comments) in preparation for merging.

PR #2500, opened 2016-05-02
Fix YearRange#ends_on for year beginning on leap day (29 Feb) - rebased version

floehopper:
## Description

Supersedes #2485.

In this [Trello card][1], @erikse pointed out that a year range beginning on 29th February should end on 28th February the following year and not 27th February as in the previous implementation. Also a year range should always include either 365 or 366 days depending on whether it includes a leap day.

In working on this I realised that there was a related problem with a year beginning on 1st March in the year before a leap year. Intuitively I think that this should end on 29th February in the leap year.

However, for both of the identified problems I can see reasons why you might not want to include the leap day (29th Feb). For now always including the leap day seems like the most unsurprising behaviour, so that what's I've gone with.

It's possible that this behaviour might need to be policy/legislation-specific, so we should bear that in mind when making use of the `YearRange` class elsewhere in the code.

This commit adds more test coverage to highlight the problems as well as fixing the implementation.

Although as far as we know this inconsistency/bug hasn't resulted in any user-facing problems, it's good to get it fixed before we roll out the use of `YearRange` to other flows.

## External changes

The `YearRange` class is only currently used within the `part-year-profit-tax-credits` flow. Although it's possible that the external behaviour of this flow has changed in some very specific circumstances, I couldn't immediately come up with such a scenario.

[1]: https://trello.com/c/M17mqlSN

### DIFF
--- a/lib/smart_answer/year_range.rb
+++ b/lib/smart_answer/year_range.rb
@@ -1,7 +1,8 @@
 module SmartAnswer
   class YearRange < DateRange
     def initialize(begins_on:)
-      super(begins_on: begins_on, ends_on: begins_on.to_date + 1.year - 1)
+      ends_on = begins_on - 1 + 1.year
+      super(begins_on: begins_on, ends_on: ends_on)
     end
   end
 end
--- a/lib/smart_answer/year_range.rb
+++ b/lib/smart_answer/year_range.rb
@@ -2,6 +2,9 @@ module SmartAnswer
   class YearRange < DateRange
     def initialize(begins_on:)
       ends_on = begins_on - 1 + 1.year
+      if ends_on.month == 2 && ends_on.day == 28 && ends_on.leap?
+        ends_on += 1
+      end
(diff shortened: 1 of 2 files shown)


floehopper:
Note that @erikse added a comment to the [original PR][1] which seems to have been lost:

> There are some 29th of Feb triggered exceptions logged in Errbit that may be worth checking, I think they were for another smart answer, but if this resolves the issue, maybe that smart answer can use the YearRange class as well.

I've done a bit of digging and the only relevant exception I can find is the one which was fixed in [this earlier PR][2].

I think there are numerous places in the code (including the one above) which calculate date ranges and I suspect some/all of them could benefit from the use of `DateRange` and its sub-classes. However, I think making those changes is out-of-scope for this PR. I've created a [new Trello card][3] to capture this instead.

[1]: https://github.com/alphagov/smart-answers/pull/2485
[2]: https://github.com/alphagov/smart-answers/pull/2041
[3]: https://trello.com/c/w0t7EATU

PR #2501, opened 2016-05-02
Split up RatesQueryTest

floehopper:
## Description

The focus of a number of these tests is actually testing the underlying data in the YAML files rather than the functionality of the `DataQuery` class itself. As such I think it's better that these tests are split off into separate test cases. This also has the benefit of making the `DataQueryTest` a more manageable size.

In preparation for splitting the tests out into separate files, I removed the unnecessary top-level context in `DataQueryTest`.

After splitting the tests out into separate files, I noticed a bunch of duplication which I decided to remove, even though this wasn't strictly necessary as part of this PR.

## External changes

None. This is purely an internal refactoring and in any case only affects test code.

### DIFF
--- a/test/unit/calculators/married_couples_allowance_rates_query_test.rb
+++ b/test/unit/calculators/married_couples_allowance_rates_query_test.rb
@@ -1,0 +1,33 @@
+require_relative "../../test_helper"
+
+module SmartAnswer::Calculators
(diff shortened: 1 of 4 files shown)


pmanrubia:
👍

floehopper:
@pmanrubia: Thanks for reviewing.

floehopper:
I'm going to rebase against master and force-push in preparation for merging.

PR #2509, opened 2016-05-06
Avoid using Kernel#eval for "shared logic"

floehopper:
## Description

Supersedes #2491.

Trello card: https://trello.com/c/II2iSzR2

Originally all the flows were defined like the shared logic (i.e. outside any class or method) and evaluated, but quite some time ago we moved them into `define` methods on sub-classes of `Flow` (e.g. see [this commit][1]). At the time, the main motivation behind this was so that we could use `rcov` to obtain coverage data to help us select suitable responses for the regression tests.

Using Ruby's `Kernel#eval` is generally considered to be a Bad Thing™ and we've had to disable the `Lint/Eval` cop in the implementation of `Flow#use_shared_logic` to satisfy Rubocop. So it feels like a step forward to move the shared logic in the same direction, if only to make things more consistent.

This PR introduces a new `Flow#append(flow)` method which allows you to add the nodes of another flow within the definition of another one as follows:

### Before

```ruby
# lib/smart_answer_flows/shared_logic/shared-logic-file-name.rb
value_question :question_two?
outcome :outcome_one

# lib/smart_answer_flows/my_flow.rb
module SmartAnswer
  class MyFlow < Flow
    def define
      value_question :question_one?
      use_shared_logic 'shared-logic-file-name'
      outcome :outcome_two
    end
  end
end

# => [:question_one?, :question_two?, :outcome_one, :outcome_two]
```

### After

```ruby
# lib/smart_answer_flows/shared/another_flow.rb
module SmartAnswer
  module Shared
    class AnotherFlow < Flow
      def define
        value_question :question_two?
        outcome :outcome_one
      end
    end
  end
end  

# lib/smart_answer_flows/my_flow.rb
require 'smart_answer_flows/shared/another_flow'

module SmartAnswer
  class MyFlow < Flow
    def define
      value_question :question_one?
      append(Shared::AnotherFlow.build)
      outcome :outcome_two
    end
  end
end
```

Nodes from the other flow are appended at the point where the `Flow#append` method is invoked. As well as avoiding the undesirable calls to `Kernel#eval`, having these flow definitions within classes will make it much easier to test them.

I've moved the "shared logic" for `maternity-paternity-calculator` to be classes nested within that flow's directory. This is because they are only used in this one flow.

I've moved the `minimum_wage` and `redundancy_pay` "shared logic" to the `lib/smart_answer_flows/shared` directory. It seemed odd to have both a `shared_logic` and a `shared` directory, and the latter seems more appropriate, because we could be shared more than just "logic".

I've renamed the all the files for these "shared logic" flows to match their class names.

### Ideas considered

See the discussions in the comments for #2491.

* I did consider alternative names for `Flow#append`. Originally I used `Flow#add`, but I thought `append` better explained that the nodes of the other flow are added to the end of the list of nodes at the point where the method is called. I also considered `Flow#add_flow` and `Flow#include_flow`, however, since in the places where the method is invoked, it's quite clear that a flow is being passed in, I thought this was unnecessarily verbose. Also I think the word `include` has quite a specific meaning in Ruby which could be confusing.

* I considered better distinguishing these "shared logic" flows from fully-fledged flows. I considered naming the classes with a `SubFlow` or `PartialFlow` suffix and/or introducing a `SubFlow` or `PartialFlow` parent class. However, it struck me that many of these flows could actually be fully-fledged flows - it's actually the "main" flow in each case which is often not a fully-defined flow in its own right. Also it wasn't obvious to me what the difference in the behaviour of these flows should be. The only really obvious thing I could think of was to ensure they aren't loaded by the `FlowRegistry`, but this won't happen anyway, because they are not in the `lib/smart_answer_flows` directory.

* I think moving the "shared logic" for the `maternity-paternity-calculator` into nested classes make it pretty clear these are not fully-fledged flows.

## External changes

[1]: 09214a4378dac5ec408b5c47c36e74d1113b169a


floehopper:
Kicking the build - intermittent failure.

floehopper:
Note that GitHub orders the commits chronologically rather than topologically - this might cause some confusion when reviewing.

lib/smart_answer/flow.rb
```diff
@@ -16,18 +16,18 @@ def initialize(&block)
       instance_eval(&block) if block_given?
     end
 
+    def append(flow)
+      flow.nodes.each do |node|
+        node.flow = self
```
chrisroos: I wondered whether this would play nicely with the `FlowRegistry` caching that happens in production. I enabled `preload_flows` in my local version of this branch and accessed both am-i-getting-minimum-wage and minimum-wage-calculator-employers and they both appear to work as expected. Note that I didn't run the server in production mode as I was having problems connecting to the asset server in integration.

test/unit/flow_test.rb
```diff
@@ -348,52 +348,24 @@ class FlowTest < ActiveSupport::TestCase
...
-end
-
-outcome :sweet_tooth
-outcome :savoury_tooth
-    EOT
 
-    s = SmartAnswer::Flow.new do
-      use_shared_logic 'test_flow_logic'
+    should 'have nodes from other flow after nodes in this flow' do
+      assert_equal %i(question? outcome another_outcome), @flow.nodes.map(&:name)
```
chrisroos: I hadn't come across [`%i()`](http://ruby-doc.org/core-2.1.1/doc/syntax/literals_rdoc.html#label-Percent+Strings) before. Nice!

lib/smart_answer_flows/am-i-getting-minimum-wage.rb
```diff
@@ -49,7 +51,7 @@ def define
         end
       end
 
-      use_shared_logic "minimum_wage"
+      append(Shared::MinimumWageFlow.build)
```
chrisroos: Why have you used parenthesis in this method call? I'm interested because I think I would've naturally written `append Shared::MinimumWageFlow.build`.

chrisroos:
Unrelated to this PR but while looking at the changes to am-i-getting-minimum-wage and minimum-wage-calculator-employers, I noticed that Q1 and Q3 in both flows were identical. I've added a [Trello card to look at moving these questions to the minimum-wage shared logic](https://trello.com/c/eW53Fgax/242-dry-up-redundancy-pay-smart-answers).

chrisroos:
In commit 1aea0b463ae4c297cf6317ebf6b2b645d529dc54, you mention updating .rubocop_todo.yml based on the changes to the maternity-paternity-calculator shared logic files. Presumably there weren't equivalent changes for the minimum-wage and redundancy-pay flows?

chrisroos:
Very minor typo in commit note for b4dbef30c7a15883fd2eecc8458daedf3684bee2: It looks as though "MinimumWageFlow match" should be "MinimumWageFlow to match".

chrisroos:
Very minor typo in commit note for 2a962255ed26da4fee989509884df43264a8db9e: "apend" should be "append".

chrisroos:
Very minor typo in commit note for 8260aa537a94fef320ef55a5362c40bef87327f8: It looks as though "...RedundancyPayFlow match..." should be "...RedundancyPayFlow to match...".

chrisroos:
This all looks great to me, @floehopper. Good work.

lib/smart_answer/flow.rb
floehopper: I'm not really clear why you think this might not play nicely with the `FlowRegistry`.

The appending of the nodes from another "inner" flow is just an implementation detail of the normal `Flow.build` method. I think the "inner" flow instance would just exist temporarily during the building of the "outer" flow after which it would be available for garbage collection. The "outer" flow would be cached in the `FlowRegistry` as usual.

Can you be any more specific about your concern?

test/unit/flow_test.rb
floehopper: 👍

lib/smart_answer_flows/am-i-getting-minimum-wage.rb
floehopper: In general, I think I tend to use parentheses whenever I pass any parameters to a method. Do you know why you wouldn't have used them in this case?

Is it because the call to `use_shared_logic` didn't use parentheses?

Or are you thinking that `Flow#append` is part of an internal DSL and so you want to apply [this section of the Ruby Style Guide][1]:

> Omit parentheses around parameters for methods that are part of an internal DSL (e.g. Rake, Rails, RSpec), methods that have "keyword" status in Ruby (e.g. `attr_reader`, `puts`) and attribute access methods. Use parentheses around the arguments of all other method invocations.

In any case, I don't think this is a big deal, so I'm not going to change it now.

[1]: https://github.com/bbatsov/ruby-style-guide#syntax

floehopper:
> In commit 1aea0b463ae4c297cf6317ebf6b2b645d529dc54, you mention updating `.rubocop_todo.yml` based on the changes to the maternity-paternity-calculator shared logic files. Presumably there weren't equivalent changes for the minimum-wage and redundancy-pay flows?

Yes, that's correct.

floehopper:
@chrisroos: Thanks for reviewing. I'm going to do an interactive rebase to fix the typos in the commit notes, then rebase against master and force-push in preparation for merging.

floehopper:
Commit note typos addressed in:

* 868ff4d26630dae0e5c3ae16cbfd3fc5d688d797
* cef770fdf7bf17c798d42aba792c5d78186bec0c
* 19decf8350536b3bed6a115142fd680cca92e9cd

Closing in favour of #2513.

lib/smart_answer_flows/am-i-getting-minimum-wage.rb
chrisroos: > Do you know why you wouldn't have used them in this case?

@floehopper: In this specific case I think I would've left them out to be consistent with calls to the other DSL methods within the `define` method (e.g. `content_id`, `name`, `multiple_choice`). In the more general case, I'm not sure I'd be able to explain how I decide whether to use them or not.

I didn't know about that section of the Ruby Style Guide so I'll bear that in mind for the future.

lib/smart_answer/flow.rb
chrisroos: > Can you be any more specific about your concern?

@floehopper: Without really thinking about it; I wondered whether there might be a problem with `FlowRegistry` caching and setting `node.flow = self`. I wondered whether it might be possible for the inner flow's nodes to have the parent flow cached such that it caused problems when the inner flow was used by two separate parent flows.

I'm not sure that makes much/any sense but I'm not too worried as your explanation sounds good to me :-)

floehopper: @chrisroos: 👍

PR #2514, opened 2016-05-09
Update stubbed time in some regression tests to current tax year

floehopper:
## Description

Trello card: https://trello.com/c/w0XNKHTr

This PR introduces a way to configure the current date/time used in the regression tests. You can provide a default value to be used across all flows, but you can override this for individual flows.

While investigating possible date-related bugs in `calculate-married-couples-allowance` I noticed that the date we have been using, `2015-01-01`, was in the 2014-2015 tax year i.e. *two* tax years ago. This doesn't seem very sensible.

I've applied the new mechanism to a selection of flows:

* `additional-commodity-code`
* `benefit-cap-calculator`
* `calculate-your-child-maintenance`
* `check-uk-visa`

The only one of these which appears to be dependent on the current date/time is `calculate-married-couples-allowance`, i.e. the one I'm actually interested in updating.

Note that unfortunately the "Last updated" value on the question pages has to be updated for any flow for which we change the current date/time. I did start investigating making the regression tests insensitive to the "Last updated" change, but I decided it would take quite a while and so I've created a new [Trello card][1] to capture the idea.

Obviously it would also be good to update the current date/time for the regression tests for *all* flows, but  I decided it would take quite a while and so I've created a new [Trello card][2] to capture the idea.

## External changes

None. This is just a change to the tests - no production code has changed.

[1]: https://trello.com/c/Qnf4uX7b
[2]: https://trello.com/c/nkShUFdm


lib/smart_answer_test_helper.rb
```diff
@@ -11,10 +11,24 @@ def self.responses_and_expected_results
...
 
+  def self.configurations
+    yaml = File.read(data_path.join('configurations.yml'))
+    YAML.load(yaml)
+  end
+
+  def self.default_configuration
+    configurations.fetch('default')
+  end
+
```
pmanrubia: Is there any reason for making `SmartAnswerTestHelper.configurations` and `SmartAnswerTestHelper.default_configuration` public?

It seems that `SmartAnswerTestHelper#configuration` will be the method that will be exposed to the regression tests. What about adding more context about how it is going to be used in the commit notes?

pmanrubia:
Great work extracting the configuration for the regression tests into a config file.
LGTM

lib/smart_answer_test_helper.rb
floehopper: Since a bit awkward to make class methods private in Ruby and this is just test code (which will hopefully go away in the not too distant future), I don't think there's much benefit to making these two methods private. In fact I can't see any harm in people calling either of them from the regression test anyway.

However, I will have a look at adding more context to the commit note(s) as you suggest.

floehopper:
Superseded by #2518. Closing.

lib/smart_answer_test_helper.rb
floehopper: As you suggested, I've added more context to the commit notes in the commits in #2518.

* 927b3e55f06e1d27ffe642c7ed8935120faa0bf9
* 471152d22beb34afb760ed0a65f7293ef174e728

PR #2521, opened 2016-05-11
Render notification message for soon to be pensioners

ikennaokpala:
[Trello Story](https://trello.com/c/t08jEsHK/114-sp-age-smart-answer-new-content-for-users-within-4-months-and-4-days-of-state-pension-age) 

* Show notification message to soon to be pensioners who are 4months and 4 days from receiving State Pension (SP) age can apply to get their State Pension. (i.e. advising them that they don't have to wait till their Sate Pension age before doing this.)

## Factcheck
[Preview Link](https://smart-answers-pr-2521.herokuapp.com/state-pension-age/y/)
[GOVU.UK](https://gov.uk/state-pension-age/y/)

### Expected changes
*  Display notification to only soon to be pensioners who are 4months and 4 days from recieving State Pension (SP)

### Before

screen shot 2016-05-19 at 10 38 33

### After
 
screen shot 2016-05-19 at 10 39 25

### DIFF
--- a/lib/smart_answer/calculators/state_pension_age_calculator.rb
+++ b/lib/smart_answer/calculators/state_pension_age_calculator.rb
@@ -16,6 +16,10 @@ module SmartAnswer::Calculators
       StatePensionDateQuery.state_pension_date(dob, gender)
     end
 
+    def can_apply?
+      Date.today >= earliest_application_date
+    end
+
     def state_pension_age
       if birthday_on_feb_29?
         friendly_time_diff(dob, state_pension_date - 1.day)
@@ -50,6 +54,12 @@ module SmartAnswer::Calculators
 
     def over_16_years_old?
       dob < 16.years.ago
+    end
+
+  private
+
+    def earliest_application_date
+      state_pension_date - 4.months - 4.days
     end
   end
 end
--- a/test/unit/calculators/state_pension_age_calculator_test.rb
+++ b/test/unit/calculators/state_pension_age_calculator_test.rb
@@ -206,5 +206,45 @@ module SmartAnswer::Calculators
         end
       end
     end
+
+    context "#can_apply?" do
+      context "for women" do
+        setup do
+          @answers = { gender: "female" }
+          @calculator = StatePensionAgeCalculator.new(@answers)
+        end
+        should "return true for someone is approaching pension age in 4 months and 4 days time" do
+          Timecop.freeze(Date.parse('2020-01-11')) do
+            @calculator.stubs(dob: Date.parse("1954-06-06"))
+            assert @calculator.can_apply?
+          end
+        end
+
+        should "return false for someone is approaching pension age in more than  4 months and 4 days time" do
+          Timecop.freeze(Date.parse('2020-03-11')) do
+            @calculator.stubs(dob: Date.parse("1954-12-06"))
+            refute @calculator.can_apply?
+          end
+        end
+      end
+
+      context "for men" do
+        setup do
+          @answers = { gender: "male" }
+          @calculator = StatePensionAgeCalculator.new(@answers.merge(dob: Date.parse("1951-12-06")))
+        end
+        should "return true for someone is approaching pension age in 4 months and 4 days time" do
+          Timecop.freeze(Date.parse('2016-09-11')) do
+            assert @calculator.can_apply?
+          end
+        end
+
+        should "return false for someone is approaching pension age in more than 4 months and 4 days time" do
+          Timecop.freeze(Date.parse('2016-03-11')) do
+            refute @calculator.can_apply?
+          end
+        end
+      end
+    end
   end
 end
--- a/lib/smart_answer_flows/state-pension-age/outcomes/not_yet_reached_sp_age.govspeak.erb
+++ b/lib/smart_answer_flows/state-pension-age/outcomes/not_yet_reached_sp_age.govspeak.erb
@@ -6,6 +6,12 @@
 <% content_for :body do %>
   Your State Pension age is <%= calculator.state_pension_age %>.
 
+  <% if calculator.can_apply? %>
+    You can [claim your State Pension](/state-pension/how-to-claim) now. Your payments will start after you reach State Pension age.
+
+    Alternatively, you can [delay (defer) claiming](/deferring-state-pension).
+  <% end %>
+
   ##Pension Credit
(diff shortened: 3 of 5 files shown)


lib/smart_answer/calculators/state_pension_age_calculator.rb
```diff
@@ -16,6 +16,19 @@ def state_pension_date
       StatePensionDateQuery.state_pension_date(dob, gender)
     end
 
+    def approaching_pension_age?
+      difference = (state_pension_date - Date.today).to_i
+      if difference >= 0 && difference <= claim_can_start_within
+        true
+      else
+        false
+      end
```
floehopper: This doesn't look like very idiomatic Ruby. I think it would be better to replace the whole `if` / `else` / `end` with this:

```ruby
difference >= 0 && difference <= claim_can_start_within
```

lib/smart_answer/calculators/state_pension_age_calculator.rb
```diff
@@ -16,6 +16,19 @@ def state_pension_date
       StatePensionDateQuery.state_pension_date(dob, gender)
     end
 
+    def approaching_pension_age?
```
floehopper: I think the fact that they are approaching pension age is the reason why they are allowed to claim. So I think I would prefer if this method was named something more like `can_apply?`.

lib/smart_answer/calculators/state_pension_age_calculator.rb
```diff
@@ -16,6 +16,19 @@ def state_pension_date
...
+      if difference >= 0 && difference <= claim_can_start_within
+        true
+      else
+        false
+      end
+    end
+
+    def claim_can_start_within
+      (4.month.since.to_date - Date.today).to_i + 4
+    end
```
floehopper: I might be wrong, but I think this could all be written a lot more simply e.g. something like this:

```ruby
    def earliest_application_date
      state_pension_date - 4.months - 4.days
    end

    def can_apply?
      Date.today >= earliest_application_date
    end
```

test/unit/calculators/state_pension_age_calculator_test.rb
```diff
@@ -206,5 +206,57 @@ class StatePensionAgeCalculatorTest < ActiveSupport::TestCase
...
+        Timecop.freeze(Date.parse('2016-01-11')) do
+          calculator = StatePensionAgeCalculator.new({})
+          assert_equal calculator.claim_can_start_within, 125
+        end
+      end
+    end
+
+    context "#approaching_pension_age" do
+      setup do
+        StatePensionAgeCalculator.any_instance.stubs(:claim_can_start_within).returns(127)
```
floehopper: Since we're stubbing time in the tests anyway, I'm not convinced that stubbing`claim_can_start_within` is really helping. I'd be inclined to test the `approaching_pension_age?` method without stubbing anything other than time.

test/unit/calculators/state_pension_age_calculator_test.rb
```diff
@@ -206,5 +206,57 @@ class StatePensionAgeCalculatorTest < ActiveSupport::TestCase
...
+        end
+
+        should "return false for someone is approaching pension age in more than 4 months and 4 days time" do
+          Timecop.freeze(Date.parse('2016-03-11')) do
+            calculator = StatePensionAgeCalculator.new(@answers.merge(dob: Date.parse("1951-12-06")))
+            refute calculator.approaching_pension_age?
+          end
+        end
+      end
+    end
```
floehopper: I think these tests might be simpler if we instantiate the calculator in the relevant `setup` block with a fixed `dob` and vary the date on which we freeze time in the tests.

floehopper:
The PR title & description and the first commit note all talk about a "call-out". However, it doesn't look as if we are showing a call-out. Also [this comment][1] in Trello suggests that there was a call-out initially, but it was subsequently removed. I think the PR title & description and the first commit note should all be updated accordingly.

[1]: https://trello.com/c/t08jEsHK/114-sp-age-smart-answer-new-content-for-users-within-4-months-and-4-days-of-state-pension-age#comment-5734547169142ac0a1d69cf3

ikennaokpala:
@floehopper thanks for these comments.. much appreciated.. this is WIP at the moment.. yester @raykhan66  was sitting to my left and initially he asked me to use a callout which I have since changed upon his request as well this morning.

floehopper:
@ikennaokpala: OK. Sorry - I somehow missed the "Waiting on factcheck" label. Sorry.

test/unit/calculators/state_pension_age_calculator_test.rb
ikennaokpala: That would not work because women have varied pension ages through the years..

ikennaokpala:
@floehopper NP..

lib/smart_answer/calculators/state_pension_age_calculator.rb
ikennaokpala: Yeah but the intention I had was to 4months and 4 days from the current day that the user is actually using the smart answer. `claim_can_start_within ` returns the number of days from today. 

For months like February that are not up to 31 days I can easily return the right number of days + 4 from the day the user uses the state pension age smart answer.  

with what you have here.. `state_pension_date - 4.months - 4.days` it is basd on the proposed date when they reach state pension age,

The purpose for this is to encourage users who will be reaching state pension ages within 4 months and 4 days time to apply

lib/smart_answer/calculators/state_pension_age_calculator.rb
ikennaokpala: @floehopper I understand your context here.. the naming was influence from my discovery session with @raykhan66.. I feel it better covers the `within` context of this scenario.

test/unit/calculators/state_pension_age_calculator_test.rb
ikennaokpala: You have a point here.. I will review this..

lib/smart_answer/calculators/state_pension_age_calculator.rb
ikennaokpala: I agree.. I lost track of this in the midst of my commit review.

ikennaokpala:
@floehopper its now ready for code review.. @raykhan66 says this is really urgent and needs to go live pretty quick

pmanrubia:
LGTM 👍

PR #2463, opened 2016-04-15
Overseas passports content changes to no service countries and Getting your passport

leenagupte:
Trello Story: https://trello.com/c/BwW6E1wW/84-asap-overseas-passports-smart-answer-change-to-no-service-countries-and-getting-your-passport

# Change to no service countries

## Factcheck
[Overseas Passports - North Korea](https://smart-answers-pr-2463.herokuapp.com/overseas-passports/y/north-korea)

## Expected Changes
[URL on GOV.UK](https://www.gov.uk/overseas-passports/y/north-korea)

### Before
screen shot 2016-04-15 at 14 44 37

### After
screen shot 2016-04-15 at 14 43 06

Affected countries:

* British Indian Ocean Territory
* Iran
* Libya
* North Korea
* South Georgia and South Sandwich Islands
* Syria
* Yemen

N.B. The exact wording for each country is different.

# Change to countries with a UK Visa Application Centre

## Factcheck
[Overseas Passports - Afghanistan](https://smart-answers-pr-2463.herokuapp.com/overseas-passports/y/afghanistan/renewing_new/adult)

## Expected Changes
[URL on GOV.UK](https://www.gov.uk/overseas-passports/y/afghanistan/renewing_new/adult)

### Before
screen shot 2016-04-15 at 16 28 58

### After
screen shot 2016-04-15 at 16 29 32

Affected countries:
* Afghanistan
* Algeria
* Azerbaijan
* Bangladesh
* Belarus
* Burma
* Cambodia
* China
* Georgia
* India
* Indonesia
* Kazakhstan
* Lebanon
* Mauritania
* Morocco
* Nepal
* Pakistan
* Russia
* Thailand
* Ukraine
* Venezuela
* Vietnam
* Western Sahara

# Change to countries with a British Embassy High Commission

## Factcheck
[Overseas Passports - Cuba](https://smart-answers-pr-2463.herokuapp.com/overseas-passports/y/cuba/renewing_new/adult)

## Expected Changes
[URL on GOV.UK](https://www.gov.uk/overseas-passports/y/cuba/renewing_new/adult)

### Before
screen shot 2016-04-15 at 16 31 19

### After
screen shot 2016-04-15 at 16 33 18

Affected countries:
* Cuba
* Kyrgyzstan
* Sudan
* Tajikistan
* The Occupied Palestinian Territories (Gaza only)
* Timor Leste
* Turkmenistan
* Uzbekistan

# Change to countries with named visa application centres or embassies

## Factcheck
[...]

PR #2296, opened 2016-02-23
Update Student finance forms with 2016 - 2017 option for full time students

erikse:
Pivotal Tracker: https://www.pivotaltracker.com/story/show/113377301 and https://www.pivotaltracker.com/story/show/111708486

This PR adds new outcomes for EU and UK full time students to find forms for the 2016 to 2017 student year, and removes outcomes for the same students for the 2014 to 2015 student year. Minor updates to existing outcomes with new PDF sizes are also included, as well as a link to apply online for certain types of student finance.

## Fact check

https://smart-answers-pr-113377301.herokuapp.com/student-finance-forms

## Expected changes
 [URL on gov.uk](https://www.gov.uk/student-finance-forms/y/uk-full-time/apply-loans-grants)
  * Adds option 2016 to 2017
  * Removes option 2014 to 2015

### Before
screen shot 2016-02-23 at 3 36 54 pm

### After
screen shot 2016-02-23 at 3 37 01 pm

 [URL on gov.uk](https://www.gov.uk/student-finance-forms/y/uk-full-time/apply-loans-grants/year-1516/new-student)
  * Adds link to 'apply online'
  * Updates PN1 form size from 468KB to 465KB
  * Updates PN1 guidance notes size from 223KB to 300KB

### Before
screen shot 2016-02-23 at 3 40 32 pm

### After
screen shot 2016-02-23 at 3 40 43 pm

Note: this PR supersedes #2266

PR #2524, opened 2016-05-12
Fix responses and expected results files for regression tests

floehopper:
## Description

While looking into something else, I noticed that the `*-responses-and-expected-results.yml` files for a number of flows were out-of-date:

* `check-uk-visa`
* `student-finance-forms`
* `state-pension-through-partner`
* `overseas-passports`

This PR updates those files and identifies the PR in which they should have been updated in the first place. I haven't done any checking that these changes are sensible, but I plan to ask the authors of the relevant PRs to check for me, because they will have more context.

While fixing the above, I also came across a few other problems:

* The changes in #2498 meant that the scripts for generating both the "questions & responses" and "responses & expected results" files were trying to use a local instance of the Whitehall app rather than the production instance.
* The questions & responses file for `simplified-expenses-checker` contained some invalid YAML.
* The questions & responses file for `landlord-immigration-check` was not up-to-date with the stubbed imminence lookups in the regression test nor its responses & expected results file.

I've fixed all of the above and it would be good to get this merged as soon as possible.

Note that I've also noticed that the script for generating the responses & expected results is using the wrong stubbed value of current time for some flows. I plan to address that in a separate PR, but that PR will rely on this one.

## External changes

None. These changes only affect test code - no changes to production code.

### DIFF
--- a/script/generate-responses-and-expected-results-for-smart-answer.rb
+++ b/script/generate-responses-and-expected-results-for-smart-answer.rb
@@ -1,3 +1,5 @@
+ENV['PLEK_SERVICE_WHITEHALL_ADMIN_URI'] = 'https://www.gov.uk'
+
 require 'timecop'
 
 Timecop.freeze(Date.parse('2015-01-01'))
--- a/test/data/check-uk-visa-responses-and-expected-results.yml
+++ b/test/data/check-uk-visa-responses-and-expected-results.yml
@@ -630,7 +630,7 @@
   - kuwait
   - work
   - six_months_or_less
-  :next_node: :outcome_work_m
+  :next_node: :outcome_work_waiver
   :outcome_node: true
 - :current_node: :staying_for_how_long?
   :responses:
@@ -650,7 +650,7 @@
   - kuwait
   - study
   - six_months_or_less
-  :next_node: :outcome_visit_waiver
+  :next_node: :outcome_study_waiver
   :outcome_node: true
 - :current_node: :staying_for_how_long?
   :responses:
@@ -695,7 +695,7 @@
   :responses:
   - kuwait
   - school
-  :next_node: :outcome_visit_waiver
+  :next_node: :outcome_school_waiver
   :outcome_node: true
 - :current_node: :purpose_of_visit?
   :responses:
--- a/test/data/overseas-passports-responses-and-expected-results.yml
+++ b/test/data/overseas-passports-responses-and-expected-results.yml
@@ -752,7 +752,7 @@
 - :current_node: :which_country_are_you_in?
   :responses:
   - yemen
-  :next_node: :cannot_apply
+  :next_node: :apply_in_neighbouring_country
   :outcome_node: true
 - :current_node: :which_country_are_you_in?
   :responses:
--- a/test/data/student-finance-forms-responses-and-expected-results.yml
+++ b/test/data/student-finance-forms-responses-and-expected-results.yml
@@ -14,29 +14,6 @@
   :responses:
   - uk-full-time
   - apply-loans-grants
-  - year-1516
-  :next_node: :continuing_student?
-  :outcome_node: false
-- :current_node: :continuing_student?
-  :responses:
-  - uk-full-time
-  - apply-loans-grants
-  - year-1516
-  - continuing-student
-  :next_node: :outcome_uk_ft_1516_continuing
-  :outcome_node: true
-- :current_node: :continuing_student?
-  :responses:
-  - uk-full-time
-  - apply-loans-grants
-  - year-1516
-  - new-student
-  :next_node: :outcome_uk_ft_1516_new
-  :outcome_node: true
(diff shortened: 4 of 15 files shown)


script/generate-responses-and-expected-results-for-smart-answer.rb
```diff
@@ -1,3 +1,5 @@
+ENV['PLEK_SERVICE_WHITEHALL_ADMIN_URI'] = 'https://www.gov.uk'
```
chrisroos: It looks as though this line should be added to generate-questions-and-responses-for-smart-answer.rb too. It's not clear to me whether that should happen in this PR or not though.

chrisroos:
Apart from my single comment, this all looks good to me.

chrisroos:
I wonder whether it's worth adding tests to avoid these sort of problems in future? Do you think the effort of adding such tests would be warranted?

script/generate-responses-and-expected-results-for-smart-answer.rb
floehopper: Good point. I think I might as well add it to this PR.

floehopper:
@chrisroos:

> I wonder whether it's worth adding tests to avoid these sort of problems in future? Do you think the effort of adding such tests would be warranted?

I'm not sure. It's not obvious to me how these changes were missed. I suspect that because many PRs only involve content changes people may just get into the habit of only updating the regression test artefacts & checksums, and not think about whether these files need updating.

I think the first thing to do would be to check that the documentation makes it clear when you should update these files. Then we could consider adding some kind of check or test to make sure they do get updated. I did wonder whether we could just always run the generate script before running the regression test for that flow. Anyway, I think it's out of scope for this PR, but I've created a [Trello card][1] to capture the idea.

[1]: https://trello.com/c/duHLnGEO

test/data/check-uk-visa-responses-and-expected-results.yml
```diff
@@ -630,7 +630,7 @@
   - kuwait
   - work
   - six_months_or_less
-  :next_node: :outcome_work_m
+  :next_node: :outcome_work_waiver
```
floehopper: It looks as if this change was the result of this commit: 8053b80eb930b8bab176d2af43b148a4656fdda0 which was part of #2524. I'm therefore happy to include it.

test/data/check-uk-visa-responses-and-expected-results.yml
```diff
@@ -650,7 +650,7 @@
   - kuwait
   - study
   - six_months_or_less
-  :next_node: :outcome_visit_waiver
+  :next_node: :outcome_study_waiver
```
floehopper: It looks as if this change was the result of this commit: 5443c5429c9ba9522b87e5f4d062e6d43cb99fdd which was part of #2524. I'm therefore happy to include it.

test/data/check-uk-visa-responses-and-expected-results.yml
```diff
@@ -695,7 +695,7 @@
   :responses:
   - kuwait
   - school
-  :next_node: :outcome_visit_waiver
+  :next_node: :outcome_school_waiver
```
floehopper: It looks as if this change was the result of this commit: 76b6d5cfaf66cabd4183d0bb4c9c0827880c4ccb which was part of #2524. I'm therefore happy to include it.

test/data/student-finance-forms-responses-and-expected-results.yml
```diff
@@ -337,42 +337,42 @@
...
   :responses:
   - eu-full-time
-  - year-1617
+  - year-1516
   - new-student
-  :next_node: :outcome_eu_ft_1617_new
+  :next_node: :outcome_eu_ft_1516_new
   :outcome_node: true
 - :current_node: :type_of_student?
   :responses:
```
floehopper: It looks as if all these changes to responses & expected results should have been included in this commit: abd3fe1a29b67d134a7f0d0de4f73135c21ef5b2 which was part of #2296. It's odd that both the questions & responses and responses & expected results files were updated in that commit, but the latter did not match the former.

Anyway now that I understand a bit more about what happened, I'm happy to include these changes.

test/data/overseas-passports-responses-and-expected-results.yml
```diff
@@ -752,7 +752,7 @@
 - :current_node: :which_country_are_you_in?
   :responses:
   - yemen
-  :next_node: :cannot_apply
+  :next_node: :apply_in_neighbouring_country
```
floehopper: I can see that this change was caused by this commit: 9c71c568436ad3bf4a5010104083d204b3f4d622 as part of #2463. So I'm happy to include it in this PR.

floehopper:
I'm going to rebase this against master and open a fresh PR so as not to disrupt the comments on this one.

floehopper:
Superseded by #2527. Closing.

PR #2529, opened 2016-05-13
Refactor calculate-married-couples-allowance and fix 4 bugs

floehopper:
## Description

This started of as an investigation of [this Trello card][1] (originally an old GitHub issue) relating to two large comments in the `calculate-married-couples-allowance` code. That card talks about separating the policy rules for the Married Couples Allowance and the Personal Allowance. While I've made a little progress on that front, by separating out the data relating to the Personal Allowance, this is by no means finished.

During this investigation, I found understanding the code very hard and so I decided to refactor the flow towards the [new style][2] to make things clearer. As I went on, I identified a number of bugs which are outlined below. The most significant bug is the first one - the other three fix behaviour for tax years prior to the current one (2015-2016).

### Income limit bug

Trello cards: https://trello.com/c/I5X3nmow & https://trello.com/c/Zse6DH8v

Prior to this PR the flow was always using a hard-coded income limit of £27,000 to decide whether or not to ask follow-up questions about pension & gift aid contributions. However, the [income limit was increased to £27,700 from the 2015-2016 tax year onwards][3].

Note that this is a relatively subtle bug, because the actual calculations of the allowance does take the correct income limit into account.

### Age vs date-of-birth bug

Trello card: https://trello.com/c/QRHxaXDi

The age-related Personal Allowance was only supposed to be based on age up until the end of the 2012-2013 tax year. [Since then it should have been based on the person's date of birth][6], but the latter was never implemented in the code. This PR chooses the correct mechanism for determining the age-related Personal Allowance depending on whether today's date is before or after the end of the 2012-2013 tax year.

Since the 2015-2016 tax year started, the Personal Allowance has been the same for all ages/dates-of-birth and so it's not possible to see a change in behaviour unless we setup a Heroku instance with the date set in the past. I'm not convinced this is worthwhile, but I do think it's worth fixing the code so that if/when we remove the age-based mechanism, it's more obvious what was going on.

### Age calculation bug

Trello card: https://trello.com/c/w6gmL50V

Someone's age on a given date depends on whether or not they've had their birthday in the current year. Also [someone who's birthday falls on 29th Feb in a leap year, is legally considered to have their birthday on 1st March in a non-leap year][4]. Prior to this commit, neither of these were being taken into account.

Given the fix for the "Age vs date-of-birth bug" (above) and the fact that we are already well past 2013-04-05 which is when the changeover to use date-of-birth was supposed to happen, this fix has no externally-visible effect. We could setup a Heroku instance with the date set in the 2012-2013 tax year, but I'm not convinced this is worthwhile. However, I do think it's worth fixing the code so that if/when we simplify the code, it's more obvious what was going on.

### Age at end of current tax year bug

Trello card: https://trello.com/c/qvduSzB6

Prior to this PR the age was being calculated on *5th April in the current calendar year*. I'm pretty confident this is incorrect and that the age should be calculated *at the end of the current tax year*. I haven't been able to find any direct (GOV.UK) evidence that this is correct, but it seems logical and I have found [this indirect evidence][5].

Given the fix for the "Age vs date-of-birth bug" (above) and the fact that we are already well past 2013-04-05 which is when the changeover to use date-of-birth was supposed to happen, this fix has no externally-visible effect. We could setup a Heroku instance with the date set in the 2012-2013 tax year, but I'm not convinced this is worthwhile. However, I do think it's worth fixing the code so that if/when we simplify the code, it's more obvious what was going on.

## External changes

Much of the work in this PR is refactoring or fixing of caclulations for previous tax years. The only externally-visible change is the fix for the "income limit" bug which can be tested as follows:

#### Before

* Visit https://www.gov.uk/calculate-married-couples-allowance/y/yes/yes/1950-01-01/27001.0
* See you are on "Are you paying into a pension?" question page.

#### After

* Visit https://smart-answers-pr-2529.herokuapp.com/calculate-married-couples-allowance/y/yes/yes/1950-01-01/27001.0
* See you are on "The husband qualifies for Married Couple's Allowance" outcome page.

## Possible further work

* Separate the policy rules for the Married Couples Allowance and the Personal Allowance
* Move the tests for `calculate-married-couples-allowance` towards the [new-style testing approach][7].

[1]: https://trello.com/c/XLE6L7s8
[2]: https://github.com/alphagov/smart-answers/blob/master/doc/refactoring.md
[3]: https://www.gov.uk/government/publications/rates-and-allowances-income-tax/income-tax-rates-and-allowances-current-and-past#personal-allowances-for-people-born-before-6-april-1948
[4]: https://en.wikipedia.org/wiki/February_29#Legal_status
[5]: http://www.pwc.co.uk/assets/pdf/budget-2012-tax-datacard-2012.pdf
[6]: https://www.gov.uk/government/publications/rates-and-allowances-income-tax/income-tax-rates-and-allowances-current-and-past#personal-allowances-for-people-born-before-6-april-1948
[7]: https://github.com/alphagov/smart-answers/blob/master/doc/new-style-testing.md


floehopper:
I've added the "Don't merge" label because I haven't quite finished writing the PR description.

floehopper:
I've just rebased this against master and updated the commit timestamps so they appear in a sensible order on GitHub.

floehopper:
I've just rebased the branch against master, force-pushed, updated the PR description and removed the "Don't merge" label. This PR is now ready for review. I'm sorry it looks quite big (in terms of the number of commits), but note that many of them are quite small and hopefully it should be relatively easy to follow what's going on.

chrisroos:
__Commit titled "Update regression test responses, expected results & artefacts" (b225746fd044c5699075778fc1b507662c881df5)__

Great explanation of the changes to the artefacts :-)

chrisroos:
__Commit titled "Remove unnecessary default values for income adjustment amounts" (829e4717202115ae5a6484893931cfb0f4500a9f)__

It looks like there's some text missing from the commit note: "Since `nil.to_f`, none...".

chrisroos:
__Commit titled "Invert income_equal_to_or_more_than_limit_for_personal_allowance? logic" (8482ff0659b444c68c253cb49a1a8d881e7c9274)__

Not wanting to send you down any more rabbit holes (and I realise you haven't changed the existing behaviour), but do you know if this logic (`income.to_f < income_limit_for_personal_allowances`) is correct? I notice that in `#calculate_allowance` there's a check for `income > income_limit_for_personal_allowances` and wondered whether there's an edge case when `income` is equal to `income_limit_for_personal_allowances`.

chrisroos:
__Commit titled "Fix assertions in personal allowance tests and fix implementation" (00587411e04b04befe2629c2cdf42ea7a5a892ae)__

How did you decide that the test names described the correct behaviour? It might be worth explaining in the commit note.

chrisroos:
__Commit titled "Add tests for personal allowance in post-2013-2014 & fix implementation" (27a1da31d2c69d2767bb16fc0e6dc9d47be6acf3)__

 It looks as though the test named "return the 2nd higher allowance for someone born on 6th April 1938" should include "5th April" so that it matches the implementation of the test.

chrisroos:
__Commit titled "Add tests for personal allowance in post-2013-2014 & fix implementation" (27a1da31d2c69d2767bb16fc0e6dc9d47be6acf3)__

I don't really understand why the regression test artefacts don't change after this commit. If I understand correctly then we're now using the new in this commit date of birth mechanism to decide on the allowance figures so I think I would've expected to see some changes in the artefacts.

chrisroos:
__Commit titled "Make SmartAnswer namespace available in PersonalAllowanceCalculator" (b684c294d438cb627ec6ab8405c7ddfa9f3c0472)__

Very minor. It might be good to explain why you want to make it accessible, and that using `ignore-all-space` helps show the actual change.

chrisroos:
__Commit titled "Make age_at_end_of_current_tax_year method private" (6df58202e0a65c85468944bb7e87e13bfcf51295)__

Incredibly minor. You might want to explain why you're making it private. Presumably because it's not being used outside the class.

chrisroos:
> I've just rebased this against master and updated the commit timestamps so they appear in a sensible order on GitHub.

How did you do this, out of interest?

chrisroos:
I've left a few comments but they shouldn't take away from the fact that this looks like a great improvement!

floehopper:
Superseded by #2545. Closing.

floehopper:
> It looks like there's some text missing from the commit note: "Since `nil.to_f`, none...".

Addressed in f73c62f6cebd22bf46de30a85f6ceccf7aba4c0d.

floehopper:
> Not wanting to send you down any more rabbit holes (and I realise you haven't changed the existing behaviour), but do you know if this logic (`income.to_f < income_limit_for_personal_allowances`) is correct? I notice that in `#calculate_allowance` there's a check for `income > income_limit_for_personal_allowances` and wondered whether there's an edge case when `income` is equal to `income_limit_for_personal_allowances`.

I hadn't noticed that and it's certainly confusing. However, I'm pretty confident there isn't a serious bug, because the `calculate_allowance` method works OK when your income is exactly the same as the limit. The only bad thing might be that you are asked the extra questions (e.g. pension contributions) unnecessarily.

I plan to revisit the `calculate_allowance` code as part of this [Trello card][1], so I've added a comment to that card to remind me about this discrepancy.

[1]: https://trello.com/c/XLE6L7s8

floehopper:
> How did you decide that the test names described the correct behaviour? It might be worth explaining in the commit note.

Addressed in 3b56b69fd4d649230b3e158a73ffe18ac4579b98.

floehopper:
> It looks as though the test named "return the 2nd higher allowance for someone born on 6th April 1938" should include "5th April" so that it matches the implementation of the test.

Addressed in 7419a7ae5af202007c295b42e69b6af80e3fb977.

floehopper:
> I don't really understand why the regression test artefacts don't change after this commit. If I understand correctly then we're now using the new in this commit date of birth mechanism to decide on the allowance figures so I think I would've expected to see some changes in the artefacts.

I've expanded the commit note in 7419a7ae5af202007c295b42e69b6af80e3fb977 to try to explain this.

floehopper:
> Very minor. It might be good to explain why you want to make it accessible, and that using ignore-all-space helps show the actual change.

I've expanded the commit note in 2ce77c854a3fa5b5b44191073c6c6688b37079dc to try to explain this.

floehopper:
> Incredibly minor. You might want to explain why you're making it private. Presumably because it's not being used outside the class.

I've expanded the commit note in 1e704e84dfa1913557a9ca7983bd8638072b062b to try to explain this.

floehopper:
>> I've just rebased this against master and updated the commit timestamps so they appear in a sensible order on GitHub.

> How did you do this, out of interest?

`git rebase --ignore-date master`

PR #2532, opened 2016-05-16
Avoid using worldwide api helpers in integration tests

chrisroos:
Trello card: https://trello.com/c/T69SfCaC

Stubbing `WorldLocation.all` and `WorldLocation.find`, instead of using the `GdsApi::TestHelpers::Worldwide` helper methods, appears to shave about 50-60 seconds off the time it takes to run `rake` (i.e. all but the regression tests) on my machine.

We're still using the `GdsApi::TestHelpers::Worldwide` methods in the country_and_date_questions_test, the regression tests and a couple of unit tests. I had hoped that making a similar change to the regression tests would give us a similar performance improvement but that doesn't seem to be the case based on my initial investigation. I plan to dig into this further at another time. I haven't updated the country_and_date_questions_test because making the same change caused the tests to fail. I didn't investigate in any detail but I think this might because of the JavaScript tests in that file. I didn't update the unit tests because I think this is a good enough improvement to get merged without spending more time on the branch.

### DIFF
--- a/lib/smart_answer_flows/register-a-birth.rb
+++ b/lib/smart_answer_flows/register-a-birth.rb
@@ -170,7 +170,7 @@ module SmartAnswer
           location = WorldLocation.find(registration_country)
           raise InvalidResponse unless location
           organisations = [location.fco_organisation]
-          if organisations.present?
+          if organisations && organisations.any?
             service_title = 'Births and Deaths registration service'
             organisations.first.offices_with_service(service_title)
           else
--- a/test/data/register-a-birth-files.yml
+++ b/test/data/register-a-birth-files.yml
@@ -1,5 +1,5 @@
 ---
-lib/smart_answer_flows/register-a-birth.rb: 232ef65452cefa8a62711ee7a0bb4c4b
+lib/smart_answer_flows/register-a-birth.rb: 4cc6aeba5cd5958bcb951ed703281810
 test/data/register-a-birth-questions-and-responses.yml: c59cb1d0c11f3ad29f905b632f7c58bb
 test/data/register-a-birth-responses-and-expected-results.yml: 3ac4372034c86d8cc87665db59ff11d5
 lib/smart_answer_flows/register-a-birth/outcomes/_fees.govspeak.erb: a2bc6c611c6809b333ab42066a805f71
--- a/test/integration/smart_answer_flows/check_uk_visa_test.rb
+++ b/test/integration/smart_answer_flows/check_uk_visa_test.rb
@@ -1,16 +1,14 @@
 require_relative '../../test_helper'
 require_relative 'flow_test_helper'
-require 'gds_api/test_helpers/worldwide'
 
 require "smart_answer_flows/check-uk-visa"
 
 class CheckUkVisaTest < ActiveSupport::TestCase
   include FlowTestHelper
-  include GdsApi::TestHelpers::Worldwide
 
   setup do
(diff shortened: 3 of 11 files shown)


floehopper:
I noticed that you've remove a bunch of calls to `worldwide_api_has_organisations_for_location` which were loading organisation fixture data. It looks as if this has been replaced by stubbing `WorldLocation#fco_organisation` to always return `nil`. I think it would be worth explaining somewhere why the organisation data previously being made available is not needed. Presumably none of the tests were actually relying on it?

floehopper:
> We're still using the `GdsApi::TestHelpers::Worldwide` methods in the `country_and_date_questions_test`, the regression tests and a couple of unit tests.

Why did you decide not to do the same in the remaining non-regression tests? If there's a good reason not to do it, I think you should explain the reason. If not, then I think we should probably do it in this PR to keep the code as consistent as possible.

floehopper:
>  I had hoped that making a similar change to the regression tests would give us a similar performance improvement but that doesn't seem to be the case based on my initial investigation.

One thing to be aware of is that in the regression tests (when run outside the default `rake` task), the setup of the stubs is only done *once* at the beginning of the whole suite. I wonder if this might explain the lack of a performance improvement.

floehopper:
Other than my comments above, this looks good to me. Good work! 👍

chrisroos:
> I noticed that you've remove a bunch of calls to `worldwide_api_has_organisations_for_location` which were loading organisation fixture data. It looks as if this has been replaced by stubbing `WorldLocation#fco_organisation` to always return `nil`. I think it would be worth explaining somewhere why the organisation data previously being made available is not needed. Presumably none of the tests were actually relying on it?

@floehopper: I tried to explain this in the commit notes of the commits that removed those calls (e.g. https://github.com/alphagov/smart-answers/pull/2532/commits/3b45bdbd646ad739ab7d38d68b38e57c9cd02ddc). I explained the removal of `worldwide_api_has_no_organisations_for_location` in the same way (e.g. https://github.com/alphagov/smart-answers/pull/2532/commits/aba724aba3a45902b8bfaec1ffe640060040911b). Did you see those? Do you think it'd be better to include this information in the PR description?

chrisroos:
>>  I had hoped that making a similar change to the regression tests would give us a similar performance improvement but that doesn't seem to be the case based on my initial investigation.

> One thing to be aware of is that in the regression tests (when run outside the default `rake` task), the setup of the stubs is only done *once* at the beginning of the whole suite. I wonder if this might explain the lack of a performance improvement.

I suspect you're right, @floehopper. When updating the regression tests to use Mocha instead of WebMock I found that I had to instantiate the stubs at the beginning of each test, rather than at the beginning of the whole suite.

chrisroos:
>> We're still using the `GdsApi::TestHelpers::Worldwide` methods in the `country_and_date_questions_test`, the regression tests and a couple of unit tests.

> Why did you decide not to do the same in the remaining non-regression tests? If there's a good reason not to do it, I think you should explain the reason. If not, then I think we should probably do it in this PR to keep the code as consistent as possible.

I haven't updated the `country_and_date_questions_test` because making the same change caused the tests to fail. I didn't investigate in any detail but I _think_ this might because of the JavaScript tests in that file. I didn't update the unit tests because I thought this was a good enough improvement to get merged without spending more time on the branch. I'll update the PR description with this information and add Trello cards to track the remaining tasks.

chrisroos:
I've updated https://trello.com/c/T69SfCaC/86-stub-gds-apis-at-ruby-client-level-vs-net-http-webmock-level with a link to this PR, and a checklist of the remaining tests to convert.

chrisroos:
Thanks for reviewing, @floehopper. I'm going to close this, rebase on master and get it merged.

PR #2516, opened 2016-05-10
Opposite sex marriages in Rwanda to Mirror other Commonwealth countries

ikennaokpala:
[Trello Story](https://trello.com/c/IFID2fDT/119-rwanda-marriage-tool-content-change-request)

* Content changes for all opposite sex marriages in Rwanda

## Factcheck
[Preview Link]( https://smart-answers-pr-2516.herokuapp.com/marriage-abroad/y/rwanda/ceremony_country/partner_local/opposite_sex)
[GOV.UK]( https://gov.uk/marriage-abroad/y/rwanda/ceremony_country/partner_local/opposite_sex)

### Expected changes
* Outcome for (all) opposite sex marriages in Rwanda to mirror that of other Commonwealth countries.

### Before
screen shot 2016-05-10 at 12 40 51

### After
screen shot 2016-05-10 at 12 40 31

#### Other affected pages

 https://smart-answers-pr-2516.herokuapp.com/marriage-abroad/y/rwanda/ceremony_country/partner_british/opposite_sex
 https://smart-answers-pr-2516.herokuapp.com/marriage-abroad/y/rwanda/ceremony_country/partner_local/opposite_sex
 https://smart-answers-pr-2516.herokuapp.com/marriage-abroad/y/rwanda/ceremony_country/partner_other/opposite_sex
 https://smart-answers-pr-2516.herokuapp.com/marriage-abroad/y/rwanda/third_country/partner_british/opposite_sex
 https://smart-answers-pr-2516.herokuapp.com/marriage-abroad/y/rwanda/third_country/partner_local/opposite_sex
 https://smart-answers-pr-2516.herokuapp.com/marriage-abroad/y/rwanda/third_country/partner_other/opposite_sex
 https://smart-answers-pr-2516.herokuapp.com/marriage-abroad/y/rwanda/uk/partner_british/opposite_sex
 https://smart-answers-pr-2516.herokuapp.com/marriage-abroad/y/rwanda/uk/partner_local/opposite_sex
 https://smart-answers-pr-2516.herokuapp.com/marriage-abroad/y/rwanda/uk/partner_other/opposite_sex

PR #2536, opened 2016-05-17
Rwanda common wealth

pmanrubia:
This PR explores the steps to take to include Rwanda in the list of Commonwealth countries.

This is a continuation of the work on PR #2516: see comments: https://github.com/alphagov/smart-answers/pull/2516#issuecomment-219454608, https://github.com/alphagov/smart-answers/pull/2516#issuecomment-218456039, https://github.com/alphagov/smart-answers/pull/2516#issuecomment-219687650

[Trello card](https://trello.com/c/U0dnbMdv/148-include-rwanda-in-the-list-of-commonwealth-countries) to track this work.
URL to [fact-check](https://smart-answers-pr-2536.herokuapp.com/marriage-abroad)

So far, it impacts on Register a Death and Register a Birth as both flows as [checking if the country is included in the list of Commonwealth countries](https://github.com/alphagov/smart-answers/blob/rwanda-common-wealth/lib/smart_answer_flows/register-a-birth.rb#L30-42).

Next steps:
- [x] Validate with the content team if this changes are valid
- [ ] Run all regression tests to validate no other tool is affected
- [x] Verify that the regression tests are covering all the options for Rwanda (as we are not getting update on the regression tests, only deletions)
- [ ] Regenerate checksums before and after adding the country to the list


floehopper:
Minor: I found the commit titles "Regenerate artefacts for *" confusing, because in these commits you are generating the artefacts for the *first* time i.e. "Generate artefacts for *" would be preferable.

floehopper:
I don't understand why lots of artefacts are being removed in the last two commits entitled "Regression tests updated for Rwanda...".

pmanrubia:
@floehopper thank you for your help here.

As I understand it, the artefacts are removed because those paths are not exercised:
```ruby
next_node_calculation :responded_with_commonwealth_country do |response|
  Calculators::RegistrationsDataQuery::COMMONWEALTH_COUNTRIES.include?(response)
end

next_node do
  if country_has_no_embassy
    outcome :no_embassy_result
  elsif responded_with_commonwealth_country
    outcome :commonwealth_result
  else
    question :who_has_british_nationality?
  end
end
```

After including Rwanda in the list, the third exit is never exercised.
image

What still I don't get is why no new artefact is generated for the new path.

_...still looking into it..._

pmanrubia:
@floehopper, thank you for your review

1. I have addressed your comment about the commit title in https://github.com/alphagov/smart-answers/pull/2536/commits/6cf798d355b24bbadfb3138c23a242831e1cc1a5, and in https://github.com/alphagov/smart-answers/pull/2536/commits/ec2e0ad993537ec23ee7fc381a7f55e8620b16d3
2. Regarding [your comment](https://github.com/alphagov/smart-answers/pull/2536#issuecomment-220056680) about the deleted artefacts, I have had a look at it, and my understanding is still the same [I explained above](https://github.com/alphagov/smart-answers/pull/2536#issuecomment-220321045). 
3. Regarding how the artefacts are affected after adding Rwanda to the Commonwealth list, it seems to be ok. After regenerating the questions and expected responses in https://github.com/alphagov/smart-answers/pull/2536/commits/9b2dd0ea22027893b7fcd984dd63be95e6612d43 and https://github.com/alphagov/smart-answers/pull/2536/commits/b2bd679ab5c9fa4120c32d687abccdc8ece37bf6, all artefacts but ```Rwanda``` are deleted.

I would say it makes sense. What are your thoughts?

floehopper:
@pmanrubia:

Thanks - that does all makes sense. I think one of the key things to understand is that once Rwanda is added to the list of Commonwealth countries, users reach an outcome page much sooner i.e. they don't get taken to a bunch of question pages. Hence the reason why many artefacts are removed in the last couple of commits. It would be good to make this clear in the commit notes.

Here are some suggestions for improvements to the commit notes:

* Adds Rwanda to register-a-death

This commit adds `rwanda` to the regression test responses for the `which_country?` question in `register-a-death`, because I plan to change the logic and I want to see the effect of this change in the regression tests. I've also updated the responses & expected results accordingly. I plan to update the regression test artefacts in a later commit.

* Add Rwanda to register-a-birth.

Similar to the previous commit, but for `register-a-birth`.

* Generate artefacts for register-a-birth

This commit updates the regression test artefacts in the light of the addition of `rwanda` to the regression test responses for the `which_country?` question in `register-a-death` in an earlier commit. As expected this has resulted in only additions to the artefacts and all the additions are related to `rwanda`.

* Generate artefacts for register-a-death

Similar to the previous commit, but for `register-a-death`.

* Add Rwanda to the Common Wealth country list.

Rwanda is a Common Wealth country [1], so it makes sense to include it within the common list for all the flows. This will result in a change to the externally visible behaviour. I plan to update the regression tests accordingly in subsequent commits.

[1]: http://thecommonwealth.org/our-member-countries/rwanda

* Regenerate questions & expected responses: register-a-birth

Artefacts need regeneration because the next node calculation depends on the `Calculators::RegistrationsDataQuery::COMMONWEALTH_COUNTRIES` and we have just added `rwanda` to this list, thus changing the behaviour.

* Regenerate questions & expected responses: register-a-death 

Similar to the previous commit, but for `register-a-death`.

* Regenerate artefacts for register-a-birth

Now when a user responds with `rwanda` to the `which_country?` question, they are taken straight to an outcome page and not to a series of question pages. This explains why there is a single addition to the artefacts for the `rwanda` version of the outcome page, and a bunch of removals for the question pages which used to follow.

* Regenerate artefacts for register-a-death

Similar to the previous commit, but for `register-a-death`.

floehopper:
Thinking about this a bit more, I wonder how valuable it is to keep `rwanda` in the responses for the `which_country?` question for the regression tests, given that it looks as if we already have `australia` in there to test the Commonwealth country path.

I think I might be inclined to remove it, although I realise that would mean there aren't any tests that require the presence of `rwanda` in the list of Commonwealth countries. I think that's already true for a bunch of other Commonwealth countries.

As a compromise, I might still add `rwanda` to the regression tests early in the PR to demonstrate the change in behaviour, but then remove it again before the end of the PR.

@chrisroos: Do you have any thoughts on this?

pmanrubia:
@floehopper: thank you for your review and for the suggested improvements on the commit notes. This was a spike, but if we finally decide to go forward with it and create a PR, I will for sure consider your notes suggestions.

I’ve just got feedback from the Content team, and both flows should only show the Commonwealth outcome for the following countries: Canada, South Africa, Ireland, New Zealand or Australia

Once [we have amended both flows][2], and the dependency with the Commonwealth list has been removed we will be able to include Rwanda in the Commonwealth list [as intended per this conversation in Spike][1].

I am closing this spike as it seems that it served its purpose.

[1]: https://github.com/alphagov/smart-answers/pull/2536
[2]: https://trello.com/c/WWExcqcM/157-register-a-birth-and-register-a-death-remove-commonwealth-list-check

PR #2542, opened 2016-05-19
Benefit cap calculator: Add a second flow to show upcoming changes to the benefit cap

leenagupte:
Supersedes  #2481
Paired with @ikennaokpala 

Trello Story: https://trello.com/c/IVrsJIfM/34-22-april-requires-fix-but-hold-deployment-benefit-cap-calculator-2016-2017-changes

## Motivation

## Factcheck
[Benefit Cap Calculator](https://smart-answers-pr-2481.herokuapp.com/benefit-cap-calculator/y/)

## Expected Changes
[URL on GOV.UK](https://gov.uk/benefit-cap-calculator/y/)
* Add a first question to beginning of the flow to ask users which benefit cap they want to view.
* Add Carer's Allowance and Guardians Allowance to the list of exempt benefits in the autumn 2016 flow
* Remove Carer's Allowance and Guardians Allowance from the list of benefits to check in the autumn 2016 flow.

### Before

screen shot 2016-04-21 at 13 34 08

### After

screen shot 2016-05-06 at 16 30 12

## Removing the future flow (Autumn 2016)
Sometime in autumn 2016, the new benefit caps will be implemented. At this point, the future flow will become the default flow and the old default flow will need to be removed.

We have (as much as possible) tried to implement these changes to make removing a flow as easy as possible. As it currently stands, these are the changes that will need to be made in Autumn 2016:

1. Remove the first question.
2. Remove the configuration for "default" from benefit_cap_data.yml
3. Rename "future" to "default in benefit_cap_data.yml
4. Remove all references to chosen_cap in the benefit-cap-calculator flow
5. Questions that have a _future version (i.e. is duplicated):
    - Remove the original, and rename _future to the same name as the original. 
    - Consolidate the templates with the current versions.
    - Questions affected:
       * single_couple_lone_parent_future?
       * receiving_non_exemption_benefits_future?
6. Outcomes that have a _future version (i.e. is duplicated):
    - Consolidate the templates with the current versions. 
    - Outcomes affected:
        * outcome_not_affected_future
        * outcome_not_affected_no_housing_benefit_future
        * outcome_not_affected_exemptions_future
7. Outcomes that have National and London future versions:
    - Outcomes affected:
        * outcome_not_affected_less_than_cap_future_london
        * outcome_not_affected_less_than_cap_future_national
        * outcome_affected_greater_than_cap_future_london
        * outcome_affected_greater_than_cap_future_national
8. Remove the default flow from the integration tests.
9. Rename the future flow to default flow in the integration tests.
10. Remove the first question from the integration tests.
11. Regenerate all test artefacts for the regression and integration tests.
12. Remove the version parameter from the functions in benefit_cap_calculator_configuration. There will only be one version (future renamed to default) of the configuration, so it's no longer needed.
13. Update benefit_cap_calculator_configuration_test

Tests affected by the changes:
* test/regression/smart_answers_regression_test.rb
* test/integration/smart_answer_flows/benefit_cap_calculator_test.rb
* test/unit/smart_answer_flows/benefit_cap_calculator_flow_test.rb 
* test/unit/calculators/benefit_cap_calculator_configuration_test.rb


chrisroos:
It looks as though benefit-cap-calculator-responses-and-expected-results.yml might need to be updated. I ran `rails r script/generate-responses-and-expected-results-for-smart-answer.rb benefit-cap-calculator` at the HEAD of this branch and it results in changes to the aforementioned file.

Running the regression tests with this updated version of the responses-and-expected-results file then fails because not all nodes are being exercised:

```
SmartAnswersRegressionTest#test_: Smart Answer: benefit-cap-calculator should ensure all nodes are being exercised.  [test/regression/smart_answers_regression_test.rb:101]:
Not all nodes are being exercised: [:outcome_affected_greater_than_cap_future_london, :outcome_not_affected_less_than_cap_future_london].
Expected: true
  Actual: false
```

This suggests that the input data in benefit-cap-calculator-questions-and-responses.yml isn't exercising all paths of the flow. Let me know if you need a hand working out what data needs to be changed/added.

chrisroos:
I've noticed that test/data/benefit-cap-calculator-files.yml doesn't include lib/data/benefit_cap_data.yml or lib/smart_answer/calculators/benefit_cap_calculator_configuration.rb. You should add these so that changes to either of these files cause the regression tests to run.

You can add them using:

```
rails r script/generate-checksums-for-smart-answer.rb \
benefit-cap-calculator \
lib/data/benefit_cap_data.yml \
lib/smart_answer/calculators/benefit_cap_calculator_configuration.rb
```

These should've been added in PR #2452 but I missed it when reviewing.

ikennaokpala:
@chrisroos thanks for your observations.. I am looking into updating files you have mentioned.

test/integration/smart_answer_flows/benefit_cap_calculator_test.rb
```diff
@@ -11,6 +11,12 @@ class BenefitCapCalculatorTest < ActiveSupport::TestCase
   end
 
   context "Benefit cap calculator" do
+    setup do
+      WebMock.stub_request(:get, "#{Plek.new.find('imminence')}/areas/IG6%202BA.json").
+        to_return(body: File.open(fixture_file('imminence/london.json')))
+      WebMock.stub_request(:get, "#{Plek.new.find('imminence')}/areas/B1%201PW.json").
+        to_return(body: File.open(fixture_file('imminence/national.json')))
```
chrisroos: I think you should be using `imminence_has_areas_for_postcode` here instead of using `WebMock`. Unless you've discounted it; in which case you should mention why in the commit note.

You'll need something like:

```
require 'gds_api/test_helpers/imminence'

include GdsApi::TestHelpers::Imminence

imminence_has_areas_for_postcode("IG6%202BA",  [{ slug: 'london', country_name: 'England' }])
imminence_has_areas_for_postcode("B1%201PW",   [{ slug: "birmingham-city-council", country_name: 'England' }])
```

test/regression/smart_answers_regression_test.rb
```diff
@@ -66,6 +66,10 @@ def teardown_hooks_installed?
         next if self.class.setup_has_run? && !self.class.teardown_hooks_installed?
         stub_content_api_default_artefact
         WebMock.stub_request(:get, WorkingDays::BANK_HOLIDAYS_URL).to_return(body: File.open(fixture_file('bank_holidays.json')))
+        WebMock.stub_request(:get, "#{Plek.new.find('imminence')}/areas/IG6%202BA.json").
+          to_return(body: File.open(fixture_file('imminence/london.json')))
+        WebMock.stub_request(:get, "#{Plek.new.find('imminence')}/areas/B1%201PW.json").
+          to_return(body: File.open(fixture_file('imminence/national.json')))
```
chrisroos: I don't think you need these two lines. The `imminence_has_areas_for_postcode("B1%201PW",   [{ slug: "birmingham-city-council", country_name: 'England' }])` line below this seems to do what you need.

test/regression/smart_answers_regression_test.rb
```diff
@@ -66,6 +66,10 @@ def teardown_hooks_installed?
         next if self.class.setup_has_run? && !self.class.teardown_hooks_installed?
         stub_content_api_default_artefact
         WebMock.stub_request(:get, WorkingDays::BANK_HOLIDAYS_URL).to_return(body: File.open(fixture_file('bank_holidays.json')))
+        WebMock.stub_request(:get, "#{Plek.new.find('imminence')}/areas/IG6%202BA.json").
+          to_return(body: File.open(fixture_file('imminence/london.json')))
```
chrisroos: I think these two lines should be replaced with something like `imminence_has_areas_for_postcode("IG6%202BA",  [{ slug: 'london', country_name: 'England' }])`. See the other uses of `imminence_has_areas_for_postcode` a few lines below.

test/fixtures/imminence/london.json
```diff
@@ -0,0 +1,69 @@
+{
```
chrisroos: I believe this file can be deleted once you switch to using `imminence_has_areas_for_postcode`.

test/fixtures/imminence/national.json
```diff
@@ -0,0 +1,51 @@
+{
```
chrisroos: I believe this file can be deleted once you switch to using `imminence_has_areas_for_postcode`.

lib/smart_answer_flows/benefit-cap-calculator.rb
```diff
@@ -108,7 +108,7 @@ def define
 
         next_node do |response|
           if response == "none"
-            outcome :outcome_not_affected
+            outcome :housing_benefit_amount?
```
chrisroos: This should be `question` instead of `outcome`.

lib/smart_answer_flows/benefit-cap-calculator.rb
```diff
@@ -127,7 +127,7 @@ def define
 
         next_node do |response|
           if response == "none"
-            outcome :outcome_not_affected_future
+            outcome :housing_benefit_amount?
```
chrisroos: This should be `question` instead of `outcome`.

test/unit/calculators/benefit_cap_calculator_configuration_test.rb
```diff
@@ -136,6 +137,49 @@ class BenefitCapCalculatorConfigurationTest < ActiveSupport::TestCase
...
+                code: 200,
+                results: [
+                  {
+                    slug: "birmingham-city-council",
+                    country_name: "England"
+                  }
+                ]
+              )
+            )
+          end
```
chrisroos: I think you should probably use `imminence_has_areas_for_postcode` instead of stubbing `GdsApi::Imminence`. Unless you've decided against doing that, in which case you should mention it in the commit note.

You'll need to use something like:

```
require 'gds_api/test_helpers/imminence'

include GdsApi::TestHelpers::Imminence

imminence_has_areas_for_postcode("B1%201PW", [{slug: 'birmingham-city-council', country_name: 'England'}])
```

test/unit/calculators/benefit_cap_calculator_configuration_test.rb
```diff
@@ -136,6 +137,49 @@ class BenefitCapCalculatorConfigurationTest < ActiveSupport::TestCase
...
+                    country_name: "England"
+                  },
+                  {
+                    slug: "london",
+                    country_name: "England"
+                  }
+                ]
+              )
+            )
+          end
```
chrisroos: I think you should probably use `imminence_has_areas_for_postcode` instead of stubbing `GdsApi::Imminence`. Unless you've decided against doing that, in which case you should mention it in the commit note.

You'll need to use something like:

```
require 'gds_api/test_helpers/imminence'

include GdsApi::TestHelpers::Imminence

imminence_has_areas_for_postcode("WC2B%206SE", [{slug: 'london', country_name: 'England'}])
```

test/unit/calculators/benefit_cap_calculator_configuration_test.rb
```diff
@@ -1,4 +1,5 @@
 require_relative "../../test_helper"
+require 'gds_api/test_helpers/content_api'
```
chrisroos: Is this necessary?

chrisroos:
If at all possible; I'd love to see this branch rewritten so that each commit passes all the tests, including the regression tests. It's very hard to understand the effect of all the changes in its current state. I actually don't think this would be as much work as it might sound (hours rather than days, I suspect). I'd happily do this rewrite in another branch if it'd serve as a useful demonstration of what I mean; although I won't be able to do it for another day or two. 

I feel as though we haven't necessarily done a very good job of communicating the purpose of the regression tests and associated files. We added them to give us the confidence to make changes to the system despite the lack of a good, more traditional, test suite. Being able to use them in this way relies on us being able to describe the changes to the regression test artefacts. We should be aiming to update the regression test artefacts in every commit so that the effect of the change can be clearly observed and described. The only time we should be batching up changes to the regression test artefacts is when they take such a long time to run that it's not particularly productive to wait for them to run after each commit (marriage-abroad is a good example of this). Even when the regression test artefact updates contain changes from multiple commits, we should still be describing those changes.

In its current state I'm afraid I won't be able to give this a thumbs up without spending a lot more time reviewing it. And I won't be able to do that until tomorrow at the earliest.

@floehopper: Can I hand this over to you to continue reviewing, please?

lib/smart_answer_flows/benefit-cap-calculator.rb
```diff
@@ -108,7 +108,7 @@ def define
 
         next_node do |response|
           if response == "none"
-            outcome :housing_benefit_amount?
+            question :housing_benefit_amount?
```
floehopper: This change seems unrelated to the commit note. I think it belongs in a separate commit.

lib/smart_answer_flows/benefit-cap-calculator.rb
```diff
@@ -127,7 +127,7 @@ def define
...
```
floehopper: This change seems unrelated to the commit note. I think it belongs in a separate commit.

ikennaokpala: I have fixed this up to 091ded2

lib/smart_answer_flows/benefit-cap-calculator.rb
ikennaokpala: commit note updated

lib/smart_answer_flows/benefit-cap-calculator.rb
floehopper: 👍

lib/smart_answer_flows/benefit-cap-calculator.rb
floehopper: @ikennaokpala has fixed this in 091ded26f5e7f9190ac09b755d1ee8769beea61f.

test/regression/smart_answers_regression_test.rb
floehopper: @ikennaokpala has addressed this in a new commit titled "Reuse imminence_has_areas_for_postcode in place of web mock" (174d0813933904f23d2e4bfe7852fbc5da36b58d).

test/regression/smart_answers_regression_test.rb
floehopper: @ikennaokpala has addressed this in a new commit titled "Reuse imminence_has_areas_for_postcode in place of web mock" (174d0813933904f23d2e4bfe7852fbc5da36b58d).

test/integration/smart_answer_flows/benefit_cap_calculator_test.rb
floehopper: @ikennaokpala has addressed this in a new commit titled "Reuse imminence_has_areas_for_postcode in place of web mock" (174d0813933904f23d2e4bfe7852fbc5da36b58d).

floehopper: @ikennaokpala has addressed this in a new commit titled "Reuse imminence_has_areas_for_postcode in place of web mock" (174d0813933904f23d2e4bfe7852fbc5da36b58d).

floehopper:
@pmanrubia: Following on from our Slack conversation, here's a rough sketch of the series of commits I might write. I hope it makes some kind of sense. It's pretty hard to come up with an accurate set of commits without actually doing the work, so I may well have missed some things. However, the main idea is to keep the changes as atomic/focussed as possible with all the tests passing after each commit. In places where I know the regression test files are going to have to change, I've made a real effort to keep the changes as small as possible in order to make it easier to explain the changes to the regression test files. Another heuristic that I apply is not to make changes until it becomes obvious that I need them i.e. don't make a bunch of changes early on just because we _think_ we're going to need them. In this case I'd probably be inclined to start a new branch from scratch and pull in changes (using `git cherry-pick` as appropriate) from the old branch as necessary, rather than using an interactive rebase, but it's up to you how you'd prefer to do it.

@chrisroos: If you have time, would you be able to have a look through this sketch to see whether it makes sense to you?

## Suggested commits

1. Add the new first multiple choice question and its template with only the pre-autumn-2016 option; whatever the response always go to the original Q1, i.e. receive_housing_benefit?; update questions & responses to add the new question with the one response; update responses & expected results based on this change; update regression test artefacts; see whether we can easily explain the changes (mostly they should just have an extra path segment for the new question/response); if we can then update checksums (ref: 60303ff972ef46c468dd4c57f0ee57a1ee0a2fb0)
2. Add the extra option to the new first question, but still always go to the original Q1 (as above); add the response to the questions & responses file; update responses & expected results based on this change; update regression test artefacts; see whether we can easily explain the changes (they should be a duplicate set); if we can then update checksums
3. Move data benefit_cap_data.yml under a pre-autumn-2016 namespace, but do not add post-autumn-2016 namespace data; also change BenefitCapCalculatorConfiguration#data to cope with this change. (ref: 5fc48d741d6d96a6f59f28473ae3db184806694e)
4. Change BenefitCapCalculatorConfiguration to allow multiple namespaces, but still defaulting to the pre-autumn-2016 namespace (ref: d7c68ef1df02fe19180356c8310dd953ff26f6cf & 3f8186f4cb9ef2a4a8e717a4906b378d2cee1a1a)
5. In the new first question, use a state variable to store the selected namespace, i.e. hard-coded to pre-autumn-2016; use this state variable to call the methods on the BenefitCapCalculatorConfiguration.
6. Use the response to the new first question to store the correct namespace; duplicate the pre-autumn-2016 namespace data in benefit_cap_data.yml into a new post-autumn-2016 namespace; do not change the values to the post-autumn-2016 values yet; thus the behaviour should be unchanged.
7. Introduce an all_questions method to BenefitCapCalculatorConfiguration; and use it in the config.questions loop (labelled #Q5a-o); this will become important when the list of questions in the two namespaces differs, because we always need to define *all* questions, whether they are used or not (ref: d43f6c0521684f332c22ce5bd6b2aa24b4e46cc4)
8. Duplicate Q4 (and its template) and have Q3's next_node block route to the version of Q4 for the relevant namespace; hard-code the value passed to config.benefits in each of the question definitions; update questions & responses to include the extra question and duplicate the responses of the original; update responses & expected results accordingly; update artefacts; see whether we can easily explain the changes; I don't think anything should change at this point, but I might be wrong; if we can then update checksums (ref: 73243a6a40ab9e4ded04aeac0fdfe3913135279b)
9. Duplicate Q6 (and its template) and have Q5p's next_node block route to the one for the relevant namespace; hard-code the value passed to config.weekly_benefit_caps in each of the question definitions; update questions & responses to include the extra question and duplicate the responses of the original; update responses & expected results accordingly; update artefacts; see whether we can easily explain the changes; I don't think anything should change at this point, but I might be wrong; if we can then update checksums (ref: 73243a6a40ab9e4ded04aeac0fdfe3913135279b)
10. In the post-autumn-2016 data in benefit_cap_data.yml, move the two relevant benefits ("Carer's Allowance" & "Guardian's Allowance") from the benefits list to the exempt_benefits list; update the questions & responses to include these options for the response which includes all the options; update responses & expected results; update artefacts; see whether we can explain the differences; if we can then update checksums; if the diff is big, we could split this into two commits, one per benefit (ref: 5fc48d741d6d96a6f59f28473ae3db184806694e)
11. Introduce a new postcode_question after the post-autumn-2016 version of Q6; always route from this Q6 to the new postcode question; move the benefit_cap calculation from Q6 to the new postcode_question and use that to route to the appropriate outcome as Q6 used to do; do not use the postcode question response for now; update questions & responses to include the new question and a single postcode response; update responses & expected results; update artefacts; see whether we can explain all the changes; if we can then update checksums; I'd imagine all the paths of the post-autumn-2016 artefacts after Q6 would have an extra path segment for the postcode (ref: 9f02fcb0c847407fdc035f25cae60b914b548323)
12. Update the weekly benefit caps methods on BenefitCapCalculatorConfiguration to accept a region argument and move data for both pre-autumn-2016 & post-autumn-2016 under a default "national" namespace; call the relevant benefit caps methods from within the flow with a hard-coded "national" region (ref: 5e9ed447c30cc54cdb01e99ab7e4d9151276d904)
14. Introduce a way to lookup the region based on postcode; lookup the region in the postcode question and use it to obtain the relevant weekly benefit caps; copy the "national" weekly benefits cap data for post-autumn-2016 to a "london" region section in benefit_cap_data.yml; there should be no need to update regression test files, because weekly benefit caps should still all be the same whatever region is (ref: 9f02fcb0c847407fdc035f25cae60b914b548323)
15. Add a London postcode as an extra response to questions & responses; update responses & expected results; update artefacts; see whether we can explain the differences; if we can then update checksums; I'd expect just additions for the new paths
16. Update the "london" data for weekly benefit caps in the post-autumn-2016 namespace; update responses & expected results (if necessary); update artefacts; see whether we can explain the differences; if we can then update checksums; I would expect some differences; it might be worthwhile tweaking the earlier benefit amounts in questions & responses to highlight the change in this commit (ref: 5fc48d741d6d96a6f59f28473ae3db184806694e)
17. For each outcome, see whether we can parameterize the ERB template rather than duplicating each outcome; ideally one outcome per commit; and updating regression test files as appropriate; if we can't do this then resort to duplicating outcomes as in the original PR; if we do the latter, then have one commit which duplicates the outcome and another commit which makes the relevant changes; that way we can make it clearer what the differences are (ref: 573e7f6b0f63b2bc8a1e1b755364b43b0c7da69e)
18. Remove duplicated blocks from the flow in a series of small focussed commits; none of the regression test files should need to change (ref: df7657c9780a755daf232f103f5921c4c703c532, 352b30816cefaba8e69c414022c295ab883e6e78, 5b2a4e115051ec368265cc8f26b56602522042d8, cc85bc088c0efd57cf89115135dc50031363f049, 4a6f66ef875f85ca34f12fcb55ee417974ca7dc3)

## Notes

1. Keep all tests passing after each commit; I've only mentioned when I think the regression tests might need updating; unit tests & integration tests should be added/removed/changed as necessary within the same commit that the change to the implementation is made.
2. I think it might be better to lookup weekly benefit caps by postcode not region, i.e. to keep the postcode lookup internal to BenefitCapCalculatorConfiguration, unless we definitely need the region in the flow.
3. Going through the code in this much detail has reminded me that the [Refactor Benefit Cap Calculator pull request][1] ended up not being as useful as we hoped. I think the current PR has been influenced by the presence of that code and has ended up not being as simple as it could be. However, I'm happy that addressing that concern is out-of-scope for the current PR.

[1]: https://github.com/alphagov/smart-answers/pull/2452

chrisroos:
It looks as though my comments about the responses-and-expected-results file being out of date (see https://github.com/alphagov/smart-answers/pull/2542#issuecomment-220920226) haven't been addressed. I'm going to investigate.

chrisroos:
Thanks for outlining the structure of commits, @floehopper. They all look good to me.

In the hope that it allows us to get this deployed sooner rather than later, I'm planning to create a new branch to demonstrate how I would've approached these changes. I imagine it's going to end up part way between what we have already and what you've outlined in your [comment](https://github.com/alphagov/smart-answers/pull/2542#issuecomment-221055997).

pmanrubia:
Superseded by #2547

PR #2545, opened 2016-05-23
Refactor calculate-married-couples-allowance and fix 4 bugs (with feedback addressed)

floehopper:
## Description

Supersedes #2529.

This started of as an investigation of [this Trello card][1] (originally an old GitHub issue) relating to two large comments in the `calculate-married-couples-allowance` code. That card talks about separating the policy rules for the Married Couples Allowance and the Personal Allowance. While I've made a little progress on that front, by separating out the data relating to the Personal Allowance, this is by no means finished.

During this investigation, I found understanding the code very hard and so I decided to refactor the flow towards the [new style][2] to make things clearer. As I went on, I identified a number of bugs which are outlined below. The most significant bug is the first one - the other three fix behaviour for tax years prior to the current one (2015-2016).

### Income limit bug

Trello cards: https://trello.com/c/I5X3nmow & https://trello.com/c/Zse6DH8v

Prior to this PR the flow was always using a hard-coded income limit of £27,000 to decide whether or not to ask follow-up questions about pension & gift aid contributions. However, the [income limit was increased to £27,700 from the 2015-2016 tax year onwards][3].

Note that this is a relatively subtle bug, because the actual calculations of the allowance does take the correct income limit into account.

### Age vs date-of-birth bug

Trello card: https://trello.com/c/QRHxaXDi

The age-related Personal Allowance was only supposed to be based on age up until the end of the 2012-2013 tax year. [Since then it should have been based on the person's date of birth][6], but the latter was never implemented in the code. This PR chooses the correct mechanism for determining the age-related Personal Allowance depending on whether today's date is before or after the end of the 2012-2013 tax year.

Since the 2015-2016 tax year started, the Personal Allowance has been the same for all ages/dates-of-birth and so it's not possible to see a change in behaviour unless we setup a Heroku instance with the date set in the past. I'm not convinced this is worthwhile, but I do think it's worth fixing the code so that if/when we remove the age-based mechanism, it's more obvious what was going on.

### Age calculation bug

Trello card: https://trello.com/c/w6gmL50V

Someone's age on a given date depends on whether or not they've had their birthday in the current year. Also [someone who's birthday falls on 29th Feb in a leap year, is legally considered to have their birthday on 1st March in a non-leap year][4]. Prior to this commit, neither of these were being taken into account.

Given the fix for the "Age vs date-of-birth bug" (above) and the fact that we are already well past 2013-04-05 which is when the changeover to use date-of-birth was supposed to happen, this fix has no externally-visible effect. We could setup a Heroku instance with the date set in the 2012-2013 tax year, but I'm not convinced this is worthwhile. However, I do think it's worth fixing the code so that if/when we simplify the code, it's more obvious what was going on.

### Age at end of current tax year bug

Trello card: https://trello.com/c/qvduSzB6

Prior to this PR the age was being calculated on *5th April in the current calendar year*. I'm pretty confident this is incorrect and that the age should be calculated *at the end of the current tax year*. I haven't been able to find any direct (GOV.UK) evidence that this is correct, but it seems logical and I have found [this indirect evidence][5].

Given the fix for the "Age vs date-of-birth bug" (above) and the fact that we are already well past 2013-04-05 which is when the changeover to use date-of-birth was supposed to happen, this fix has no externally-visible effect. We could setup a Heroku instance with the date set in the 2012-2013 tax year, but I'm not convinced this is worthwhile. However, I do think it's worth fixing the code so that if/when we simplify the code, it's more obvious what was going on.

## External changes

Much of the work in this PR is refactoring or fixing of caclulations for previous tax years. The only externally-visible change is the fix for the "income limit" bug which can be tested as follows:

#### Before

* Visit https://www.gov.uk/calculate-married-couples-allowance/y/yes/yes/1950-01-01/27001.0
* See you are on "Are you paying into a pension?" question page.

#### After

* Visit https://smart-answers-pr-2529.herokuapp.com/calculate-married-couples-allowance/y/yes/yes/1950-01-01/27001.0
* See you are on "The husband qualifies for Married Couple's Allowance" outcome page.

## Possible further work

* Separate the policy rules for the Married Couples Allowance and the Personal Allowance
* Move the tests for `calculate-married-couples-allowance` towards the [new-style testing approach][7].

[1]: https://trello.com/c/XLE6L7s8
[2]: https://github.com/alphagov/smart-answers/blob/master/doc/refactoring.md
[3]: https://www.gov.uk/government/publications/rates-and-allowances-income-tax/income-tax-rates-and-allowances-current-and-past#personal-allowances-for-people-born-before-6-april-1948
[4]: https://en.wikipedia.org/wiki/February_29#Legal_status
[5]: http://www.pwc.co.uk/assets/pdf/budget-2012-tax-datacard-2012.pdf
[6]: https://www.gov.uk/government/publications/rates-and-allowances-income-tax/income-tax-rates-and-allowances-current-and-past#personal-allowances-for-people-born-before-6-april-1948
[7]: https://github.com/alphagov/smart-answers/blob/master/doc/new-style-testing.md


floehopper:
This has been reviewed by @chrisroos and is ready to merge from a technical point-of-view.

floehopper:
I've added the "Waiting on factcheck" label while I solicit feedback from the Content Team.

floehopper:
Rebased against master to resolve conflicts and force-pushed. Ready for review.

chrisroos:
I'm adding the LGTM label so that I can see at a glance that this has been reviewed and given the OK.

floehopper:
I've just rebased this against master and force-pushed to keep it up-to-date while I wait for feedback from the Content Team.

floehopper:
I've just rebased this against master (again) and force-pushed to keep it up-to-date while I wait for feedback from the Content Team.

floehopper:
I've just heard back from @raykhan66:

> The department have confirmed that it should be £27,700 (not £27,000) that triggers the calculator questions about pensions and gift aid contributions, for the current financial year.

So I'm going to get this merged.

PR #2546, opened 2016-05-24
Move DateHelper into the SmartAnswer namespace

floehopper:
This brings the module into line with the other classes & modules defined in files in the `lib/smart_answers` directory. It also is more idiomatic Ruby/Rails in that the directory structure and namespace structure should match each other.

It also feels important to distinguish this helper from `ActionView::Helpers::DateHelper` which is used elsewhere in the codebase.

Since this commit includes indentation changes, it's easier to see what's going on in the diff by using the `--ignore-all-space` option.

I've also updated the regression test checksums for `maternity-paternity-calculator` since all the tests are passing.


floehopper:
Rebased against master to resolve conflicts and force-pushed. Ready for review.

chrisroos:

floehopper:
@chrisroos: Thanks.

PR #2547, opened 2016-05-24
Add Autumn 2016 changes to benefit-cap-calculator

chrisroos:
This is a reworked version of PR #2542. I've made use of pretty much everything in that branch but rewritten the history (mostly following @floehopper's suggestions in https://github.com/alphagov/smart-answers/pull/2542#issuecomment-221055997) so that the changes are built up incrementally. This makes it possible to use the changes to the regression test artefacts to verify the corresponding changes in behaviour.

I used `diff` to compare the test/artefacts/benefit-cap-calculator directory in this branch to the one in PR #2452. Apart from the use of a different London postcode in this branch (WC2B 6SE instead of IG6 2BA), the artefacts appear to be identical. This gives me confidence that I haven't missed anything while creating this branch.

@floehopper: Can you review this branch, please? I'm afraid that I'm unlikely to be able to spend any more time on this before we want it deployed (assuming we do want to deploy this version) so can I ask you to make any necessary changes, please?


test/data/benefit-cap-calculator-files.yml
```diff
@@ -30,3 +30,5 @@ lib/smart_answer_flows/benefit-cap-calculator/questions/widowed_mother_amount.go
 lib/smart_answer_flows/benefit-cap-calculator/questions/widowed_parent_amount.govspeak.erb: 7355088c4cbc00513807d90b1930ff5c
 lib/smart_answer_flows/benefit-cap-calculator/questions/widows_aged_amount.govspeak.erb: 56ddf0a2aed4310db694dafeaea8e1db
 lib/smart_answer_flows/benefit-cap-calculator/questions/working_tax_credit.govspeak.erb: d96c2ff97c1bb068389fbefc25b9ed1a
+lib/data/benefit_cap_data.yml: 91b7e907643ed0403884a0b71544e25f
+lib/smart_answer/calculators/benefit_cap_calculator_configuration.rb: 6114fb32efba47d5686a3e53595745c8
```
floehopper: > These two additional files should've been added to the checksum data in PR

The PR number is missing.

lib/smart_answer_flows/benefit-cap-calculator.rb
```diff
@@ -8,6 +8,16 @@ def define
...
+        option :future
+
+        next_node do
+          question :receive_housing_benefit?
+        end
+      end
+
       # Q1
       multiple_choice :receive_housing_benefit? do
         option :yes
```
floehopper: > TODO: Consider adding only single response to the first question for now.

This TODO should be removed from the commit note.

This commit is quite big, so it might be worth trying the suggestion in the TODO.

lib/smart_answer_flows/benefit-cap-calculator.rb
```diff
@@ -127,7 +127,7 @@ def define
 
         next_node do |response|
           if response == "none"
-            outcome :outcome_not_affected_future
+            question :housing_benefit_amount?
```
floehopper: Given that this fix affects both the "default" and "future" flows, it might've been simpler to apply this change before introducing the "future" flow functionality.

floehopper: While I understand the changes to `benefit-cap-calculator-responses-and-expected-results.yml`, I'm not 100% sure I understand the changes to the regression test artefacts. Ideally I think we'd try to improve the explanation in the commit note.

lib/smart_answer_flows/benefit-cap-calculator/questions/enter_postcode.govspeak.erb
```diff
@@ -1,8 +1,12 @@
 <% content_for :title do %>
-  Enter your postcode:
+  Enter your postcode
 
 <% end %>
 
+<% content_for :body do %>
+  eg SW1A 2AA
+<% end %>
+
```
floehopper: Ideally the changes in this commit would've been split up a bit more with the corresponding changes to the regression test artefacts included in each commit.

test/artefacts/benefit-cap-calculator/default.html
```diff
@@ -78,13 +78,13 @@ <h3 class="previous-answers-title">
...
-  <td class="previous-question-title">The benefit cap is changing.</td>
+  <td class="previous-question-title">Which period do you want to check for?</td>
     <td class="previous-question-body">
-    The current benefit cap</td>
+    Now</td>
 
   <td class="link-right">
     <a href="/benefit-cap-calculator/y?previous_response=default">
-      Change<span class="visuallyhidden"> answer to "The benefit cap is changing."</span>
+      Change<span class="visuallyhidden"> answer to "Which period do you want to check for?"</span>
```
floehopper: Ideally the changes to the regression test artefacts in this commit would've been included in a series of commits alongside the changes which caused them. See my comment on the previous commit.

test/unit/calculators/benefit_cap_calculator_configuration_test.rb
```diff
@@ -127,7 +127,7 @@ class BenefitCapCalculatorConfigurationTest < ActiveSupport::TestCase
           end
         end
         context "merge questions" do
-          should "get all benefit questions from multiple configuration sets" do
+          should "get all unique benefit questions from multiple configuration sets" do
```
floehopper: Ideally this change would've been incorporated into 70ba9b321c03bc4da0faf0cac9d53bea80c7675b where the test was originally added.

lib/smart_answer/calculators/benefit_cap_calculator_configuration.rb
```diff
@@ -42,10 +42,10 @@ def all_questions
     end
 
     def region(postcode)
-      is_london?(postcode) ? :london : :national
+      london?(postcode) ? :london : :national
     end
 
-    def is_london?(postcode)
+    def london?(postcode)
```
floehopper: Ideally this would've been included in 226d3496da925c4bdfa6f89f8911ee40a3a085fc where the method was originally added. Also the commit note could explain that removing the `is` prefix makes the method name more idiomatic Ruby.

floehopper: I found it slightly surprising that no changes to the flow were required in this commit. That appears to be because we use `region == :london` in the flow, i.e. the `#london?` method is only used within this class. As such it might be clearer if it was made private. Alternatively we could use the method in the conditionals within the flow. However, presumably that would mean extra requests to the external Imminence endpoint.

floehopper:
I've just finished reviewing all the commits in this PR. I found it much easier to understand what was changing at each stage, particularly why the regression test artefacts were changing.

There are still two commits where I want to double-check the changes to the regression tests:

* fab1112db9ba9bec743f754834120774b65c3a7c
* 0d89a967be4039c49100ca3c88217d4832d33495

Also I want to compare the final state of the regression test artefacts between this PR and #2542.

floehopper:
It's worth noting that the commit notes in this PR sometimes refer to commits in #2542 for further details. While I think this is great for comparing the approach in this PR with that in #2542, it's not completely obvious to me that the commit notes are in the most useful form for merging to `master`. However, I don't think this is important enough to prevent this PR from being merged.

floehopper:
I'm now about 99% sure I understand the changes in the regression test artefacts.

I've also now just compared the final state of the regression test artefacts between this PR and #2542 and the only differences are related to the different London postcode being used.

Given the latter, I'm confident this could be merged as it stands.

floehopper:
I've added a new commit to clarify that the Benefit Cap Calculator cares whether you are inside/outside *Greater* London and not just London. This came to light when @leenagupte asked why the postcode in the regression test had been changed from IG6 2BA in #2542 to WC2B 6SE in this PR.

floehopper:
This now looks good enough to merge to me.

lib/data/benefit_cap_data.yml
```diff
@@ -65,3 +65,69 @@ default:
...
+      description: "Widowed Mother's Allowance"
+    widowed_parent:
+      question: :widowed_parent_amount?
+      description: "Widowed Parent's Allowance"
+    widow_pension:
+      question: :widow_pension_amount?
+      description: "Widow's Pension"
+    widows_aged:
+      question: :widows_aged_amount?
+      description: "Widow's Pension (age related)"
```
floehopper: This comment is just a test to see whether the commit authorship determines who receives notifications.

test/data/benefit-cap-calculator-files.yml
chrisroos: Sorry. It should've been PR #2452. I think it disappeared when I added newlines to the commit note: the number (along with the preceding hash) were moved to a new line and git treated it as a comment.

lib/smart_answer_flows/benefit-cap-calculator.rb
chrisroos: With hindsight, it probably would've lead to slightly easier to understand commits if I'd split this into two. Having said that, I'm confident that I understand the changes in this large commit so am not too worried about it being combined.

Irrespective, I should've removed the TODO from the commit note. Sorry!

lib/smart_answer_flows/benefit-cap-calculator.rb
chrisroos: > Given that this fix affects both the "default" and "future" flows, it might've been simpler to apply this change before introducing the "future" flow functionality.

I agree. My aim in this branch was to reuse as many commits as possible from PR #2542, which is why I included this one as it was.

lib/smart_answer_flows/benefit-cap-calculator/questions/enter_postcode.govspeak.erb
chrisroos: I agree and did consider splitting it. I decided against it partly due to time available and partly as I wanted to reuse as many commits as possible from PR #2542.

test/artefacts/benefit-cap-calculator/default.html
chrisroos: I agree and did consider making that change. I decided against it partly due to time available and partly as I wanted to reuse as many commits as possible from PR #2542.

test/unit/calculators/benefit_cap_calculator_configuration_test.rb
chrisroos: I agree. I decided against it partly due to time available and partly as I wanted to reuse as many commits as possible from PR #2542.

chrisroos:
> It's worth noting that the commit notes in this PR sometimes refer to commits in #2542 for further details. While I think this is great for comparing the approach in this PR with that in #2542, it's not completely obvious to me that the commit notes are in the most useful form for merging to master. However, I don't think this is important enough to prevent this PR from being merged.

Yeah, the commits definitely aren't as good as they could be. The main reasons for this branch were to sanity check the changes in PR #2542 and to demonstrate an alternative approach that helps make the review process much easier. To that end, I've reused and referenced lots of earlier commits in this branch. If there was more time then I think we would've created yet another branch that takes even more of the original feedback into account before getting it merged.

chrisroos:
> I've added a new commit to clarify that the Benefit Cap Calculator cares whether you are inside/outside Greater London and not just London. This came to light when @leenagupte asked why the postcode in the regression test had been changed from IG6 2BA in #2542 to WC2B 6SE in this PR.

Thanks @floehopper. I'm afraid I'd completely missed the purpose of using a Greater London postcode. I chose to change it to WC2B 6SE as we were already using that elsewhere.

chrisroos:
Thanks for reviewing and merging, @floehopper!

PR #2551, opened 2016-05-26
Update artefacts for marriage-abroad

pmanrubia:
Complements https://github.com/alphagov/smart-answers/pull/2516, by including missing artefacts for marriage-abroad in: https://github.com/alphagov/smart-answers/pull/2516/commits/dfe3298ea55cc0d462f624f90ee92bf270e0542a
See [comment](https://github.com/alphagov/smart-answers/pull/2516#issuecomment-221726904) from @floehopper.

### DIFF
--- a/test/artefacts/marriage-abroad/rwanda/ceremony_country/partner_british/opposite_sex.txt
+++ b/test/artefacts/marriage-abroad/rwanda/ceremony_country/partner_british/opposite_sex.txt
@@ -1,6 +1,6 @@
 Marriage in Rwanda
 
-Contact the [Embassy of Rwanda](/government/publications/foreign-embassies-in-the-uk) to find out about local marriage laws, including what documents you’ll need.
+Contact the relevant local authorities in Rwanda to find out about local marriage laws, including what documents you’ll need.
 
 ^You should [get legal advice](/government/collections/list-of-lawyers) before making any plans.^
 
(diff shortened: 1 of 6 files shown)


floehopper:
As a general rule, I think we should be trying to explain any changes to the regression test artefacts in the relevant commit note. This will help someone reviewing the PR understand whether the changes are all expected changes. Otherwise I think there's a danger we just run the tests and accept all of the changes without really looking at them. I note that there is no explanation in either the [original commit][1] or [the commit in this PR][2].

[1]: dfe3298ea55cc0d462f624f90ee92bf270e0542a
[2]: 06dd399ec3910f15839c3c84e67f555c3934a20f

pmanrubia:
I am not sure if I am getting you correctly. If the changes are small enough I can't see a big benefit in describing the changes in the commit notes. And if they are big then we would need to split them so they are easy to follow.  Could you post an example so I understand you better?

floehopper:
I think the important thing is not to describe the changes, but to explain how you have satisfied yourself that they are all changes that you were expecting because of changes you have made to the flow and/or its templates. I'm not sure I'm explaining myself very well.

Given that we need to fix the build, as long as you are happy that the changes to the artefacts are as you expect then I think it's fine to merge this.

pmanrubia:
Thank you @floehopper for the clarification; I have understood you and you have a point in what you suggest, it will simplify and will add security to our reviews. I will try to follow a similar approach in incoming PRs.

PR #2552, opened 2016-05-26
Update gds-api-adapters from v25.1.0 to v30.6.0 (the latest version)

floehopper:
## Description

I noticed that we were using quite an old version of gds-api-adapters and I thought it would be sensible to get more up-to-date.

Also I'm contemplating pushing some functionality out of the smart-answers app and into gds-api-adapters. Being up-to-date with gds-api-adapters will make this easier.

This PR updates gds-api-adapters from v25.1.0 to v30.6.0 (the latest version). It also updates one of its dependencies, plek, from v1.7.0 to v1.12.0 (the latest version). Strictly speaking it was only necessary to update plek to v1.9.0, but I decided I might as well update it to the latest version as well.

I've run all the regression tests locally and they all pass.

See individual commit notes for more details.

## External changes


chrisroos:

floehopper:
@chrisroos: Thanks for reviewing. Rebasing against master and force-pushing in preparation for merging.

PR #2555, opened 2016-05-31
Stub WorldLocation directly instead of using GdsApi::TestHelpers::Worldwide

floehopper:
## Description

Trello cards:

* https://trello.com/c/Ps4kj7ME
* https://trello.com/c/5Ruw6swd

Previously many of the tests were using methods on `GdsApi::TestHelpers::Worldwide` to stub calls to the Worldwide API. Each of these helper methods set up a bunch of `WebMock` stubs to stub at the `Net::HTTP`/`RestClient` level.

It seems simpler and clearer to stub at the `WorldLocation` level in the `smart-answers` tests and rely on the `gds-api-adapters` tests to verify that the lower level behaviour is correct.

This PR follows on from #2540 and converts most of the remaining occurrences of `GdsApi::TestHelpers::Worldwide` methods. The only places left are:

* `WorldLocationTest` - It doesn't make sense to stub `WorldLocation` in this unit test. However, I think it would make sense to stub `Services.worldwide_api` instead of using `GdsApi::TestHelpers::Worldwide` methods. I did briefly look into doing this, but it wasn't completely straightforward, so I've left it in favour of getting the other changes in this PR merged.

* `SmartAnswersRegressionTest` - I did [some work on this][1] but ran into problems, so I've left it in favour of getting the other changes in this PR merged. However, I have made some changes which should make this easier to do in the future.

## External changes

None. This is just an internal refactoring and no production code has changed.

[1]: https://github.com/alphagov/smart-answers/compare/stub-world-location-directly-in-regression-tests

### DIFF
--- a/test/fixtures/smart_answer_flows/country-and-date-sample.rb
+++ b/test/fixtures/smart_answer_flows/country-and-date-sample.rb
@@ -4,7 +4,7 @@ module SmartAnswer
       name 'country-and-date-sample'
       status :draft
 
-      country_select :which_country_do_you_live_in?, exclude_countries: %w(afghanistan united-kingdom) do
+      country_select :which_country_do_you_live_in? do
         save_input_as :country
         next_node do
           question :what_date_did_you_move_there?
@@ -25,7 +25,7 @@ module SmartAnswer
         end
       end
 
-      country_select :which_country_were_you_born_in?, include_uk: true do
+      country_select :which_country_were_you_born_in? do
         save_input_as :birth_country
         next_node do
           outcome :ok
--- a/test/integration/engine/changing_answer_test.rb
+++ b/test/integration/engine/changing_answer_test.rb
@@ -1,12 +1,9 @@
 require_relative 'engine_test_helper'
-require 'gds_api/test_helpers/worldwide'
 
 class ChangingAnswerTest < EngineIntegrationTest
-  include GdsApi::TestHelpers::Worldwide
-
   with_and_without_javascript do
     should "be able to change country and date answers" do
-      worldwide_api_has_selection_of_locations
+      stub_worldwide_locations(%w(argentina belarus))
 
       visit "/country-and-date-sample/y"
 
@@ -28,10 +25,10 @@ class ChangingAnswerTest < EngineIntegrationTest
         assert page.has_selector? :select, "response", selected: "Belarus"
       end
 
-      select "South Korea", from: "response"
+      select "Argentina", from: "response"
       click_on "Next step"
 
-      assert_current_url "/country-and-date-sample/y/south-korea"
+      assert_current_url "/country-and-date-sample/y/argentina"
 
       select "10", from: "Day"
       select "June", from: "Month"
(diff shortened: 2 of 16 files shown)


chrisroos:
Commit: Clarify that WorldLocation.all doesn't return delegations/missions (91c8ab27881ff06b8186213c7678e1be05584bc6)

I don't completely understand this comment/change. The change suggests that `worldwide_api_has_locations` sets the format to "International delegation" if the slug contains "delegation" or "mission", and to "World location" otherwise. Yet the comment says something slightly different.

chrisroos:
Aside from a query about the message of a single commit, this looks good to me.

Good work finding and making the necessary changes to allow you to use `stub_world_locations` in the country_and_date_questions_test.

floehopper:
> Commit: Clarify that WorldLocation.all doesn't return delegations/missions (91c8ab27881ff06b8186213c7678e1be05584bc6)
>
> I don't completely understand this comment/change. The change suggests that `worldwide_api_has_locations` sets the format to "International delegation" if the slug contains "delegation" or "mission", and to "World location" otherwise. Yet the comment says something slightly different.

Hmm. It is confusing. There's two things going on here. Firstly the thing you said about `worldwide_api_has_locations`, but *also* `WorldLocation.all` then filters out any locations which do not have their format set to "World location". The following are the two relevant lines of code:

* [gds-api-adapters/lib/gds_api/test_helpers/worldwide.rb:100](https://github.com/alphagov/gds-api-adapters/blob/29118ff686ecdb05ea338f6dd788bfa7015d69bd/lib/gds_api/test_helpers/worldwide.rb#L100)
* [smart-answers/app/models/world_location.rb:17](https://github.com/alphagov/smart-answers/blob/c265170b8c94984daa83924bc50b3ec796b2b70d/app/models/world_location.rb#L17)

Does that make sense?

While you could argue the test should only be testing `WorldLocation.all`, I think it also relies on some non-obvious behaviour of `worldwide_api_has_locations`. I'll have a go at re-writing the commit and/or commit note to make this clearer.

floehopper:
Superseded by #2563. Closing.

floehopper:
> Commit: Clarify that WorldLocation.all doesn't return delegations/missions (91c8ab27881ff06b8186213c7678e1be05584bc6)
>
> I don't completely understand this comment/change. The change suggests that `worldwide_api_has_locations` sets the format to "International delegation" if the slug contains "delegation" or "mission", and to "World location" otherwise. Yet the comment says something slightly different.

I've tweaked the commit and expanded the commit note in [this new commit][1] in #2563. I hope that's a bit clearer.

[1]: https://github.com/alphagov/smart-answers/commit/9499bf74a15837dda0892fd66b52a412d2fe0598

chrisroos:
> I've tweaked the commit and expanded the commit note in [this new commit][1] in #2563. I hope that's a bit clearer.

The new commit note seems much clearer to me. Thanks, @floehopper!

[1]: https://github.com/alphagov/smart-answers/commit/9499bf74a15837dda0892fd66b52a412d2fe0598

PR #2561, opened 2016-06-01
Refactor calculate-agricultural-holiday-entitlement to new-style flow

floehopper:
## Description

This moves the flow towards the style described in the [refactoring documentation][1]. I've concentrated on moving logic out of the flow and into the calculator. I have not spent any time trying to improve the code in the calculator (e.g. by extracting domain concepts).

I've added some unit tests for the validation that I've extracted, because this behaviour is not covered in the regression tests. However, otherwise I have not moved the tests closer to how it's described in the [new-style testing documentation][2].

So, although there is more work that could be done on this flow and its associated classes, I think the changes in this PR are useful in their own right and if we can do just this much for all the flows, I think we'll be able to dramatically simplify the DSL.

Note that in doing this work I came across some suspect looking code, but rather than get sucked into sorting that out, I've captured the possible problems in a couple of Trello cards:

* https://trello.com/c/0WNxygY4
* https://trello.com/c/HXgHxbpp

## External changes

None. This is just an internal refactoring.

[1]: https://github.com/alphagov/smart-answers/blob/master/doc/refactoring.md
[2]: https://github.com/alphagov/smart-answers/blob/master/doc/new-style-testing.md

PR #2562, opened 2016-06-01
Refactor additional-commodity-code to new-style flow

floehopper:
## Description

Similar to #2561, this moves the flow towards the style described in the [refactoring documentation][1]. I've concentrated on moving logic out of the flow and into the calculator. This is a fairly simple flow - mostly just a decision tree and so it may not be obvious why I want to do this. However, if we can move all the flows to use this approach, then I'm confident we can significantly simplify the DSL.

## External changes


chrisroos:

floehopper:

PR #2563, opened 2016-06-02
Stub WorldLocation directly instead of using GdsApi::TestHelpers::Worldwide (with feedback addressed)

floehopper:
## Description

Supersedes #2555.

Trello cards:

* https://trello.com/c/Ps4kj7ME
* https://trello.com/c/5Ruw6swd

Previously many of the tests were using methods on `GdsApi::TestHelpers::Worldwide` to stub calls to the Worldwide API. Each of these helper methods set up a bunch of `WebMock` stubs to stub at the `Net::HTTP`/`RestClient` level.

It seems simpler and clearer to stub at the `WorldLocation` level in the `smart-answers` tests and rely on the `gds-api-adapters` tests to verify that the lower level behaviour is correct.

This PR follows on from #2540 and converts most of the remaining occurrences of `GdsApi::TestHelpers::Worldwide` methods. The only places left are:

* `WorldLocationTest` - It doesn't make sense to stub `WorldLocation` in this unit test. However, I think it would make sense to stub `Services.worldwide_api` instead of using `GdsApi::TestHelpers::Worldwide` methods. I did briefly look into doing this, but it wasn't completely straightforward, so I've left it in favour of getting the other changes in this PR merged. I've created a [new Trello card][1] to capture this.

* `SmartAnswersRegressionTest` - I did [some work on this][2] but ran into problems, so I've left it in favour of getting the other changes in this PR merged. However, I have made some changes which should make this easier to do in the future.

## External changes

None. This is just an internal refactoring and no production code has changed.

[1]: https://trello.com/c/M4Y09qFy
[2]: https://github.com/alphagov/smart-answers/compare/stub-world-location-directly-in-regression-tests

### DIFF
--- a/test/fixtures/smart_answer_flows/country-and-date-sample.rb
+++ b/test/fixtures/smart_answer_flows/country-and-date-sample.rb
@@ -4,7 +4,7 @@ module SmartAnswer
       name 'country-and-date-sample'
       status :draft
 
-      country_select :which_country_do_you_live_in?, exclude_countries: %w(afghanistan united-kingdom) do
+      country_select :which_country_do_you_live_in? do
         save_input_as :country
         next_node do
           question :what_date_did_you_move_there?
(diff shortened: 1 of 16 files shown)


floehopper:
Since @chrisroos marked the previous version of this PR (#2555) as LGTM and I have addressed the one bit of feedback, I'm going to merge this.

PR #2548, opened 2016-05-24
Revert handling of template exceptions when generating indexable content

floehopper:
## Description

Trello card: https://trello.com/c/W7EDHYFB

This PR (mostly) reverts [this commit][1]. Note, however,  that I have *not* reverted the introduction of `IndexableContentTest`, because it's still a useful test to make sure we don't introduce any more code which results in raising exceptions.

The reverted commit was originally introduced so that we could generate `indexable_content` for the `outcome_os_other_countries` outcome in `marriage-abroad`. We had to add this special case to handle the potential exception raised by the outcome.

This [later pull request][2] updates the `outcome_os_other_countries` outcome such that it no longer raises an exception, which means that we can simplify the code by reverting the commit mentioned above.

However, in the meantime there have been some other changes which also cause template exceptions to be raised. Initially I was thinking that I might just not bother implementing the idea in this card, but then I realised there's a danger that we may unknowingly keep introducing problematic code thereby reducing the indexable content that we can generate.

To this end, all of the commits in this PR except for the last one fix place where template exceptions were being raised during the generation of indexable content. Once all of these were applied, I was able to do revert the commit as planned.

## External changes

It's likely that the changes in this PR will mean that more indexable content is supplied to Panopticon (and thence Rummager?) for the flows in which template exceptions were previously being raised (`register-a-birth`, `register-a-death` & `marriage-abroad`). However, there's no easy way to test this directly and in any case such changes are very low risk.

[1]: https://github.com/alphagov/smart-answers/commit/50ab9b8d4963530eae7142f230d07ae75d04fb93
[2]: https://github.com/alphagov/smart-answers/pull/2504

PR #2565, opened 2016-06-02
Revert handling of template exceptions when generating indexable content (with feedback addressed)

floehopper:
## Description

Supersedes #2548.

Trello card: https://trello.com/c/W7EDHYFB

This PR (mostly) reverts [this commit][1]. Note, however,  that I have *not* reverted the introduction of `IndexableContentTest`, because it's still a useful test to make sure we don't introduce any more code which results in raising exceptions.

The reverted commit was originally introduced so that we could generate `indexable_content` for the `outcome_os_other_countries` outcome in `marriage-abroad`. We had to add this special case to handle the potential exception raised by the outcome.

This [later pull request][2] updates the `outcome_os_other_countries` outcome such that it no longer raises an exception, which means that we can simplify the code by reverting the commit mentioned above.

However, in the meantime there have been some other changes which also cause template exceptions to be raised. Initially I was thinking that I might just not bother implementing the idea in this card, but then I realised there's a danger that we may unknowingly keep introducing problematic code thereby reducing the indexable content that we can generate.

To this end, all of the commits in this PR except for the last one fix place where template exceptions were being raised during the generation of indexable content. Once all of these were applied, I was able to do revert the commit as planned.

In order to identify the problems that needed fixing I did the following:

1. Reverted the commit
2. Ran the `IndexableContentTest`
3. Fixed the first problem that I encountered
4. Repeat from step 2 until no more problems
5. Moved revert commit to the end of the commits so that the tests pass for every commit in the branch

## External changes

It's likely that the changes in this PR will mean that more indexable content is supplied to Panopticon (and thence Rummager?) for the flows in which template exceptions were previously being raised (`register-a-birth`, `register-a-death` & `marriage-abroad`). However, there's no easy way to test this directly and in any case such changes are very low risk.

[1]: https://github.com/alphagov/smart-answers/commit/50ab9b8d4963530eae7142f230d07ae75d04fb93
[2]: https://github.com/alphagov/smart-answers/pull/2504


floehopper:
@chrisroos: Please can you check that I've addressed your feedback from #2548? Thanks.

chrisroos:
> @chrisroos: Please can you check that I've addressed your feedback from #2548? Thanks.

I've checked the new commits mentioned in https://github.com/alphagov/smart-answers/pull/2548#issuecomment-223311997. Aside from what appears to be a duplicate test ("'use description within ERB tags as to_s'") added in af46c0e9fbb0fcef09b418a0b0d8bc4a51aed4e8, they all look good to me.

floehopper:
> I've checked the new commits mentioned in https://github.com/alphagov/smart-answers/pull/2548#issuecomment-223311997. Aside from what appears to be a duplicate test ("'use description within ERB tags as to_s'") added in af46c0e9fbb0fcef09b418a0b0d8bc4a51aed4e8, they all look good to me.

Well spotted. I'm going to remove the duplicate test, rebase against master, and force-push in preparation for merging.

floehopper:
> Aside from what appears to be a duplicate test ("'use description within ERB tags as to_s'") added in af46c0e9fbb0fcef09b418a0b0d8bc4a51aed4e8, they all look good to me.

The branch has now been rebased and force-pushed. The duplicate test was is no longer in the new version of the first commit: ee5e8e567733780e2c1eff66b577530acc540cc8.

PR #2569, opened 2016-06-03
Refactor calculate-your-child-maintenance to new style flow

floehopper:
## Description

There are still one `calculate` block and a bunch of `precalculate` blocks towards the end of the flow which are horribly tangled up with both calculation and presentational concerns. I couldn't see an easy way to untangle them while retaining the same rounding, number formatting, etc. My suspicion is that there is some unintentional behaviour in here, but at this stage I want to avoid getting sucked into needing any fact-checking for this PR.

I haven't really done any significant work on the calculator itself e.g. extracting domain concepts. Also I haven't made any significant changes to the tests - we'll still need to move them towards the [new-style approach][2] at some point.

## External changes


chrisroos:
__Commit 3a578b5477a892024d8494ea8523be52f5aa8c79 ("Inline rate_type_formatted into outcome template")__

It wasn't immediately obvious to me why you were able to use `humanize` when inlining the `rate_type_formatted` precalculate block. Presumably none of the other values of `rate_type` will be changed by this method?

chrisroos:
Aside from my single comment this all looks great to me, @floehopper!

floehopper:
Improving commit comment based on feedback, rebasing against master, and force-pushing in preparation for merging.

floehopper:
> It wasn't immediately obvious to me why you were able to use `humanize` when inlining the `rate_type_formatted` precalculate block. Presumably none of the other values of `rate_type` will be changed by this method?

I've expanded the commit note to explain my thinking here: cd118ace4062524e3dbed9153b9857387a4b91e2.

chrisroos:
> I've expanded the commit note to explain my thinking here: cd118ac.

Awesome. Thanks, @floehopper.

PR #2570, opened 2016-06-03
Refactor calculate-agricultural-holiday-entitlement to new style flow (after rebase)

floehopper:
## Description

Supersedes #2561.

This moves the flow towards the style described in the [refactoring documentation][1]. I've concentrated on moving logic out of the flow and into the calculator. I have not spent any time trying to improve the code in the calculator (e.g. by extracting domain concepts).

I've added some unit tests for the validation that I've extracted, because this behaviour is not covered in the regression tests. However, otherwise I have not moved the tests closer to how it's described in the [new-style testing documentation][2].

So, although there is more work that could be done on this flow and its associated classes, I think the changes in this PR are useful in their own right and if we can do just this much for all the flows, I think we'll be able to dramatically simplify the DSL.

Note that in doing this work I came across some suspect looking code, but rather than get sucked into sorting that out, I've captured the possible problems in a couple of Trello cards:

* https://trello.com/c/0WNxygY4
* https://trello.com/c/HXgHxbpp

Once #2565 was merged, I found that I needed to make some changes to `MethodMissingHelper` and `MethodMissingObject` in order to keep `IndexableContentTest` passing:

* "Ensure MethodMissingObject#description is always a string" (c5c43494586b1dbcc1fb4431cb8889aac3030086)
* "Change MethodMissingObject overrides to return actual object" (d4af6921061e3941909494c07cf920b63b0cce83)

These changes allowed me add an "override" for "holiday_entitlement_days" and "calculator.holiday_entitlement_days" in the following two commits respectively:

* "Move formatting of holiday_entitlement_days into templates" (721ee3b78f92ee00a95ea30826c8f2efcec445ca)
* "Inline holiday_entitlement_days calculate blocks into templates" (65b72982f246875fe6acf436119a064589b76e04)

## External changes


chrisroos:
I'm adding the LGTM label given that I reviewed this in #2561 and have also now reviewed the additional commits that have been added.

floehopper:

PR #2573, opened 2016-06-06
Refactor estimate-self-assessment-penalties to new-style flow

floehopper:
## Description

I haven't really done any significant work on the calculator itself e.g. extracting domain concepts. I added a comment to [this Trello card][2] to say that this flow has some logic where it would make sense to use the `TaxYear` class. The money formatting logic in the templates could probably also be considerably simplified, but this wasn't possible without affecting the regression test artefacts, so I've left that for now. Also I haven't made any significant changes to the tests - we'll still need to move them towards the [new-style approach][3] at some point.

## External changes

### DIFF
--- a/lib/smart_answer/calculators/self_assessment_penalties.rb
+++ b/lib/smart_answer/calculators/self_assessment_penalties.rb
@@ -26,6 +26,10 @@ module SmartAnswer::Calculators
       filing_date >= start_of_next_tax_year
     end
 
+    def valid_payment_date?
+      filing_date <= payment_date
+    end
+
     def paid_on_time?
       (filing_date <= filing_deadline) && (payment_date <= payment_deadline)
     end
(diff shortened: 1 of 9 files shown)


chrisroos:
This looks great to me, @floehopper.

Out of interest, do you know whether it's correct that the tool only works for 2011/12, 2012/13 and 2013/14? If not, do you think we should add a Trello card to discuss it with the content team?

floehopper:
@chrisroos:

Thanks for reviewing.

> Out of interest, do you know whether it's correct that the tool only works for 2011/12, 2012/13 and 2013/14? If not, do you think we should add a Trello card to discuss it with the content team?

I've added a [new Trello card][1] to point this out.

[1]: https://trello.com/c/Kwzrq7jX

floehopper:
Rebasing against master & force-pushing in preparation for merging.

PR #2579, opened 2016-06-13
Refactor energy-grants-calculator to new-style flow

floehopper:
## Description

I haven't really done any significant work on the calculator itself e.g. extracting domain concepts and some of the logic could almost certainly be considerably simplified. Since some of the logic in the calculator is quite complicated, I ended up adding quite a few unit tests to check I wasn't breaking anything. However, there is still quite a bit of work to do to move them towards the [new-style approach][3].

This flow was a classic example where complex logic had been implemented in the flow and the emergent behaviour was very hard to understand. Some of the logic which has ended up in the calculator looks more complicated than the original code. However, I think this is because some implicit logic has been made explicit (which is a good thing).

It would also be considerably improved if we reduced duplication and extracted domain concepts i.e. explaining what the various combinations of conditionals mean and why they exist. However, I'm going to stop at this point, because the PR already has a *lot* of commits.

One other thing I haven't tackled is moving validation into the calculator. This is because there is validation on a checkbox question which gives different error messages for different combinations of options, and I couldn't work out a good way to do this using calculator method(s).

I should also note that in going through this exercise I have identified a couple of bugs:

* [Possible bug in Energy Grants Calculator (Winter Fuel Payment threshold date)](https://trello.com/c/VlCZCmxt)
* [Energy Grants Calculator - Warm Home Discount bug?](https://trello.com/c/oNzNLYSM)

I'm also suspicious that a bunch of logic around the various checkbox questions may be incorrect. I plan to write up one or more Trello cards to capture my concerns.

## External changes


test/unit/calculators/energy_grants_calculator_test.rb
```diff
@@ -0,0 +1,292 @@
...
+      end
+    end
+
+    context '#bills_help?' do
+      should 'return true if which_help is the help_with_fuel_bill option' do
+        @calculator.which_help = 'help_with_fuel_bill'
+        assert @calculator.bills_help?
+      end
+
+      should 'return false if which_help is not the help_energy_efficiency option' do
```
chrisroos: This test name doesn't seem to match the implementation. It looks as though it returns false if `which_help` _is_ the `help_energy_efficiency` option.

test/unit/calculators/energy_grants_calculator_test.rb
```diff
@@ -0,0 +1,292 @@
...
+      end
+    end
+
+    context '#both_help?' do
+      should 'return true if which_help is the all_help option' do
+        @calculator.which_help = 'all_help'
+        assert @calculator.both_help?
+      end
+
+      should 'return false if which_help is not the help_energy_efficiency option' do
```
chrisroos: This test name doesn't seem to match the implementation. It looks as though it returns false if `which_help` _is_ the `help_energy_efficiency` option.

chrisroos:
It looks like there's a redundant "redundant" in the commit note for https://github.com/alphagov/smart-answers/pull/2579/commits/177cb62c5c7bc6e2b03ad1de3fe3b89a7afad776 :-)

chrisroos:
It looks as though this Smart Answer was in a real tangled mess! This PR looks good to me. Great job, @floehopper!

floehopper:
I'm going to use an interactive rebase to address the feedback, rebase against master and force-push in preparation for merging.

test/unit/calculators/energy_grants_calculator_test.rb
floehopper: I've corrected the test name in the new version of this commit: c3fd80672d6194b18fbb2e48c01e6402793c4301.

test/unit/calculators/energy_grants_calculator_test.rb
floehopper: I've corrected the test name in the new version of this commit: 4e2f32f0785c7a487adc24cc1000a8aee995bf22.

floehopper:
> It looks like there's a redundant "redundant" in the commit note for 177cb62 :-)

I've corrected the commit note in the new version of this commit: 4ade49a8896d74533e0bc948c15a799d1534c808.

floehopper:
Note that I also did some work to split up the "Use benefits_claimed attribute in all which_benefits? blocks" commit with regard to the `next_node_calculation(:disabled_or_have_children_question)` block:

* 7e4c6bbe0034ff8fa8dc27541855797f0ad4f6db
* 21149910bae5961db4c9e411de5de3523b03e207
* 2de4d12185a1d51bb1e54cac5d8f2360551d9e6e
* 2c88d01e2990c5136ba0f09d1c9613da507b2c27

PR #2583, opened 2016-06-15
Add document about development workflow

chrisroos:
I've just been reviewing some PRs and realised that I'm not completely clear on the current process.

I started writing an email to ask questions about the process when I realised that it might be easier to document my understanding and use that as a way of discussing it.

Does what I've written match your understanding of our process, @floehopper, @leenagupte, @ikennaokpala and @pmanrubia?

### DIFF
--- a/doc/development-workflow.md
+++ b/doc/development-workflow.md
@@ -1,22 +1,35 @@
 # Development workflow
 
-## PR author
+## Creating a Pull Request
+
+As the Pull Request author:
 
 1. Open a PR with the changes you want to merge.
 
-2. Add the "Waiting on factcheck" label if the changes require feedback from someone other than developers.
+2. Add the "Needs factcheck" label if the changes require feedback from someone other than developers.
 
-3. Move the associated Trello card from "In progress" to "Code review".
+3. Add the "Ready for code review" label.
 
-## PR reviewer
+4. Move the associated Trello card from "In progress" to "Code review".
 
-1. Assign yourself to the PR.
+## Reviewing a Pull Request
 
-2. Review the PR.
+As the Pull Request reviewer:
 
-3. If/when you're happy with the PR:
-  * Add a comment to say that the PR looks good.
-  * Add the "Passed code review" label to allow people to see at a glance that the PR has been reviewed.
+1. Assign yourself to the Pull Request.
 
-4. Move the associated Trello card from "Code review" to "Fact check".
+2. Review the Pull Request.
 
+3. Add a comment to say that the PR looks good if/when you're happy with it.
+
+4. Add the "Passed code review" label to allow people to see at a glance that the PR has been reviewed.
+
+## Fact-checking a Pull Request (Optional)
+
+As the Pull Request author:
+
+1. Replace the "Needs factcheck" label with "Waiting on factcheck".
+
+2. Move the associated Trello card from "Code review" to "Fact check".
+
+3. Add a comment to the associated Trello card to let the relevant people know it's ready to check.
--- a/doc/development-workflow.md
+++ b/doc/development-workflow.md
@@ -16,7 +16,7 @@
 
 3. If/when you're happy with the PR:
   * Add a comment to say that the PR looks good.
-  * Add the "LGTM" label to allow people to see at a glance that the PR has been reviewed.
+  * Add the "Passed code review" label to allow people to see at a glance that the PR has been reviewed.
 
 4. Move the associated Trello card from "Code review" to "Fact check".
 
--- a/doc/development-workflow.md
+++ b/doc/development-workflow.md
@@ -6,7 +6,13 @@ As the Pull Request author:
 
 1. Open a PR with the changes you want to merge.
 
-2. Add the "Needs factcheck" label if the changes require feedback from someone other than developers.
+2. Add the appropriate label(s):
+
+  * Documentation - if this change only affects docs within the app.
+  * Refactoring - if this change doesn't affect the external behaviour of the app.
+  * Spike - if this is an exploratory change to see whether a certain approach works.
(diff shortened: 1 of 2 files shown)


doc/development-workflow.md
```diff
@@ -0,0 +1,22 @@
+# Development workflow
+
+## PR author
+
+1. Open a PR with the changes you want to merge.
+
+2. Add the "Waiting on factcheck" label if the changes require feedback from someone other than developers.
+
+3. Move the associated Trello card from "In progress" to "Code review".
```
leenagupte: @chrisroos The order of the flow changed recently. We're sending PRs to code review before factcheck.

floehopper: I _think_ @chrisroos already knows about the change in order of the workflow, but is unsure if/when we should add the "Waiting on factcheck label". Is that right, Chris?

Since the PR author is the person who knows whether or not the work needs fact-checking, it makes sense to me that the label is added in point 2 as Chris has suggested (i.e. when the PR is created). Otherwise cards might end up with the "LGTM" label and get merged before someone realises it needs fact-checking.

Maybe we should change the label to "Needs fact-checking"?

Having said all that, I don't really mind how we do this as long as we all have a shared understanding.

floehopper:
This process closely mirrors my current understanding.

floehopper:
I think it might be good to extend this a bit and include what happens when something comes back from fact-check.

doc/development-workflow.md
leenagupte: Then I suggest removing the LGTM label and replacing it with a 'Passed Code Review' label.

pmanrubia: @chrissross: thank you for sharing this document. It is true that we have not sorted out yet how to organize labels and workflow. 

Another option could be following the convention of having the tags described below before deploying to production:

##### If it is not a refactoring (need fact-check)
1. Label with LGTM
2. Label with Fact-checked

##### If it is a refactoring (no fact-check needed)
1. Label with LGTM
2. Label with Refactoring

A draft of the process could be something like:

1. Open a PR with the changes you want to merge.
2. Add the “Refactoring” label if the changes does not require feedback from someone other than developers.
3. Move the associated Trello card from "In progress" to "Code review".
4. Once reviewed, If it is a refactoring, move to ready for deploy.
5. Once reviewed, If it is not a refactoring, move the card to the Fact-check
6. Once fact-checked, move the card to ready for deploy.

pmanrubia:
> I think it might be good to extend this a bit and include what happens when something comes back from fact-check.

Yes, it would be good to clarify it as well

doc/development-workflow.md
leenagupte: @pmanrubia That's more confusing.

I'd do the following:

1. Open a PR with the changes.
2. Add a "Ready for Code Review" label when it's ready for review.
3. The code reviewer should then change the label to "Passed code review" when they and the developer have agreed that the PR is good to go.
4. If the PR needs to be looked at by someone other than a dev, the developer should then:
   i) Set the "Waiting for factcheck" label.
   ii) Remove the "Waiting for factcheck" label when the change is signed off from factcheck.
5. If the PR doesn't need a factcheck, or has successfully come back from a factcheck, the developer should then merge the PR.

I would still ditch the LGTM label entirely. It's too confusing to keep around. That label makes you think a PR is ready to be deployed, when that is not necessarily true any longer.

chrisroos: > I think @chrisroos already knows about the change in order of the workflow, but is unsure if/when we should add the "Waiting on factcheck label". Is that right, Chris?

That is correct, @floehopper / @leenagupte. My concern, as you've identified, is that we end up accidentally merging PRs that have been code reviewed because they don't have the "Waiting on factcheck" label.

chrisroos:
I've renamed the "LGTM" label to "Passed code review" as suggested by @leenagupte. I've also updated the document in this PR to reflect that change.

chrisroos:
> I think it might be good to extend this a bit and include what happens when something comes back from fact-check.

What are you thinking about in particular, @floehopper? How to incorporate any changes that come back from fact-check?

chrisroos:
Thanks for your feedback, @floehopper, @leenagupte and @pmanrubia. I've tried to incorporate your suggestions in the document. I'd appreciate you all taking another look when you have a moment.

doc/development-workflow.md
```diff
@@ -0,0 +1,53 @@
...
+
+3. Add a comment to the associated Trello card to let the relevant people know it's ready to check.
+
+## Merging a Pull Request
+
+As the Pull Request author:
+
+1. Rebase the branch on master.
+
+2. Force push the branch to GitHub.
```
chrisroos: Note that I'm not really convinced about this but thought it was a good enough starting point for discussion.

pmanrubia:
Yes, good starting point 👍 LGTM

doc/development-workflow.md
leenagupte: Adding a `Waiting for factcheck` label whilst the PR is in code review and not currently being factchecked is not representative of the status of the PR though.

pmanrubia:
@chrisroos I would be happy with this PR unless there are pending comments from @floehopper @leenagupte 
We can iterate on top of it in case we would need to amend / add something. 
Thank you for your great work 👍

floehopper:
@pmanrubia: I'm happy with this PR as it stands. The most important thing is that we all have a shared understanding of how things work. As you say, if we need to change something, we can iterate. 😄

chrisroos:
I've rebased on master and force pushed this in preparation for merging to master.

chrisroos:
I've added the "Needs factcheck" and "Ready for code review" GitHub labels as described in this documentation.

PR #2576, opened 2016-06-09
Denmark marriage tool update - Content change request

pmanrubia:
Supersedes #2520 

### Trello card
* [125-denmark-marriage-tool-update-content-change-request](https://trello.com/c/9mdERL9m/125-denmark-marriage-tool-update-content-change-request)

## Description
This story is linked to this [Zendesk ticket](https://govuk.zendesk.com/agent/tickets/1317289). 

Updates outcomes copy where a citizen lives in Uk or in Denmark [according to this document](https://docs.google.com/document/d/1QMLTHznjIMN7__cMhlp_4GCrTMa5MNgkUFHaOMLu04U/).

## Factcheck
[denmark-marriage-tool-update-content-change-request](https://smart-answers-pr-2520.herokuapp.com/marriage-abroad)

## Expected changes (1)

* [URL on GOV.UK](https://gov.uk/marriage-abroad/y/denmark/uk/partner_local/same_sex)

### Before

screen shot 2016-06-07 at 17 13 58

### After

screen shot 2016-06-07 at 17 13 48

## Expected changes (2)

* [URL on GOV.UK](https://gov.uk/marriage-abroad/y/denmark/denmark/partner_local/same_sex)

### Before

screen shot 2016-06-07 at 17 17 49

### After

screen shot 2016-06-07 at 17 13 48

PR #2591, opened 2016-06-20
Update regression tests first for marriage-abroad Denmark change

chrisroos:
This PR demonstrates an approach to making marriage-abroad content changes by updating the regression test artefacts _first_.

We currently make these sort of changes by updating the flows/templates and then regenerating the regression test artefacts to confirm that the changes have had the desired effect. One of the problems with this approach is that it's not always easy/possible to see the effect of the changes by looking at the diff of the regression test artefacts. This, in combination with the overly complicated marriage-abroad code, makes it possible to commit unintentional changes to the rendered outcomes.

Updating the regression test artefacts first, and then updating the flows/templates, gives us confidence that the changes we make _only_ affect the things we want.

I chose a recent marriage-abroad content change (PR #2576) and have reimplemented it in this branch by updating the regression test artefacts first.

The branch contains a number of preparatory commits that aren't interesting when it comes to understanding this approach. These commits refactor the code (i.e. don't change behaviour and therefore don't result in changes to the regression test artefacts) to make the actual change easier.

The last 6 commits are the only ones of real interest:

1. Update marriage-abroad denmark/uk opposite sex test artefacts (https://github.com/alphagov/smart-answers/pull/2591/commits/72a4bd440953b2468ca6e0d5863c285aae9380c2)
  * Manually updates 3 regression test artefacts based on the requirements in the Google Doc.
  * Results in 3 failing regression tests. For paths containing "denmark/uk/*/opposite_sex".
2. Update marriage-abroad denmark/ceremony_country opposite sex test artefacts (https://github.com/alphagov/smart-answers/pull/2591/commits/1592f2b3b38acf541f1679deea3c5dc955363f55)
  * Manually updates 3 regression test artefacts based on the requirements in the Google Doc
  * Results in 3 more failing regression tests. For paths containing "denmark/ceremony_country/*/opposite_sex".
3. Make changes required by opposite sex cereonies in Denmark (https://github.com/alphagov/smart-answers/pull/2591/commits/de3b90942dae22d9b9de7c8d9de2fbe43a4db825)
  * Update the outcome template to fix the 6 failing regression tests.
4. Update marriage-abroad denmark/uk same-sex test artefacts (https://github.com/alphagov/smart-answers/pull/2591/commits/59d2c7756ec869baef584dbcdbcb52b4e8c498b8)
  * Manually updates 3 regression test artefacts based on the requirements in the Google Doc.
  * Results in 3 failing regression tests. For paths containing "denmark/uk/*/same_sex".
5. Update marriage-abroad denmark/ceremony_country same-sex test artefacts (https://github.com/alphagov/smart-answers/pull/2591/commits/a38a3a6664b18e1a049bf3914b7c505c3140e564)
  * Manually updates 3 regression test artefacts based on the requirements in the Google Doc.
  * Results in 3 more failing regression tests. For paths containing "denmark/ceremony_country/*/same_sex".
6. Make changes for same sex ceremonies in Denmark (https://github.com/alphagov/smart-answers/pull/2591/commits/42a424445a3cf88b7bc15788b61b697a8a47c191)
  * Update the flow, data and outcome to fix the 6 failing regression tests.

I think there's an additional benefit of this approach in that I can imagine helping the content team/department to make changes to the regression test artefacts directly; allowing us to remove the intermediate step of translating the requirements in a Google Doc.

What do you think, @ikennaokpala, @pmanrubia, @leenagupte and @floehopper? Does this seem like a good approach to you?

Note that while I've focussed on marriage-abroad, I can't see any reason we can't use this approach for all these sort of content changes.

### DIFF
--- a/lib/data/marriage_abroad_services.yml
+++ b/lib/data/marriage_abroad_services.yml
@@ -364,7 +364,11 @@ denmark:
       - :receiving_notice_of_marriage
       - :issuing_cni_or_nulla_osta
   same_sex:
-    default:
+    ceremony_country:
+      default:
+      - :receiving_notice_of_marriage
+      - :issuing_cni_or_nulla_osta
+    third_country:
       default:
       - :issuing_cni_or_nulla_osta
   payment_partial_name: pay_by_cash_or_credit_card_no_cheque
--- a/lib/smart_answer_flows/marriage-abroad.rb
+++ b/lib/smart_answer_flows/marriage-abroad.rb
@@ -132,6 +132,9 @@ module SmartAnswer
           elsif calculator.partner_is_opposite_sex?
             if calculator.ceremony_country == 'hong-kong'
               outcome :outcome_opposite_sex_marriage_in_hong_kong
+            elsif calculator.ceremony_country == 'denmark' &&
+                (calculator.resident_of_uk? || calculator.resident_of_ceremony_country?)
+                outcome :outcome_opposite_sex_marriage_in_denmark_when_residing_in_uk_or_denmark
             elsif calculator.ceremony_country == 'germany'
               outcome :outcome_opposite_sex_marriage_in_germany
             elsif calculator.ceremony_country == 'georgia'
@@ -269,6 +272,7 @@ module SmartAnswer
       outcome :outcome_opposite_sex_marriage_in_commonwealth_countries
       outcome :outcome_opposite_sex_marriage_in_consular_cni_countries_when_residing_in_third_country
       outcome :outcome_opposite_sex_marriage_in_consular_cni_countries_when_residing_in_uk_or_ceremony_country
+      outcome :outcome_opposite_sex_marriage_in_denmark_when_residing_in_uk_or_denmark
       outcome :outcome_opposite_sex_marriage_in_germany
       outcome :outcome_opposite_sex_marriage_in_georgia
       outcome :outcome_opposite_sex_marriage_in_hong_kong
--- a/lib/smart_answer_flows/marriage-abroad.rb
+++ b/lib/smart_answer_flows/marriage-abroad.rb
@@ -113,6 +113,9 @@ module SmartAnswer
         next_node do
           if calculator.ceremony_country == 'brazil' && calculator.resident_outside_of_uk?
             outcome :outcome_marriage_in_brazil_when_residing_in_brazil_or_third_country
+          elsif calculator.ceremony_country == 'denmark' &&
+              (calculator.resident_of_uk? || calculator.resident_of_ceremony_country?)
+            outcome :outcome_ceremonies_in_denmark_when_residing_in_uk_or_denmark
           elsif calculator.ceremony_country == "netherlands"
             outcome :outcome_ceremonies_in_netherlands_or_marriage_via_local_authority_countries
           elsif calculator.ceremony_country == "portugal"
@@ -132,9 +135,6 @@ module SmartAnswer
           elsif calculator.partner_is_opposite_sex?
             if calculator.ceremony_country == 'hong-kong'
               outcome :outcome_opposite_sex_marriage_in_hong_kong
-            elsif calculator.ceremony_country == 'denmark' &&
-                (calculator.resident_of_uk? || calculator.resident_of_ceremony_country?)
-                outcome :outcome_opposite_sex_marriage_in_denmark_when_residing_in_uk_or_denmark
             elsif calculator.ceremony_country == 'germany'
               outcome :outcome_opposite_sex_marriage_in_germany
             elsif calculator.ceremony_country == 'georgia'
@@ -250,6 +250,7 @@ module SmartAnswer
         end
       end
 
+      outcome :outcome_ceremonies_in_denmark_when_residing_in_uk_or_denmark
       outcome :outcome_ceremonies_in_dominican_republic
       outcome :outcome_ceremonies_in_ireland
       outcome :outcome_ceremonies_in_netherlands_or_marriage_via_local_authority_countries
@@ -272,7 +273,6 @@ module SmartAnswer
       outcome :outcome_opposite_sex_marriage_in_commonwealth_countries
       outcome :outcome_opposite_sex_marriage_in_consular_cni_countries_when_residing_in_third_country
       outcome :outcome_opposite_sex_marriage_in_consular_cni_countries_when_residing_in_uk_or_ceremony_country
-      outcome :outcome_opposite_sex_marriage_in_denmark_when_residing_in_uk_or_denmark
       outcome :outcome_opposite_sex_marriage_in_germany
       outcome :outcome_opposite_sex_marriage_in_georgia
       outcome :outcome_opposite_sex_marriage_in_hong_kong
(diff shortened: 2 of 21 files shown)


leenagupte:
@chrisroos I'm confused. Perhaps you could make the commit statements match the bullet points in the PR description. I don't see how you have update the test artefacts before changing the code if you aren't adding Denmark as another option to be tested in the `questions and responses` YAML file.

chrisroos:
Hey @leenagupte.

[Denmark was already in marriage-abroad-questions-and-responses.yml](https://github.com/alphagov/smart-answers/blob/master/test/data/marriage-abroad-questions-and-responses.yml#L31) which is why I didn't have to add it in this branch.

I don't quite follow what you mean by making "the commit statements match the bullet points in the PR description". The 6 top-level bullet points in the PR description contain the titles of, and links to, commits in this PR. The indented bullet points below each of those 6 points contain a summary of the commit.

I've just updated the top level bullet points to a numbered list (1 to 6) in case that helps make it clearer that these represent the 6 commits I talk about in the description.

I've also amended the commit notes of those 6 commits:

* In commits 1 and 2, I've made it clearer that each of them breaks 3 regression tests.
* In commit 3, I've made it clearer that it fixes the 6 broken regression tests introduced in commits 1 and 2.
* In commits 4 and 5, I've made it clearer that each of them breaks 3 regression tests.
* In commit 6, I've made it clearer that it fixes the 6 broken regression tests introduced in commits 4 and 5.

Does any of that make things any clearer?

leenagupte:
@chrisroos This statement is confusing:
> Updating the regression test artefacts first, and then updating the flows/templates, gives us confidence that the changes we make only affect the things we want.

As the Regresssion test artefacts are not updated until commit 19.

chrisroos:
@leenagupte: I've updated the PR description in the hope that it's clearer that you can ignore all the preparatory commits. These just refactor the code in order to make the actual change easier to implement.

The approach I'm trying to describe/promote in this branch can be seen in the last 6 commits, which are all linked from the description.

Does that help?

> @chrisroos This statement is confusing as the Regresssion test artefacts are not updated until commit 19.:

>>  Updating the regression test artefacts first, and then updating the flows/templates, gives us confidence that the changes we make only affect the things we want.

leenagupte:
@chrisroos My confusion stems from the fact that you are manually updating the test artefacts.

That's fine if you only have to change 3 artefacts, but a change I made to a partial template used in outcome page affect over 2,500 test artefacts and it would be madness to have updated all of those manually. See PR: https://github.com/alphagov/smart-answers/pull/2427

Unless, are you actually just committing the changes in reverse order? 
e.g. 
1. updating the templates
2. regenerating the test artefacts
3. commit the test artefacts
4. commit the templates

chrisroos:
> @chrisroos My confusion stems from the fact that you are manually updating the test artefacts.
> 
> That's fine if you only have to change 3 artefacts, but a change I made to a partial template used in outcome page affected over 2,500 test artefacts and it would be madness to have updated all of those manually. See PR: #2427

Ah, OK. I should've made it clearer that this approach almost certainly won't work for all types of content changes. Although, having just looked, I do think it would've been possible to use this test-first approach in PR #2427. By automatically replacing all occurrences of the removed line ("You can [claim back sick pay]...") from the test artefacts, we would've been left with failing regression tests that can then be fixed by removing the same line from "entitled_to_sick_pay.govspeak.erb".

> Unless, are you actually just committing the changes in reverse order? 
> i.e.
> 1. updating the templates
> 2. regenerating the test artefacts
> 3. commit the test artefacts
> 4. commit the templates

Nope. The order is:

1. Update regression test artefacts so that they contain the results of the content changes. This gives us a failing regression test.
2. Update the Smart Answer flow/templates to fix the failing test introduced in 1.

The current workflow (at least for marriage-abroad) is something like:

1. Content team/department describe changes to outcomes in a Google Doc.
2. Developers make the changes to the system that we think will result in the changes described in the Google Doc.
3. Developers inspect the changes to the regression test artefacts to confirm whether the changes to the system have had the desired effect.

The main problem I'm trying to solve with the approach in this PR is to avoid having to carefully review/explain the changes to regression test artefacts in step 3. By updating the artefacts first, and then changing the system until the tests pass, we can be confident that our changes are correct.

Does that make it any clearer, @leenagupte?

leenagupte:
> By automatically replacing all occurrences of the removed line ("You can [claim back sick pay]...") from the test artefacts

Automatically how? Relying on your IDE to do a find and replace?

The problem is that there are so many test artefacts generated. And in the case of #2427 it's impossible for a human being to review them all.

I think that we should take into account when a change modifies the flow and when they just update the copy.

I think that for flow changes, or more specifically, adding a new option, then yes, update the questions and responses file first and then regenerate the test artefacts and start with failing tests, but for parts of the change that are just changes to the copy in templates, it's fine to do those changes first, and the regenerate the test artefacts to match them.

The copy in the templates is not as not as important to me as the structure of the file path that used to mimic the order of the questions and the URL.

pmanrubia:
Hi @chrisroos, I find it very interesting. Content designers are more used to work with documents than with reusable templates. 

So..., two nice wins here: 
1. __If a Content designer edits the artefacts__: they would just need to update text documents in markdown. It fits their mental model and their normal procedures.
2. __If a developer edits the artefact__: we have a failing test that matches the document provided by the content designer. No need to explain the artefacts, and we would save time.

I am happy to explore this approach for incoming PRs, but I would delay it a few weeks; the main reason is that we would need to get up to speed in creating PRs, organising commits / PRs, tests and delivering stories. Starting a new exploratory path right now seems a bit too early for me, to be honest. Does it sound ok to you?

As a side note, I would be worth discussing about:
1. This idea might be related to this Trello card: https://trello.com/b/bVfnXnEc/content-development. A tool to get the list of the pages / documents / artefacts that would need to be modified when updating a flow.
2. I guess we all agree that this new approach is not ideal when changes could impact in dozens, hundreds of files (another argument in favor of what I described above)
3. Current coverage of our artefacts.

So to summarise my feeling about it:

1. A very interesting idea to explore that could apply to a subset of the smart-answers considering the tools we have so far.
2. Happy to give it try in a few weeks once we are up to speed.

When it comes to involving content designers, I would try first with developers until the process is fully understood and agreed among all of us.

floehopper:
@leenagupte:

>> By automatically replacing all occurrences of the removed line ("You can [claim back sick pay]...") from the test artefacts
>
> Automatically how? Relying on your IDE to do a find and replace?

I imagine Chris means that you would craft one or more command lines (e.g. using `sed`) to do the replacement in an automated way.

> The problem is that there are so many test artefacts generated. And in the case of #2427 it's impossible for a human being to review them all. Either as the developer or the code reviewer.

If the commands mentioned above were included in the commit note, then there would be no need to inspect the changes in the artefacts. The reviewer could simply re-run the command(s) against the original source and check that the result is the same as that in the commit.

floehopper:
@pmanrubia: I agree with your assessment. It would be great if you could try this out in another card in the next few weeks.

floehopper:
@chrisroos: This looks good to me. It feels as if it's an extension of what I described in the [Preparatory commits section of the Regression tests documentation][1]. If I get that merged, it might be worth extending or adding to that documentation to suggest this as an alternative approach.

[1]: https://github.com/alphagov/smart-answers/blob/6c8c50a8e014af809ec3ac8c7a56786c9182ce7f/doc/regression-tests.md#preparatory-commits

pmanrubia:
@chrisroos shall we keep this PR opened? I guess it served its purpose.
As a side note, a PR #2628 following the suggested approach

chrisroos:
> @chrisroos shall we keep this PR opened? I guess it served its purpose.

I agree that it's served its purpose. I'll close it now.

> As a side note, a PR #2628 following the suggested approach

This is great! I'm glad you found the approach useful.

PR #2593, opened 2016-06-21
Avoid using worldwide API helpers in regression tests

chrisroos:
Trello card: https://trello.com/c/Hiizo5g6

This follows on from the work in PRs #2540 and #2555 and applies the same
changes to the regression tests.

Switching from WebMock to Mocha appears to reduce the time taken to run all
regression tests by about 15 minutes on my machine.

== Regression test duration before this commit:

    $ time RUN_REGRESSION_TESTS=true \
    ruby -Itest test/regression/smart_answers_regression_test.rb

    real  56m34.146s
    user  45m49.017s
    sys   9m19.234s

== Regression test duration after this commit:

    $ time RUN_REGRESSION_TESTS=true \
    ruby -Itest test/regression/smart_answers_regression_test.rb

    real  36m29.928s
    user  27m45.692s
    sys   8m33.085s

== Notes

The main difference between this and PRs #2540 and #2555 is that a number of
the regression tests rely on `WorldLocation#fco_organisation` returning a
`WorldwideOrganisation` rather than nil. I've satisfied this requirement by
adding the optional `load_fco_organisation_data` parameter to both
`#stub_world_locations` and `#stub_world_location`, and loading the FCO data
when the parameter is true.

The implementation in `#stub_world_location` uses the organisation data that
was previously being used in `#setup_worldwide_locations`. I've reimplemented
the logic from `WorldwideOrganisation#fco_sponsored?` by querying the
`organisations_data` hash. On finding an FCO organisation in the data I use
`GdsApi::Response.build_ostruct_recursively` to turn it into a "deep"
openstruct that can be passed to the `WorldwideOrganisation` initializer.

I was using the Hashugar Gem to create the "deep" openstruct before @floehopper
pointed out the code in `GdsApi::Response`.

I experimented with _always_ loading the FCO organisation data in
`stub_world_location` but this resulted in the non-regression tests taking an
additional couple of minutes to run on my machine.

### DIFF
--- a/test/regression/smart_answers_regression_test.rb
+++ b/test/regression/smart_answers_regression_test.rb
@@ -9,10 +9,12 @@ require 'webmock'
 WebMock.disable_net_connect!(allow_localhost: true)
 
 require_relative '../support/fixture_methods'
+require_relative '../support/world_location_stubbing_methods'
 
 require 'gds_api/test_helpers/content_api'
-require 'gds_api/test_helpers/worldwide'
 require 'gds_api/test_helpers/imminence'
+
+require 'mocha/api'
 
 class SmartAnswersRegressionTest < ActionController::TestCase
   i_suck_and_my_tests_are_order_dependent!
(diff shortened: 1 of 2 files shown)


floehopper:
This looks good to me. 👍

chrisroos:
I noticed that the number of assertions is different when comparing the results of running the regression tests in this branch, master, and in @floehopper's alternative implementation. I'm going to investigate that before merging this to master.

chrisroos:
I've run all regression tests in this branch 3 times and am seeing the same number of assertions as in master (22092). I'm still not 100% sure what caused the difference in the first place but now wonder whether it was as simple as the branch being out of date with master.

PR #2597, opened 2016-06-27
Split marriage-abroad outcome for Dominican Republic

chrisroos:
The current outcome for ceremonies in the Dominican Republic (outcome_ceremonies_in_dominican_republic) contained two logical outcomes:

* One for opposite sex ceremonies
* One for same sex ceremonies

The outcome body contained a single `if/else` statement that decided which logical outcome to display. Splitting these logical outcomes and moving the "routing logic" to the flow is more consistent with the majority of this Smart Answer.


floehopper:
I realise this _might_ be a bit difficult, but I think it would be useful if you could explain a bit about how you decided that there are "two logical outcomes". I think we've talked before about how at one end of the spectrum every Smart Answer could have a single outcome with lots of conditional logic in the ERB template, and at the other end of the spectrum lots of outcomes with no conditional logic in the ERB template. Do you feel as if you've started to develop some heuristics for making these decisions?

floehopper:
Very minor: The commit titled "Add same sex outcome for Dominican Republic" (cb04fd52b88fce584d1bfda53942d55818f3e1fe) says "This is a duplicate of [a template]". However, "This" sounds as if it refers to the outcome.

floehopper:
> I realise this might be a bit difficult, but I think it would be useful if you could explain a bit about how you decided that there are "two logical outcomes".

At the very least it would be useful to explain a little more about your motivation for doing this - perhaps you're trying to make things more consistent?

floehopper:
This looks good to me 👍

chrisroos:
> I realise this might be a bit difficult, but I think it would be useful if you could explain a bit about how you decided that there are "two logical outcomes".

@floehopper: I've updated the PR description to explain how I decided that there were "two logical outcomes". Can you have a read and let me know if it makes sense, please?

chrisroos:
> Very minor: The commit titled "Add same sex outcome for Dominican Republic" (cb04fd5) says "This is a duplicate of [a template]". However, "This" sounds as if it refers to the outcome.

@floehopper: I've rewritten the commit note in an attempt to make it less ambiguous. It now reads: "The new template is a duplicate of outcome_ceremonies_in_dominican_republic.govspeak.erb."

chrisroos:
Thanks for reviewing, @floehopper. I've rebased on master and force pushed in preparation for merging.

PR #2598, opened 2016-06-27
Split marriage-abroad outcome for Brazil

chrisroos:
The current outcome for ceremonies in Brazil (outcome_marriage_in_brazil_when_residing_in_brazil_or_third_country) contains two logical outcomes:

* One for residents of Brazil
* One for residents of a third country (i.e. not resident in the UK or Brazill)

The outcome body contains a single `if/else` statement that decides which logical outcome to display. Splitting these logical outcomes and moving the "routing logic" to the flow is more consistent with the majority of this Smart Answer.


floehopper:
As for #2597, I think it would be helpful to explain a little more about your motivation for making this change.

floehopper:
This looks good to me 👍

chrisroos:
> As for #2597, I think it would be helpful to explain a little more about your motivation for making this change.

I've updated the PR description to better explain why I'm making this change. I've also rewritten some of the commit notes to add more information.

chrisroos:

PR #2600, opened 2016-06-27
Refactor state-pension-through-partner to new-style flow

floehopper:
## Description

I done some work on the calculator itself, but I haven't attempted to e.g. extract domain concepts, and some of the logic could possibly be simplified. Since some of the logic in the calculator was quite complicated, I ended up adding quite a few unit tests to check I wasn't breaking anything. However, there is still quite a bit of work to do to move all the tests for this flow towards the [new-style approach][3].

This flow was another example where complex logic had been implemented in the flow and the emergent behaviour was very hard to understand. I hope that the logic which has ended up in the calculator is  simpler to understand than the original code, but the important thing is that it has been made more explicit and more node-independent.

It would also be considerably improved if we reduced duplication and extracted domain concepts i.e. explaining what the various combinations of conditionals mean and why they exist. However, I'm going to stop at this point, because the PR already has a *lot* of commits.

## External changes


chrisroos:
It looks as though commits fe90ad1d846ef6a34a57732d98d4dc26fcea0994 ("Store when_will_your_partner_reach_pension_age on calculator attribute in on_response block") and 8bc7bd584eeb78b6ae01f082870ebfc71342dfc3 ("Use when_will_you_reach_pension_age attribute instead of response") are a bit muddled. The first commit removes the `next_node_calculation` line and the second adds it back.

chrisroos:
It looks as though commit 8bc7bd584eeb78b6ae01f082870ebfc71342dfc3 ("Use when_will_you_reach_pension_age attribute instead of response") should refer to `when_will_your_partner_reach_pension_age` instead of `when_will_you_reach_pension_age`.

chrisroos:
Aside from my comment about two commits looking as though they're slightly entangled, this all looks good to me @floehopper!

floehopper:

Since all the comments are on the PR, I'm going to address the feedback using an interactive rebase and force-push this branch, rebasing against master if necessary.

floehopper:
> It looks as though commits fe90ad1d846ef6a34a57732d98d4dc26fcea0994 ("Store when_will_your_partner_reach_pension_age on calculator attribute in on_response block") and 8bc7bd584eeb78b6ae01f082870ebfc71342dfc3 ("Use when_will_you_reach_pension_age attribute instead of response") are a bit muddled. The first commit removes the `next_node_calculation` line and the second adds it back.

Well spotted - I'm not sure how that happened!

I've fixed this by not removing the `next_node_calculation` in the new version of the first commit 9354bf388e55ed5eda071a4dc28fedd4d5f08a50 ("Store when_will_your_partner_reach_pension_age on calculator attribute in on_response block").

The second commit is unchanged as far as this problem is concerned.

floehopper:
> It looks as though commit 8bc7bd584eeb78b6ae01f082870ebfc71342dfc3 ("Use when_will_you_reach_pension_age attribute instead of response") should refer to `when_will_your_partner_reach_pension_age` instead of `when_will_you_reach_pension_age`.

I have fixed the commit message title in the new version of this commit: fd7123148925acff7ca2b00275e60b65bbdbc2ca.

PR #2602, opened 2016-06-28
Refactor state-pension-topup to new-style flow

floehopper:
## Description

I done some work on the calculator itself, but I haven't attempted to e.g. extract domain concepts, and some of the logic could possibly be simplified. Since some of the logic in the calculator was quite complicated, I ended up adding quite a few unit tests to check I wasn't breaking anything. However, there is still quite a bit of work to do to move all the tests for this flow towards the [new-style approach][2].

The logic in the calculator would also be considerably improved if we extracted domain concepts i.e. make the code more intention revealing as opposed to just procedural. However, I'm going to stop at this point, because the PR already has a *lot* of commits.

## External changes

### DIFF
--- a/lib/smart_answer/calculators/state_pension_topup_calculator.rb
+++ b/lib/smart_answer/calculators/state_pension_topup_calculator.rb
@@ -19,7 +19,7 @@ module SmartAnswer::Calculators
     end
 
     def lump_sum_and_age
-      return [] if too_young?(date_of_birth, gender)
+      return [] if too_young?
       rows = []
       dob = leap_year_birthday?(date_of_birth) ? date_of_birth + 1.day : date_of_birth
       age = age_at_date(dob, TOPUP_START_DATE)
@@ -31,7 +31,7 @@ module SmartAnswer::Calculators
       rows
     end
 
-    def too_young?(date_of_birth, gender = 'female')
+    def too_young?
       case gender
       when 'female'
         date_of_birth > FEMALE_YOUNGEST_DOB
--- a/lib/smart_answer/calculators/state_pension_topup_calculator.rb
+++ b/lib/smart_answer/calculators/state_pension_topup_calculator.rb
@@ -18,6 +18,10 @@ module SmartAnswer::Calculators
       @gender ||= 'female'
     end
 
+    def valid_whole_number_weekly_amount?
+      weekly_amount.to_f % 1 == 0
+    end
+
     def lump_sum_and_age
       return [] if too_young?
       rows = []
--- a/lib/smart_answer/calculators/state_pension_topup_calculator.rb
+++ b/lib/smart_answer/calculators/state_pension_topup_calculator.rb
@@ -22,6 +22,10 @@ module SmartAnswer::Calculators
       weekly_amount.to_f % 1 == 0
     end
 
+    def valid_weekly_amount_in_range?
+      (1..25).include?(weekly_amount.to_f)
+    end
+
     def lump_sum_and_age
       return [] if too_young?
       rows = []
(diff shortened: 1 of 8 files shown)


lib/smart_answer/calculators/state_pension_topup_calculator.rb
```diff
@@ -1,16 +1,35 @@
...
 
-    def lump_sum_and_age(dob, weekly_amount, gender)
-      return [] if too_young?(dob, gender)
+    attr_accessor :date_of_birth
+    attr_accessor :gender
+    attr_accessor :weekly_amount
+
+    def initialize(attributes = {})
+      super
+      @gender ||= 'female'
```
chrisroos: Why does this default to `'female'`? Is it because we were previously defaulting `gender` to `'female'` in `#too_young?`?

lib/smart_answer_flows/state-pension-topup/questions/how_much_extra_per_week.govspeak.erb
```diff
@@ -6,6 +6,10 @@
   Enter a number between 1 and 25.
 <% end %>
 
-<% content_for :error_message do %>
-  Enter an amount in whole pounds between 1 and 25.
+<% content_for :error_not_whole_number do %>
+  Enter an amount which is a whole number.
```
chrisroos: I wonder whether this should still use "whole pounds", i.e. "Enter an amount in whole pounds"?

chrisroos:
I've added a couple of comments but they shouldn't prevent this from being merged. Looks good to me, @floehopper!

lib/smart_answer/calculators/state_pension_topup_calculator.rb
floehopper: Yes. The original behaviour is pretty odd as I noted ages ago in this [Trello card][1], but I wanted to limit the scope of the changes I was making. If I can find the relevant commit(s), I'll update the commit message(s) to make things clearer.

[1]: https://trello.com/c/jv765Pwc

lib/smart_answer_flows/state-pension-topup/questions/how_much_extra_per_week.govspeak.erb
floehopper: I thought that my wording might make it clearer that the user should not enter a "£" sign, but I'm happy to change it back if you think that's better.

floehopper:
Superseded by #2614. Closing.

lib/smart_answer/calculators/state_pension_topup_calculator.rb
floehopper: In the new version of this PR, I've moved the constructor definition (including the default value for `@gender`) from the commit titled "Use attributes vs arguments in lump_sum_and_age calculator method" (a8c67aadedb036b60ab68dfc735209886698c510) into the later commit titled "Use attributes vs arguments in too_young? calculator method" (c37cbc8cc7afd057458533df5463947fbf2dec6c) and added an explanatory paragraph to the latter's commit message.

lib/smart_answer_flows/state-pension-topup/questions/how_much_extra_per_week.govspeak.erb
floehopper: Following your suggestion, in the new version of this PR, I've changed both validation messages in the commit titled "Split weekly_amount validation into two validate blocks" (7cd20d938a893aab748148069da4ad6d046a28d5) as follows:

* "Enter an amount in whole pounds."
* "Enter an amount in pounds between 1 and 25."

I think this results in the smallest change from the existing behaviour/text.

PR #2607, opened 2016-06-30
Refactor register-a-birth to new-style flow

floehopper:
## Description

Trello card: https://trello.com/c/GPLhNzGq

In this case, I haven't done much work on the calculator itself other than extracting methods from the flow into it. Also I haven't attempted to extract any domain concepts. Some of the logic could almost certainly be simplified. There is still quite a bit of work to do to move all the *tests* for this flow towards the [new-style approach][2].

## External changes

### DIFF
--- a/lib/smart_answer/calculators/register_a_birth_calculator.rb
+++ b/lib/smart_answer/calculators/register_a_birth_calculator.rb
@@ -7,6 +7,7 @@ module SmartAnswer::Calculators
     attr_accessor :married_couple_or_civil_partnership
     attr_accessor :childs_date_of_birth
     attr_accessor :current_location
+    attr_accessor :current_country
 
     def initialize
       @reg_data_query = RegistrationsDataQuery.new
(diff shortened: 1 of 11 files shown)


chrisroos:
The changes in commit 784b4c10f99699d741e7c666d59ba32b31331ffd (removal of `country_name_query` and changes to register_a_birth_test) appear to be reversed as part of 889742808d0062343f022330072b8c67dd756545 (the next commit). The original changes are then reapplied in 97909720639117cde00e4ce2f3bacaf0feb4b667 and 25a1f39e08ccc067dcf1876e7bc7b324810e2661.

Should the first commit (784b4c10f99699d741e7c666d59ba32b31331ffd) be removed completely?

chrisroos:
I've added a single comment about a confusing commit. Aside from that everything looks good to me, @floehopper.

floehopper:
I've created a new branch based off this one, removed the commit as suggested by @chrisroos, rebased against master, resolving the conflicts with the recently merged #2610, and opened a new PR: #2619. Closing.

PR #2608, opened 2016-06-30
Update documentation to include on_response blocks

floehopper:
## Description

Trello card: https://trello.com/c/kWO8Z3i1

In #2408 we introduced `on_response` blocks with the idea of using them to build and store a calculator object in a `calculator` state variable, and to manually store the response to each question on the calculator. This is instead of using `save_input_as` which stores the response as a state variable. This is because in moving towards [new-style flows][1], we're trying to move away from storing any "state variables" other than a single calculator object.

Note that this PR is *not* an attempt to address the action in [this Trello card][2] from the benefit-cap-calculator retrospective. I plan to address that action separately. The idea of this PR is simply to bring the existing documentation up-to-date with the existence of the `on_response` method.

## External changes

None. This only changes documentation.

[2]: https://trello.com/c/XMu0U6Qc

### DIFF
--- a/doc/refactoring.md
+++ b/doc/refactoring.md
@@ -4,15 +4,15 @@ Some Smart Answer flows contain a mix of concerns: routing logic, policy/calcula
 
 These concerns should really be split. Routing belongs in the flow, policy/calculator logic belongs in the calculator and presentation belongs in the ERB templates. You might think of the flow as the Controller, the calculator as the Model and the ERB template as the View in an MVC architecture.
 
-We've refactored a small number of these Smart Answers and have a rough set of steps that we follow:
+We've refactored a number of these Smart Answers and have a rough set of steps that we follow:
 
 * Remove unused code from the flow, e.g. unused `save_input_as`, `precalculate` or `calculate` blocks. See commit [fe4014d7c007108f3daa07f1c5dd749f5de683a0](https://github.com/alphagov/smart-answers/commit/fe4014d7c007108f3daa07f1c5dd749f5de683a0) for an example.
 
 * Include `ActiveModel::Model` in the calculator so that it's easy to instantiate it with a number of attributes set. See commit [f78671a062339ae97abb5cb267b55536233316c9](https://github.com/alphagov/smart-answers/commit/f78671a062339ae97abb5cb267b55536233316c9) for an example.
 
-* Instantiate the calculator in a `next_node_calculation` block in the first question. See commit [6114c5fc55bfdaff52501e2a19dead35ebb42f13](https://github.com/alphagov/smart-answers/commit/6114c5fc55bfdaff52501e2a19dead35ebb42f13) for an example.
+* Instantiate the calculator in a `on_response` block in the first question. See commit [0df6fd7df9a00ec882edf249eeaaa68a291633c5](https://github.com/alphagov/smart-answers/commit/0df6fd7df9a00ec882edf249eeaaa68a291633c5) for an example.
 
-* Save the responses to questions on the calculator object rather than using `save_input_as`. See commit [64c60baf42e726fa234e579a04c36bc0da0fa3fc](https://github.com/alphagov/smart-answers/commit/64c60baf42e726fa234e579a04c36bc0da0fa3fc) for an example.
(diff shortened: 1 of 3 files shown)


doc/storing-data.md
```diff
@@ -50,11 +61,20 @@ multiple_choice :question_2? do
     'q2-precalculated-answer'
   end
 
+  on_response do |response|
+    # response             => 'q2_option'
+    # responses            => ['q1_option']
+    # q1_calculated_answer => 'q1-calculated-answer'
+
```
chrisroos: I think this `on_response` block would also have access to the `q2_precalculated_answer`, wouldn't it?

chrisroos:
@floehopper: Aside from my question about the `on_response` example in the docs, this all looks good to me.

doc/storing-data.md
floehopper: Yes, I think you're right. I'll fix it. Thanks.

doc/storing-data.md
```diff
@@ -1,6 +1,17 @@
...
+* `on_response` values are available in:
+  * The `precalculate` block
+  * The question template
+  * The `next_node_calculation` block
+  * The `validate` block
+  * The `next_node` block
+  * The `calculate` block
+  * All subsequent questions and outcomes
+
+__NOTE.__ `on_response` blocks are not named, because they don't automatically store a value in a state variable. In fact doing so is actively discouraged apart from when storing a calculator object in the first question of a flow.
```
leenagupte: An inline example of how to use an `on_response` block and how to access the value it stores would be useful.

doc/refactoring.md
```diff
@@ -4,15 +4,15 @@ Some Smart Answer flows contain a mix of concerns: routing logic, policy/calcula
...
+We've refactored a number of these Smart Answers and have a rough set of steps that we follow:
 
 * Remove unused code from the flow, e.g. unused `save_input_as`, `precalculate` or `calculate` blocks. See commit [fe4014d7c007108f3daa07f1c5dd749f5de683a0](https://github.com/alphagov/smart-answers/commit/fe4014d7c007108f3daa07f1c5dd749f5de683a0) for an example.
 
 * Include `ActiveModel::Model` in the calculator so that it's easy to instantiate it with a number of attributes set. See commit [f78671a062339ae97abb5cb267b55536233316c9](https://github.com/alphagov/smart-answers/commit/f78671a062339ae97abb5cb267b55536233316c9) for an example.
 
-* Instantiate the calculator in a `next_node_calculation` block in the first question. See commit [6114c5fc55bfdaff52501e2a19dead35ebb42f13](https://github.com/alphagov/smart-answers/commit/6114c5fc55bfdaff52501e2a19dead35ebb42f13) for an example.
+* Instantiate the calculator in a `on_response` block in the first question. See commit [0df6fd7df9a00ec882edf249eeaaa68a291633c5](https://github.com/alphagov/smart-answers/commit/0df6fd7df9a00ec882edf249eeaaa68a291633c5) for an example.
```
leenagupte: Can you use an example block as well as linking to a commit.

leenagupte:
@floehopper Is there anywhere that defines what these blocks are?
e.g. `on_response` is used to....? Store responses from questions? Is the type of question it can be used in limited? e.g. only multiple choice questions?

doc/storing-data.md
floehopper: I've addressed this in a fixup commit: 913368d7f7b2be84f19e88d7b66225c8e643ffaf.

floehopper:
@leenagupte:

> Is there anywhere that defines what these blocks are?
> e.g. on_response is used to....? Store responses from questions? Is the type of question it can be used in limited? e.g. only multiple choice questions?

I've added a load of information about these different block types in this later PR: #2613. I think that answers all of these questions.

doc/storing-data.md
floehopper: I've added an inline example in an extra commit: 5eed6a46a24d7a63395222496d7bee5189355f90.

doc/refactoring.md
floehopper: I've added a couple of inline example code blocks in an extra commit: 8762b6f023081311500cdcdf0bafe76b68c90573.

floehopper:
I've created a new branch based off this one, rebased against master, and done an interactive rebase to apply the fix-up commits. I've opened a new PR (#2616) for the new version of the branch. Closing.

PR #2609, opened 2016-07-01
Register a death - Libya

leenagupte:
Trello card: https://trello.com/c/QhANM5Vi/209-30-jun-register-a-death-libya-content-change-request

## Factcheck
[Register a death Libya](https://smart-answers-pr-2609.herokuapp.com/register-a-death/y/overseas/libya)

## Expected Changes

[URL on GOV.UK](https://www.gov.uk/register-a-death/y/overseas/libya)
    * Selecting Libya should send you straight to an outcome

### Before

screen shot 2016-07-01 at 11 05 09

### After

screen shot 2016-07-01 at 11 36 42


pmanrubia:
@leenagupte could you please accommodate your commits to [50 chars for the title and 72 characters per line in the description](https://github.com/alphagov/styleguides/blob/master/git.md#example-1)

test/artefacts/register-a-death/overseas/north-korea.html
```diff
@@ -0,0 +1,132 @@
+<!DOCTYPE html>
```
pmanrubia: @floehopper @chrisroos would you know why are we getting changes in North-Korea artefacts if nothing related to this country has been modified in this PR? Is there any document, or any point of the codebase we can look into?

pmanrubia:
@floehopper @chrisroos: regarding e225ed6c1736d9f3173567ab9dde087e7982fa7b,  would you know why are we getting changes in North-Korea artefacts if nothing related to this country has been modified in this PR? Is there any document, or any point of the codebase we can look into?

floehopper:
@pmanrubia:

> regarding e225ed6, would you know why are we getting changes in North-Korea artefacts if nothing related to this country has been modified in this PR? Is there any document, or any point of the codebase we can look into?

Since it's possible to visit the same question page via a number of different routes, the regression tests only capture the HTML from the *first visit* to a question page. I'm not sure whether this is documented anywhere - I'll add a Trello card to make sure that it is.

If you look at both the commits where the regression test artefacts are changed, you can see that the following question page artefacts are removed:

* `test/artefacts/register-a-death/overseas/libya.html`
* `test/artefacts/register-a-death/overseas/libya/another_country.html`

And replaced with the following question page artefacts:

* `test/artefacts/register-a-death/overseas/north-korea.html`
* `test/artefacts/register-a-death/overseas/north-korea/another_country.html`

I assume this is because the regression test no longer first (or ever) reaches these question pages via a `libya` response, but now reaches them via a `north-korea` response. Does that make sense?

floehopper:
I've just created a [new Trello card][1] to ensure how/when question page artefacts are captured is documented.

[1]: https://trello.com/c/JK9nyoaa

leenagupte:
@pmanrubia Done.

pmanrubia:
Thank you @leenagupte. Still, the lines of the description are not 72 characters max. I guess you are doing it because Github is able to wrap it nicely, but there are other editors out there that are not that smart / user-friendly (particularly terminal-based editors). So, I would follow the conventions until it is agreed that this is not necessary.

pmanrubia:
It makes sense @floehopper, thank you for the clarification.

floehopper:
> If you look at both the commits where the regression test artefacts are changed, you can see that the following question page artefacts are removed [...] and replaced with the following question page artefacts [...]

I think it would be clearer if all the changes to the question page artefacts were in the same commit, then a single commit note could explain the removals and the additions i.e. I'm not convinced it's useful to split the artefact commits by country in this particular case.

leenagupte:
@floehopper They were originally split because I had no idea what was going on and I wanted to highlight the anomaly.  However, they have squashed now and the commit message has been updated to reflect how I interpreted your explanation.

pmanrubia:
LGTM

PR #2613, opened 2016-07-05
Documentation improvements

floehopper:
## Description

Trello card: https://trello.com/c/XMu0U6Qc

This is an attempt to improve the documentation, particularly around the flow definition e.g. the scope of variables within the various blocks, the execution order of those same blocks, and hence the availability of variables within templates.

I've focussed on documenting the *current* behaviour of the app. However, where appropriate, I've also tried to highlight the following:

* peculiarities of the current behaviour
* things that are deprecated/encouraged (as we try to [move towards the new-style of flow][refactoring])

Note that where I have ascribed an intention to the current behaviour of the app, this is essentially guesswork, because neither @chrisroos nor I were involved in building the app in the first place.

I was slightly hesitant to write this documentation, because the flow definition DSL is a moving target as we [move towards the new-style of flow][refactoring]. The aim is for it to be much simpler/intuitive and not need as much documentation!

It's probably best to review this pull request by viewing the final versions of the affected pages, rather than the diffs:

* [flow-definition](https://github.com/alphagov/smart-answers/blob/improve-flow-definition-documentation/doc/flow-definition.md)
* [landing-page-template](https://github.com/alphagov/smart-answers/blob/improve-flow-definition-documentation/doc/landing-page-template.md)
* [question-templates](https://github.com/alphagov/smart-answers/blob/improve-flow-definition-documentation/doc/question-templates.md)
* [outcome-templates](https://github.com/alphagov/smart-answers/blob/improve-flow-definition-documentation/doc/outcome-templates.md)

## External changes

None. This only includes changes to documentation.

[refactoring]: https://github.com/alphagov/smart-answers/blob/improve-flow-definition-documentation/doc/refactoring.md

### DIFF
--- a/doc/flow-definition.md
+++ b/doc/flow-definition.md
@@ -1,6 +1,6 @@
 # Flow
 
-The Ruby flow file defines a subclass of `SmartAnswer::Flow` that specifies metadata, and defines all the question nodes, the outcome nodes, and the rules to control flow between the nodes.
+At the heart of each Smart Answer is a subclass of `SmartAnswer::Flow`. Subclasses like this make a DSL available for [specifying metadata](#metadata), and [defining](#definition) all the [question nodes](#question-nodes), the [outcome nodes](#outcome-nodes), and the [rules](doc/next-node-rules.md) to control routing between the nodes.
 
 ## Naming
 
--- a/doc/flow-definition.md
+++ b/doc/flow-definition.md
@@ -248,7 +248,7 @@ Each of these block types and the point at which they are executed is explained 
 ##### `precalculate(variable_name, &block)`
 
 * These blocks were intended to be used to store state variables that are needed to render a question or outcome node template.
-* These blocks execute before the question/outcome template is rendered and before the user response is parsed from the request path.
+* These blocks are the first to execute for a given question/outcome. They execute before the relevant template is rendered, so that the state variables that they define are available in question/outcome templates. They also execute before the user response is parsed from the request path.
 * The parsed response is *not* passed to the block, because it has not been received at the point the block is executed.
 * The block return value is stored on the state object as a state variable named `variable_name`.
 
--- a/doc/flow-definition.md
+++ b/doc/flow-definition.md
@@ -248,7 +248,8 @@ Each of these block types and the point at which they are executed is explained 
 ##### `precalculate(variable_name, &block)`
 
 * These blocks were intended to be used to store state variables that are needed to render a question or outcome node template.
-* It's simplest to think about these blocks executing before the user response is parsed from the request path.
+* These blocks execute before the question/outcome template is rendered and before the user response is parsed from the request path.
+* The parsed response is *not* passed to the block, because it has not been received at the point the block is executed.
 * The block return value is stored on the state object as a state variable named `variable_name`.
 
 > Note that due to an oversight in the implementation of `Flow#process`, it's not currently possible to use a `precalculate` block in the first question.
@@ -258,7 +259,7 @@ Each of these block types and the point at which they are executed is explained 
 ##### `on_response(&block)`
 
 * These blocks are intended to be used to store user responses on a `calculator` state variable. They are a [relatively new addition to the DSL][introduction-of-on-response-blocks].
-* These blocks are called  after the user response has been parsed from the request path and before any of the `next_node_calculation` blocks are executed.
+* These blocks are called after the question/outcome template has been rendered and after the user response has been parsed from the request path, but before any of the `next_node_calculation` blocks are executed.
 * The parsed response is passed to the block as the only argument and by convention is named `response`.
 * The block return value is not used and no state variable is stored.
 
--- a/doc/flow-definition.md
+++ b/doc/flow-definition.md
@@ -122,7 +122,7 @@ See [In-question blocks](#in-question-blocks) below for more information on the 
 
 The value of `self` inside the question node definition block (but outside other blocks) is the instance of the relevant question node. Thus in the example above, the calls to `#option` and `#next_node` are made on the instance of the question node.
 
-In the same way that there is a single instance of each flow class, there is only ever a single instance of each question definition in the system at any one time i.e. a flow instance has an invariant set of question node definition instances.
+In the same way that there is a single instance of each flow class, there is only ever a single instance of each question definition in the system at any one time i.e. a flow instance has a fixed set of question node definition instances.
 
 Since the same instance of a question definition is used across multiple requests, it's important that no request-specific state is stored on them.
 
--- a/doc/flow-definition.md
+++ b/doc/flow-definition.md
@@ -126,7 +126,7 @@ In the same way that there is a single instance of each flow class, there is onl
 
 Since the same instance of a question definition is used across multiple requests, it's important that no request-specific state is stored on them.
 
-The value of `self` inside the `next_node` blocks (or the other optional blocks) is an instance of `SmartAnswer::State` ([see below](#state)), i.e. *not* an instance of a question node.
+The value of `self` inside the `next_node` blocks (or the other [in-question blocks](#in-question-blocks)) is an instance of `SmartAnswer::State` ([see below](#state)), i.e. *not* an instance of a question node.
 
 ```ruby
 def define
--- a/doc/flow-definition.md
+++ b/doc/flow-definition.md
@@ -229,7 +229,9 @@ All question definition blocks, must include a single `next_node` block. A numbe
 
 The value of `self` inside all these blocks is an instance of `SmartAnswer::State` ([see above](#state)). The code inside these blocks is executed at request time, not at flow definition time.
 
-All blocks of a particular type (within a single question) are executed at particular points in the request processing sequence i.e. all `on_response` blocks are executed before all `next_node_calculation` blocks. The order in which the blocks are defined only affects the order in which they are executed within the group of blocks of the same type i.e. when two `on_response` blocks are defined, the one defined first will be executed before the one defined second; however, even if a `next_node_calculation` block is defined before both of these `on_response` blocks, it will always be executed after both of them.
+All blocks of a particular type (within a single question) are executed at particular points in the request processing sequence e.g. all `on_response` blocks are executed before all `next_node_calculation` blocks.
+
+The order in which the blocks are defined only affects the order in which they are executed within the group of blocks of the same type e.g. when two `on_response` blocks are defined, the one defined first will be executed before the one defined second; however, even if a `next_node_calculation` block is defined before both of these `on_response` blocks, it will always be executed after both of them.
 
 The block types are executed in the following order:
 
--- a/doc/flow-definition.md
+++ b/doc/flow-definition.md
@@ -225,7 +225,9 @@ second_state = first_state.transition_to(:third_node, 'second-response')
 
 #### In-question blocks
 
-All question definition blocks, must include a single `next_node` block. A number of other in-question blocks can optionally be defined by passing a block to any of the following methods on `SmartAnswer::Node` & `SmartAnswer::Question::Base`. The value of `self` inside all these blocks is an instance of `SmartAnswer::State` ([see above](#state)). The code inside these blocks is executed at request time, not at flow definition time.
+All question definition blocks, must include a single `next_node` block. A number of other in-question blocks can optionally be defined by passing a block to any of the following methods on `SmartAnswer::Node` & `SmartAnswer::Question::Base`: `precalculate`, `on_response`, `next_node_calculation`, `validate`, `next_node` & `calculate`. The `save_input_as` method is used in a similar way, but does not accept a block.
+
+The value of `self` inside all these blocks is an instance of `SmartAnswer::State` ([see above](#state)). The code inside these blocks is executed at request time, not at flow definition time.
 
 All blocks of a particular type (within a single question) are executed at particular points in the request processing sequence i.e. all `on_response` blocks are executed before all `next_node_calculation` blocks. The order in which the blocks are defined only affects the order in which they are executed within the group of blocks of the same type i.e. when two `on_response` blocks are defined, the one defined first will be executed before the one defined second; however, even if a `next_node_calculation` block is defined before both of these `on_response` blocks, it will always be executed after both of them.
 
--- a/doc/flow-definition.md
+++ b/doc/flow-definition.md
@@ -34,7 +34,7 @@ The flow is defined within the `define` instance method of the `Flow` subclass u
 > * It's confusing that blocks of code are not necessarily executed in the order that they appear in the flow definition.
 > * The code in the blocks is not easily unit-testable.
 
-Each flow is instantiated in the `FlowRegistry` via the `Flow.build` method which, in turn, instantiates the flow and invokes its `Flow#define` method.
+Each flow is instantiated in the [`FlowRegistry`][flow-registry] via the `Flow.build` method which, in turn, instantiates the flow and invokes its `Flow#define` method.
 
 Currently in production a single instance of each flow class is instantiated at application start-up (or strictly speaking on the first request) and cached in the `FlowRegistry` i.e. the same instance of a particular flow class is used to service all the user requests involving that Smart Answer. Thus it's important that no request-specific state is stored on the flow instance, otherwise this could leak into other user requests.
 
@@ -306,3 +306,4 @@ See the [documentation for outcome templates](doc/outcome-templates.md).
 [object-dup]: http://ruby-doc.org/core-2.3.0/Object.html#method-i-dup
 [local-variable-in-flow-definition]: https://github.com/alphagov/smart-answers/blob/20d2d9f524e912a3edcb9256ce2d790059538641/lib/smart_answer_flows/register-a-birth.rb#L9-L12
 [introduction-of-on-response-blocks]: https://github.com/alphagov/smart-answers/pull/2408
+[flow-registry]: https://github.com/alphagov/smart-answers/blob/cc6c5050bb78d0085cffe0a40630784724aa062a/lib/smart_answer/flow_registry.rb
--- a/doc/flow-definition.md
+++ b/doc/flow-definition.md
@@ -61,9 +61,11 @@ end
 
 ### Arbitrary Ruby code
 
-Since the flow definition is just the body of a standard Ruby method, it's possible to write arbitrary Ruby code at any point within it.
+Since the flow definition is just the body of a standard Ruby method, it's possible to write arbitrary Ruby code at any point within it. Arbitrary Ruby code is pretty much anything other than calls to the DSL methods described in this document. The use of a local variable at the top-level of the `#define` method and used later within a block is an example of arbitrary Ruby code which is explained in more detail below.
 
 > It should be possible to implement the majority of flows *without* writing such arbitrary Ruby code. Any decision to introduce such arbitrary code should be very carefully considered, particularly regarding it's maintainability.
+
+> Some exceptions to the "no arbitrary Ruby" rule are: instantiating and storing a "calculator" object (in an `on_response` block), storing a response on the "calculator" object, and conditional logic within a `next_node` block. Although even the latter should be kept to a minimum by extracting predicate methods onto the "calculator" object.
 
 Since the flow definition is just the body of a `Flow` instance method (`#define`), the value of `self` at the "top-level" within the method is an instance of the flow.
 
--- a/doc/flow-definition.md
+++ b/doc/flow-definition.md
@@ -36,7 +36,7 @@ The flow is defined within the `define` instance method of the `Flow` subclass u
 
 Each flow is instantiated in the [`FlowRegistry`][flow-registry] via the `Flow.build` method which, in turn, instantiates the flow and invokes its `Flow#define` method.
 
-Currently in `development` & `production` environments (but not in `test`) a single instance of each flow class is instantiated at application start-up (or strictly speaking on the first request) and cached in the `FlowRegistry` i.e. the same instance of a particular flow class is used to service all the user requests involving that Smart Answer. Thus it's important that no request-specific state is stored on the flow instance, otherwise this could leak into other user requests.
+Currently when the app is running in the Rails `development` & `production` environments (but not in `test`) a single instance of each flow class is instantiated at application start-up (or strictly speaking on the first request) and cached in the `FlowRegistry` i.e. the same instance of a particular flow class is used to service all the user requests involving that Smart Answer. Thus it's important that no request-specific state is stored on the flow instance, otherwise this could leak into other user requests.
 
 It's important to understand this distinction between "flow definition time" and "request time" when trying to understand a flow definition. A number of the sections below refer to this distinction.
 
--- a/doc/flow-definition.md
+++ b/doc/flow-definition.md
@@ -281,10 +281,10 @@ Each of these block types and the point at which they are executed is explained 
 * The parsed response is passed to the block as the only argument and by convention is named `response`.
 * If the block return value is truth-y, then no action is taken.
 * If the block return value is false-y, then:
-  * A `SmartAnswer::InvalidResponse` exception is raised with the `message_key` set as the exception message.
-  * Although this exception is handled within the app, it prevents the transition to the next node.
-  * The `message_key` from the exception message is set on the built-in state variable, `error`.
-  * When the question template is re-rendered, the `error` state variable is used to lookup the appropriate validation error message in the [question template](doc/question-templates.md#-error_message-message-).
+  1. A `SmartAnswer::InvalidResponse` exception is raised with the `message_key` set as the exception message.
+  2. This exception is handled within the app and prevents the transition to the next node.
+  3. The `message_key` from the exception message is set on the built-in state variable, `error`.
+  4. When the question template is re-rendered, the `error` state variable is used to lookup the appropriate validation error message in the [question template](doc/question-templates.md#-error_message-message-).
 
 > The use of these blocks is encouraged. However, they should call `valid_xxx?` methods on the `calculator` state variable and not rely on the `response` argument passed into the block.
 
--- a/doc/flow-definition.md
+++ b/doc/flow-definition.md
@@ -173,11 +173,34 @@ end
 The state object is intended to store all request-specific state, keeping that away from the instance of the flow which is reused across multiple requests. `SmartAnswer::State` inherits from `OpenStruct` and uses [`BasicObject#method_missing`][method-missing]
 to allow arbitrary "state variables" to be written and read.
 
+```ruby
+state = SmartAnswer::State.new(:first_node)
+state.example_state_variable = 123
+state.example_state_variable # => 123
+```
+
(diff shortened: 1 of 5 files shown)


doc/flow-definition.md
```diff
@@ -1,11 +1,308 @@
-# The flow definition
+# Flow
 
-The Ruby flow file defines a `Flow` class that contains all the questions, outcomes and rules to control flow between.
+The Ruby flow file defines a subclass of `SmartAnswer::Flow` that specifies metadata, and defines all the question nodes, the outcome nodes, and the rules to control flow between the nodes.
```
leenagupte: What is the `The Ruby flow file`? Can I have an example?

doc/flow-definition.md
```diff
@@ -1,11 +1,308 @@
...
+The flow is defined within the `define` instance method of the `Flow` subclass using a DSL to specify metadata, question nodes and outcome nodes.
+
+> Unfortunately this DSL is overly complex and makes it all too easy to create code which is very hard to follow. The DSL makes heavy use of Ruby's [`BasicObject#instance_eval`][instance-eval] & [`#instance_exec`][instance-exec]. This means that:
+>
+> * It's not obvious what variables are in scope within a given block of code.
+> * It's all too easy to write arbitrary custom code which can make each flow a bit of a special case. This doesn't help on the maintainability front.
+> * It's confusing that blocks of code are not necessarily executed in the order that they appear in the flow definition.
+> * The code in the blocks is not easily unit-testable.
+
+Each flow is instantiated in the `FlowRegistry` via the `Flow.build` method which, in turn, instantiates the flow and invokes its `Flow#define` method.
```
leenagupte: Link to the Flow class?

doc/flow-definition.md
```diff
@@ -1,11 +1,308 @@
...
+
+Currently in production a single instance of each flow class is instantiated at application start-up (or strictly speaking on the first request) and cached in the `FlowRegistry` i.e. the same instance of a particular flow class is used to service all the user requests involving that Smart Answer. Thus it's important that no request-specific state is stored on the flow instance, otherwise this could leak into other user requests.
+
```
leenagupte: This happens in all environments, right? Not just production?

leenagupte: What could a `request-specific state` look like?

doc/flow-definition.md
```diff
@@ -1,11 +1,308 @@
...
+  end
+end
+```
+
+### Arbitrary Ruby code
+
+Since the flow definition is just the body of a standard Ruby method, it's possible to write arbitrary Ruby code at any point within it.
+
+> It should be possible to implement the majority of flows *without* writing such arbitrary Ruby code. Any decision to introduce such arbitrary code should be very carefully considered, particularly regarding it's maintainability.
+
```
leenagupte: What does arbitrary ruby refer to? Is it functions that are used by the flow but aren't a question or outcome?

doc/flow-definition.md
```diff
@@ -1,11 +1,308 @@
...
+Since the flow definition is just the body of a `Flow` instance method (`#define`), the value of `self` at the "top-level" within the method is an instance of the flow.
+
+#### Local variables
+
+Some flows (e.g. [register-a-birth][local-variable-in-flow-definition]) define local variables towards the top of the flow definition. Standard Ruby scoping means that these are then available within question & outcome blocks, as well as within blocks nested inside these blocks. However, note that the value of the local variable is set at flow definition time and *not* at request time. Since a single flow instance is cached in the `FlowRegistry` at application start-up time, this means the local variable is _only assigned once_.
+
```
leenagupte: Are the local variables supposed to be an example of arbitrary code?

doc/flow-definition.md
```diff
@@ -1,11 +1,308 @@
...
+There is an implicit assumption that the question definition which appears first in the flow definition is the first question of the flow. All other "routing" between nodes is [done explicitly](doc/next-node-rules.md) in a single `next_node` block per question node.
+
+By convention, question nodes are usually listed roughly in the order that a user would visit them. Although this isn't always straightforward when there are multiple paths through the flow.
+
+Since all "routing" is done explicitly within `next_node` blocks, the order of the question definitions (other than the first one) is functionally unimportant.
+
+By convention, all question nodes are defined *before* any of the outcome nodes. However, again the order is functionally unimportant.
+
+Question nodes are defined by calls to one of the various [question-type methods](doc/question-types.md). Since the value of `self` at the "top-level" within the `#define` method is an instance of the flow, these question-type methods are defined on the `SmartAnswer::Flow` base class.
+
```
leenagupte: Can you add `self` to the example below.

doc/flow-definition.md
```diff
@@ -1,11 +1,308 @@
...
+```
+
+See [In-question blocks](#in-question-blocks) below for more information on the optional blocks mentioned in a comment in the example.
+
+#### Scope
+
+The value of `self` inside the question node definition block (but outside other blocks) is the instance of the relevant question node. Thus in the example above, the calls to `#option` and `#next_node` are made on the instance of the question node.
+
+In the same way that there is a single instance of each flow class, there is only ever a single instance of each question definition in the system at any one time i.e. a flow instance has an invariant set of question node definition instances.
+
```
leenagupte: I would change `invariant` to `fixed` for readability.

doc/flow-definition.md
```diff
@@ -1,11 +1,308 @@
...
+Since the same instance of a question definition is used across multiple requests, it's important that no request-specific state is stored on them.
+
```
leenagupte: Add a link to the page where this is explained in more detail.

doc/flow-definition.md
```diff
@@ -1,11 +1,308 @@
...
+The value of `self` inside the `next_node` blocks (or the other optional blocks) is an instance of `SmartAnswer::State` ([see below](#state)), i.e. *not* an instance of a question node.
+
```
leenagupte: optional blocks?

doc/flow-definition.md
```diff
@@ -1,11 +1,308 @@
...
+```
+
+#### State
+
+The state object is intended to store all request-specific state, keeping that away from the instance of the flow which is reused across multiple requests. `SmartAnswer::State` inherits from `OpenStruct` and uses [`BasicObject#method_missing`][method-missing]
+to allow arbitrary "state variables" to be written and read.
+
+Since the request path only includes the user's responses and *not* the question keys, and the app is stateless, *every* request has to be processed by walking through the question definition nodes starting at the first one. As a request is processed, the state is duplicated using [`Object#dup`][object-dup] in each "transition" to a new node.
+
+It's important to note that `Object#dup` does not do a "deep" copy. Thus any "state variables" set on the state which are references to other objects will continue to reference the same instances of those other object - those objects will *not* themselves be duplicated. The *only* exceptions to this are two built-in state variables, `responses` & `path` ([see below](#built-in-state-variables)) which are themselves duplicated using `Object#dup` in `State#initialize_copy`.
```
leenagupte: An example of the affected state variables would be good here.

doc/flow-definition.md
```diff
@@ -1,11 +1,308 @@
...
+* `path` - array of symbol keys for nodes previously processed
+* `responses` - user responses parsed from request path; usually strings (?)
+* `response` - always `nil` (?)
+* `error` - key for validation error message to display; usually a string (?)
+
+> Note that some of the application code (e.g. illegal multiple choice response) erroneously sets the error key to the validation error message *string*. Since this string is not the *key* to an error message, the default error message is displayed.
+
+#### In-question blocks
+
+All question definition blocks, must include a single `next_node` block. A number of other in-question blocks can optionally be defined by passing a block to any of the following methods on `SmartAnswer::Node` & `SmartAnswer::Question::Base`. The value of `self` inside all these blocks is an instance of `SmartAnswer::State` ([see above](#state)). The code inside these blocks is executed at request time, not at flow definition time.
```
leenagupte: following methods?

doc/flow-definition.md
```diff
@@ -1,11 +1,308 @@
...
+
+All blocks of a particular type (within a single question) are executed at particular points in the request processing sequence i.e. all `on_response` blocks are executed before all `next_node_calculation` blocks. The order in which the blocks are defined only affects the order in which they are executed within the group of blocks of the same type i.e. when two `on_response` blocks are defined, the one defined first will be executed before the one defined second; however, even if a `next_node_calculation` block is defined before both of these `on_response` blocks, it will always be executed after both of them.
```
leenagupte: I would break this up and add a section/line to the information about block about where and when they are executed.

doc/flow-definition.md
```diff
@@ -1,11 +1,308 @@
...
+Each of these block types and the point at which they are executed is explained in more detail below:
+
+##### `precalculate(variable_name, &block)`
+
+* These blocks were intended to be used to store state variables that are needed to render a question or outcome node template.
+* It's simplest to think about these blocks executing before the user response is parsed from the request path.
+* The block return value is stored on the state object as a state variable named `variable_name`.
+
+> Note that due to an oversight in the implementation of `Flow#process`, it's not currently possible to use a `precalculate` block in the first question.
+
```
leenagupte: Isn't it true that it's also not possible to use the value stored in a precalculate block in the template of the question the block is defined in?

doc/flow-definition.md
```diff
@@ -1,11 +1,308 @@
...
+* These blocks are intended to be used to store user responses on a `calculator` state variable. They are a [relatively new addition to the DSL][introduction-of-on-response-blocks].
+* These blocks are called  after the user response has been parsed from the request path and before any of the `next_node_calculation` blocks are executed.
+* The parsed response is passed to the block as the only argument and by convention is named `response`.
+* The block return value is not used and no state variable is stored.
+
+> The use of these blocks is encouraged; however, they should only ever be used to store a single, `calculator` state variable (in the first question definition); otherwise they should only be used to store user responses on that `calculator` object.
+
+##### `next_node_calculation(variable_name, &block)`
+
+* These blocks were intended to be used to store state variables that are needed in the `next_node` block.
```
leenagupte: How does it differ from `on_response`?
When should it be used?

doc/flow-definition.md
```diff
@@ -1,11 +1,308 @@
...
+* These blocks are executed after all the `on_response` blocks have been executed and before any of the `validate` blocks are executed.
+* The parsed response is passed to the block as the only argument and by convention is named `response`.
+* The block return value is stored on the state object as a state variable named `variable_name`.
+
```
leenagupte: Should they be used when you need a named state variable?

doc/flow-definition.md
```diff
@@ -1,11 +1,308 @@
...
+> The use of these blocks is deprecated and should never be necessary. Define methods on a `calculator` object instead.
```
leenagupte: Maybe start with this.

doc/flow-definition.md
```diff
@@ -1,11 +1,308 @@
...
+
+##### `validate(message_key, &block)`
+
+* These blocks are intended to be used to validate the user response.
+* These blocks are executed after all the `next_node_calculation` blocks have been executed and before the `next_node` block is executed.
+* The parsed response is passed to the block as the only argument and by convention is named `response`.
+* If the block return value is truth-y, then no action is taken.
+* If the block return value is false-y, then:
+  * A `SmartAnswer::InvalidResponse` exception is raised with the `message_key` set as the exception message.
+  * Although this exception is handled within the app, it prevents the transition to the next node.
```
leenagupte: This is part of the previous sub-bullet, and should be moved up.

doc/flow-definition.md
```diff
@@ -1,11 +1,308 @@
...
+  * The `message_key` from the exception message is set on the built-in state variable, `error`.
+  * When the question template is re-rendered, the `error` state variable is used to lookup the appropriate validation error message in the [question template](doc/question-templates.md#-error_message-message-).
```
leenagupte: The link to the question template doesn't work.

doc/flow-definition.md
```diff
@@ -1,11 +1,308 @@
...
+##### `next_node(&block)`
+
+* There must only be one of these blocks per question definition.
+* This block is intended to determine which node comes next based on the user responses so far.
+* These blocks are executed after all the `validate` blocks have been executed and before any `save_input_as` blocks are executed.
+* The built-in state variables, `path`, `current_node` & `responses`, are updated if this block returns successfully.
+* The block return value must be the result of calling the `#question` or `#outcome` methods passing in the key of the next node - see the [next node documentation](doc/next-node-rules.md) for more details.
+
+> The use of this block type is *required*. However, it should call methods on the `calculator` state variable and not rely on the `response` argument passed into the block.
+
```
leenagupte: Are the methods on the "`calculator` state variable" another way of say a "question or outcome method defined in the flow"?

doc/flow-definition.md
```diff
@@ -1,11 +1,308 @@
...
+#### Templates
+
+See the [documentation for question templates](doc/question-templates.md).
+
+### Outcome nodes
+
+These are very similar to question nodes, however, only the following methods are available within the node definition:
+
+* [`precalculate`](#precalculatevariable_name-block)
+* [`on_response`](#on_responseblock)
```
leenagupte: You can use `on_response` in an outcome? How? I understood that on_response stores user input?

doc/flow-definition.md
```diff
@@ -1,11 +1,308 @@
...
+* [`next_node_calculation`](#next_node_calculationvariable_name-block)
+* [`calculate`](#calculatevariable_name-block)
+
+Having said that, no response is usually processed by an outcome node and so it only makes any sense to use `precalculate` blocks. Furthermore if any attempt is made to process a response when the current node is an outcome node (e.g. by hacking the URL path), an exception will be raised.
+
+> The use of `precalculate` blocks in an outcome node definition is deprecated and should never be necessary; it should always be possible to call methods on the `calculator` state variable from within the templates instead.
```
leenagupte: This contradicts the line above. Should precalculate blocks be used in an outcome or not?

doc/question-templates.md
```diff
@@ -2,9 +2,67 @@
...
+### `title(text)`
+
+* Valid for all question types
+* `text` argument is a String
+* Used as the heading (currently "h2")
+
+### `options(hash)`
+
+* Valid for multiple choice & checkbox question types
+* `hash` argument is a `Hash` of option key Strings versus option text Strings
```
leenagupte: I would change 'versus' to 'and' for readability.

doc/question-templates.md
```diff
@@ -2,9 +2,67 @@
...
+* Used to "translate" options keys for multiple choice & checkbox questions into human-friendly text
+
+### `hint(text)`
+
+* Valid for all question types
+* `text` argument is a String
+* Used as a "hint" paragraph
+* Also used to generate a label for value & postcode questions
```
leenagupte: Are postcode questions defined anywhere in the documentation? If they are, they should be linked to from here.

doc/question-templates.md
```diff
@@ -2,9 +2,67 @@
...
+
+* Valid for value & money questions
+* `text` argument is a String
+* Used as the label (following the input control)
+
+### `body(govspeak)`
+
+* Valid for all question types
+* `govspeak` argument is a String in [Govspeak][] format
+* Used to generate the main content (appearing above the input control)
```
leenagupte: Change: (appearing above the input control)
To: (appearing above the input control, e.g. the start button)

doc/flow-definition.md
floehopper: I've addressed this in a fix-up commit: e21ec46055eeab8cf9dd4720615d92a3200fedc2.

doc/flow-definition.md
floehopper: I've addressed this in a fix-up commit: 8a69bde8b464a6ccdb20edc07ffba290c21c9e75.

doc/flow-definition.md
floehopper: > This happens in all environments, right? Not just production?

Flows are preloaded in both `development` & `production`, but not in `test`. I've made this clearer in a fix-up commit: 353f8a2d0982a24aa5510048aee91a50fb22a8f1.

floehopper: > What could a request-specific state look like?

When I say "state", I mean the state (in the [generic sense][1]) of any object (e.g. instance variables). When I say "request-specific state", I mean _any_ state that is specific to an individual user request.

It is actually pretty hard, but not impossible, to access the flow instance from within the blocks which are processed at request time. I just wanted to make it clear that the _same_ flow instance is used for _every_ request and so you need to be careful that information related to the request from one user does not leak into the request from another user via the flow instance.

> Is this related to the save input as changes in #2608?

I assume you mean the plan to use `on_response` in preference to `save_input_as`. If so, then no, this paragraph is not related to that.

I've tried, but failed to improve this paragraph. Since there's not anything _incorrect_ about the paragraph as it stands, I suggest that we merge it as it stands and then chat about it as a group. Then if we think it's necessary, we can make improvements.

[1]: https://en.wikipedia.org/wiki/State_(computer_science)

doc/flow-definition.md
floehopper: I've tried to explain this a bit more in this fix-up commit: 1c4acfa1f6b5b439361e0c1fc7b943326ea53618.

doc/flow-definition.md
floehopper: Yes. I've tried to made this clearer in this fix-up commit: 1c4acfa1f6b5b439361e0c1fc7b943326ea53618.

doc/flow-definition.md
floehopper: I've added this in this fix-up commit: a55ddea57789968c18709a97535d4d2e3ab433ac.

doc/flow-definition.md
floehopper: I've made this change in a fix-up commit: e9606f11cefe238b93c67d85f4f4baf75e8b5dda.

doc/flow-definition.md
floehopper: I mean the in-question blocks. I've changed the wording and linked it to that section in a fix-up commit: 26cc2a126cea49dc739da4cd388eb8fdc852a019.

doc/flow-definition.md
floehopper: I've added three inline code blocks to show some examples of how this works. See fix-up commit: 0ea902b3081700211ac1de61ad12139b5f74c530.

doc/flow-definition.md
floehopper: I've addressed this in a fix-up commit: ed94910095934feca749a94084b8570d0172d318.

doc/flow-definition.md
floehopper: I've addressed this in a fix-up commit: e15b858263866c733b199b81e708e89ace6ee4aa.

doc/flow-definition.md
floehopper: No that's not correct. In fact I think _the whole point_ of `precalculate` blocks is to define state variables needed to render the template for that question/outcome.

I wonder if you fell foul of the problem described in the note, i.e. `precalculate` doesn't work *at all* for the first question.

Anyway, I know this stuff is all very confusing, so I've re-worked the bullet points for `precalculate` and `on_response` to explain the point at which the question/outcome template is rendered. See this fix-up commit: d99874902bd13a352886f011a90f39fd895badb1. I hope that helps.

doc/flow-definition.md
floehopper: > How does it differ from `on_response`?

If you compare the bullet points and notes for the two block types, you should be able to see the differences. I'm reluctant to spell the differences out explicitly, because then I feel as if I'd need to do the same comparison for all block types and that would quickly get confusing.

The main difference is that `next_node_calculation` automatically stores the return value from the block in a state variable with the name specified by the symbol passed in to the `next_node_calculation` method; whereas `on_response` does not do this.

> When should it be used instead of `on_response`?

I feel as if this is fairly well covered by the notes on the two block types. I made an attempt at introducing some inline code examples to try to explain this better. However, having written them I felt as if they were a bit of a distraction in this (already long) document and I think they would fit better into the "refactoring" document. I plan to add them shortly.

doc/flow-definition.md
floehopper: > Should they be used when you need a named state variable?

Yes, but we are trying to move away from having *any* state variables *other* than a single "calculator" object. This is why the note for `next_node_calculation` says their use is deprecated.

doc/flow-definition.md
floehopper: > This is part of the previous sub-bullet, and should be moved up.

I'm not sure this is right. The sub-bullets are intended to explain the rough sequence of events. I've tweaked the text and numbered the bullet points to try to make this clearer. See this fix-up commit: 4b087edc9a3559ba23037a4585b4c139a4a3229b.

doc/flow-definition.md
floehopper: Well spotted! It turns out that none of the links to other pages were working, so I've fixed them in this fix-up commit: b745d051a4b994e367df242edd3574cdaeb2d122.

doc/flow-definition.md
floehopper: No. These are instance methods on an instance of a `XxxCalculator` class. The instance of this class is stored as a state variable, named `calculator`.

An example would be the [`already_getting_maternity_pay?` method][1] which is called from [this `next_node` block][2].

This is something that is explained a bit in the [documentation about new-style flows][3], albeit fairly briefly.

I think it would be better to address your question by having a session where Chris & I explain the new-style flow in person. If necessary we can then update the new-style flow documentation accordingly.

[1]: https://github.com/alphagov/smart-answers/blob/b2fe9c04efdb4efa37f2d0589ef1e536314e0537/lib/smart_answer/calculators/statutory_sick_pay_calculator.rb#L49-L51
[2]: https://github.com/alphagov/smart-answers/blob/b2fe9c04efdb4efa37f2d0589ef1e536314e0537/lib/smart_answer_flows/calculate-statutory-sick-pay.rb#L24
[3]: https://github.com/alphagov/smart-answers/blob/master/doc/refactoring.md

doc/flow-definition.md
leenagupte: Is this correct:
`on_response` and `next_node_calculation` use the response entered by the user, but they do different things with them.
`on_response` blocks can only be used to store the value of the response in a variable defined in a calculator model.
`next_node_calculation` take the response entered by the user and can apply some logic (a function etc) to it and stores it in a variable called `variable_name`. This variable is then available in the `next_node` blocks defined in the question.
?

doc/flow-definition.md
leenagupte: > Yes, but we are trying to move away from having any state variables other than a single "calculator" object. This is why the note for next_node_calculation says their use is deprecated.

This should be included in the docs somewhere. I don't recall reading it anywhere, and I wasn't aware that this is where the code is heading.

doc/flow-definition.md
leenagupte: Better, but it still doesn't read quite right. I can edit this in a later PR.

doc/flow-definition.md
leenagupte: Ok. But I think that fact I misunderstood means that it should be added to the docs.

I also we are suffering from too many uses and meanings of "calculator".

doc/flow-definition.md
floehopper: > You can use on_response in an outcome?

You can (as you can `next_node_calculation` & `calculate`), even though as you rightly say, it makes no sense.

> I understood that on_response stores user input?

Yes, that is correct. Or at least that's what it's intended to be used for.

I was trying to document the fact that currently these 4 methods (including `on_response`) are all _technically_ available in an outcome, even though only `precalculate` blocks make any sense.

I've re-jigged the section to try to make this clearer in this fix-up commit: aab40ec39b5f4b03ffb87d7d4f078dd8d7d33666.

doc/flow-definition.md
floehopper: I don't think it does contradict it. To be clear, `precalculate` blocks **can be** and **are being** used in outcomes, but they are **deprecated**.

They are deprecated, because we want to call methods on the `calculator` state variable from within the templates instead. This commit shows an example of how we've been making this kind of change: 9bd3dd9f12b098ec89f3d6e9ce42a9b9ce71792f.

I think the changes I made in this fix-up commit should make things clearer: aab40ec39b5f4b03ffb87d7d4f078dd8d7d33666.

doc/question-templates.md
floehopper: I've addressed this in this fix-up commit: ad99603552d97c19289928372d4d2c4a50336187.

doc/question-templates.md
floehopper: I've converted the bullet points in `question-types.md` into headings so they are linkable and I've added the links you suggested in this fix-up commit: 4c15b16c80de46819a0f0d7bd52e59177e195ed3.

doc/question-templates.md
floehopper: The start button only appears on landing page templates which are [documented separately][1]. However, I've added a more appropriate example of an input control in this fix-up commit: e24c2f54434349a820d6ff3ea30ec5eda8a39032. 

[1]: https://github.com/alphagov/smart-answers/blob/improve-flow-definition-documentation/doc/landing-page-template.md#bodygovspeak

doc/flow-definition.md
floehopper: Do you mean the one about [storing data][1] or another document? I'm not convinced that the "storing data" document is particularly relevant to this section, so I'm going to leave this for now.

[1]: https://github.com/alphagov/smart-answers/blob/improve-flow-definition-documentation/doc/storing-data.md

doc/flow-definition.md
floehopper: That's pretty much correct, although I think it would be clearer to call the "variable" on the calculator an "attribute" to distinguish it from the concept of a "state variable".

The `on_response` block type is something we have introduced, because we're trying to move to the situation where only a _single_ state variable (`calculator`) is *ever* set on the magic state object. Once we've done that for all flows, we plan to get rid of the magic state object altogether and use the calculator object instead.

I think this thinking is an example of something that would be better explained in person with the whole team. Once we've chatted about it, perhaps the "wiki elves" can improve what I've written here...?

doc/flow-definition.md
floehopper: > This should be included in the docs somewhere. I don't recall reading it anywhere, and I wasn't aware that this is where the code is heading.

I think this would be better addressed by improving the [refactoring documentation][1]. I want to keep this PR more focussed on the current behaviour and the [flow-definition documentation][2]. And as I've said elsewhere, I think this is something that would benefit from discussing with the whole team ideally in-front of a white-board.

[1]: https://github.com/alphagov/smart-answers/blob/improve-flow-definition-documentation/doc/refactoring.md
[2]: https://github.com/alphagov/smart-answers/blob/improve-flow-definition-documentation/doc/flow-definition.md

doc/flow-definition.md
floehopper: I understand what you mean, but, for the moment, I prefer to have the current behaviour first followed by notes about deprecation, etc.

floehopper:
I've now addressed all the feedback on this PR and I'm going to create a new branch, rebase interactively to apply all the fix-up commits, rebase against master, and get it merged into master.

floehopper:
Superseded by #2625. Closing.

doc/question-templates.md
leenagupte: Ok, that makes more sense. Thanks.

PR #2614, opened 2016-07-05
Refactor state-pension-topup to new-style flow (with feedback addressed)

floehopper:
## Description

Supersedes #2602.

I done some work on the calculator itself, but I haven't attempted to e.g. extract domain concepts, and some of the logic could possibly be simplified. Since some of the logic in the calculator was quite complicated, I ended up adding quite a few unit tests to check I wasn't breaking anything. However, there is still quite a bit of work to do to move all the tests for this flow towards the [new-style approach][2].

The logic in the calculator would also be considerably improved if we extracted domain concepts i.e. make the code more intention revealing as opposed to just procedural. However, I'm going to stop at this point, because the PR already has a *lot* of commits.

## External changes

### DIFF
--- a/lib/smart_answer/calculators/state_pension_topup_calculator.rb
+++ b/lib/smart_answer/calculators/state_pension_topup_calculator.rb
@@ -18,6 +18,10 @@ module SmartAnswer::Calculators
       @gender ||= 'female'
     end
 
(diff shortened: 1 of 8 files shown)


floehopper:

PR #2622, opened 2016-07-06
Split marriage-abroad outcome for Poland

chrisroos:
The current outcome template (outcome_opposite_sex_marriage_in_poland_when_residing_in_uk_or_poland) contains two logical outcomes:

* One for residents of UK
* One for residents of Poland

The outcome body contains multiple `if/else` statements that, when combined,
essentially results in one of two logical outcomes being displayed.  Splitting
these logical outcomes and moving the "routing logic" to the flow is more
consistent with the majority of this Smart Answer. Note that there is some
common content between these logical outcomes but there is more that's
different so I still think it makes sense to split the template.


floehopper:
LGTM 👍

chrisroos:
I've rebased this branch on master and force pushed in preparation for merging.

PR #2626, opened 2016-07-07
Split marriage-abroad outcome for opposite sex ceremonies in affirmation countries

chrisroos:
This branch extracts 15 country-specific outcome templates from the rather complicated outcome_opposite_sex_marriage_in_affirmation_countries template. I think this probably results in a small amount of duplicated content (that isn't already in partials) between the new templates. I'm hoping that we can gradually remove that over time.

The following list (corresponding to the `OS_AFFIRMATION_COUNTRIES` array) contains the 20 countries that previously lead to the outcome_opposite_sex_marriage_in_affirmation_countries template. I've indicated (with `(*)`) the countries that now have their own outcomes:

* Belgium (*)
* Cambodia
* Colombia
* China (*)
* Ecuador (*)
* Egypt (*)
* Lebanon
* Finland
* Macao (*)
* Mongolia
* Morocco (*)
* Norway (*)
* Peru (*)
* Philippines (*)
* Qatar (*)
* South Korea (*)
* Thailand (*)
* Turkey (*)
* United Arab Emirates (*)
* Vietnam (*)

There are quite a few commits but I don't think they should be _too_ hard to follow.

The first two commits inline two partials to the outcome_opposite_sex_marriage_in_affirmation_countries template in preparation for extracting country-specific outcomes.

There are then 4 commits for each of the 15 countries:

1. Add separate condition for the country to marriage-abroad flow

2. Add new outcome template for the country. The template is a duplicate outcome_opposite_sex_marriage_in_affirmation_countries.

3. Remove country-specific logic and content from outcome_opposite_sex_marriage_in_affirmation_countries

4. Remove all logic and content not related to the country from the outcome template created in step 2.

I generated regression test artefacts for all countries before making this change and used those to double check that I haven't altered the behaviour of the system.


lib/smart_answer_flows/marriage-abroad/outcomes/outcome_opposite_sex_marriage_in_qatar.govspeak.erb
```diff
@@ -7,321 +7,55 @@
...
-  <% if calculator.ceremony_country == 'united-arab-emirates' %>
-    ^There are no civil marriages in the United Arab Emirates, but you can get married through a religious ceremony at a church or Sharia court.^
-
-    You can only get married in the United Arab Emirates if you or your partner are resident there.
-  <% end %>
-
-  <% if %w(turkey egypt china).include?(calculator.ceremony_country) %>
-    <%= render partial: 'what_you_need_to_do.govspeak.erb',
-        locals: { calculator: calculator } %>
-  <% elsif calculator.opposite_sex_21_days_residency_required? %>
```
floehopper: Minor: It's hard to see this in the diff without the `--ignore-all-spaces` option enabled, but some "What you need to do" content related to the `opposite_sex_21_days_residency_required` is retained in this commit. It wasn't immediately obvious to me why this was the case. Having investigated it a bit, I see that `qatar` is in the list of `OS_21_DAYS_RESIDENCY_REQUIRED_COUNTRIES` in the Marriage Abroad data query class. I think it would be worth mentioning this in the commit note.

test/data/marriage-abroad-files.yml
```diff
@@ -123,14 +125,25 @@ lib/smart_answer_flows/marriage-abroad/outcomes/outcome_opposite_sex_marriage_in
...
 : e4ff64b9911bd6b8d2d980f6311db678
+lib/smart_answer_flows/marriage-abroad/outcomes/outcome_opposite_sex_marriage_in_south_korea.govspeak.erb: 9e8bce8e41660249e19361fae264462b
+lib/smart_answer_flows/marriage-abroad/outcomes/outcome_opposite_sex_marriage_in_thailand.govspeak.erb: 35b669954974ce4ae10b92d0ac3a8479
+lib/smart_answer_flows/marriage-abroad/outcomes/outcome_opposite_sex_marriage_in_turkey.govspeak.erb: 30f42fd8b80b4a25fc464c6e5b4f44d9
+lib/smart_answer_flows/marriage-abroad/outcomes/outcome_opposite_sex_marriage_in_united_arab_emirates.govspeak.erb: a959cbe9a2171aeb49ff2a786f54a9bd
+lib/smart_answer_flows/marriage-abroad/outcomes/outcome_opposite_sex_marriage_in_vietnam.govspeak.erb: 7de688fb25bf157202b76e1dc960a07c
 lib/smart_answer_flows/marriage-abroad/outcomes/outcome_opposite_sex_marriage_in_yemen.govspeak.erb: 75931823b0856f97977823a467d25ec9
 lib/smart_answer_flows/marriage-abroad/outcomes/outcome_same_sex_civil_partnership.govspeak.erb: 32a91cace111cb531a1e895307785f71
 lib/smart_answer_flows/marriage-abroad/outcomes/outcome_same_sex_civil_partnership_in_affirmation_countries.govspeak.erb: 959fc1ca1916dbf62a8f2e1e8aed0d39
```
floehopper: When I spoke to you earlier, you said you'd run the regression tests for all countries before and after the changes in this PR. If that's the case, I think you should remove the "TODO" in the commit note and explain what you've done.

floehopper:
Do you know whether the non-exceptional path, i.e. the `outcome_opposite_sex_marriage_in_affirmation_countries.govspeak.erb` template is still needed? I'm guessing that it is, but I think it would be worth saying so in the PR description.

floehopper:
Do you think this has resulted in much _extra_ duplication across the new templates? If not, I think it would be worth saying so. If so, I think it might be worth explaining how you'd imagine that duplication being removed.

floehopper:
Overall I think this looks like a big improvement. Great work!

Other than the "TODO" in the last commit note, I'm happy that this is merged as it stands. 👍

floehopper:
I also meant to say that although I tried hard to check all the changes in the diffs, I found the commits titled "Update <country> outcome for opposite sex ceremonies" quite hard to follow, even with space ignoring switched on. However, I'm still happy for this to be merged on the basis that the regression tests _for all countries_ pass.

lib/smart_answer_flows/marriage-abroad/outcomes/outcome_opposite_sex_marriage_in_qatar.govspeak.erb
chrisroos: I've updated the commit note to say:

> Note that Qatar is in the `OS_21_DAYS_RESIDENCY_REQUIRED_COUNTRIES` array so
> I've retained the "##What you need to do" section within the `elsif
> calculator.opposite_sex_21_days_residency_required?` condition.

test/data/marriage-abroad-files.yml
chrisroos: I've removed the TODO and copied the following text from the PR description to the commit note:

> I generated regression test artefacts for all countries before making this change and used those to double check that I haven't altered the behaviour of the system.

chrisroos:
> Do you know whether the non-exceptional path, i.e. the `outcome_opposite_sex_marriage_in_affirmation_countries.govspeak.erb` template is still needed? I'm guessing that it is, but I think it would be worth saying so in the PR description.

I've updated the PR description to explain that this template is still required. I've listed the 20 countries that used to lead to this template, and highlighted the 15 that now have their own templates.

chrisroos:
> Do you think this has resulted in much extra duplication across the new templates? If not, I think it would be worth saying so. If so, I think it might be worth explaining how you'd imagine that duplication being removed.

I've add the following two sentences to the PR description:

> I think this probably results in a small amount of duplicated content (that isn't already in partials) between the new templates. I'm hoping that we can gradually remove that over time.

chrisroos:
> I also meant to say that although I tried hard to check all the changes in the diffs, I found the commits titled "Update outcome for opposite sex ceremonies" quite hard to follow, even with space ignoring switched on. However, I'm still happy for this to be merged on the basis that the regression tests for all countries pass.

Thanks for reviewing this PR, @floehopper. I appreciate that it wasn't the easiest thing to review!

chrisroos:

PR #2629, opened 2016-07-12
Refactor register-a-death to new-style flow

floehopper:
## Description

Trello card: https://trello.com/c/jjPO9Leq

Similar to #2619, this moves the flow towards the style described in the [refactoring documentation][1]. I've concentrated on moving logic out of the flow and into the calculator and presentational concerns into the templates. The idea is to change all the flows to use this approach so we can significantly simplify the DSL.

In this case, I haven't done much work on the calculator itself other than extracting methods from the flow into it. Also I haven't attempted to extract any domain concepts. Some of the logic could almost certainly be simplified. There is still quite a bit of work to do to move all the *tests* for this flow towards the [new-style approach][2].

In extracting logic into the new calculator class, I've tried to keep things as similar as possible to the calculator class for `register-a-birth`. In fact this refactoring has made it clear that there is quite a bit of duplication between the two which could almost certainly be removed.

Although the PR has a lot of commits, the vast majority of them are very small and only affect a few files. Also because of the similarities between this flow and `register-a-birth`, I've tried to structure the commits in a similar way as when I refactored `register-a-birth`.

## External changes

### DIFF
--- a/lib/smart_answer/calculators/register_a_death_calculator.rb
+++ b/lib/smart_answer/calculators/register_a_death_calculator.rb
@@ -36,5 +36,9 @@ module SmartAnswer::Calculators
     def responded_with_commonwealth_country?
       RegistrationsDataQuery::COMMONWEALTH_COUNTRIES.include?(country_of_death)
     end
+
+    def in_the_uk?
+      current_location == 'in_the_uk'
+    end
   end
 end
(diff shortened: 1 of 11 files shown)


floehopper:
I've used an interactive rebase to fix the Rubocop offence in the relevant commit and have force-pushed.

floehopper:
And now I've amended the final commit to update the checksums and force-pushed. Hopefully the latest build should pass.

chrisroos:

floehopper:

PR #2630, opened 2016-07-12
Consistently display overseas-passport courier info

pmanrubia:
Consistently display the courier information in overseas-passports. The cost is now always on a separate line to the information about what you're paying for. This brings the display of this information inline with that for Tajikistan, Turkmenistan and Uzbekistan. This should help us remove duplication in future.

We spoke to Ben Mortimer about making this change and he's happy for it to be deployed. No fact check required.

### DIFF
--- a/test/artefacts/overseas-passports/falkland-islands/applying/adult/afghanistan.txt
+++ b/test/artefacts/overseas-passports/falkland-islands/applying/adult/afghanistan.txt
@@ -15,7 +15,9 @@ Don’t book travel until you have a valid passport. If you need to travel more 
 
 ## Cost
 
-You’ll have to pay a fee for your passport and a courier fee of £19.86. The courier fee pays for your passport and supporting documents to be sent back to you securely.
+You’ll have to pay a fee for your passport and a courier fee of £19.86.
+
+The courier fee pays for your passport and supporting documents to be sent back to you securely.
 
 | Passport type | Passport fee | Total to pay (including courier fee) |
 |---------------|--------------|--------------------------------------|
--- a/test/artefacts/overseas-passports/falkland-islands/replacing/adult.txt
+++ b/test/artefacts/overseas-passports/falkland-islands/replacing/adult.txt
@@ -17,7 +17,9 @@ Don’t book travel until you have a valid passport. If you need to travel more 
 
 ## Cost
 
-You’ll have to pay a fee for your passport and a courier fee of £9.70. The courier fee pays for your passport to be sent back to you securely.
+You’ll have to pay a fee for your passport and a courier fee of £9.70.
+
+The courier fee pays for your passport to be sent back to you securely.
 
 | Passport type | Passport fee | Total to pay (including courier fee) |
 |---------------|--------------|--------------------------------------|
--- a/lib/smart_answer_flows/overseas-passports/outcomes/_cost.govspeak.erb
+++ b/lib/smart_answer_flows/overseas-passports/outcomes/_cost.govspeak.erb
@@ -36,7 +36,9 @@
   <%= render partial: 'cost/payment_instructions.govspeak.erb', locals: { calculator: calculator } %>
 
 <% elsif calculator.replacing? and calculator.ips_number == '1' and calculator.ips_docs_number == '1' %>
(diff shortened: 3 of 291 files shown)


test/artefacts/overseas-passports/falkland-islands/replacing/adult.txt
```diff
@@ -17,7 +17,9 @@ Don’t book travel until you have a valid passport. If you need to travel more
 
 ## Cost
 
-You’ll have to pay a fee for your passport and a courier fee of £9.70. The courier fee pays for your passport to be sent back to you securely.
+You’ll have to pay a fee for your passport and a courier fee of £9.70.
+
+The courier fee pays for your passport to be sent back to you securely.
```
floehopper: Minor: I assume `falkland-islands` and `the-occupied-palestinian-territories` are the only regression test artefacts to change, because they are the only "IPS 1" countries. However, since I have no idea what "IPS 1" means, this wasn't at all obvious. It might be worth explaining in the commit note.

test/artefacts/overseas-passports/falkland-islands/applying/adult/afghanistan.txt
```diff
@@ -15,7 +15,9 @@ Don’t book travel until you have a valid passport. If you need to travel more
 
 ## Cost
 
-You’ll have to pay a fee for your passport and a courier fee of £19.86. The courier fee pays for your passport and supporting documents to be sent back to you securely.
+You’ll have to pay a fee for your passport and a courier fee of £19.86.
+
+The courier fee pays for your passport and supporting documents to be sent back to you securely.
```
floehopper: Minor: It seems slightly surprising to me that the artefacts for `falkland-islands` are changing in this commit as well as in the one for "IPS 1" countries.

floehopper:
Ideally I think it would be good if the message for commits involving changes in behaviour of _groups_ of countries (e.g. "IPS 1") could list the countries that we expect to be affected and how this list was obtained. I don't think it's reasonable to expect a reviewer to know what "IPS 1" means.

floehopper:
Other than my comments, this looks good to me. 👍

floehopper:
Oh, also it would be nice to say something in the PR description about the missing regression test coverage you found and whether you are planning to do anything about it.

test/artefacts/overseas-passports/falkland-islands/replacing/adult.txt
pmanrubia: I'm afraid we (@chrisroos and I) also don't know what "IPS1" countries are. We're happy to accept the changes to the regression test artefacts as they match the change we've made to the template. We've added the following message to each of the commit notes that mention IPS1/2/3:

> We don't know what defines IPS1 countries. This just seemed like a
natural way to split the changes in the branch. We're happy to accept
the changes to the regression test artefacts because they match the
change we made to the template.

test/artefacts/overseas-passports/falkland-islands/applying/adult/afghanistan.txt
pmanrubia: We're fairly confident that this is because the original changes were related to replacing a passport, and these are related to applying for a passport.

pmanrubia:
> Ideally I think it would be good if the message for commits involving changes in behaviour of groups of countries (e.g. "IPS 1") could list the countries that we expect to be affected and how this list was obtained. I don't think it's reasonable to expect a reviewer to know what "IPS 1" means.

As per our comment above, we (@chrisroos and I) also don't know what the "IPS1" countries are. We've decided that it's not important to understand in this branch given that the changes to the artefacts match the changes to the templates. We've updated the commit notes to explain this. Apologies for not doing so earlier.

pmanrubia:
> Oh, also it would be nice to say something in the PR description about the missing regression test coverage you found and whether you are planning to do anything about it.

We've expanded the commit notes of the three commits that didn't result in changes to the regression test artefacts:

> NOTE. No regression tests failed after making this change. We've done
> some initial investigation and think that this path of code might be
> unreachable. We're not going to investigate any further at the moment as
> our primary goal is to make the content consistent so that we can remove
> duplication.

pmanrubia:
Thanks for reviewing, @floehopper. We've rebased this on master and force pushed in preparation for merging.

PR #2639, opened 2016-07-13
Document use of the "Passed factcheck" label

chrisroos:
@pmanrubia and I added this label to GitHub yesterday; mirroring the existing "Passed code review" label.


floehopper:
This looks good to me 👍

PR #2642, opened 2016-07-14
SPIKE: Links from content to template/partial on GitHub

floehopper:
## Description

Trello card: https://trello.com/c/X7t256s9

I've prototyped a way to get from a piece of rendered content to the source template (or partial template) on GitHub. Currently this works as follows:

* Chunks of Govspeak-generated HTML are wrapped in a `<div>` element.
  * CSS classes are used to make the elements easily identifiable.
  * Data attributes are used to make the template path and content key available in the DOM
* Govspeak generated from a partial template is wrapped in a `<div>` element
  * CSS classes are used to make the elements easily identifiable.
  * Data attributes are used to make the template path and content key available in the DOM
  * The `markdown="1"` attribute means that Govspeak/Kramdown will still convert the wrapped text.
* Some JavaScript in the main application highlights the `<div>` elements mentioned above to the user.
  * Border, padding & margin are added to each element.
  * A link to the source template on GitHub is added (`release` branch).

## Heroku

I've deployed this to Heroku so people can have a play with it and see what they think. Here are a few URLs which show it in action:

* https://smart-answers-pr-2642.herokuapp.com/marriage-abroad
* https://smart-answers-pr-2642.herokuapp.com/marriage-abroad/y/italy/uk/partner_local/opposite_sex
* https://smart-answers-pr-2642.herokuapp.com/calculate-married-couples-allowance/y/yes/yes/1950-01-01/27701.0/yes
* https://smart-answers-pr-2642.herokuapp.com/calculate-married-couples-allowance/y/yes/yes/1950-01-01/27701.0/yes/123.0/2342.0/2342.0

## To Do

- [ ] Is it worth making it work for single-line/non-Govspeak content?
- [ ] Should probably only be available in some environments or require a query string parameter.
- [ ] Is the `release` branch the right place to links to?
- [ ] Would it better to have the JS in a bookmarklet i.e. not part of the main app?
- [ ] Possibly put the JS in the [GOV.UK Toolkit Chrome extension](https://github.com/alphagov/govuk-toolkit-chrome)?
- [ ] Prepending the `PartialTemplateWrapper` module into `ActionView::PartialRenderer` probably isn't very robust and the initializer should probably fail fast if the version of Rails changes significantly.
- [ ] Fix failing tests
- [ ] Add tests for new code

### Minimum wage calculator for employers

2016-07-14 at 16 08

### Marriage abroad

2016-07-14 at 16 09

### Calculate your Married Couple's allowance (question page)

2016-07-14 at 16 12

### Calculate your Married Couple's allowance (outcome page)

2016-07-14 at 16 10

### DIFF
--- a/app/assets/javascripts/smart-answers.js
+++ b/app/assets/javascripts/smart-answers.js
@@ -164,3 +164,25 @@ $(document).ready(function() {
   contentPosition.init();
 
 });
+
+$(document).ready(function() {
+  $('.debug.template').each(function() {
+    var element = $(this);
+    var path = element.data('path');
+    var filename = path.split('/').pop();
+    var key = element.data('key');
+    var url = 'https://github.com/alphagov/smart-answers/blob/release/' + path;
+    var anchor = $('<a>Template on GitHub (key: ' + key + ')</a>').attr('href', url).attr('style', 'color: deeppink;').attr('title', filename);
+    element.prepend(anchor);
+    element.attr('style', 'border: 3px solid deeppink; padding: 10px; margin: 3px');
+  });
+  $('.debug.partial-template').each(function() {
+    var element = $(this);
+    var path = element.data('path');
+    var filename = path.split('/').pop();
+    var url = 'https://github.com/alphagov/smart-answers/blob/release/' + path;
+    var anchor = $('<a>Partial template on GitHub</a>').attr('href', url).attr('style', 'color: hotpink;').attr('title', filename);
+    element.prepend(anchor);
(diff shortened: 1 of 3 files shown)


floehopper:
I've just pushed up some even more hack-y changes to have something similar working for non-govspeak / single-line content. I've also deployed those changes to the Heroku app. /cc @chrisroos

floehopper:
Note to self: maybe use the `deployed-to-production` branch instead of `release`...?

floehopper:
I showed this to a group of Content Designers on Thursday last week and they thought it would be useful, so I'm going to work on production-ising the code. In the meantime, I'm closing this PR.

PR #2035, opened 2015-10-26
Convert pay-leave-for-parents to Ruby

chrisroos:
This PR converts the final Smartdown flow (pay-leave-for-parents) to Ruby. Although it’s really quite large I hope the commits are relatively easy to follow. I’m fairly confident that the converted Smart Answer mirrors the original as the only regression test artefact changes are due to spaces and blank lines.

I’m convinced this code can be improved but I’ve tried to make this as much of a one-to-one conversion from Smartdown as possible.

Here’s an overview of the changes:

* First 7 commits update the Smartdown conversion tool to cope with additional requirements of the pay-leave-for-parents Smart Answer.

* Next 2 commits use the Smartdown conversion tool to generate the Ruby version of pay-leave-for-parents, and to rename the Smartdown version so that the names don’t clash.

* Next 7 commits manually copy the next node rules from the Smartdown questions to the Flow.

* Next 2 commits reorder the flow so that `employment_status_of_mother` and `employment_status_of_partner` appear in the code in the order they appear in the flow.

* Next commit adds `save_input_as` to all questions. This should make the conversion easier at it replicates Smartdown functionality.

* Next commit updates the flow class name.

* Next commit adds a placeholder calculator to the flow.

* Next 11 commits add `precalculate` blocks to questions to ensure question text interpolation works as expected. This involves copying methods to the calculator where appropriate.

* Next 7 commits replace the commented out `next_node` logic with equivalent Ruby code.

* Next 7 commits add permitted next nodes to the questions that use `next_node` with a block.

* Next 7 commits replace Smartdown conditionals (e.g. `$IF`) with ERB in outcome partials.

* Next 5 commits replace YAML interpolation tags with ERB in outcome partials.

* Next 2 commits add a script to convert the Smartdown scenarios to integration tests, and the converted integration tests.

* Next commit updates responses and expected results.

* Next commit moves the outcomes to their own directories. This is a change that’s happened in master since this branch was created.

* Next commit updates checksum data.

* Next commit reverts change to preload flows if running regression tests for pay-leave-for-parents. This is no longer necessary now that it’s been converted to Ruby.

* Next 13 commits manually rename pay-leave-for-parents test artefact directories. This should make it easier to see if there are any important differences when the regression tests run.

* Next 3 commits manually update pay-leave-for-parents test artefacts for questions, to find/replace date and salary amounts. This should make it easier to see if there are any important differences when the regression tests run.

[...]

PR #2643, opened 2016-07-15
Quick improvements to views: Remove unused partials and improve folder structure

selfthinker:
[Trello card](https://trello.com/c/yszihFOa/221-easy-frontend-fixes-and-cleanup)

While looking at the code I identified some quick things to improve.
I **removed 3 partials** which don't seem to be used. I searched the code base for them and couldn't find any reference. I found only a 100% [proof for `optional_date_question`](https://github.com/alphagov/smart-answers/commit/6275d7689549de9763c3b2ff2987555f525f3687). Although I found [where `country_question` and `text_question` were removed](https://github.com/alphagov/smart-answers/pull/2050), it was difficult to find evidence for their substitutes.
I also **improved the folder structure of the views** by moving all question types into their own folder.

## Expected change

Nothing visible/noticeable for the user. I ran the regression tests and they passed.


floehopper:
> Although I found where country_question and text_question were removed, it was difficult to find evidence for their substitutes.

Well done for tracking things down in the git history.

That PR (#2035) removed all Smartdown related functionality. The motivations behind the removal of Smartdown are documented in [ADR 1: Remove Smartdown code](https://github.com/alphagov/smart-answers/blob/master/doc/arch/adr-001-removing-smartdown.md).

There were only two published Smartdown flows:

* [Convert landlord-immigration-check from Smartdown to Smart Answer](https://github.com/alphagov/smart-answers/pull/1900)
* [Convert pay-leave-for-parents to Ruby](https://github.com/alphagov/smart-answers/pull/2035)

I imagine that the `country_question` and `text_question` were not used by either of these flows and so there was no need to create the non-Smartdown equivalent when they were converted.

I'm not sure whether I've explained that very well, so do ask me if it doesn't. In any case, it's not hugely important - the main thing is that the partials you have removed are not currently used.

pmanrubia:
@selfthinker did you run the regression tests after the last commit?

selfthinker:
As I mentioned in the description, I did run the regression tests and they passed. But now that there is a conflict, I can resolve it and run the tests again afterwards.

pmanrubia:
LGTM

PR #2644, opened 2016-07-18
Improve documentation about creating a new Smart Answer

chrisroos:
I've added example content for the landing, question and outcome page templates. I think this leaves us with a _slightly_ more realistic Smart Answer.


floehopper:
LGTM 👍

PR #2648, opened 2016-07-20
Add "Deploy to Heroku" button to README

floehopper:
## Description

Trello card: https://trello.com/c/X40d9w9t

This PR adds a [Heroku Deploy button][1] to the repo README. The idea is to make it easier for developers and content designers to deploy a version of the app for previewing and/or fact-checking purposes.

2016-07-20 at 15 59

The URL for this button [uses an implicit template][2] which means that Heroku determines the fork and branch of the repo to be deployed from the "Referer" HTTP header. This should mean that content designers can use this button to deploy changes from a fork of the canonical repo.

In order to make the button work, I've added an `app.json` file (based on earlier work by @chrisroos) to the root directory of the repo. This file is used by Heroku to configure the app. The environment variable settings are copied from the `startup_heroku.sh` script. At some point we might be able to do away with the latter script and just use the button, but that's a job for another day.

I've also introduced a home page to the app which renders a simple list of links to the landing pages of all the flows. This means we can sensibly set the `success_url` in `app.json` to be the root path. The user can then navigate to the flow which they are interested in.

2016-07-20 at 16 03

The new home page should only be visible when the app is running standalone i.e. not behind an instance of the GOV.UK router. This is because the URL is not registered with Panopticon. Thus you will not be able to see it in integration, staging or production.

## External changes

None, although it's worth bearing in mind that the Heroku Deploy button should be visible near the top of the README on the GitHub repo.

[1]: https://devcenter.heroku.com/articles/heroku-button
[2]: https://devcenter.heroku.com/articles/heroku-button#using-an-implicit-template

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,8 @@
 
 Read more in [Lisa Scott's GDS blog post](https://gds.blog.gov.uk/2012/02/16/smart-answers-are-smart/).
 
+[![Deploy](https://www.herokucdn.com/deploy/button.svg)](https://heroku.com/deploy)
+
 ## Screenshots
 
 ![Student Finance Forms screenshot](./doc/assets/govuk-student-finance-forms.png)
--- a/app.json
+++ b/app.json
@@ -1,0 +1,16 @@
+{
+  "name": "Smart Answers",
+  "website": "https://www.gov.uk/",
+  "repository": "https://github.com/alphagov/smart-answers",
+  "success_url": "/",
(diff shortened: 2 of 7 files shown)


chrisroos:
I'm going to review this now.

chrisroos:

floehopper:
Thanks. Rebasing against master and force-pushing in preparation for merging.

PR #2649, opened 2016-07-20
Enlarge hit area of checkboxes

selfthinker:
[Trello card](https://trello.com/c/FM41X6a8/227-accessibility-bug-checkboxes-have-no-large-hit-area)

This gives checkbox questions the standard design and makes it consistent with multiple choice questions. It increases the checkboxes' hit area so that it's easier to hit the otherwise tiny boxes.

Any flow with a checkbox question will be affected:

* benefit-cap-calculator
* energy-grants-calculator
* legalisation-document-checker
* maternity-paternity-calculator
* simplified-expenses-checker
* student-finance-calculator

I ran the regression tests after the commit again and they passed.

## Expected changes

Checkboxes will change their design from a non-standard to a standard design, which increases the hit area.
You can test it out on e.g. `/calculate-statutory-sick-pay/y`

Before:
screen shot 2016-07-20 at 16 58 21

After:
screen shot 2016-07-20 at 16 57 43

### DIFF
--- a/app/views/smart_answers/inputs/_checkbox_question.html.erb
+++ b/app/views/smart_answers/inputs/_checkbox_question.html.erb
@@ -1,7 +1,7 @@
 <ul class="options">
   <% question.options.each.with_index do |option, i| %>
     <li>
-      <label for="response_<%= i %>">
+      <label for="response_<%= i %>" class="selectable">
         <%= check_box_tag "response[]", option.value, prefill_value_includes?(question, option.value), id: "response_#{i}" %>
         <%= option.label %>
       </label>
--- a/test/artefacts/benefit-cap-calculator/default/yes/no/no.html
+++ b/test/artefacts/benefit-cap-calculator/default/yes/no/no.html
@@ -42,91 +42,91 @@
 
       <ul class="options">
     <li>
-      <label for="response_0">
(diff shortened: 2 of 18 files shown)


chrisroos:
If possible, it'd be good to link to a styleguide/reference page that explains that we should be using this `selectable` class for checkboxes.

Aside from my minor comment, this looks good to me @selfthinker.

selfthinker:
Unfortunately, we don't seem to have a styleguide for [static](https://github.com/alphagov/static) where these styles are coming from. (At least I couldn't find any and asked a couple of days ago in the frontend Slack channel, but got no reply.)
The **pattern** however is described in [Elements](http://govuk-elements.herokuapp.com/form-elements/#form-checkboxes), but we don't use that (yet) and it uses different HTML.

Shall I add a reference to Elements in the commit message anyway, even though Smart Answers doesn't use it? I guess it would make more sense to reference the radio buttons (i.e. the `multiple_choice` question type) within Smart Answers as that is using the same class.

chrisroos:
Hey @selfthinker. I think you could almost copy your entire reply to me into the commit message. I think that would give plenty of context for someone coming to this in future.

selfthinker:
I updated the commit message accordingly and also rebased with origin/master.

PR #2656, opened 2016-07-22
Add draft Smart Answer with all question types

selfthinker:
[Trello card](https://trello.com/c/tO7drViD/230-create-example-flow-with-all-question-types)

This is not intended to ever be live, it was specifically done to make it easier to work on all the front-end elements. Therefore no tests are necessary.

I would also like to add simple error validation to all elements. That might be added in a separate PR.
I added four different date questions, because it is very likely we will have more than one type of date question in the future.

## Expected changes

None on production. On other environments a new Smart Answer gets added under `/all-smart-answer-questions`.

### DIFF
--- a/lib/smart_answer_flows/all-smart-answer-questions.rb
+++ b/lib/smart_answer_flows/all-smart-answer-questions.rb
@@ -1,0 +1,123 @@
+# not a real smart answer, this is here to make testing and changing all front-end elements easier
+module SmartAnswer
+  class AllSmartAnswerQuestionsFlow < Flow
+    def define
+      name 'all-smart-answer-questions'
+      status :draft
+
+      additional_countries = [OpenStruct.new(slug: "mordor", name: "Mordor")]
+
+      checkbox_question :which_checkboxes? do
+        option :radagast
+        option :mithrandir
+        option :olorin
+        option :tharkun
+        option :elrohir
+
+        next_node do
+          question :which_country?
+        end
+      end
+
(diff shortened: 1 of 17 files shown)


floehopper:
A quick thought: it might be worth adding a link to each question page from the [recently added](550fcdb6d4fd7205286c1a34176e7226983527ae) [index page](https://github.com/alphagov/smart-answers/blob/master/app/views/smart_answers/index.html.erb) which is only available in development.

selfthinker:
I'm not sure if the index page would be a good place, it'd be muddling the purposes a bit. I wonder if it makes sense to link all the questions on the landing page for this flow?
Then again, all the pages are already linked from the results page via the "Previous answers" section. I could add a link to that on the landing page? E.g. `/all-smart-answer-questions/y/mithrandir,olorin,tharkun/mordor/2016-07-27/1970-01-01/2020-02-29/0000-12-25/9999999.0/yes/WC2B%206NH/500000.0-month/world/0/0.5`

selfthinker:
FYI, I just made a change and force-pushed. The change is adding a second multiple choice question (`which_boolean_choice`) as a Yes/No question will look different to others, and changing the previous `which_choice` question to not be a Yes/No/Maybe question.

floehopper:
This looks good to me. 👍

selfthinker:
Thanks, I just rebased with origin/master and also added a link to the results on the landing page to make it easy to navigate to all the different questions individually.

PR #2659, opened 2016-07-25
Use text version of question pages as regression test artefacts

floehopper:
## Description

Trello card: https://trello.com/c/P6tuDjcg

This introduces a text version of question pages and converts all existing regression test artefacts for question pages to the new text version.

Note that we lose a little "end-to-end" test coverage of the "previous answers" section on question pages. However, this was only ever incidental test coverage and it's not the right place to test such functionality. I suggest this is better addressed by making sure we have sufficient coverage before making changes in that area of the code.

## Motivation

The idea is to make the regression tests less brittle for question pages, e.g. when @selfthinker is making accessibility improvements to the HTML.

## External changes

The new text version of question pages should be accessible in the integration environment, e.g. https://www-origin.integration.publishing.service.gov.uk/calculate-statutory-sick-pay/y.txt

However, the new text version of question pages will *not* be visible in staging or production environments.


chrisroos:
I'm starting to review this now.

test/artefacts/calculate-agricultural-holiday-entitlement/different-number-of-days.txt
```diff
@@ -0,0 +1,12 @@
...
+
+
+This is the first day that you will take off work as leave
+
+
+
+
+
+
+
```
chrisroos: It doesn't look as though `date_question` options are appearing in the question page artefacts. Is that intentional?

chrisroos:
I found an example question page artefact for each type of question to help me check the changes:

* [`multiple_choice` example](https://github.com/alphagov/smart-answers/blob/use-text-version-of-question-pages-as-regression-test-artefacts/test/artefacts/additional-commodity-code/0.txt)
* [`country_select` example](https://github.com/alphagov/smart-answers/blob/use-text-version-of-question-pages-as-regression-test-artefacts/test/artefacts/check-uk-visa/y.txt)
* [`date_question` example](https://github.com/alphagov/smart-answers/blob/use-text-version-of-question-pages-as-regression-test-artefacts/test/artefacts/calculate-agricultural-holiday-entitlement/different-number-of-days.txt)
* [`value_question` example](https://github.com/alphagov/smart-answers/blob/use-text-version-of-question-pages-as-regression-test-artefacts/test/artefacts/am-i-getting-minimum-wage/current_payment/not_an_apprentice.txt)
* [`money_question` example](https://github.com/alphagov/smart-answers/blob/use-text-version-of-question-pages-as-regression-test-artefacts/test/artefacts/calculate-married-couples-allowance/yes/yes/1950-01-01.txt)
* [`salary_question` example](https://github.com/alphagov/smart-answers/blob/use-text-version-of-question-pages-as-regression-test-artefacts/test/artefacts/pay-leave-for-parents/yes/2012-02-01/employee/employee/yes/yes.txt)
* [`checkbox_question` example](https://github.com/alphagov/smart-answers/blob/use-text-version-of-question-pages-as-regression-test-artefacts/test/artefacts/calculate-statutory-sick-pay/y.txt)
* [`postcode_question` example](https://github.com/alphagov/smart-answers/blob/use-text-version-of-question-pages-as-regression-test-artefacts/test/artefacts/landlord-immigration-check/y.txt)

Aside from the `date_question` options that I've already asked about, the only other question I have is whether we should display the options (week, month, year) for a `salary_question`? Maybe it doesn't make sense if the options are _always_ shown?

test/artefacts/calculate-agricultural-holiday-entitlement/different-number-of-days.txt
floehopper: It wasn't intentional. I only included things that were already being rendered in the templates. However, it feels as if we should be rendering the `date_question` options. I'm going to investigate how hard that will be.

floehopper:
I've chatted to @chrisroos about the issues he has raised about options for `date_question` & `salary_question` not appearing in the new text version of the question page artefacts.

The `date_question` artefacts are generated using Rails helper methods (i.e. `select_day`, `select_month`, `select_year`) within the `app/views/smart_answers/inputs/_date_question.html.erb` template. Rendering these options as part of the text version of the artefact would be awkward, because (a) there are up to 3 sets of options to display, and (b) the options to display are not readily available from the `DateQuestionPresenter`. Given the amount of work involved and the resultant additional complexity in the code, I'm not convinced rendering the options is worthwhile.

The `salary_question` artefacts are generated from a hard-coded array of strings in the `app/views/smart_answers/inputs/_salary_question.html.erb` template. We could move the hard-coded options into an `options` method on `SalaryQuestionPresenter` - that would mean they would show up in the text version of the artefact. However, there are only 3 instances of `salary_question` in published flows (all in `pay-leave-for-parents`) and I believe they are all redundant and in need of removal. Given this and the fact that they are hard-coded options (i.e. how risk of breaking), I'm not convinced rendering the options is worthwhile.

Stepping back slightly, both Chris & I are wondering how useful the question page regression test artefacts are. We might get enough value by changing the regression tests to just check that that each question page renders without an error i.e. stop checking that the content matches exactly.

Anyway, I plan to rebase this against master, resolving conflicts, and force-push in preparation for merging.

floehopper:
I've added an extra commit to update the docs which explain how to view the Govspeak version of pages.

PR #2662, opened 2016-07-26
Overseas passport - DRY up cost info

pmanrubia:
Paired with @chrisroos.

We've DRYed up the cost/fee table and associated information that were previously rendered in multiple places in both the ips_application_result_online template and the _cost partial.

We've grouped logic relating to a particular concern in the following partials:
- _courier_fee_information
- _fee_information
- _passport_fee_table
- _payment_instructions

This grouping will hopefully make it easier to remove country specific branches from the partials as they are not spread in multiples files.
## Expected changes

None. This is a refactoring and doesn't affect the external behaviour of the app.


floehopper:
Since you paired on this, perhaps it doesn't need to be code reviewed?

chrisroos:
> Since you paired on this, perhaps it doesn't need to be code reviewed?

I'll leave this for you to decide, @pmanrubia. I'd be happy for you to merge this without review given that we essentially reviewed it as we went.

pmanrubia:
Rebasing with master, and force pushing in preparation for merging

pmanrubia:
Hi @chrisroos, this card finally went live today. 
Sorry it took so much time, but we forgot about this PR 😊 

Thank you for your work on this 👍

PR #2663, opened 2016-07-26
Fix commented out code in flow_preview initializer

floehopper:
## Description

This commented out code is intended to allow a developer to use the app to view fixture flows. Since the code is commented out, inevitably it has become out-of-date. I'm not totally convinced it's worth keeping this hanging around, but in the meantime I thought it was worth fixing it.

## External changes

None. This is change is only relevant to developers who might want to un-comment out this code.


chrisroos:

floehopper:
Rebasing against master in preparation for merging.

PR #2664, opened 2016-07-26
Add content_id for all-smart-answer-questions draft flow

floehopper:
## Description

Even draft flows must have a `content_id`, because they are registered with the Publishing API in the integration environment.

I've added a test to catch this problem, but it would be better to modify the `Flow` class to effectively make it impossible to create a flow without a `content_id`.

## External changes

None in production. However, this should fix [the exception][1] that was happening in `rake publishing_api:publish` when the integration build was deployed.

[1]: https://errbit.integration.publishing.service.gov.uk/apps/533c2ee40da115303f0129a5/problems/5797b6606578632af8d47f00


chrisroos:

pmanrubia:
👍

PR #2667, opened 2016-07-28
Make template path available as data attribute

floehopper:
## Description

Trello card: https://trello.com/c/X7t256s9

This PR makes the relative path to a template available in each of the `NodePresenter` sub-classes, i.e.  `StartNodePresenter`, `QuestionPresenter` & `OutcomePresenter`. This is then added to an appropriate HTML element as a data attribute, `debug-template-path`.

The plan is to use this in [a bookmarklet to annotate content with a link to the underlying ERB template in the smart-answers repo][1]. This should make it easier for Content Designers to edit the templates directly themselves.

I hope to add functionality to do something similar for _partial_ templates, but I thought that this PR is useful in its own right.

## External changes

There should not be any visible changes. However, if you view the source of any Smart Answers page, you should be able to see a data attribute named `debug-template-path`.

[1]: https://gist.github.com/floehopper/045408d7fe2f6386c0d1457f53481bb0

### DIFF
--- a/app/presenters/outcome_presenter.rb
+++ b/app/presenters/outcome_presenter.rb
@@ -25,4 +25,8 @@ class OutcomePresenter < NodePresenter
   def next_steps(html: true)
     @renderer.content_for(:next_steps, html: html)
   end
+
+  def relative_erb_template_path
+    @renderer.relative_erb_template_path
+  end
 end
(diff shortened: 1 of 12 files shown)


chrisroos:
I'm going to review this now.

chrisroos:

floehopper:
Note that the bookmarklet link can be found on [this page][1].

[1]: http://bl.ocks.org/floehopper/91b990d66bfae1dbdbaa88aa4772f1a9

PR #2669, opened 2016-07-28
Add links to visualise Smart Answers to index

chrisroos:
The ability to visualise flows is a little bit hidden at the moment.
This should make it more discoverable.


floehopper:
This looks good to me 👍

PR #2675, opened 2016-08-01
Fix question layout when increasing font size

selfthinker:
[Trello card](https://trello.com/c/rd2r84Pf/226-accessibility-bug-increasing-font-size-breaks-question-page-layout)
This is a better fix and an alternative version of #2654.

The question page breaks when the font size gets increased,
making the content too narrow, showing only 1-2 words per line.
This is because the grid of the question page does not use
the standard grid and it's using `em` instead of `px` or `%`.

This implements the stardard grid. A side effect is that
the width of question page content changes from 7/12
to our standard of 2/3 and that width now includes the h1.

## Expected behaviour

For users who increase their browser's font size, the copy will become much more readable. A side effect of the change as an effort to streamline it with our standard design, the content area of question pages will be a bit wider.

On the example of `/benefit-cap-calculator/y/default/yes/no/no/bereavement`...

### With 200% increased font size (fixes bug)
Before:
screen shot 2016-07-21 at 20 51 29

After:
screen shot 2016-07-21 at 20 52 10

### Normal font size (fixes inconsistencies)
Before:
screen shot 2016-08-01 at 20 28 45

After:
screen shot 2016-08-01 at 20 31 10

## Testing help

* When reviewing the "Files changed", remember you cand add `?w=1` to the URL to repress any whitespace changes.
* Increasing the font size without zooming is a bit hidden in browsers. These screenshots I found on the web show it in all your favourite browsers:

Safari:
image

Firefox:
image

Chrome:
image

### DIFF
--- a/app/assets/stylesheets/smart_answers.scss
+++ b/app/assets/stylesheets/smart_answers.scss
@@ -231,14 +231,9 @@ $is-ie: false !default;
 
   .step.current {
     background-color: #fff;
-    margin-right: 15em;
-    padding: 0 10em 1em 0;
+    margin-right: 0;
+    padding: 0 0 1em 0;
     position: relative;
-
-    @include media-down(tablet) {
-      margin-right: 0;
-      padding: 0 0 1em;
-    }
   }
 
   @include ie(6) {
@@ -315,18 +310,6 @@ $is-ie: false !default;
   /*
    * results / outcome
    */
-
-  article.outcome {
-    @extend %grid-row;
-
-    .result-info {
-      @include grid-column( 2/3 );
-    }
-    .related-wrapper {
-      @include grid-column( 1/3 );
-    }
-  }
-
 
   .next-steps {
     right: -24em;
--- a/app/views/smart_answers/show.html.erb
+++ b/app/views/smart_answers/show.html.erb
@@ -1,19 +1,20 @@
 <% content_for :title do %><%= @presenter.title %><% end %>
 <% content_for :head do %>
-<%= javascript_include_tag "smart-answers", defer: true %>
+  <%= javascript_include_tag "smart-answers", defer: true %>
 <% end %>
-<% unless @presenter.started? %><div class="grid-row"><% end %>
-<main id="content" role="main">
-  <div id="js-replaceable" <% if @presenter.started? %> class="smart-answer-questions group"<% end %>>
-    <%= render partial: "content" %>
-  </div>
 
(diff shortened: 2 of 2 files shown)


ikennaokpala:
@selfthinker it appears that the addition of the grid-row class has broken text layout as follows:

## With grid-row class

screen shot 2016-08-03 at 10 08 50

## Without grid-row class

screen shot 2016-08-03 at 10 09 03

selfthinker:
Yes, that is intentional. Our general design standard is that we should never show copy in full width on a desktop. That is because it makes the copy more difficult to read. Our standard width for the main copy is 2/3 of the full 960px on desktop. (On mobile it's obviously full width.)
I double checked with the designers in the design channel on Monday. My question and their response:

>Anika Henke [7:58 PM]  
>what do we usually do with pages which don’t have anything in their sidebar? I presume we show the content in 2/3 width as well and just leave the sidebar empty? or would we show it in full width or try to put something into the sidebar?

>Chris Heathcote [8:01 PM]  
>the first. it’s mainly about keeping line lengths short and readable. sidebar is a bonus.

selfthinker:
As just discussed, the outcome page is indeed not correct... I will look at it.

selfthinker:
I just fixed the outcome page but introduced another difference to what's currently live: The main content width (2/3) now includes the header.

Before:
screen shot 2016-08-03 at 13 05 27

After:
screen shot 2016-08-03 at 13 05 41

I double checked in the design channel again:

> Anika Henke [12:43 PM]  
> [...] when a page has no sidebar and we want the copy to not be 100% width but 2/3 width (on a desktop), do we usually include or exclude the page header (h1 etc)?
>i.e. is the header 100% width and the main copy 2/3? or is everything, including the header, 2/3 width?

>Alex Torrance [12:49 PM]  
>It changes between formats :scream:
>But my preference is to include the header

[...]

> Anika Henke [12:55 PM]  
> is there a consensus in the case there is a sidebar? the only pages I found so far, always include the header in the 2/3 bit

>Alex Torrance [12:56 PM]  
>I think mainstream is always in the 2/3 bit
> So I would go with that

ikennaokpala:
@selfthinker I have setup this [heroku preview](https://smart-answers-pr-2675.herokuapp.combenefit-cap-calculator/y/default/yes/no/no/bereavement)..

ikennaokpala:
@selfthinker This LGTM

PR #2682, opened 2016-08-09
Disable AJAX for some Smart Answers

selfthinker:
[Trello card](https://trello.com/c/CBDnKPys)

This disables the JS which loads questions via AJAX.
We would like to remove the AJAX altogether as it is not making things much quicker and it introduces maintability issues and bugs. To be reassured that this will not worsen the user experience, we only disable it on some Smart Answers.
The list of Smart Answers was provided by Dan Gilbert who chose some which are not visited too often or too seldom.

I also restructured the JS files to follow standard Rails conventions and to make sure `smart-answers.js` only contains JS relevant to the AJAX.

## Expected changes

The pages `energy-grants-calculator`, `calculate-employee-redundancy-pay` and `register-a-death` will not use any AJAX calls in the background. The users will most probably not notice any difference. (We made this change to verify that assumption.) They will experience less bugs and very slightly longer delays when the page reloads.
As a developer you can check changes in Firefox by watching the console for those AJAX calls (which will be missing for the three Smart Answers mentioned above but be there for all other Smart Answers). In Chrome it can be checked in the console via the Network tab and filter by XHR.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -1,0 +1,7 @@
+//= require smart-answers
+//= require helpers
+//= require start-button-ab-test-july-2016
+
+$(document).ready(function() {
+  $('#current-error').focus();
+});
--- a/app/assets/javascripts/helpers.js
+++ b/app/assets/javascripts/helpers.js
@@ -1,0 +1,21 @@
+var SmartAnswer = SmartAnswer || {};
+SmartAnswer.isStartPage = function(slug) { // Used mostly during A/B testing
+  return window.location.pathname.split("/").join("") == slug;
+}
+
+function linkToTemplatesOnGithub() {
+  $('*[data-debug-template-path]').each(function() {
+    var element = $(this);
+    var path = element.data('debug-template-path');
+    var filename = path.split('/').pop();
+    var host = 'https://github.com';
+    var organisation = 'alphagov';
+    var repository = 'smart-answers'
+    var branch = 'deployed-to-production';
+    var url = [host, organisation, repository, 'blob', branch, path].join('/');
+    var anchor = $('<a>Template on GitHub</a>').attr('href', url).attr('style', 'color: deeppink;').attr('title', filename);
+    element.prepend(anchor);
+    element.attr('style', 'border: 3px solid deeppink; padding: 10px; margin: 3px');
+    element.removeAttr('data-debug-template-path');
+  });
+};
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -2,4 +2,12 @@ module ApplicationHelper
   def last_updated_date
     File.mtime(Rails.root.join('REVISION')).to_date rescue Date.today
   end
+
+  def ajax_enabled_for?(name)
+    %w(
+      energy-grants-calculator
+      calculate-employee-redundancy-pay
+      register-a-death
+    ).exclude?(name)
+  end
 end
--- a/app/presenters/flow_presenter.rb
+++ b/app/presenters/flow_presenter.rb
@@ -37,6 +37,10 @@ class FlowPresenter
     @current_state ||= @flow.process(all_responses)
   end
 
+  def name
+    @flow.name
+  end
+
   def collapsed_questions
     @flow.path(all_responses).map do |name|
       presenter_for(@flow.node(name))
--- a/app/views/smart_answers/show.html.erb
+++ b/app/views/smart_answers/show.html.erb
@@ -1,6 +1,6 @@
 <% content_for :title do %><%= @presenter.title %><% end %>
 <% content_for :head do %>
-  <%= javascript_include_tag "smart-answers", defer: true %>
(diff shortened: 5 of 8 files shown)


app/assets/javascripts/application.js
```diff
@@ -0,0 +1,24 @@
+//= require smart-answers
+//= require helpers
+//= require start-button-ab-test-july-2016
+
+$(document).ready(function() {
+  $('#current-error').focus();
+});
+
+function linkToTemplatesOnGithub() {
```
ikennaokpala: @selfthinker I think this function should move as well to the helpers.js. This function is principally helps highlight content and provide a link to the template on github.

app/assets/javascripts/helpers.js
```diff
@@ -0,0 +1,5 @@
+
```
ikennaokpala: @selfthinker this line appears to be redundant.. can you take it away..

app/views/smart_answers/show.html.erb
```diff
@@ -1,5 +1,13 @@
 <% content_for :title do %><%= @presenter.title %><% end %>
 <% content_for :head do %>
+  <script>
+    var SmartAnswer = SmartAnswer || {};
+    <% if ajax_enabled_for?(@presenter.name) %>
+      SmartAnswer.AJAX_ENABLED = true;
+    <% else %>
+      SmartAnswer.AJAX_ENABLED = false;
+    <% end %>
+  </script>
```
ikennaokpala: @selfthinker 

You `ajax_enabled_for?(name)` is a predicate method returning booleans..

You could try

```javascript
  <script>
    var SmartAnswer = SmartAnswer || {};
    SmartAnswer.AJAX_ENABLED = <%= ajax_enabled_for?(@presenter.name) %>;
  </script>
```

selfthinker: Ah, I didn't know that was possible (because it's Ruby boolean vs JavaScript boolean).

app/helpers/application_helper.rb
```diff
@@ -2,4 +2,12 @@ module ApplicationHelper
   def last_updated_date
     File.mtime(Rails.root.join('REVISION')).to_date rescue Date.today
   end
+
+  def ajax_enabled_for?(name)
+    %w(
+      energy-grants-calculator
+      calculate-employee-redundancy-pay
+      register-a-death
+    ).exclude?(name)
```
ikennaokpala: Building this list to me is brittle.. an ideal solution will for the `Flow` object to provide is ajax enabled capability. 

This way we avoid having to building this list and avoid future increase in elements of the array here.

so just like `@presenter` provides the name of the `Flow`.

It ideal for a `Flow` ajax status to be set in the `define` method.
For example

```ruby
  content_id "yuewyeuw-ueiwueiw-uier76"
  name 'energy-grants-calculator'
  status :published
  satisfies_need "100982"
  ajax_enabled false      
```

And then in the presenter

```ruby
 def ajax_enabled
    @flow.ajax_enabled
 end
```

```javascript
<script>
    var SmartAnswer = SmartAnswer || {};
    SmartAnswer.AJAX_ENABLED = <%= @presenter.ajax_enabled %>;
  </script>
```

@selfthinker This is an ideal scenario

selfthinker: This is actually how I did it at first. But then I had a chat with @pmanrubia and he said it probably shouldn't be part of a flow.

selfthinker: Also, please keep in mind that we don't plan to maintain that list. It is only temporary (for 2-3 weeks). That knowledge might change your approach.

selfthinker:
@ikennaokpala, everything you suggested except adding the status to the Flow object makes sense to me and I just made the corresponding changes, rebased and pushed.
Let's discuss the other suggested change with Pablo tomorrow.

app/presenters/flow_presenter.rb
```diff
@@ -37,6 +37,10 @@ def current_state
     @current_state ||= @flow.process(all_responses)
   end
 
+  def name
```
pmanrubia: Very minor, but could you please add a test for this method?

pmanrubia:
@selfthinker: could you please update the card description with the steps we need to take to validate that the AJAX is disabled after it is deployed? This helps in our releases.
Something similar to:

> ### Expected Changes
> Visit these three URLs open the inspector and check…

pmanrubia:
@selfthinker  @ikennaokpala 
Thank you for your work on this PR. I would prefer not to modify the workflow DSL to include UI specific concerns. 

I would try to keep the workflow agnostic of the data presentation if possible. This is a bit more clear to me if I think of multiple devices with different presentation technologies.

pmanrubia:
@selfthinker
Well done creating the `application.js` 👍 . It is a good first step towards getting our JS sorted out.

ikennaokpala:
@pmanrubia that is fine by me

selfthinker:
@pmanrubia, I have added the test with the help of @ikennaokpala and updated the Trello card description.

PR #2684, opened 2016-08-11
Removed inaccurate hint text

keithiopia:
Deleted "... and any children where the deceased had a parental role."
Requested in Zen #1317975.
https://trello.com/c/3I2ufDzz/205-remove-misleading-text-from-intestacy-tool


robinjam:
See #2694

PR #2698, opened 2016-08-22
Remove July's Start button AB testing code.

ikennaokpala:
[Trello card](https://trello.com/c/1KgczS5u/290-remove-start-button-label-a-b-testing-code-from-smart-answers)

### Description 

This PR removes the start button AB testing code for the month of July.

This will in effect, stop the creation of cookies and allocation of users to cohorts respectively.

Affecting only the following smart answers:

* Overseas passports 
* Calculate your child maintenance
* Marriage abroad

After the removal of the files and require entry in the javascript manifest file,

I ran all the regression, unit and flow tests and I did not record any errors.


selfthinker:
Can you add a link to the Trello card?

You did not remove `SmartAnswer.isStartPage` from `helpers.js` on purpose, right? I'm a bit in two minds about that. On the one hand it's dead code now, but on the other hand we know we will need it again the next time we run an A/B test. I'm happy to leave it as it is, just wanted to make sure you were aware of it.

LGTM :+1:

ikennaokpala:
@selfthinker I have added the Trello card link.. Yes `SmartAnswer.isStartPage` remains a part of the code base.. there is a comment there to explain

PR #2699, opened 2016-08-22
Stop relying on slugs in areas returned by the imminence api

h-lame:
For: https://trello.com/c/PKAohBzb/468-stop-adding-slugs-to-areas-returned-from-mapit-by-imminence-2

Imminence is removing the slugs because they are based on transforming the
names, which aren't static pieces of data and so shouldn't be relied on.
The only real change here is how the benefit-cap-calculator determines if
the claimant is based in London or not.  Instead of looking for an area
with the slug 'london' we look for an area with the type EUR (European
parliament constituency) and the name 'London'.  This is no more robust
than checking against the slug, as the names of the constituencies could
change, or the boundaries could change, or we could leave Europe and no
longer need these areas, but slug is going away so we have to do
something.

A better solution might be to check for known london councils, or use
mapit directly instead of imminience and check for different area types
(GLA for example) that are more permanent and idicative of a postcode
being in London.  We may also need to refer to the legislation to see
how it defines what "in London" means as it may not equate to "votes in
the London European Parliament Consitituency".

See https://github.com/alphagov/imminence/pull/136 for the imminence change

### DIFF
--- a/lib/smart_answer/calculators/benefit_cap_calculator_configuration.rb
+++ b/lib/smart_answer/calculators/benefit_cap_calculator_configuration.rb
@@ -46,7 +46,7 @@ module SmartAnswer::Calculators
     end
 
     def london?(postcode)
-      area(postcode).any? { |result| result[:slug] == "london" }
+      area(postcode).any? { |result| result[:type] == 'EUR' && result[:name] == "London" }
     end
 
     def area(postcode)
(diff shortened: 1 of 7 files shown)


emmabeynon:
👍

pmanrubia:
Thank you for your work on it, and for the detailed description in the commit notes 👍.

I have created a [Trello Card](https://trello.com/c/6Jm4MV8Q/296-benefit-cap-calculator-clarify-with-a-content-designer-what-inside-london-means) to clarify the definition of `inside London` with a Content Designer.

PR #2704, opened 2016-08-23
Make it easier to change labels

selfthinker:
[Trello card](https://trello.com/c/ol5EiaNC)

I've been working on adding/changing more labels to make the question pages more accessible. But that work is currently blocked, that's why I cherry-picked these changes which can be released without being blocked.

It's mainly rewriting tests to make them less dependent on the HTML. And it removes redundant and problematic code.

## Expected changes

None.


pmanrubia:
Good work! LGTM

PR #2711, opened 2016-08-30
Fix only block elements in partials

selfthinker:
[Trello card](https://trello.com/c/qwShHfW6)

This is to prepare for linking all partials to GitHub on Integration.
That would need to wrap each partial in a `div` which would create invalid HTML in these cases and might lead to unexpected browser behaviour.

Coincidentally, fixing this also leads to better code, making sure not to break up any elements which logically belong together (table head belongs to the rest of the table, list item belongs to the whole list, etc).

## Expected changes

None


lib/smart_answer_flows/marriage-abroad/outcomes/_consular_fees_table_items.govspeak.erb
```diff
@@ -1,3 +1,7 @@
+<% if service_counter == 0 %>
```
pmanrubia: I am curious about why not using:
```ruby
<% if calculator.services.any?  == 0 %>
```
It is not obvious to me where is ```service_counter`` coming from.

pmanrubia:
I am not totally sure we need e976e0fc402c716a1f35f0a974812e2f5c597469 if we insert ```ul > li``` as described in this [Trello card](https://trello.com/c/qwShHfW6/273-investigate-work-done-to-date-linking-content-to-partial-in-github-for-smart-answer-updates). Would you say we have now `better HTML`?

pmanrubia:
It looks good to me @selfthinker. Feel free to merge if you think that e976e0fc402c716a1f35f0a974812e2f5c597469 moves us to a better place.

lib/smart_answer_flows/marriage-abroad/outcomes/_consular_fees_table_items.govspeak.erb
selfthinker: `service_counter` is a bit of Rails magic, see [Layouts and Rendering in Rails > Local Variables](http://guides.rubyonrails.org/layouts_and_rendering.html#local-variables):

> Rails also makes a counter variable available within a partial called by the collection, named after the member of the collection followed by _counter. For example, if you're rendering @products, within the partial you can refer to product_counter to tell you how many times the partial has been rendered. This does not work in conjunction with the as: :value option.

I personally dislike all kind of "magic" but in this case I wouldn't know what else to use instead. This seems to be made for exactly this case.
`calculator.services.any?` wouldn't work as that would show the table header all the time.

pmanrubia: Thank you @selfthinker. I totally forgot about that counters for collection 😊

selfthinker:
As discussed, although my changes for the list items don't result in much better code, I think it's a little bit better than before and makes sense. It's keeping all list items together in one view, although not the list item *content*.

I have run the regression and other tests after every change before pushing.

pmanrubia:
Great work @selfthinker!
LGTM

PR #2724, opened 2016-09-13
Fix flow visualisation in Chrome

selfthinker:
[Trello card](https://trello.com/c/YN09P43l)

The flow visualisation was broken in Chrome since version 48 as they removed support for `getTransformToElement` and it was also removed from the SVG spec (see https://www.w3.org/TR/SVG2/changes.html#types).

Although it was fixed upstream with a polyfill, I only copied that polyfill to patch up the old library. An update of the whole library would need us to update jQuery as well. @floehopper tried to update it and its dependencies and was running into problems.

See also http://jointjs.com/blog/get-transform-to-element-polyfill.html

## Expected results ##

Visualise URLs are now showing flow visualisations in Chrome again.

* [Example on production (broken)](https://www.gov.uk/register-a-birth/visualise)
* [Example on Heroku (fixed)](https://smart-answers-pr-2724.herokuapp.com/register-a-birth/visualise)


leenagupte:
@selfthinker This is amazing! 👍 👍 👍

PR #2735, opened 2016-09-20
Use smart answer name to run regression tests from jenkins.sh script

ikennaokpala:
## Description 

Given the lengthy period it takes to run the regression tests within this repo.
It appears that the jenkins.sh script has omitted running the regression tests.

This idea came up during the on of our developer retro.

The Idea is to add to the jenkins script ```RUN_REGRESSION_TESTS=<smart-answer> ruby test/regression/smart_answers_regression_test.rb``` where <smart-answer> can be obtained from the git branch name.

The git branch name should ideally be structured like this. 

``` <smart-answer>/<description-of-task-or-feature> ```

The unix script will then read the <smart-answer> part before the forward slash and then use this to run the regresstion test for the affected smart answer.

Hence this PR seeks to include the regression test as part of the jenkins script.

This is reliant on the git branch being setup with the name of the smart-answer first followed by a
forward slash and the the actually name of the branch (i.e some git work flow conventions have referred to this as a feature branching pattern).

This convention in effect provides the basis for sensible assumption to be made, that the part
before the forward slash is indeed the name of the smart answer flow (i.e ideal scenario).

This is then used to provide RUN_REGRESSION_TESTS environment variable with the required.

if the branch is named without the forward slash, the branch name is returned preceded by an
asterisk (*) from the parse_git_branch function.

if the value return from the parse_git_branch function does not match a smart answer flow

The aforementioned (i.e edge cases) will not invoke or run any regression tests (i.e gracefully).

### DIFF
--- a/jenkins.sh
+++ b/jenkins.sh
@@ -2,6 +2,16 @@
 
 set -e
 
+# 1. This function first lists the branches available.
+# 2. Finds the current branch (i.e starts with an asterisk).
+# 3. Finds the value/group between the asterisk and forward slash.
+# 4. Returns the value/group.
+# For example for a branch named <smart-answer-name>/<description-of-task-or-feature>,
+# This function will return <smart-answer-name> only.
+parse_git_branch(){
+  git branch 2> /dev/null | sed -e '/^[^*]/d' -e 's/* \(.*\)\(\/.*\)/\1/'
+}
+
 # Try to merge master into the current branch, and abort if it doesn't exit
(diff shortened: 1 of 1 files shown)


ikennaokpala:
Following the collab session earlier today, where I demonstrated this idea to my team. We agree to further investigate and build on the finds discussed from this PR. We have also created this [trello card](https://trello.com/c/RNLoz0wT/164-add-unix-scripts-to-run-regression-test-for-the-current-smart-answer) to help track future activity around this area of work.

PR #2757, opened 2016-09-29
Send external_related_links to publishing api

MatMoore:
These used to be managed by panopticon, but we are retiring
panopticon and moving these links to the publishing api.
Hardcode the related link for the one smart answer that uses it.


jackscotti:
Tested on Integration by deploying the branch. As you can see in the screenshot, the external related links have been published to content-store. 
screen shot 2016-10-03 at 10 42 57

rubenarakelyan:
👍

PR #2761, opened 2016-10-03
Update gds-api-adapters to 36.4.0

mobaig:
Update to the latest version, particularly as we've updated the
Panopticon adapter to stop handling search-related attributes (see
alphagov/gds-api-adapters#592).


rubenarakelyan:
👍

PR #2789, opened 2016-10-20
[DO NOT MERGE]Fetch metadata from content-store

jackscotti:
https://trello.com/c/49441pPb

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -26,6 +26,7 @@ gem 'uglifier'
 gem 'uk_postcode', '~> 1.0.1'
 gem 'unicorn', '4.8.3'
 gem 'rails_stdout_logging'
+gem 'govuk_navigation_helpers', '~> 2.0'
 
 group :development do
   gem 'better_errors'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -93,6 +93,7 @@ GEM
     govuk_frontend_toolkit (4.14.0)
       rails (>= 3.1.0)
       sass (>= 3.2.0)
+    govuk_navigation_helpers (2.0.0)
     htmlentities (4.3.4)
     http-cookie (1.0.3)
       domain_name (~> 0.5)
@@ -273,6 +274,7 @@ DEPENDENCIES
   govuk-content-schema-test-helpers (~> 1.3.0)
   govuk-lint
   govuk_frontend_toolkit (= 4.14.0)
+  govuk_navigation_helpers (~> 2.0)
   htmlentities (~> 4)
   json
   logstasher (= 0.4.8)
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,10 +1,5 @@
-require "slimmer/headers"
-
 class ApplicationController < ActionController::Base
-  include Slimmer::Headers
   include Slimmer::Template
-  include Slimmer::SharedTemplates
-  before_action :set_analytics_headers
 
   rescue_from GdsApi::TimedOutException, with: :error_503
   rescue_from ActionController::UnknownFormat, with: :error_404
@@ -31,9 +26,5 @@ protected
     else
       render status: status_code, text: error_message
     end
-  end
-
-  def set_analytics_headers
-    set_slimmer_headers(format: "smart_answer")
   end
 end
--- a/app/controllers/smart_answers_controller.rb
+++ b/app/controllers/smart_answers_controller.rb
@@ -1,7 +1,10 @@
 class SmartAnswersController < ApplicationController
+  include Slimmer::SharedTemplates
+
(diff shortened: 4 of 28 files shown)


Gemfile
```diff
@@ -20,12 +18,13 @@ gem 'plek', '1.12.0'
...
 gem 'rails-i18n'
 gem 'sass-rails', '~> 5.0.0'
-gem 'slimmer', '~> 9.0.1'
+gem 'slimmer', '~> 9.6.0'
 gem 'tilt', '1.4.1'
 gem 'uglifier'
 gem 'uk_postcode', '~> 1.0.1'
 gem 'unicorn', '4.8.3'
 gem 'rails_stdout_logging'
+gem 'govuk_navigation_helpers', github: "alphagov/govuk_navigation_helpers"
```
tijmenb: This needs to be `2.0.0`

Gemfile
```diff
@@ -1,7 +1,5 @@
 source 'https://rubygems.org'
 
-ruby '2.3.0'
```
tijmenb: https://github.com/alphagov/smart-answers/pull/2796 will fix this in a better way.

Gemfile.lock
```diff
@@ -87,8 +92,8 @@ GEM
       sanitize (~> 2.1.0)
     govuk-content-schema-test-helpers (1.3.0)
       json-schema (~> 2.5.1)
-    govuk-lint (0.7.0)
-      rubocop (~> 0.35.0)
+    govuk-lint (1.2.1)
```
tijmenb: Not sure if you meant to update `govuk-lint`?

app/controllers/application_controller.rb
```diff
@@ -32,8 +27,4 @@ def error(status_code, exception = nil)
       render status: status_code, text: error_message
     end
   end
-
-  def set_analytics_headers
-    set_slimmer_headers(format: "smart_answer")
```
tijmenb: Since we are now using the meta tags component to set analytics tags `govuk:format` will change from `smart_answer` (set here) to `smartanswer_document` (the [document type of all smart answers](https://www.gov.uk/api/content/additional-commodity-code)).

To not disrupt analytics reporting we need to change the publishing-api document type into `smart_answer`.

app/controllers/smart_answers_controller.rb
```diff
@@ -90,10 +91,6 @@ def redirect_response_to_canonical_url
     end
   end
 
-  def set_header_footer_only
```
tijmenb: It looks like this determines the layout of the pages, so I think this is not supposed to be deleted.

app/views/smart_answers/show.html.erb
```diff
@@ -18,7 +20,7 @@
       <%= render :partial => 'smartanswer_metadata' %>
     </div>
   </main>
-  <% unless @presenter.started? %>
```
tijmenb: Won't this now show related items for all pages, and not only the start pages?

test/functional/smart_answers_controller_test.rb
```diff
@@ -55,112 +56,111 @@ def teardown
...
 
     should "404 Not Found if request is for an unknown format" do
       @controller.stubs(:respond_to).raises(ActionController::UnknownFormat)
 
-      get :show, id: 'smart-answers-controller-sample'
+      get :show, id: @smartanswer_name
       assert_response :not_found
     end
 
-    should "send slimmer analytics headers" do
```
tijmenb: 👍 this is good to delete.

app/views/smart_answers/show.html.erb
jackscotti: good point, I should not have deleted this

app/controllers/application_controller.rb
jackscotti: With `the publishing-api document type` you mean the content-schema?

tijmenb: This one: https://github.com/alphagov/smart-answers/blob/8d1b172c7f16ec627d68429286f668d13eacb98c/app/presenters/flow_content_item.rb#L16

Gemfile.lock
jackscotti: automatically pushed up by bundler:

```
Bundler could not find compatible versions for gem "rubocop":
  In snapshot (Gemfile.lock):
    rubocop (= 0.39.0)

  In Gemfile:
    govuk-lint was resolved to 0.7.0, which depends on
      rubocop (~> 0.35.0)
```

tijmenb:
This PR is now ready for review by @rubenarakelyan. I would like it to be reviewed & merged by @pmanrubia too so that it's not a surprise change.

tijmenb:
Superseded by https://github.com/alphagov/smart-answers/pull/2805.

PR #2790, opened 2016-10-20
Changes to PIP Checker Middle Group

ikennaokpala:
[Trello card](https://trello.com/c/a20W0vWM/375-pip-checker-fixing-logic-for-17-and-18-year-old-outcomes)

## Description 

The aim of this PR is to make amends to the middle group in the pip-checker smart answer to cater for users who recently turned 16.

## Factcheck

[Preview link](//smart-answers-pr-2790.herokuapp.com//pip-checker/y/yes/1998-07-06)
[GOVUK](https://www.gov.uk/pip-checker/y/yes/1998-07-06)

### Expected changes
* User who recently turned 16 now get the expected outcomes

### Before
screen shot 2016-10-20 at 12 50 46

### After

screen shot 2016-10-20 at 12 51 34

### DIFF
--- a/lib/smart_answer/calculators/pip_dates.rb
+++ b/lib/smart_answer/calculators/pip_dates.rb
@@ -12,7 +12,7 @@ module SmartAnswer::Calculators
     end
 
     def in_middle_group?
-      self.dob > GROUP_65_CUTOFF && self.dob < MIDDLE_GROUP_CUTOFF
+      self.dob > GROUP_65_CUTOFF && self.dob <= 16.years.ago
     end
 
     def turning_16_before_oct_2013?
--- a/test/integration/smart_answer_flows/pip_checker_test.rb
+++ b/test/integration/smart_answer_flows/pip_checker_test.rb
@@ -55,29 +55,29 @@ class PIPCheckerTest < ActiveSupport::TestCase
   context "getting DLA" do
     setup do
       add_response 'yes'
-      Timecop.travel('2014-02-03')
+      Timecop.travel('2013-02-03')
     end
 
-    should "ask for your date of birth (03/03/14)" do
+    should "ask for your date of birth (03/03/13)" do
       assert_current_node :what_is_your_dob?
     end
 
-    should "be over 65 by 2013, if born on or before 08-04-1948 (03/03/14)" do
+    should "be over 65 by 2013, if born on or before 08-04-1948 (03/03/13)" do
       add_response "1948-04-08"
       assert_current_node :over_65_in_2013
     end
 
-    should "qualify for pip, if born between 08-06-1997 and 06-10-1997 (03/03/14)" do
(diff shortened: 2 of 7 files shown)


lib/smart_answer/calculators/pip_dates.rb
```diff
@@ -12,7 +12,7 @@ def in_group_65?
     end
 
     def in_middle_group?
-      self.dob > GROUP_65_CUTOFF && self.dob < MIDDLE_GROUP_CUTOFF
+      self.dob > GROUP_65_CUTOFF && self.dob <= 16.years.ago(Date.today)
```
pmanrubia: Not sure if `Date.today` is needed here.

test/integration/smart_answer_flows/pip_checker_test.rb
```diff
@@ -55,29 +55,29 @@ class PIPCheckerTest < ActiveSupport::TestCase
   context "getting DLA" do
     setup do
       add_response 'yes'
-      Timecop.travel('2014-02-03')
+      Timecop.travel('2013-02-03')
```
pmanrubia: How can we be sure that the time set by `Timecop` is reset after each test?
Shouldn't we have a `teardown` section?

lib/smart_answer/calculators/pip_dates.rb
ikennaokpala: This is required and caters for users who have recently turned 16

test/integration/smart_answer_flows/pip_checker_test.rb
pmanrubia: No worries @ikennaokpala. It [seems to be reset here for all tests](https://github.com/alphagov/smart-answers/blob/e275244a0464576805108edd337fdf665a802ffb/test/test_helper.rb#L26-L26).

ikennaokpala: Thanks @pmanrubia for point this out.. Much appreciated

lib/smart_answer/calculators/pip_dates.rb
```diff
@@ -12,7 +12,7 @@ def in_group_65?
     end
 
     def in_middle_group?
-      self.dob > GROUP_65_CUTOFF && self.dob < MIDDLE_GROUP_CUTOFF
+      self.dob > GROUP_65_CUTOFF && self.dob <= 16.years.ago
```
leenagupte: @ikennaokpala As you are no longer using variable MIDDLE_GROUP_CUTOFF here, and it's not being used anywhere else, I think it should be removed completely.

leenagupte:
@ikennaokpala I've added a couple of minor inline comments. 

The changes to the code LGTM 👍 

The whole concept of the `middle group` is confusing, but as it pre-dates this change we should leave it be. I do think though, that if we get further changes for this smart answer, we should consider renaming the `in_middle_group?` function to something that tells us what the middle group is.

PR #2796, opened 2016-10-25
Bump ruby to 2.3.1

tijmenb:


Gemfile
```diff
@@ -1,6 +1,6 @@
 source 'https://rubygems.org'
 
-ruby '2.3.0'
+ruby File.read(".ruby-version").chomp
```
tijmenb: For Heroku: https://devcenter.heroku.com/articles/ruby-versions

jackscotti: I see.

PR #2797, opened 2016-10-25
Refactor consumption of API response via gds-api-adapter gem

ikennaokpala:
Related to #2761

[Trello card](https://trello.com/c/Gduv7tZL/355-upgrade-gds-adapters-to-latest-version-in-smart-answers)

## Why? 

The reason for these changes is to reduce the occurrence of error-prone nil response from OpenStruct objects that were previously returned by the gds-api-adapters. OpenStruct returns nil even when an attribute doesn’t exist by default, which is confusing and error-prone (and makes the code more complicated to debug).

## Description

This PR makes changes to the following api calls: 
- worldwide_api
- imminence_api

Changes made within affect:
- mainly how the response is consumed. 
- a few cosmetic changes (mostly renaming for clarity)
- fix of lint errors 
- regression artefacts

**NB:**
Previously a response object was actually returned (pre gds-api-adapter changes 37.x.x ). Following the recent changes, response are returned as hashes (with the help of the new to_hash message).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -43,6 +43,7 @@ group :test do
   gem 'capybara', '2.1.0'
   gem 'ci_reporter'
   gem 'minitest', '~> 5.1'
+  gem 'minitest-focus', '~> 1.1', '>= 1.1.2'
   gem 'mocha', '1.1.0', require: false
   gem 'poltergeist', '1.6.0'
   gem 'shoulda', '~> 3.5.0'
--- a/Gemfile
+++ b/Gemfile
@@ -6,7 +6,7 @@ gem 'rails', '~> 4.2.7'
 
 gem 'airbrake', '4.1.0'
 gem 'ast'
-gem 'gds-api-adapters', '~> 36.4.0'
+gem 'gds-api-adapters', '~> 37.2.0'
 gem 'govspeak', '~> 3.3.0'
 gem 'govuk-content-schema-test-helpers', '~> 1.3.0'
 gem 'govuk_frontend_toolkit', '4.14.0'
--- a/app/models/world_location.rb
+++ b/app/models/world_location.rb
@@ -22,7 +22,10 @@ class WorldLocation
 
   def self.find(location_slug)
     cache_fetch("find_#{location_slug}") do
-      data = Services.worldwide_api.world_location(location_slug)
+      response = Services.worldwide_api.world_location(location_slug)
+      if response
+        data = OpenStruct.new(response.to_hash)
+      end
       self.new(data) if data
     end
   end
@@ -53,16 +56,21 @@ class WorldLocation
     @data = data
   end
 
+  def slug
+    self.details["slug"]
+  end
+
   def ==(other)
-    other.is_a?(self.class) && other.slug == self.slug
+    other.is_a?(self.class) && other.slug == slug
   end
 
   def_delegators :@data, :title, :details
-  def_delegators :details, :slug
+  # def_delegators :details, :slug
   alias_method :name, :title
 
   def organisations
-    @organisations ||= WorldwideOrganisation.for_location(self.slug)
+    # binding.pry
+    @organisations ||= WorldwideOrganisation.for_location(slug)
   end
 
   def fco_organisation
--- a/app/models/world_location.rb
+++ b/app/models/world_location.rb
@@ -1,7 +1,7 @@
 require 'lrucache'
 
 class WorldLocation
-  extend Forwardable
+  attr_reader :title, :details, :slug
 
   def self.cache
     @cache ||= LRUCache.new(max_size: 250, soft_ttl: 24.hours, ttl: 1.week)
@@ -13,8 +13,11 @@ class WorldLocation
 
   def self.all
     cache_fetch("all") do
-      world_locations = Services.worldwide_api.world_locations.with_subsequent_pages.map do |l|
-        new(l) if l.format == "World location" && l.details && l.details.slug.present?
+      world_locations = Services.worldwide_api.world_locations.with_subsequent_pages.map do |response|
+        location = response.to_hash
+        if valid_world_location_format?(location)
+          self.new(location)
+        end
       end
       world_locations.compact
     end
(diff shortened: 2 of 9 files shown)


app/models/world_location.rb
```diff
@@ -13,20 +13,21 @@ def self.reset_cache
...
+            location.slug.present?
+          self.new(location)
+        end
       end
       world_locations.compact
     end
   end
 
   def self.find(location_slug)
     cache_fetch("find_#{location_slug}") do
```
pmanrubia: @ikennaokpala could you please if we need to cache information locally. It seems that every adapter is already doing it? If so, could you prepare another commit to remove it?

app/models/world_location.rb
```diff
@@ -13,20 +13,21 @@ def self.reset_cache
...
-      response = Services.worldwide_api.world_location(location_slug)
-      if response
-        data = OpenStruct.new(response.to_hash)
-      end
-      self.new(data) if data
+      response = Services.worldwide_api.world_location(location_slug)&.to_hash
+      self.new(Location.new(response)) if response
```
pmanrubia: If we get an OpenStruct directly from the adapter, we should be able to rewrite this method with:

```ruby

def self.find(location_slug)
    cache_fetch("find_#{location_slug}") do
      location = Services.worldwide_api.world_location(location_slug)
      self.new(location)
    end
  end
```

We should also be able to remove most of the delegators

lib/smart_answer/calculators/benefit_cap_calculator_configuration.rb
```diff
@@ -46,12 +46,14 @@ def region(postcode)
     end
 
     def london?(postcode)
-      area(postcode).any? { |result| result[:type] == 'EUR' && result[:name] == "London" }
+      area(postcode).any? do |result|
```
pmanrubia: Please, update this commit with the OpenStruct direct access once we are using the OpenStruct response object from the GdsAdapters.

Gemfile
```diff
@@ -6,7 +6,7 @@ gem 'rails', '~> 4.2.7'
 
 gem 'airbrake', '4.1.0'
 gem 'ast'
-gem 'gds-api-adapters', '~> 36.4.0'
```
ikennaokpala: Not sure this is relevant to this commit.. maybe has already been split

config/initializers/gds_api_adapters.rb
```diff
@@ -1 +0,0 @@
-GdsApi.config.always_raise_for_not_found = false
```
ikennaokpala: it appears either way fails.. will look into this a bit more

app/models/world_location.rb
ikennaokpala: I am not sure.. will take a look

app/models/world_location.rb
```diff
@@ -52,28 +53,28 @@ def self.cache_fetch(key)
...
-
-  def slug
-    self.details["slug"]
+  def initialize(location)
+    @location = location
   end
 
   def ==(other)
-    other.is_a?(self.class) && other.slug == slug
+    other.is_a?(self.class) && other.slug == self.slug
```
ikennaokpala: The intent was to mirror the way it was prior to the initial change and somewhat be a bit explicit

app/models/world_location.rb
pmanrubia: @ikennaokpala did you have a look at this cache?

ikennaokpala: This is a bit tricky one for me as I don't have a whole picture to form a view

test/unit/world_location_test.rb
```diff
@@ -211,7 +211,7 @@ class WorldLocationTest < ActiveSupport::TestCase
     should "allow accessing required top-level attributes" do
```
ikennaokpala: @pmanrubia I have made changes that no longer require defining the slug method anymore.. 

See https://github.com/alphagov/smart-answers/blob/3c8ff7a363617d959145b70ac58a7d934757b1ec/app/models/world_location.rb#L61

app/models/world_location.rb
```diff
@@ -13,20 +13,21 @@ def self.reset_cache
...
   end
 
   def self.find(location_slug)
     cache_fetch("find_#{location_slug}") do
-      response = Services.worldwide_api.world_location(location_slug)
-      if response
-        data = OpenStruct.new(response.to_hash)
-      end
-      self.new(data) if data
+      response = Services.worldwide_api.world_location(location_slug)&.to_hash
```
ikennaokpala: @pmanrubia the location sruct no longer exists.. see https://github.com/alphagov/smart-answers/blob/3c8ff7a363617d959145b70ac58a7d934757b1ec/app/models/world_location.rb#L61

app/models/world_location.rb
ikennaokpala: yes true..

app/models/world_location.rb
```diff
@@ -53,16 +56,21 @@ def initialize(data)
...
+  end
+
   def ==(other)
-    other.is_a?(self.class) && other.slug == self.slug
+    other.is_a?(self.class) && other.slug == slug
   end
 
   def_delegators :@data, :title, :details
-  def_delegators :details, :slug
+  # def_delegators :details, :slug
```
ikennaokpala: This has been removed,.. see https://github.com/alphagov/smart-answers/blob/3c8ff7a363617d959145b70ac58a7d934757b1ec/app/models/world_location.rb#L61

app/models/world_location.rb
```diff
@@ -53,16 +56,21 @@ def initialize(data)
...
   alias_method :name, :title
 
   def organisations
-    @organisations ||= WorldwideOrganisation.for_location(self.slug)
+    # binding.pry
```
ikennaokpala: This has been removed in later commits and fixed up.. see https://github.com/alphagov/smart-answers/blob/3c8ff7a363617d959145b70ac58a7d934757b1ec/app/models/world_location.rb#L61

app/models/world_location.rb
```diff
@@ -52,28 +53,28 @@ def self.cache_fetch(key)
...
 
-  def_delegators :@data, :title, :details
-  # def_delegators :details, :slug
+  def_delegators :@location, :title, :details, :slug
   alias_method :name, :title
 
   def organisations
-    # binding.pry
-    @organisations ||= WorldwideOrganisation.for_location(slug)
+    @organisations ||= WorldwideOrganisation.for_location(self.slug)
```
ikennaokpala: Yeah, I have change slug to be an instance variable see https://github.com/alphagov/smart-answers/blob/3c8ff7a363617d959145b70ac58a7d934757b1ec/app/models/world_location.rb#L61

test/unit/world_location_test.rb
ikennaokpala: @pmanrubia I have made changes that no longer require defining the slug method anymore..

See https://github.com/alphagov/smart-answers/blob/3c8ff7a363617d959145b70ac58a7d934757b1ec/app/models/world_location.rb#L61

pmanrubia:
@ikennaokpala good to merge. Could you please test that all the calls through the adapters are working properly?

ikennaokpala:
@pmanrubiaI have reviewed calls made through the gds-adapters and all responded properly.. Kindly change the labels and requested reviews

PR #2804, opened 2016-11-03
Move environment variables to test helper

tijmenb:
These two environment variables need to live in the test helper so that they're set every time, regardless wether or not the test is running on CI.

Also, it's better to use `test.gov.uk`, since it won't exist and won't cause inadvertent real requests.


rubenarakelyan:
👍

PR #2805, opened 2016-11-03
Use content store for breadcrumbs, sidebar & metadata

tijmenb:
This commit changes the way we render breadcrumbs, sidebar and content-store. It is part of similar work in all "mainstream"-like applications on GOV.UK (https://trello.com/c/mOeDK914).

Instead of the content-api, we now use the content-store as the data source for everything. We also use the `govuk_navigation` gem to generate GOV.UK component data.

https://trello.com/c/49441pPb

### DIFF
--- a/app/controllers/smart_answers_controller.rb
+++ b/app/controllers/smart_answers_controller.rb
@@ -1,7 +1,10 @@
 class SmartAnswersController < ApplicationController
+  include Slimmer::SharedTemplates
+
   before_action :find_smart_answer, except: %w(index)
   before_action :redirect_response_to_canonical_url, only: %w{show}
   before_action :set_header_footer_only, only: %w{visualise}
+  before_filter :setup_navigation_helpers_and_content_item, except: %w(index)
 
   rescue_from SmartAnswer::FlowRegistry::NotFound, with: :error_404
   rescue_from SmartAnswer::InvalidNode, with: :error_404
@@ -11,7 +14,6 @@ class SmartAnswersController < ApplicationController
   end
 
   def show
-    set_slimmer_artefact(@presenter.artefact)
     respond_to do |format|
       format.html { render }
       format.json {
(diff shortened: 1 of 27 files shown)


app/controllers/smart_answers_controller.rb
```diff
@@ -95,13 +97,23 @@ def set_header_footer_only
   end
 
   def set_expiry(duration = 30.minutes)
-    # if the artefact returned from the Content API is blank, or if
-    # the request to the Content API fails, set a very short cache so
-    # we don't cache an incomplete page for a while
-    duration = 5.seconds if @presenter.present? && @presenter.artefact.blank?
-
     if Rails.configuration.set_http_cache_control_expiry_time
       expires_in(duration, public: true)
```
rubenarakelyan: Does this also need to be removed now that `duration` is no longer set?

tijmenb: `duration` is passed in as an argument to this method.

rubenarakelyan: Silly me

rubenarakelyan:
👍  Looks great, one for the smart answers team?

app/controllers/smart_answers_controller.rb
```diff
@@ -95,13 +97,23 @@ def set_header_footer_only
...
     end
   end
+
+  def setup_navigation_helpers_and_content_item
+    @content_item = Services.content_store.content_item!("/" + params[:id]).to_hash
+
+    # Remove the organisations from the content item - this will prevent the
+    # govuk:analytics:organisations meta tag from being generated until there is
```
mobaig: Why don't we want this meta tag set? Reading this without any prior context, this comment is a little mysterious. Could do with a little more detail.

app/controllers/smart_answers_controller.rb
```diff
@@ -95,13 +97,23 @@ def set_header_footer_only
...
+    # a better way of doing this.
+    if @content_item["links"]
+      @content_item["links"].delete("organisations")
+    end
+
+    @navigation_helpers = GovukNavigationHelpers::NavigationHelper.new(@content_item)
+  rescue GdsApi::HTTPNotFound, GdsApi::HTTPGone
+    @navigation_helpers = @content_item = nil
```
mobaig: Surprised the linter didn't complain 😉 . Reckon this'd be clearer on 2 lines.

PR #2810, opened 2016-11-06
Fix regression tests and unit tests

pmanrubia:
- Fixes a regression test issue introduced in #2805.
- Fixes a unit test issue introduced in #2790 

@ikennaokpala: could you please have a look at the issue in #2790. We need to fix it properly.


pmanrubia:
@ikennaokpala @leenagupte 
I am merging this PR without a code review as we need to do an important deployment tomorrow.
Please, kind review it in case you spot any issue.

@ikennaokpala please, have a look at the issue introduced in #2790

ikennaokpala:
@pmanrubia Will take a look at it..

PR #2814, opened 2016-11-10
Update content store environment variable

pmanrubia:
In #2805 we changed the way we render breadcrumbs and sidebars.  Instead of using the [content-api][1] we use the [content-store][2] as the data source for all the information. We are also using [govuk_navigation][3] gem to generate GOV.UK component data.

This PR update documents and scripts that still refer to [content-store][2]:

- Heroku app
- common-errors.md
- development-without-vm
- startup_heroku.sh

[1]: https://github.com/alphagov/govuk_content_api
[2]: https://github.com/alphagov/content-store
[3]: https://github.com/alphagov/govuk-content-navigation

### DIFF
--- a/app.json
+++ b/app.json
@@ -5,7 +5,7 @@
   "success_url": "/",
   "env": {
     "GOVUK_APP_DOMAIN": "integration.publishing.service.gov.uk",
-    "PLEK_SERVICE_CONTENTAPI_URI": "https://www.gov.uk/api",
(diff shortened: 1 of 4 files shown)


ikennaokpala:
@pmanrubia This LGTM 👍

pmanrubia:
@ikennaokpala thank you for the review!

PR #2815, opened 2016-11-10
PIP Checker: Fix Timecop vs ::ActiveSupport::TimeWithZone issue

ikennaokpala:
Trello cards
- [#409](https://trello.com/c/wPgE2Iuf/409-pipchecker-fix-failing-test-that-depends-on-time)
- [#414](https://trello.com/c/buSLWlnU/414-pip-checker-fix-timecop-vs-activesupport-timewithzone-issue)

## Description

It appears to me the main reason this fails (i.e middle_group?),
This is because the 16.year.ago (which is from ActiveSupport::Duration#ago [1])
uses :Time.current [2], which underneath returns ::Time.zone.now.

::Time.zone.now is an instance of ActiveSupport::TimeWithZone which
Timecop [3] appears to ignore.

According to this comment [3] Time.cop only caters for the following:

- Date.today
- Time.now
- DateTime.now

Years (i.e 16.years) is only a measurement of time .. see its
implementation here [4]

This is the reason why it is now important to include pass Date.today
to ago.

[1] https://github.com/rails/rails/blob/03cd88a07221a1511c9d06fe4173052bb5554d0b/activesupport/lib/active_support/duration.rb#L69
[2] https://github.com/rails/rails/blob/b775987e72260233c66080483b3c964f9549d094/activesupport/lib/active_support/core_ext/time/calculations.rb#L45

[3] https://github.com/travisjeffery/timecop/blob/master/lib/timecop/timecop.rb#L10

[4]  https://github.com/rails/rails/blob/b451de0d6de4df6bc66b274cec73b919f823d5ae/activesupport/lib/active_support/core_ext/integer/time.rb#L35

### DIFF
--- a/lib/smart_answer/calculators/pip_dates.rb
+++ b/lib/smart_answer/calculators/pip_dates.rb
@@ -12,7 +12,7 @@ module SmartAnswer::Calculators
     end
 
     def in_middle_group?
-      self.dob > GROUP_65_CUTOFF && self.dob <= 16.years.ago
+      self.dob > GROUP_65_CUTOFF && self.dob <= 16.years.ago(Date.today)
     end
 
     def turning_16_before_oct_2013?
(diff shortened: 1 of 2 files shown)


pmanrubia:
@ikennaokpala  Thank you for fixing it 👍 

LGTM

PR #2817, opened 2016-11-11
Change plek service content store uri

ikennaokpala:
## Description

This PR changes the plek service content store uri from [1] to [2]

This is becuase [1] returns GdsApi::HTTPUnauthorized response when
smartanswers tries to use the content store from heroku, but at the
moment [2] appears to grant access to the content store.

[1]  https://www-origin.integration.publishing.service.gov.uk/api/
[2]  https://www.gov.uk/api/


pmanrubia:
👍

PR #2821, opened 2016-11-18
Add good and bad examples of validation

chao-xian:
A comment in the doc outlines practices for using validation but did not show examples. I've added a good and bad example.


leenagupte:
👍  LGTM

PR #2800, opened 2016-11-01
Create one outcome per path for Italy in marriage-abroad

leenagupte:
Trello card: https://trello.com/c/81U1zVl0
Supersedes PR #2799 
Builds on PR #2767 

## Motivation

Marriage abroad is the most complex Smart Answer to update. It's also one of the most frequent Smart Answers to be updated. We want to simplify the updates so they are quicker, less complex and reduce the need for Developers to make the update. By creating one template per outcome for a country we greatly reduce the mount of logic in the templates.

## Expected changes.
As this is a refactor of Italy, and not a change to the content or flow of the smart-answer, users should not experience any changes.  The changes that have occurred to the code are:

1. Create a new generic country outcome.
2. Create one outcome per path through the flow for Italy. This results in 18 new outcomes.
3. Remove all Italy specific partials. 
4. Remove references to Italy in other outcomes.
5. Remove Italy from all country lists.

[URL on GOV.UK](https://www.gov.uk/marriage-abroad/y/italy)
[Preview Link](https://smart-answers-pr-2800.herokuapp.com/marriage-abroad/y/italy)

PR #2825, opened 2016-11-25
Create one outcome per path for Sweden in marriage-abroad

leenagupte:
Trello card: https://trello.com/c/bb0m6XM1
Follows on from: PR #2800 

## Motivation

## Expected changes

As this is a refactor of Sweden, and not a change to the content or flow of the smart-answer, users should not experience any changes. The changes that have occurred to the code are:

Create one outcome per path through the flow for Sweden. This results in 18 new outcomes.
Remove all Sweden specific partials.
Remove references to Sweden in other outcomes.
Remove Sweden from all country lists.


leenagupte:
Note to self: Split commit 997070f9542dd29160a7952bc82baffc6a3a38c1

pmanrubia:
LGTM

PR #2822, opened 2016-11-21
Change Portugal link to list of translators

chao-xian:
### Trello card 
[Register a death: Fix Portugal translators link](https://trello.com/c/6zuLYkMl/405-register-a-death-fix-portugal-translators-link)

## Description
Fix link on Register A Death that points to list of lawyers when it should be pointing to a list of translators.

## Factcheck
[Factcheck version](https://smart-answers-pr-2822.herokuapp.com/register-a-death/y/overseas/portugal/in_the_uk)

## Expected changes
[URL on GOV.UK](https://www.gov.uk/register-a-death/y/overseas/portugal/in_the_uk)

#### Before

https://www.gov.uk/register-a-death/y/overseas/portugal/in_the_uk

image

#### After
https://smart-answers-pr-2822.herokuapp.com/register-a-death/y/overseas/portugal/in_the_uk

image

PR #2826, opened 2016-11-28
Generate test artefacts for Portugal

leenagupte:
Generate test artefacts for Portugal in register-a-birth

The regression tests were failing for Portugal in register-a-birth.
See: https://ci.dev.publishing.service.gov.uk/job/govuk_smart_answers_regressions/6112/console

I think that when PR #2822 was rebased over PR #2820 the shared change
to the GRO text for North Korea needed to be regenerated.


pmanrubia:
LGTM

PR #2831, opened 2016-12-02
How to generate checksums example

chao-xian:
Add examples of how to (re-)generate checksums.

When I tried to run the script, I didn't know I had to use the rails
command or what parameters looked like. After being shown, I have added
examples to the docs.


ikennaokpala:
Just like you have explained.. found the single file option inconsistent.. barring that think this is a useful piece of information to have.. thanks for doing this.. 👍  LGTM

PR #2837, opened 2016-12-12
Update _same_sex.govspeak.erb

raykhan66:
Replaced

'It’s not possible to get legal recognition for your same-sex relationship in Mozambique.'

With

Brand new content to reflect that same-sex marriage is now recognised.


raykhan66:
Closed because I don't care about my work!

PR #2839, opened 2016-12-13
Create outcome per path for France Marriage Abroad

chao-xian:
## Trello card
https://trello.com/c/tKI6NBOS

## Motivation

France already has one outcome per path, but it shares them with the French Territories. This PR add templates only used for France in a structure similar to the outcomes for Italy and Sweden. 

## Factcheck
[marriage-abroad France](https://smart-answers-pr-2839.herokuapp.com/marriage-abroad/y/france)

## Expected changes

[URL on GOV.UK](https://www.gov.uk/marriage-abroad/y/france)
* Change the question asked after the user selects `France`.
* Change how the outcome templates for France are structured in the code base.

### Before

screen shot 2016-12-16 at 12 53 24

### After

screen shot 2016-12-16 at 12 53 08


leenagupte:
@ikennaokpala I've updated the commit structure as you recommended, and expanded some of the commit notes.

ikennaokpala:
I am updating this PR from its last passing commit 4b2e852 and update from @sistephens 1b87e37

PR #2847, opened 2016-12-22
Create an outcome per path for Vietnam in marriage-abroad

leenagupte:
Trello card: https://trello.com/c/UYyWUwS6

## Motivation

## Expected changes

[URL on GOV.UK](https://www.gov.uk/marriage-abroad/y/vietnam)

As this is a refactor of Vietnam, and not a change to the content or flow of the smart-answer, users should not experience any changes. The changes that have occurred to the code are:

* Create one outcome per path through the flow for Vietnam. This results in 18 new outcomes.
* Remove old outcome templates previously only used by Vietnam
* Remove Vietnam related logic and content from other shared outcome templates
* Remove Vietnam from all country lists.


leenagupte:
@ikennaokpala I thought about changing `(%w(cambodia).exclude?(country_slug))` in c3df2aa to `country-slug != "cambodia"` but thought it would make the rest of the conditional harder to read.

ikennaokpala:
@leenagupte I can understand that..Hopefully as we continue to flatten country outcome.. this will eventually go away.

pmanrubia:
- I am removing the label `waiting for factcheck` as this is not needed for this PR
- Rebasing with master, removing commit e60791bb5b3a18733784bd94e1343bce672b82e3, and force pushing in preparation for merging.

PR #2849, opened 2016-12-29
Flatten marriage aboard flow for the Philippines and use only two question/outcomes

ikennaokpala:
[Trello card](https://trello.com/c/TRzFnwG3/466-marriage-abroad-improvement-philippines-2-outcomes)

## Motivation

## Expected changes

[URL on GOV.UK](https://www.gov.uk/marriage-abroad/y/philippines)
[Preview link](https://smart-answers-pr-2849.herokuapp.com/marriage-abroad/y/philippines)

As this is a refactor of Philippines flow, and not a change to the content or flow of the smart-answer, users should not experience any changes. The changes that have occurred to the code are:

* Create one outcome per path through the flow for Philippines. This results in 20 file changes.
* Remove old outcome templates previously only used by Philippines
* Remove Philippines related logic and content from other shared outcome templates
* Remove Philippines from all country lists.
* Consolidation of opposite and same sex content.
* Change Philippines flow to ask only two questions

**NB:** There should be noticeable changes in the fees tables but regression artefacts don't reflect this because the configuration.yml date set to 2015-01-01, hence the old fees are being used.


lib/smart_answer_flows/marriage-abroad.rb
```diff
@@ -29,7 +29,7 @@ def define
         end
 
         next_node do
-          if calculator.ceremony_country == 'ireland'
+          if %w(ireland philippines).include?(calculator.ceremony_country)
```
pmanrubia: @leenagupte I would like to be explicit on line 32 about the country being migrated to 2 outcomes. 
I would like to have something like:

```ruby
if calculator.is_two_outcomes_country?
# ...
```

As there are a few countries already in fact-check [I have created a Trello card][1] to perform the refactor.

[1]: https://trello.com/c/V1T9dLEg/481-refactor-for-marriage-abroad-two-outcomes

lib/smart_answer/calculators/marriage_abroad_calculator.rb
```diff
@@ -334,6 +334,10 @@ def path_to_outcome
                    'opposite_sex'
                  end
 
+      if ceremony_country == 'philippines'
```
pmanrubia: I would like to refactor this method to something like:

```ruby
if calculator.is_two_outcomes_country?
  [ceremony_country_path, same_sex_path]
else
  [ceremony_country_path, ceremony_location_path, partner_nationality_path, same_sex_path]
```

As there are a few countries already in fact-check [I have created a Trello card][1] to perform the refactor.

[1]: https://trello.com/c/V1T9dLEg/481-refactor-for-marriage-abroad-two-outcomes

leenagupte: Ireland only has two outcomes, but it's templates are at the top-level of the `outcomes` folder. We'll need to refactor the templates before we can add it to any `two outcomes` list.

ikennaokpala:
Update from 	30fcc68

PR #2851, opened 2017-01-03
Fix state pension topup tests

leenagupte:
The state-pension-topup tests check how much you need to pay in
each calendar year.

As we are in a new year, the number of payments has decreased.

This scheme ends in April this year, and this calculator will be retired.
Rather make radical changes to the tests, I've frozen the date to a time when
they did still pass.

These changes should have been incorporated in https://trello.com/c/ymI446hb


chao-xian:
LGTM

PR #2857, opened 2017-01-05
Put body text into hint for benefits cap calculator postcode question

theleebriggs:
## Trello

Part of card https://trello.com/c/SQwpXvCF/339-sa-questions-page-restructure-quick-tasks and https://trello.com/c/fEeJdpVr/506-smart-answer-accessibility-changes

## Motivation

Make this question consistent with the rest of gov.uk  - the "body" copy is more appropriate in the question hint.

## Expected change

Question content previously in the body section should now be in the hint.

### Before

[gov.uk](https://www.gov.uk/benefit-cap-calculator/y/yes/no/no/child_benefit/90.0/90.0/parent)

old

### After

new

### DIFF
--- a/lib/smart_answer_flows/benefit-cap-calculator/questions/enter_postcode.govspeak.erb
+++ b/lib/smart_answer_flows/benefit-cap-calculator/questions/enter_postcode.govspeak.erb
@@ -3,7 +3,7 @@
 
 <% end %>
 
-<% content_for :body do %>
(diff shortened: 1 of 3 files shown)


selfthinker:
LGTM :+1:

As discussed earlier, this doesn't need factcheck, so can be merged at once. Only let Kati know when it's ready to go live.

PR #2854, opened 2017-01-04
Create an outcome per path for Latvia and copy changes

leenagupte:
Trello card: https://trello.com/c/TVGfheIY

## Factcheck
[marriage-abroad Latvia](https://smart-answers-pr-2854.herokuapp.com/marriage-abroad/y/latvia/uk/partner_british/opposite_sex)

## Expected changes

* [URL on GOV.UK]
   * One outcome per path
   * Copy changes for opposite sex marriages where the user is in the UK.

### Before

screen shot 2017-01-04 at 10 01 19

### After

screen shot 2017-01-04 at 10 00 55

PR #2884, opened 2017-01-16
latvia uk other table

JamesButlerGDS:
Replaced table with partial for opposite sex outcome when user is in UK and partner is from other country

https://trello.com/c/TVGfheIY/370-5-latvia-marriage-tool-opposite-sex-marriages-now-through-affirmations-content-change-request


ikennaokpala:
Superseded by #2854

PR #2894, opened 2017-01-25
Marriage abroad hungary

pmanrubia:
[Trello card](https://trello.com/c/jyKt4Iwx/489-3-create-one-outcome-per-path-for-hungary-marriage-abroad)

### Motivation

### Expected changes

As this is a refactor of Hungary, and not a change to the content or flow of the smart-answer, users should not experience any changes. The changes that have occurred to the code are:

Create one outcome per path through the flow for Hungary. This results in 18 new outcomes.
Remove references to Hungary in other outcomes.
Remove Hungary from all country lists.


pmanrubia:
Good spot @ikennaokpala. Thank you!

Rebasing and forcepushing addressing your changes.

PR #2897, opened 2017-01-27
One outcome per path China

theleebriggs:
# Need
https://trello.com/c/VhskEufa/488-3-create-one-outcome-per-path-for-china-marriage-abroad

# Description
Refactor marriage abroad in China to use the new "outcome per path" structure to allow content designers to more easily make updates to content.

The content output should not be affected by this change, it is purely structural.


theleebriggs:
@ikennaokpala I've made those newline changes, rebased as well.

PR #2899, opened 2017-01-27
Display taxon breadcrumbs for Education A/B test

alecgibson:
Show taxon breadcrumbs when content is tagged to a taxon, and the user is in the correct A/B testing bucket.

https://trello.com/c/6TLMKxot/337-changes-to-education-related-content-pages-as-part-of-the-new-navigation-in-smartanswers


app/controllers/smart_answers_controller.rb
```diff
@@ -123,4 +126,13 @@ def setup_navigation_helpers_and_content_item
...
   end
+
+  def breadcrumbs
+    return {} if @navigation_helpers.nil?
+    if should_present_new_navigation_view?
+      @navigation_helpers.taxon_breadcrumbs
+    else
+      @navigation_helpers.breadcrumbs
+    end
+  end
```
alecgibson: (See discussion here: https://github.com/alphagov/frontend/pull/1109#discussion_r98249455)

tijmenb:
Requests that are A/B tested need to return a `Vary` header so Fastly knows to vary on our AB header.

But I'd rather have that this is fixed by using the client library, so that the test helpers can detect this problem. Marking DNM until we've started that.

alecgibson:
@tijmenb this is now using the officially published A/B Testing gem.

test/functional/smart_answers_controller_test.rb
```diff
@@ -193,6 +194,54 @@ def teardown
         assert_select "pre.debug", false, "The page should not render debug information"
       end
     end
+
+    context "A/B testing" do
+      setup do
+        # Switch on feature flag
+        ENV['ENABLE_NEW_NAVIGATION'] = 'yes'
```
mobaig: Unless ENV is being reset somewhere this is leaking state. Probably best to stub access to that key.

alecgibson: Yup, about to push a teardown

alecgibson:
@carvil I've also bumped the versions of `govuk_navigation_helpers` and `govuk_ab_testing` on this. Also deployed to int, and can confirm that breadcrumbs work as expected.

carvil:
@alecgibson thanks, could you please add the beta label as well?

alecgibson:
@carvil done

PR #2905, opened 2017-02-06
Remove ajax

theleebriggs:
## Trello

https://trello.com/c/j7tWmjbO/324-remove-ajax-from-smart-answers

## Desire

Using AJAX to navigate through smart answers does not have any positive effects.

1. A performance test resulted in quite insignificant speed improvements
   with AJAX when compared against the same without AJAX.
2. When we removed the AJAX from 3 specific Smart Answers,
   it didn't have any impact on user behaviour at all.

## Outcome

JSON rendering functions have been removed from the main smart answers controller; tests for these have been removed and in some cases refactored to test the standard html flow.

Smart answer views have be refactored into a set of partials with a smart answer specific layout file.

The updates do not affect the actual content of smart answers.


theleebriggs:
@pmanrubia I've reworded the commits and did a slightly messy rebase with master - all tests pass after i resolved the conflicts (mainly around A/B testing code). Let me know what you think.

fofr:
My interpretation of this:
> A performance test resulted in quite insignificant speed improvements
with AJAX when compared against the same without AJAX.
> When we removed the AJAX from 3 specific Smart Answers,
it didn't have any impact on user behaviour at all.

Is that AJAX provides a small benefit? How small were those benefits?
Did we find that the AJAX ever impeded users?

(I can't see the Trello card linked to as it's private)

pmanrubia:
Hi @fofr, thank you for your feedback (I have added you to our board)

Our main reason for deleting the AJAX support is the way it is implemented. It is increasing the complexity of the codebase, which is one of the biggest issues with smart-answers, so far. Happy to follow a different approach if you have any other suggestion.

nickcolley:
I've been doing some thinking around how we can improve the 'One thing per page' pattern, really the main impact I can see is reducing the amount of round trips a user has to go through.

For example if we cached X pages in the journey with a Service Worker etc.

This AJAX approach does not reduce the journey in that way so I think it's reasonable to make this codebase simpler and explore the above later.

nickcolley:
CI is failing: https://ci.integration.publishing.service.gov.uk/job/smartanswers/job/remove-ajax/6/console

PR #2909, opened 2017-02-08
Fix state pension topup 2017 birthday bug

leenagupte:
Trello card: https://trello.com/c/M542k4eW

## Motivation

The state pension popup calculator is supposed to show you how large of a lump sum payment you need to make to popup your weekly pension.

This lump sum changes depending on how old you are. So, you would pay smaller lump sum when you are 70 years old than when you are 69.

The potential lump sums displayed to the user, shows them a list of payments they can make at each age.

e.g.

£18,450 when you’re 72
£17,975 when you’re 73

The code to generate this churns out one payment amount per calendar year. This seemed ok because you only have one birthday a year, and could therefore only make one payment.

There was a problem with this though. The final year of the scheme, 2017, only runs from 1st January to 5th April. If you birthday is after 6th April, everything is fine. You can only make a payment based on your current age.

However, if your birthday happens to fall between these 3 months, you  could potentially make a lower lump sum payment on or after your next birthday.

The calculator did not handle this last scenario.

I’ve added an ugly hack to get around this problem. The calculator is being retired in 2 months, so it doesn’t seem worth the effort to come up with a more elegant solution.

## Factcheck
[state-pension-topup](https://smart-answers-pr-2909.herokuapp.com/state-pension-topup/y/1948-04-05/male/20)

## Expected changes
*  [URL on GOV.UK](https://www.gov.uk/state-pension-topup/y/1948-04-05/male/20)

### Before
screen shot 2017-02-08 at 11 48 04

### After
screen shot 2017-02-08 at 11 48 35

### DIFF
--- a/lib/smart_answer/calculators/state_pension_topup_calculator.rb
+++ b/lib/smart_answer/calculators/state_pension_topup_calculator.rb
@@ -36,7 +36,7 @@ module SmartAnswer::Calculators
         rows << { amount: lump_sum_amount(age, weekly_amount), age: age } if age >= retirement_age(gender)
         age += 1
       end
-      if (TOPUP_END_DATE.year == Date.today.year) && (dob.month < 5) && !birthday_after_topup_end?(dob, age)
+      if (TOPUP_END_DATE.year == Date.today.year) && (dob.month < 5) && (birthday(dob) > Date.today) && !birthday_after_topup_end?(dob, age)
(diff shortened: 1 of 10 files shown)


leenagupte:
@ikennaokpala I copied the line to append a row from the loop directly above it. As it's a short-term hack I didn't put any thought into what the guard is needed for.

pmanrubia:
Agree with @ikennaokpala's point, but we are going to retire this Smart Answer in a few weeks... let's move to fact-check.

PR #2911, opened 2017-02-09
Migrate smart-answers to new CI environment

leenagupte:
Trello card: https://trello.com/c/94VbayRY

## Motivation

The old CI environment is being turned off, so we need to start building smart-answers on the new Jenkins2 environment.

Jenkins2 looks for the presence of a `Jenkinsfile` and runs that by default.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,53 @@
+#!/usr/bin/env groovy
+
+REPOSITORY = 'smart-answers'
+
+node {
+  def govuk = load '/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy'
+
+  try {
+    stage('Checkout') {
+      checkout scm
+      govuk.cleanupGit()
+      govuk.mergeMasterBranch()
+      govuk.contentSchemaDependency()
+      govuk.setEnvar("GOVUK_CONTENT_SCHEMAS_PATH", "tmp/govuk-content-schemas")
+      govuk.setEnvar("DISPLAY", ":99")
+    }
+
+    stage('Bundle') {
+      govuk.bundleApp()
+    }
+
+    stage('Linter') {
+      govuk.rubyLinter()
+    }
+
+    stage("Assets") {
+      govuk.precompileAssets()
+    }
+
+    stage('Tests') {
+      govuk.runTests()
+    }
+
+    if (env.BRANCH_NAME == 'master') {
+      stage('Push release tag') {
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -0,0 +1,51 @@
...
+      govuk.bundleApp()
+    }
+
+    stage('Linter') {
+      govuk.rubyLinter()
+    }
+
+    stage("Assets") {
+	    govuk.precompileAssets()
+	  }
```
cbaines: Some inconsistent indentation here.

Jenkinsfile
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    stage('Push release tag') {
+      govuk.pushTag(REPOSITORY, BRANCH_NAME, 'release_' + BUILD_NUMBER)
+    }
+
+    stage('Deploy to Integration') {
+      // Deploy on Integration (only master)
+      govuk.deployIntegration(REPOSITORY, BRANCH_NAME, 'release', 'deploy')
+    }
```
cbaines: I'd recommend wrapping the 'Push release tag' and 'Deploy to Integration' stages to only run when building the master branch (e.g. like [1]), as this makes the display in Jenkins less surprising.

1: https://github.com/alphagov/publishing-api/blob/master/Jenkinsfile#L55

Jenkinsfile
```diff
@@ -0,0 +1,51 @@
...
+
+node {
+  def govuk = load '/var/lib/jenkins/groovy_scripts/govuk_jenkinslib.groovy'
+
+  try {
+    stage('Checkout') {
+      checkout scm
+      govuk.cleanupGit()
+      govuk.mergeMasterBranch()
+      govuk.contentSchemaDependency()
```
cbaines: If smart-answers depends on the govuk-content-schemas, this might be a good time to add the functionality to run the smart-answers tests against changes to the govuk-content-schemas, there are some docs here:
https://github.gds/pages/gds/opsmanual/infrastructure/testing/application-testing.html?highlight=schemas#content-schema-tests

leenagupte: @cbaines There is another ticket to run the smart-answers regression tests. I'll add this task to that ticket. Thanks!

Jenkinsfile
leenagupte: Thanks! Will update.

Jenkinsfile
leenagupte: @cbaines Thanks! I'll do that and fix-up.

PR #2926, opened 2017-02-22
Bump slimmer to use Rails' cache

tijmenb:
This bumps slimmer to make it use the rails cache and send a user agent:

- https://github.com/alphagov/slimmer/pull/184
- https://github.com/alphagov/slimmer/pull/183

It will prevent this app from making many needless requests to `static`.

https://trello.com/c/D9HmkJwI


Gemfile.lock
```diff
@@ -300,13 +300,16 @@ DEPENDENCIES
...
-  slimmer (~> 10.0.0)
+  slimmer (~> 10.1.1)
   tilt (= 1.4.1)
   timecop
   uglifier
   uk_postcode (~> 1.0.1)
   unicorn (= 4.8.3)
   webmock (= 1.20.4)
 
+RUBY VERSION
```
carvil: Did you mean to include this here @tijmenb ?

PR #2949, opened 2017-03-07
Only add A/B testing Vary header and meta tag to education pages

suzannehamilton:
Ensure that only pages in the navigation A/B test (those which are tagged to taxons) are cached separately and tracked in the A/B test analytics.

This change touches a lot files because we now have to set up a content item in all the tests which call `SmartAnswersController`. I think that's the correct thing to do, though, short of major refactoring to make the tests more isolated from the controller logic.

https://trello.com/c/Xehe42eM/464-only-add-a-b-meta-tag-to-pages-in-the-navigation-a-b-test

### DIFF
--- a/test/functional/smart_answers_controller_test.rb
+++ b/test/functional/smart_answers_controller_test.rb
@@ -50,6 +50,10 @@ class SmartAnswersControllerTest < ActionController::TestCase
   end
 
   context "GET /<slug>" do
+    setup do
+      stub_smart_answer_in_content_store("smart-answers-controller-sample")
+    end
+
     should "respond with 404 if not found" do
       @registry = stub("Flow registry")
       @registry.stubs(:find).raises(SmartAnswer::FlowRegistry::NotFound)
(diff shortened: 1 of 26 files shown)


test/functional/smart_answers_controller_test.rb
```diff
@@ -196,65 +200,83 @@ def teardown
...
-
-      should "show normal breadcrumbs by default" do
-        get :show, id: 'smart-answers-controller-sample'
-
-        assert_match(/NormalBreadcrumb/, response.body)
-        refute_match(/TaxonBreadcrumb/, response.body)
-        sidebar = Nokogiri::HTML.parse(response.body).at_css(".related-container")
-        refute_match(/A Taxon/, sidebar)
-      end
+      context "pages under A/B test" do
```
alecgibson: NB: I recently swapped this pattern out for a neater one in [`frontend`](https://github.com/alphagov/frontend/commit/f03ea746049f65d144844b065865022a6977ae6e).

Obviously fine to leave unless you feel like being a particularly good citizen.

PR #3000, opened 2017-04-12
Add implementation and tests to render and validate start button text and link

ikennaokpala:
[Trello card](https://trello.com/c/7AFqtTjE/598-important-overseas-passports-add-integration-test-to-validate-link-forwards-to-external-service)

## Description 

This PR includes implementation, unit and integration test for the stage page button text and link.

This ensures that the expected button text and link are rendered and validated each time the page is rendered or the tests are run against recent changes.

It also updates the start button link href with the start_button_href helper method.

The logic  in start_button_href helper has led to the update of the test to reflect the correct external link being used.

### DIFF
--- a/test/integration/start_page_test.rb
+++ b/test/integration/start_page_test.rb
@@ -1,0 +1,32 @@
+require_relative "../integration_test_helper"
+
+class StartPageTest < ActionDispatch::IntegrationTest
+  context "start page" do
+    RegisterableSmartAnswers.new.flow_presenters.each do |flow_presenter|
+      slug = flow_presenter.slug
+      context "when the smart answer is `#{slug}`" do
+        setup do
+          @view = mock("view")
+          @view.stubs(:smart_answer_path).returns("/#{slug}/y")
+        end
+
+        should "have button text and href belonging to this smart answer" do
+          start_button = SmartAnswer::StartButton.new(slug, @view)
+          Services.content_store.stubs(:content_item)
+            .with("/#{slug}")
+            .returns({})
+
+          visit "/#{slug}"
+
+          assert_current_url "/#{slug}"
+          within ".intro" do
+            assert page.has_link?(
+              start_button.text,
+              href: start_button.href
+            )
+          end
+        end
+      end
+    end
+  end
+end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -4,21 +4,10 @@ module ApplicationHelper
   end
 
   def start_button
-    case @name.to_s
-    when "overseas-passports"
-      "Continue"
-    when "calculate-your-child-maintenance"
-      "Calculate your child maintenance"
-    else
-      "Start now"
-    end
+    SmartAnswer::StartButton.new(@name, self).text
   end
 
   def start_button_href
-    if @name.to_s == "overseas-passports"
-      "https://www.passport.service.gov.uk/filter"
-    else
-      smart_answer_path(@name, started: 'y')
-    end
+    SmartAnswer::StartButton.new(@name, self).href
   end
 end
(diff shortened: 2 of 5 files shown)


test/helpers/fixture_flows_helper.rb
```diff
@@ -1,4 +1,16 @@
 module FixtureFlowsHelper
+  BUTTON_TEXT_AND_LINK_FIXTURE = {
```
pmanrubia: `FixtureFlowsHelper` is included in most of the functional tests, so I don't think it is the right place to place `BUTTON_TEXT_AND_LINK_FIXTURE` and `button_text_and_link`

The constant and the method are just expectations you want to meet, it is nothing related to fixtures; I would place this code close to the test where it is being used.

test/helpers/fixture_flows_helper.rb
```diff
@@ -1,4 +1,16 @@
...
+    "overseas-passports": {
+      start_button_text: "Continue",
+      href: "https://www.passport.service.gov.uk/filter"
+    },
+    "calculate-your-child-maintenance": {
+      start_button_text: "Calculate your child maintenance",
+      href: "/calculate-your-child-maintenance/y"
+    }
+  }.with_indifferent_access
```
pmanrubia: If we add `with_indifferent_access` to a hash that is set up in a test, it has to be because it mimics the data in the code we are testing. I don't think this is happening here. 

It would be better to use [deep_stringify_keys][1] or [deep_symbolize_keys][2].

[1]: https://github.com/rails/rails/blob/8b2d978d2892abe85859c75dfcbdfe218b98de81/activesupport/lib/active_support/hash_with_indifferent_access.rb#L260
[2]: https://github.com/rails/rails/blob/8b2d978d2892abe85859c75dfcbdfe218b98de81/activesupport/lib/active_support/hash_with_indifferent_access.rb#L264

test/helpers/fixture_flows_helper.rb
```diff
@@ -21,4 +33,15 @@ def stub_smart_answer_in_content_store(smart_answer_id)
       .with("/#{smart_answer_id}")
       .returns({})
   end
+
+  def button_text_and_link(smart_answer_slug)
```
pmanrubia: Same as the comment above, why do we have this method here? It is duplicating the behaviour specified in the [application helper][1]. 

[1]: https://github.com/alphagov/smart-answers/blob/432830cf9b3e9dbafbe4c63d886b29b900947151/app/helpers/application_helper.rb#L14-L31

test/integration/start_page_test.rb
```diff
@@ -0,0 +1,26 @@
+require_relative "../integration_test_helper"
+require_relative "../helpers/fixture_flows_helper"
+
+class StartPageTest < ActionDispatch::IntegrationTest
```
pmanrubia: This integration test is not describing the behaviour of the different types of start buttons we have. Everything is hidden behind FixtureFlowsHelper.

I would expect to have something like:

```ruby
describe "start_page" do
  context "when the smart answer is `overseas-passports`" do
  end
  context "when the smart answer is `calculate-your-child-maintenance`" do
  end
  context "otherwise" do
  end
end
```

test/helpers/fixture_flows_helper.rb
ikennaokpala: Fair point..

test/helpers/fixture_flows_helper.rb
ikennaokpala: Yeah I agree.. I have made some changes..

ikennaokpala:
Based on the observations from @pmanrubia review.. we think it is best to close this and raise another to properly reflect the agenda.

PR #3002, opened 2017-04-13
Revert "Remove ajax"

ikennaokpala:
Reverts alphagov/smart-answers#2905


fofr:
Why was this reverted?

ikennaokpala:
@fofr we are currently investigating an anomaly and also given how regularly this repo gets updated.. We have thought is wise to revert this to allow other PR to be merged without this in the way..

PR #3006, opened 2017-04-18
Revert "Revert "Remove ajax""

ikennaokpala:
This PR is dependent on #3007 

Reverts alphagov/smart-answers#3002


ikennaokpala:
This will re-introduce the changes from #2905. This also depends on changes from #3000

PR #3028, opened 2017-05-04
Sort the files relating to a smart answer

cbaines:
This means that the return value of this method is no longer dependant
on the ordering on the underlying file system. Therefore, this helps
any functionality using all_paths or anything derived from it be more
deterministic.

The motivation for this change is the checksum generation output, as
with this change, the order of the files in the generated yaml file is
deterministic.


cbaines:
From talking to @ikennaokpala, I've realised that this isn't just a nicety, but important for doing development of Smart Answers on a file system that does not guarantee a consistent order for file traversal.

cbaines:
I've put up https://github.com/alphagov/smart-answers/pull/3036 which easily allows updating all the files.

PR #3038, opened 2017-05-08
Use redirect unpublishing

cbaines:
Just putting this up before I forget.

### DIFF
--- a/lib/tasks/retire.rake
+++ b/lib/tasks/retire.rake
@@ -7,8 +7,9 @@ namespace :retire do
 
     content_item_publisher = ContentItemPublisher.new
 
-    content_item_publisher.unpublish(args.content_id)
-    content_item_publisher.publish_redirect(args.base_path, args.destination)
+    content_item_publisher.unpublish_with_redirect(
+      args.content_id, args.base_path, args.destination
+    )
   end
 
   desc "Unpublish a smart answer from publishing-api"
--- a/app/services/content_item_publisher.rb
+++ b/app/services/content_item_publisher.rb
@@ -21,6 +21,25 @@ class ContentItemPublisher
     )
   end
 
+  def unpublish_with_redirect(content_id, path, destination)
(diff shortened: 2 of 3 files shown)


chao-xian:
@cbaines did you forget? 😄

cbaines:
> @cbaines did you forget? 😄

The forgetting was more about about publishing anything.

I think this would be an improvement, but I haven't thought through it fully, or done any testing. Are you looking to retire more smart answers soon?

chao-xian:
There is one that came out of Firebreak to convert it to a Simple Smart Answer (which I think is a format in publisher). We're waiting on the Content Team to get green light from SME and once they do we'll retire the smart answer and re-point the base path to the new simple smart answer.

chao-xian:
Would this actually help in the case I've just described?

cbaines:
> Would this actually help in the case I've just described?

This change is mostly a neatness thing, it doesn't really help or hinder as far as I can remember.

lib/tasks/retire.rake
```diff
@@ -7,8 +7,9 @@ namespace :retire do
 
     content_item_publisher = ContentItemPublisher.new
 
-    content_item_publisher.unpublish(args.content_id)
-    content_item_publisher.publish_redirect(args.base_path, args.destination)
+    content_item_publisher.unpublish_with_redirects(
+      args.content_id, args.path, args.destination
```
AlanGabbianelli: Should `args.path` be `args.base_path`?

bevanloon:
@cbaines is this still relevant or can we close the pr?

cbaines:
> @cbaines is this still relevant or can we close the pr?

I believe it's still relevant, well, just for smart answer retiring, which doesn't happen very regularly.

lib/tasks/retire.rake
cbaines: Yep, good spot, I've updated the commit.

PR #3048, opened 2017-05-18
Fix header problem with A variant

sihugh:
This wasn't setting the meta tag or the Vary header correctly for the A variant which caused problems running the test.

The logic for whether the page is under test should not be the same as whether the B variant should be shown.  :man_facepalming: 

We live and learn


chao-xian:
Ah this is useful - and timely for me 😄

sihugh:
Thanks @chao-xian 👍

PR #3051, opened 2017-05-25
A/B test the title and start button on calculate-your-child-maintenance

issyl0:
- This sets up the A/B test for the child maintenance page title and the
  start button text.
- We're not editing the `href` for the start button,
  so don't add `ab_start_button_href` anywhere - it's still going to the
  same place (`/calculate-your-child-maintenance/y`).

**Do not merge this until Tuesday - it deletes (most? some?) of the code for the ongoing BenchmarkInlineLink test, and will require a rebase when we revert the changes from #3046 and #3048 on Tuesday morning. 😱**


issyl0:
Oh, sorry. I've squashed them now. :-)

issyl0:
Thanks, @chao-xian. :1st_place_medal:

issyl0:
I've just rebased this on top of master (the removal of the old one), so it should be ready to go soon.

PR #3054, opened 2017-05-31
Upgrade to Rails 5

mgrassotti:


mgrassotti:
All errors were fixed. There are still 3 deprecation warnings: 
1)
```
DEPRECATION WARNING: Passing strings or symbols to the middleware builder is deprecated, please change
them to actual class references.  For example:

  "ActionDispatch::DebugExceptions" => ActionDispatch::DebugExceptions

 (called from <top (required)> at /var/govuk/smart-answers/config/environment.rb:5)
```
2)
```
warning: parser/current is loading parser/ruby23, which recognizes
warning: 2.3.3-compliant syntax, but you are running 2.3.1.
warning: please see https://github.com/whitequark/parser#compatibility-with-ruby-mri.
```
3)
```
DEPRECATION WARNING: #original_exception is deprecated. Use #cause instead. (called from block (2 levels) in <class:InputValidationTest> at /var/govuk/smart-answers/test/integration/engine/input_validation_test.rb:87)
```

But they don't seem directly related to this repository, so I am not sure if we should fix them here.

cbaines:
@mgrassotti The build is still failing, in case you can't see the logs, its a linting issue:
```
lib/friendly_time_diff.rb:32:7: C: Style/NumericPredicate: Use amount.positive? instead of amount > 0.  (https://github.com/bbatsov/ruby-style-guide#predicate-methods)
      amount > 0 ? [amount, label.pluralize(amount)].join(" ") : nil
```

mgrassotti:
@cbaines I think it's ready for review

config/cable.yml
```diff
@@ -0,0 +1,9 @@
+development:
+  adapter: async
+
+test:
+  adapter: async
+
+production:
+  adapter: redis
+  url: redis://localhost:6379/1
```
cbaines: Is this config file needed? I'm unsure what it is. but it sounds web sockets related, which I'm not aware of smart-answers using.

lib/friendly_time_diff.rb
```diff
@@ -29,7 +29,7 @@ def date_parts
     end
 
     def date_part(label, amount)
-      amount > 0 ? pluralize(amount, label) : nil
+      amount > 0 ? [amount, label.pluralize(amount)].join(" ") : nil
```
cbaines: I think something like `"#{amount} #{label.pluralize(amount)}"` would be easier to read here. What do you think?

mgrassotti: @cbaines I agree

config/cable.yml
mgrassotti: @cbaines ok yes it's one new Rails 5 feature. I can remove it.

mgrassotti:
the rails dependency of the gem `govuk_frontend_toolkit` brings the `actioncable` gem which leads to the vulnerability report https://snyk.io/test/github/alphagov/smart-answers/b12dea733f9ec3633f23f59909a90039a08186c2?fromStatus=true.

pmanrubia:
@mgrassotti we can ignore the vulnerability issue with `Action Cable` for the time being.

mgrassotti:
@cbaines @pmanrubia I think it's ready for review again

cbaines:
> @cbaines @pmanrubia I think it's ready for review again

Unfortunately, I think this needs rebasing due to changes in the ` test/functional/smart_answers_controller_test.rb` file.

cbaines:
I've just rebased this to remove a conflict with some code removal that occurred master.

cbaines:
I've just tested this locally again, and everything that I checked looks fine :) I've looked through all the commits again, and the only one I have questions about is the final one a49192b31d3e648bf6e31b870c4db16cfe8e5b7f . If this is still needed, can you add the motivation to the commit message @mgrassotti ?

mgrassotti:
@cbaines sure, this is needed to avoid a vulnerability issue raised by Synk

mgrassotti:
@cbaines done 👍

PR #3056, opened 2017-06-02
Don't run open if it doesn't exist

cbaines:
Just echo the URL instead.

### DIFF
--- a/startup_heroku.sh
+++ b/startup_heroku.sh
@@ -32,13 +32,21 @@ echo
 export CURRENT_BRANCH_NAME=`git branch | grep "^\*" | cut -d" " -f2`
 git push $HEROKU_REMOTE $CURRENT_BRANCH_NAME:master
 
-echo
-echo "# Opening Smart Answers"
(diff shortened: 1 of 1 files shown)


startup_heroku.sh
```diff
@@ -38,7 +38,11 @@ echo "*NOTE.* You may have to refresh as the app can be slow to start"
 echo
 export HEROKU_URL=`heroku apps:info --app $APP_NAME | grep "Web URL" | cut -c16-`
 export SMART_ANSWER_TO_OPEN="marriage-abroad"
-open $HEROKU_URL$SMART_ANSWER_TO_OPEN
+if type open 2>/dev/null; then
+    open $HEROKU_URL$SMART_ANSWER_TO_OPEN
+else
+    echo $HEROKU_URL$SMART_ANSWER_TO_OPEN
```
chao-xian: Can we also have a message saying we didn't open the page and why? As the previous echo statement says that we're doing it. If nothing happens it might seem that something is broken.

cbaines: I've changed it now such that the message reflects what is going to happen.

PR #3058, opened 2017-06-02
Handle imminence errors

cbaines:
imminence-failure


cbaines:
As discussed in person @ikennaokpala I've now made some changes to the test suite.

PR #3062, opened 2017-06-06
Clean addresses

cbaines:
Clean addresses, removing unnecessary trailing whitespace. 
Remove any whitespace prior to a newline. This catches trailing
spaces, as well as carriage returns.

The carriage returns cause issues when flattening countries to use one
outcome per path, as when converted to an erb template, the carriage
returns do not show up in the output, causing a discrepancy between
the new and old test artefacts.


cbaines:
> Squashing the two commits would be ideal IMHO

I've squashed, and merged :)

PR #2976, opened 2017-03-24
[Do not merge] Remove hardcoded sidebar overrides

alecgibson:
Now that we have the ability to [override taxonomy sidebar content](https://trello.com/c/vM6Kaw9B/546-ias-and-content-designers-can-curate-related-links-for-any-content-type), we no longer need this hardcoded list of overrides.

### Trello 

https://trello.com/c/CikV9fMW/544-remove-hack-in-frontend-and-smart-answers-about-showing-old-related-links-for-some-mainstream-content

PR #3067, opened 2017-06-06
[DO NOT MERGE]Display curated links in sidebar

Rosa-Fox:
[Trello](https://trello.com/c/J8ofEnrt/167-display-curated-related-links-in-the-sidebar)

- No longer override sidebar for mainstream content
- Overriding will be allowed by the component (within `govuk_navigation_helpers`)

smartanswers-sidebar-example


cbaines:
This sounds fine, but I have no idea what will change for users. Is there a page I can look at to see what this change will look like?

Rosa-Fox:
Hi Chris,

I have been waiting for my data to replicate so that I can view it locally. Was going to screen grab when done, though have just discovered Alec opened a [PR](https://github.com/alphagov/smart-answers/pull/2976/files) in March to do this but it looks slightly better... so will probably close this one.

cbaines:
> I have been waiting for my data to replicate so that I can view it locally. 

Just on this point, you might be able to get away with running it without local data, if you setup the environment as described in https://github.com/alphagov/smart-answers/blob/master/doc/developing-without-vm.md . Ignore the bit about the VM, as I think it is not relevant, it should work in or outside the VM.

cbaines:
The screenshot and diff looks good. You might want to compare what Alec did in https://github.com/alphagov/smart-answers/pull/2976 .

app/controllers/concerns/education_navigation_ab_testable.rb
```diff
@@ -4,10 +4,7 @@ def should_present_new_navigation_view?(content_item)
   end
 
   def present_taxonomy_sidebar?(content_item)
```
alecgibson: Given that this method is now essentially an alias, it would be neater to remove the entire method and change any references to it to point to `should_present_new_navigation_view?` instead.

Rosa-Fox: Thanks @alecgibson have updated.

carvil:
Closing as this has been removed from smart-answers recently: https://github.com/alphagov/smart-answers/pull/3134

PR #3075, opened 2017-06-12
Update text on holiday entitlement landing page

deborahchua:
The additional text helps the user to understand what this service provides before deciding to start the process


sihugh:
We're spinning this up on integration so that it can be 👍  by a content designer.

sihugh:
Approved!

PR #3076, opened 2017-06-13
Remove the router_api service

cbaines:
It is unused.


cbaines:
> LGTM, but tests are not passing

I ran them again, and they passed... thanks for your review :D

PR #3080, opened 2017-06-14
Create an outcome per path for Luxembourg

cbaines:
Create an outcome per path for the Luxembourg content in the Marriage
Abroad smart answer.

This involved:
 - copying the test artefacts
 - renaming them
 - stripping the titles
 - creating the title partial
 - and adding luxembourg to the list of countries with an outcome per
   path.
 - updating the checksums
 - updating the expected responses


cbaines:
> Integration tests to prove luxembourg flow don't exist.

I think I get what you mean. To clarify, are you saying that there are some tests that you'd like to see added to the marriage_abroad_test.rb file, that step through each path to assert the outcome?

> Line 27 of _synonyms_of_cp.govspeak.erb has a Luxembourg content

I've removed this now.

ikennaokpala:
@cbaines yes exactly each flattening should ideally come with integration tests. See previous flattening.

cbaines:
> @cbaines yes exactly each flattening should ideally come with integration tests. See previous flattening.

I've added these now.

PR #3081, opened 2017-06-15
Update child maintenance landing text

sihugh:
User research has shown that some people find the structure of the current page
a little confusing.  We've kept most of the content the same, just reordered it
slightly.

We've also tweaked the disclaimer text which is used on a number of pages.  I split
that into a separate commit given that it's a bit more widespread. 

The default landing page text now matches that in the B variant of the test that is currently
running.

The last commit is removing the last test where we AB tested titles.

https://trello.com/c/ko4cXOYP/144-cm-button-position-develop-and-deploy-ab-test


app/views/smart_answers/landing.html.erb
```diff
@@ -28,11 +24,7 @@
       <% end %>
 
         <p class="get-started">
-          <% if should_show_benchmarking_variant? %>
-            <a rel="nofollow" href="<%= start_button_href %>" class="big button"><%= ab_start_button %></a>
```
issyl0: `ab_start_button` needs removing too.

PR #2977, opened 2017-03-24
Flatten Greece and Add Support for three questions flow

ikennaokpala:
[Trello card](https://trello.com/c/WysStRHr/580-5-marriage-abroad-improvement-greece-6-outcomes)

## Description 

The goal of this PR is to migrate the flow for Greece to use the outcome per path strategy and also add support for three questions flow.

As you will  observe (in the first commit ) the title only contains opposite sex title, this
because the same sex title doesn't use the content title block.  It uses the markdown syntax, this will be fix in a7cfd83.

## Factcheck

[Preview link](https://smart-answers-pr-2977.herokuapp.com/marriage-abroad/y/greece/uk)
[GOVUK](https://www.gov.uk/marriage-abroad/y/greece/uk/)

## Expected changes

- Flatten content for Greece
- Enabling outcome per path strategy
- Removed partner nationality question for Greece
- Enabling  three questions flow

### Before 

screen shot 2017-03-27 at 16 52 00

### After

screen shot 2017-03-27 at 16 53 53

PR #3083, opened 2017-06-15
Marriage abroad: Add support for three question flow

ikennaokpala:
[Trello card ](https://trello.com/c/g8pBgHkG/651-add-support-for-three-question-flow)

## Motivation 

The Greece #2977 PR contains implementation (i.e 2cbd8e6bb952325859a6696d18110a7194a34ffa) that is required for converting Poland flow to use three question flow. 

This PR seeks to separate this from #2977 and make it available for Poland et al.

This has become necessary to make concrete the support for three questions country flow ASAP.

Also this presents the opportunity to further refactor the way in which marriage data is retrieved.

This PR #3085  shows an example of how the implementation here can be applied. 

### Expected changes
- Retrieve 18,  2 and 6 outcome country list from YAML file 
- Add support for three question country flow


lib/smart_answer/calculators/marriage_abroad_data_query.rb
```diff
@@ -152,13 +152,22 @@ def ss_unknown_no_embassies?(country_slug)
     def os_consular_cni_in_nearby_country?(country_slug)
       OS_CONSULAR_CNI_IN_NEARBY_COUNTRY.include?(country_slug)
     end
-    
+
+    def countries_with_18_outcomes
+      countries = marriage_data.fetch(:countries_with_18_outcomes)
+      valid_outcomes_country_data_structure?(countries) ? countries : []
```
cbaines: I think raising an exception might be better here if the data is invalid? Otherwise if something goes wrong, a valid, albeit probably wrong value of `[]` will be returned.

lib/smart_answer/calculators/marriage_abroad_data_query.rb
```diff
@@ -152,13 +152,22 @@ def ss_unknown_no_embassies?(country_slug)
...
+    end
+
     def marriage_data
       @marriage_data ||= YAML.load_file(path_to_data_file).with_indifferent_access
     end
 
   private
 
+    def valid_outcomes_country_data_structure?(countries)
+      countries.is_a?(Array) && countries.any? { |country| country.is_a?(String) }
```
cbaines: I think this wants to be `countries.all?` rather than `countries.any?`, as I think you want to say, "its an array, and all elements are strings".

lib/smart_answer/calculators/marriage_abroad_data_query.rb
ikennaokpala: yeah I have made the changes that we discussed yesterday.

lib/smart_answer/calculators/marriage_abroad_data_query.rb
ikennaokpala: Sure.. the ```all?``` reads nicer as well

cbaines: Have you pushed this change, as I'm still seeing `any?`?

ikennaokpala: @cbaines Somehow this was omitted.. I have amended it.

cbaines: Great, I can see it now.

PR #3089, opened 2017-06-16
Add annual leave entitlement AB test

sihugh:
Our user research suggests that giving the context of when a leave year starts
closer to the place where a user needs that information could help completion.

The text we are testing is this:

 -----
 Leave year dates
  The leave year will usually be in an employee’s contract. If it isn’t, their leave year will start:

  - on their first day (if the job began after 1 October 1998)
  - on 1 October (if the job began on or before 1 October 1998)
-----

The A variant keeps this text on the [landing page](https://www.gov.uk/calculate-your-holiday-entitlement) where it is currently.

The B variant does not show this text on the landing page, but instead on:

- https://www.gov.uk/calculate-your-holiday-entitlement/y/days-worked-per-week
- https://www.gov.uk/calculate-your-holiday-entitlement/y/shift-worker
- https://www.gov.uk/calculate-your-holiday-entitlement/y/hours-worked-per-week

To minimise duplication and to make it more obvious where the text is displayed,
I've put this markup directly in the question view rather than using Govspeak
in the node body (which is the approach we've typically used for the landing
pages)

This test will run until at least 26th June 2017.

https://trello.com/c/GgnYfigi/161-annual-leave-description-develop-and-deploy-change-to-form-flow

##Screenshots of the A and B variants of the landing page
screen shot 2017-06-16 at 14 30 32
screen shot 2017-06-16 at 14 30 12

##Screenshots of A and B variants of a question page
screen shot 2017-06-16 at 14 30 46
screen shot 2017-06-16 at 14 31 01

### DIFF
--- a/app/controllers/concerns/benchmark_holiday_entitlement_ab_testable.rb
+++ b/app/controllers/concerns/benchmark_holiday_entitlement_ab_testable.rb
@@ -1,0 +1,36 @@
+module BenchmarkHolidayEntitlementABTestable
+  BENCHMARKING_PATHS = [
+    '/calculate-your-holiday-entitlement',
+    '/calculate-your-holiday-entitlement/y/days-worked-per-week',
+    '/calculate-your-holiday-entitlement/y/hours-worked-per-week',
+    '/calculate-your-holiday-entitlement/y/shift-worker'
+  ].freeze
+
+  def should_show_holiday_entitlement_variant?
+    # Use GOVUK-ABTest-BenchmarkALDescription1=B header in dev to test this
+    benchmark_holiday_entitlement_variant.variant_b? &&
+      is_holiday_entitlement_tested_path?
+  end
+
+  def is_holiday_entitlement_tested_path?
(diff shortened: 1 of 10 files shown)


app/controllers/concerns/benchmark_holiday_entitlement_ab_testable.rb
```diff
@@ -0,0 +1,36 @@
+module BenchmarkHolidayEntitlementABTestable
+  BENCHMARKING_PATHS = [
+    '/calculate-your-holiday-entitlement',
+    '/calculate-your-holiday-entitlement/y/days-worked-per-week',
+    '/calculate-your-holiday-entitlement/y/hours-worked-per-week',
+    '/calculate-your-holiday-entitlement/y/shift-worker'
+  ].freeze
+
+  def should_show_holiday_entitlement_variant?
+    # Use GOVUK-ABTest-BenchmarkCmButton1=B header in dev to test this
```
binaryberry: Could you update the cookie name in the comment?

PR #3093, opened 2017-06-19
Update marriage-abroad-questions-and-responses.yml [SEAL IGNORE]

cbaines:
By running:

  rails r script/generate-questions-and-responses-for-smart-answer.rb marriage-abroad


ikennaokpala:
The `test/data/marriage-abroad-questions-and-responses.yml` file is used to test example flows in the case of marriage abroad the example are run per country. A majority of the countries (i.e example flows) add here have already been covered and are redundant in this instance. Also adding this many countries will inturn increase the time taken to run regresion tests. This PR should be closed for the aofrementioned reasons.

PR #3085, opened 2017-06-15
Marriage abroad: Support 6 outcomes/3 question flow for Poland

ikennaokpala:
This PR needs to be merged after #3083 

[Trello card](https://trello.com/c/JvyZgtto/625-5-marriage-abroad-improvement-poland-6-outcomes)

## Description

The goal of this PR is to convert the flow for Poland to use the outcome per path strategy and also add support for three questions flow.

## Factcheck

[Preview link](https://smart-answers-pr-3085.herokuapp.com/marriage-abroad/y/poland/uk)
[GOVUK](https://www.gov.uk/marriage-abroad/y/poland/uk/)

## Expected changes
- Enabling  three questions flow for Poland
- Number of outcomes have been reduces to 6

### Before

screen shot 2017-06-15 at 17 24 01

### After
screen shot 2017-06-15 at 17 23 49

PR #3100, opened 2017-06-22
Improve Poland content

sistephens:
I've added content written as part of the ongoing marriage abroad improvement project.

We don't need a Heroku preview. The dept has already fact-checked so this can be merged as soon as poss 😃


ikennaokpala:
@sistephens thanks for the factcheck update. I have moved your changes to #3085.

PR #3109, opened 2017-06-26
[DO NOT MERGE] Publish "/part-year-profit-tax-credits" start page as a transaction

binaryberry:
This modifies the code that publishes smart-answers on deploy so it will publish the start page (aka "landing page") of "/part-year-profit-tax-credits" as a transaction that has the same content-id as the current "/part-year-profit-tax-credits" content item, and an exact route. Its "Start now" button will still point to the "/part-year-profit-tax-credits/y" path.

The pages of "/part-year-profit-tax-credits" other than the start page get republished as a StartPageLessContentItemFlow that have a new content-id and prefix routes. Their base path is "/part-year-profit-tax-credits/y".

Considerations:
- This is a proof of concept, all smart answers are meant to eventually get their start pages converted into transactions in the Publisher app.
- The code related to the publishing of the start page is meant to be ephemeral, and will be deleted once the transaction is managed in Publisher. 
- Marking this as [DO NOT MERGE] until other PRs are ready.
- I have not touched the StartNodePresenter, but once all the smart answer start pages have been converted to transactions, we should consider whether it is still useful.

cc @fofr and @pmanrubia if you/someone on your team gets a chance to have a look

This is part of the work to convert all start pages across GOV.UK publishers to transactions in the Publisher app, so they can be consolidated.

[Trello](https://trello.com/c/arlZczQO/423-2-republish-smart-answer-part-year-profits-to-finalise-your-tax-credits-start-page-as-a-transaction-page)

### DIFF
--- a/app/services/content_item_publisher.rb
+++ b/app/services/content_item_publisher.rb
@@ -1,6 +1,11 @@
 class ContentItemPublisher
   def publish(flow_presenters)
     flow_presenters.each do |smart_answer|
+      if smart_answer.slug == "part-year-profit-tax-credits"
+        content_item = StartPageLessFlowContentItem.new(smart_answer)
+      else
+        content_item = FlowContentItem.new(smart_answer)
+      end
       Services.publishing_api.put_content(content_item.content_id, content_item.payload)
       Services.publishing_api.publish(content_item.content_id, 'minor')
     end
--- a/app/services/content_item_publisher.rb
+++ b/app/services/content_item_publisher.rb
@@ -215,6 +215,7 @@ private
       schema_name: :transaction
     }
 
+    reserve_path_for_publishing_app(base_path, publishing_app)
     create_and_publish_via_publishing_api(payload, content_id)
   end
 
--- a/app/services/content_item_publisher.rb
+++ b/app/services/content_item_publisher.rb
@@ -1,7 +1,6 @@
 class ContentItemPublisher
   def publish(flow_presenters)
     flow_presenters.each do |smart_answer|
-      content_item = FlowContentItem.new(smart_answer)
       Services.publishing_api.put_content(content_item.content_id, content_item.payload)
       Services.publishing_api.publish(content_item.content_id, 'minor')
     end
--- a/app/services/content_item_publisher.rb
+++ b/app/services/content_item_publisher.rb
@@ -1,21 +1,33 @@
 class ContentItemPublisher
   def publish(flow_presenters)
-    start_page_less_smart_answer = flow_presenters.select { |smart_answer| smart_answer.slug == "part-year-profit-tax-credits/y" }.first
-    start_page_smart_answer = flow_presenters.select { |smart_answer| smart_answer.slug == "part-year-profit-tax-credits" }.first
+    find_smart_answers_without_start_pages(flow_presenters)
     flow_presenters.each do |smart_answer|
-      content_item = case smart_answer
-                     when start_page_less_smart_answer
-                       StartPageLessFlowContentItem.new(smart_answer)
-                     when start_page_smart_answer
-                       nil
-                     else
-                       FlowContentItem.new(smart_answer)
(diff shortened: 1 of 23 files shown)


binaryberry:
Hi @chrislo @floehopper @chrisroos  if you have any thoughts on this, we could use your feedback.

app/services/content_item_publisher.rb
```diff
@@ -1,10 +1,48 @@
...
+  end
+
+  def find_smart_answers_without_start_pages(flow_presenters)
+    @start_page_less_smart_answer = flow_presenters.select { |smart_answer| smart_answer.slug == "part-year-profit-tax-credits/y" }.first
+    @start_page_smart_answer = flow_presenters.select { |smart_answer| smart_answer.slug == "part-year-profit-tax-credits" }.first
+  end
+
+  def assign_content_item(smart_answer)
+    case smart_answer
+    when @start_page_less_smart_answer
```
fofr: I wonder if `@smart_answer` and `@smart_answer_without_start_page` would be easier to read.

binaryberry: Happy to rename @start_page_less_smart_answer to @smart_answer_without_start_page if it's more readable - will need to also rename a few classes and variables.

Don't want to turn smart_answer into an instance variable however as I'd rather keep its scope to just that method.

binaryberry: Renamed in cb12897.

lib/tasks/retire.rake
```diff
@@ -64,6 +64,29 @@ namespace :retire do
     )
   end
 
+  desc "Publish transaction start page via publishing api"
```
ikennaokpala: This needs to be moved to `lib/tasks/publishing_api.rake` retire namespace is reserved for rake task that are part or concern retiring smart answers. 

Also we require tests to be add to rake tasks

app/services/content_item_publisher.rb
```diff
@@ -1,10 +1,48 @@
...
+      content_item = assign_content_item(smart_answer)
+      if content_item
+        Services.publishing_api.put_content(content_item.content_id, content_item.payload)
+        Services.publishing_api.publish(content_item.content_id, 'minor')
+      end
     end
+    republish_start_page
+  end
+
+  def find_smart_answers_without_start_pages(flow_presenters)
```
ikennaokpala: By defining a predicate an extract attr accessor field and may be a predicate method in the `lib/smart_answer/flow.rb` that will identify a smart answer with a start page 

So you flow will look something like:

```ruby
module SmartAnswer
  class PartYearProfitTaxCreditsStartPageFlow < Flow
    def define
      start_page = true
      name 'part-year-profit-tax-credits'

      status :published
      satisfies_need "103438"
      content_id "de6723a5-7256-4bfd-aad3-82b04b06b73e"

```

````ruby

module SmartAnswer
  class Flow
    attr_reader :nodes
    attr_accessor :status, :need_id :start_page

    def start_page?
       start_page
    end 
```

app/services/content_item_publisher.rb
```diff
@@ -1,10 +1,48 @@
...
+    @smart_answer_without_start_page = flow_presenters.select { |smart_answer| smart_answer.slug == "part-year-profit-tax-credits/y" }.first
+    @start_page_smart_answer = flow_presenters.select { |smart_answer| smart_answer.slug == "part-year-profit-tax-credits" }.first
+  end
+
+  def assign_content_item(smart_answer)
+    case smart_answer
```
ikennaokpala: I am not sure why this exists but from the other example that I have given, this can be made redundant. 

```ruby

if smart_answer.start_page? 
...
else
...
end
```

chrisroos:
Sorry for the late reply, @binaryberry. I'll take a look at this now.

binaryberry:
Closing in favour of #3122

PR #3110, opened 2017-06-27
Marriage abroad: Derive outcome path countries

ikennaokpala:
[Trello card](https://trello.com/c/JRHLjCS4/663-derive-outcome-path-country-list)

## Motivation 

This PR seeks to avoid the unnecessary duplication introduced by the 18 outcome country list.

In this group, countries belonging to other country list exist.

This PR seeks to derive the total list of countries that have been flattened. 

Also correct a few omission of tests 

These changes have not effect on regression test artefacts or any other smart answers.  See [this for more details](https://ci.integration.publishing.service.gov.uk/job/smartanswers/job/marriage-abroad%252Frefactor-has-outcome-path/13/)

## Expected changes

- Each outcome per path countries contains only countries that belong to that list and not in other
- Refactor has outcome to use a derived list from:
      - countries_with_18_outcomes list
      - countries_with_6_outcomes list
      - countries_with_2_outcomes list
- Update tests to include omitted cases


cbaines:
I think you have put "to" rather than "two" in the second commit, as in "This commit changes two tests".

PR #3112, opened 2017-06-28
Change Child Maintenance results page text for basic rates

issyl0:
- Following on from Benchmarking's user research, this was shown as the
  best wording, so we're changing it permanently.

https://trello.com/c/WfiHIC43/194-change-to-child-maintenance-results-page


sihugh:
I've got a feeling that this will fail the regression tests as we should have a bunch of test artefacts reflecting the changes on each of the affected results pages.  If you run `RUN_REGRESSION_TESTS=calculate-your-child-maintenance ruby test/regression/smart_answers_
regression_test.rb` then commit the extra files then it should be OK.

The rest looks 👍

PR #3122, opened 2017-07-03
Publish "/part-year-profit-tax-credits" start page as a transaction

binaryberry:
This publishes the start page of "/part-year-profit-tax-credits" as a mainstream publisher transaction to the publishing-api with a start_button pointing to "/part-year-profit-tax-credits/y", and sets up the ground work for doing this for the start pages of all the other smart-answers.

Here we set up a flag that identifies the smart answers whose start pages will be published as a transaction. Those start pages will retain the existing content-id of the smart answer when they are published as transactions, and we will need to create a new content-id (the "flow_content_id") for the rest of the smart answer (any page under the "smart-answer-slug/y" path) to be republished.

We use the flow registration presenter to expose the contents of the start page (title, description, body) to the rake task that will publish the transaction. The rake task publishing_api:publish will run on deploy and carry out the republishings.

This has been tested on integration and works as expected - the start page content-item has the same content-id as before, but its document_type is a transaction and publishing-app a transaction. The content and formatting are also the same. The paths under "part-year-profit-tax-credits/y" have the new flow_content_id and their content-item is has a smartanswer document_type and smartanswers as its publishing-app. 

cc @ikennaokpala @chrisroos 

https://trello.com/c/arlZczQO/423-2-republish-smart-answer-part-year-profits-to-finalise-your-tax-credits-start-page-as-a-transaction-page

And: https://trello.com/c/yhkabR7v/673-5-convert-smart-answers-start-pages-into-transactions-in-publisher

### DIFF
--- a/app/presenters/flow_content_item.rb
+++ b/app/presenters/flow_content_item.rb
@@ -26,6 +26,10 @@ class FlowContentItem
     flow_presenter.content_id
   end
 
+  def flow_content_id
+    flow_presenter.flow_content_id
+  end
+
 private
 
   def routes
@@ -36,7 +40,11 @@ private
   end
 
   def base_path
-    '/' + flow_presenter.slug
+    if flow_presenter.transaction_start_page?
+      "/#{flow_presenter.slug}/y"
+    else
+      "/#{flow_presenter.slug}"
+    end
   end
 
   def json_path
--- a/app/presenters/flow_registration_presenter.rb
+++ b/app/presenters/flow_registration_presenter.rb
@@ -15,12 +15,24 @@ class FlowRegistrationPresenter
     @flow.content_id
   end
 
+  def flow_content_id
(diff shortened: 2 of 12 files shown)


pmanrubia:
@binaryberry @ikennaokpala thank you for your work on this PR. 
What are your thoughts of #3126. It feels to me like a safer approach? Would you be happy with merging it?

binaryberry:
@pmanrubia - yes indeed #3126 does sound better, we would prefer this approach. Still a bit more work to be done on that one to preserve the custom start button text for calculate-your-child-maintenance, but Chris Roos is on it. Happy for this PR to be closed unless you have objections?

ikennaokpala:
This PR was another iteration from #3109 and now based on the learnings we have #3126 so it is quite logical to go with it and close this PR

PR #3124, opened 2017-07-05
Fix register a death for Scotland and Northern Ireland

mgrassotti:
The "register a death" start page tells users in Scotland and Northern Ireland how to register the death, but the tool itself then ask them unhelpful questions. 

So we are going to: 
- remove the initial links for Scotland and Northern Ireland
- remove the question flow for Scotland and Northern Ireland
- integrate the final results in the process flow. They will be results of the 1st question: "Where did the death happen?"

Trello: https://trello.com/c/rMuC1ego/670-2-register-a-death-fix-scotland-and-northern-ireland-outcomes

## Before
screen shot 2017-07-11 at 10 42 07

## After
screen shot 2017-07-11 at 10 42 13

## Factcheck
https://smart-answers-pr-3124.herokuapp.com/register-a-death


mgrassotti:
👍 thanks @ikennaokpala. I tried to make commits more self explaining. I also run `bundle exec rails r script/generate-responses-and-expected-results-for-smart-answer.rb register-a-death`, which I previously omitted. Let me know if it could be ok now.

ikennaokpala:
@mgrassotti I see you have generated the `test/data/register-a-death-responses-and-expected-results.yml`. This file is placed in the wrong commit.. it needs to go in 496c86c17e2cd9998e65ea14981d9eadc0162afa where the actual change was trigger.

- I noticed that some commit messages are running over the 75 character limit. 
- It is ideally easier to follow each commit with checksums within. This way it is clear which files have changed per commit.

mgrassotti:
@ikennaokpala 👍 I tried to update following your comments

ikennaokpala:
@mgrassotti This is looking better. If you move `test/integration/smart_answer_flows/register_a_death_test.rb` in af7a7a4829154ed5a8861e32abcec6064a43b55a to 039f2f14fc8fa16af18c547b7453462a4115a539

mgrassotti:
👍 it should be done

PR #3127, opened 2017-07-06
Marriage abroad Indonesia - one outcome per path

mgrassotti:
In order to simplify the maintenance of marriage abroad smart answers, we are going to create one outcome per path for Indonesia.

## Fact check
https://smart-answers-pr-3127.herokuapp.com/


lib/data/marriage_abroad_data.yml
```diff
@@ -21,6 +21,7 @@ countries_with_18_outcomes:
   - tanzania
   - tunisia
   - vietnam
+  - indonesia
```
ikennaokpala: This line doesn't follow the alphabetical order.. kindly place this above `iran`

mgrassotti: 👍 you are right. I also added some integration tests.

mgrassotti:
👍 about the latter changes disappeared rebasing onto master, so I think we don't need them here

PR #3131, opened 2017-07-10
[WIP] Add script for flattening

cbaines:
This is still a work in progress, and I'm not sure how much time I'll have to work on this now that I'll have another team after my time.

### DIFF
--- a/script/flatten.sh
+++ b/script/flatten.sh
@@ -18,6 +18,10 @@ END
 
 FEE_TABLE_PARTIAL_WITH_LITERAL_NEWLINES="${FEE_TABLE_PARTIAL//$'\n'/\\n}\\n"
 
+read -r -d '' HOW_TO_PAY_PARTIAL << END
+<%= render partial: 'how_to_pay.govspeak.erb', locals: {calculator: calculator} %>\n
+END
+
 for file in $(find $COUNTRIES/$COUNTRY/*/ -type f)
 do
     OLD_NAME="$file"
@@ -30,6 +34,11 @@ do
 
     echo "Stripping top two lines"
     sed -i -e 1,2d "$NEW_NAME"
+
+    echo "Replacing how to pay with partial in ${file/$COUNTRIES\//}"
+    perl -0777 -i -pe \
+         "s/^\^?You can( only)? pay by.*?\n.*?^$/${HOW_TO_PAY_PARTIAL}/migs" \
+         "$NEW_NAME"
 
     echo "Replacing Fees table with fees partial in ${file/$COUNTRIES\//}"
     perl -0777 -i -pe \
--- a/script/flatten.sh
+++ b/script/flatten.sh
@@ -1,0 +1,38 @@
+#!/bin/sh
+
+COUNTRY="$1"
+
+COUNTRIES="lib/smart_answer_flows/marriage-abroad/outcomes/countries"
+COUNTRY_PARTIALS="$COUNTRIES/$COUNTRY"
+
+rm -rf "$COUNTRY_PARTIALS"
+
+cp -r "test/artefacts/marriage-abroad/$COUNTRY" "$COUNTRY_PARTIALS"
+
+read -r -d '' FEE_TABLE_PARTIAL << END
+<%= render partial: 'consular_fees_table_items.govspeak.erb',
+    collection: calculator.services,
+    as: :service,
+    locals: { calculator: calculator } %>
+END
+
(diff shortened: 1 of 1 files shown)


pmanrubia:
@ikennaokpala have you tried it locally with a couple of countries? 
@chao-xian Why did you change the title to [WIP]? The PR is approved, but still the title does not match the status.

ikennaokpala:
I don't this is a script that is meant to be perfect.. given the anomalies that may come up..I see it as a script that jet start the flattening process.. Hence why it serve as a starting point. The developer that uses this script should then review the changes and make sure that it is in order before checking it in.

pmanrubia:
@ikennaokpala I don't think so. The script is supposed to flatten countries following a set of steps. It should generally work, and the developer has to review the output in any case.
I don't think we should approve this PR without running it locally for a couple of countries.

ikennaokpala:
That is my humble opinion.. There is not point trying to perfect it knowing that Marriage abroad templates may present a few anomalies

chao-xian:
@pmanrubia I added the `[WIP]` to the title because that's what @cbaines said the status is in the PR, which I did before the PR was approved.

pmanrubia:
@chao-xian could you please run the script locally and see if it is fit for purpose? if not, could you please find out if there are any steps we could take?

pmanrubia:
@ikennaokpala there is no point in approving a PR without even running it locally. This is not how we work.

ikennaokpala:
FYI.. I have run it locally and found it sufficient

pmanrubia:
@chao-xian thank you for having a look at it. 

Could you please add a comment to this PR with the countries you have run the script against? and the anomalies that you have found, if any?
If the script is running smoothly, we would only need to update the documentation before merging.

chao-xian:
@pmanrubia this seems fine. I can't see any obvious problems with it. All it does is create the outcome per path files that would normally be quite a laborious manual process - assuming you are doing an 18 outcome flattening. 

And it's definitely not complete because you are left with files that need cleaning up, so for a `_opposite_sex.erb` file it has alongside it a `_opposite_sex.erb-e` that needs removing.

I ran this against:

* Malaysia
* Singapore
* Finland

cbaines:
> And it's definitely not complete because you are left with files that need cleaning up, so for a _opposite_sex.erb file it has alongside it a _opposite_sex.erb-e that needs removing.

Interesting, I don't think I encountered this. I just ran the flatten script locally for the 3 countries you mentioned, and I didn't get any odd files generated. Perhaps the perl or sed you are using works differently to the one I'm using?

chao-xian:
@cbaines I had been running it on OS X.

Another issue is that it doesn't work on the VM at the moment 

```
vagrant@development:/var/govuk/smart-answers$ script/flatten.sh malaysia
script/flatten.sh: 12: read: Illegal option -d
script/flatten.sh: 19: script/flatten.sh: Bad substitution
```

But weirdly if I run that read command with the same options on the command line it doesn't complain.

cbaines:
> But weirdly if I run that read command with the same options on the command line it doesn't complain.

I'm running it with bash, but I guess that in the govuk-puppet development VM, you are using dash. Try running it explicitly with bash perhaps? (e.g. `bash ./scripts/flatten.sh ...`).

chao-xian:
@cbaines yeah that worked and without the `-e` erb templates.

cbaines:
> @cbaines yeah that worked and without the -e erb templates.

Great :+1:

chao-xian:
@mgrassotti have you been using this script at all?

mgrassotti:
@chao-xian yes I am

chao-xian:
@mgrassotti both on your Mac or just in the VM? Any changes that you've made? If you have been using it in anger and it's fine, then I think we need to update this PR and look to merge it in (unless that's already been done somewhere else).

What do you reckon?

chao-xian:
Hi @mgrassotti we really should merge this into master if we are actively using it. Can you please add your review to this PR. I'd be happy to update our docs if you can let me know that the script is ok.

chao-xian:
@mgrassotti hi I'm going to close this PR as we are slowing down on the number of requests we get for flattening Marriage Abroad. The only other option is to merge into `master`. I'll close it by the end of tomorrow. Cc @pmanrubia

cbaines:
Matteo is no longer active, and I'm currently trying to cut down on the number of Pull Requests I have open. I'm going to close this for now, but I'll leave the branch around. I think this script is still a real time saver when it comes to flattening.

Issue #3136, opened 2017-07-11
The instructions for getting started don't work anymore?


vickytnz:
I can use the other examples of the repo, but when I start creating one from the instructions
https://github.com/alphagov/smart-answers/blob/master/doc/creating-a-new-smart-answer.md
but as soon as I get to the 

> "Assuming you're still running rails server, visit http://localhost:3000/example-smart-answer and you should see the empty landing page of our new Smart Answer."

I just get errors
```
NoMethodError at /example-smart-answer
undefined method `dig' for nil:NilClass
```
![image](https://user-images.githubusercontent.com/1223264/28090246-9851d89c-6683-11e7-9bc1-d70df21f8e3a.png)

![image](https://user-images.githubusercontent.com/1223264/28090335-dcfd3752-6683-11e7-9612-31c269970f4e.png)

has something changed?

chao-xian:
Hi @vickytnz thanks for flagging this. There is an assumption that the SmartAnswer being created also exists in the GOV.UK Content Store, but the example used to cope with that not being the case. It appears that the navigation, which is undergoing AB testing, now always assumes that the Smart Answer is in the Content Store. 

I'm not certain that we will ever likely fix this as Smart Answers is treated as a BAU app and we are not creating new "smart answers" to add in. We are more active in retiring or moving them from out of the app. But I'll check (Cc @pmanrubia)

Also out of curiosity what are your plans with Smart Answers? It may not be the most useful for your purposes. If you want to give me a shout I'm also on the x-gov slack.

ikennaokpala:
@vickytnz Thanks for reporting this issue.. This issue has now been resolved.. Kindly pull latest changes on master.

PR #3144, opened 2017-07-17
Retrieve ContentItem via ContentItemRetriever Service

ikennaokpala:
[Trello card](https://trello.com/c/SxnTo8i8/677-5-instructions-for-getting-started-in-the-documentation-does-not-work)

## Motivation
Sequel to the issue raised in [1], it appears that recent AB testing efforts have introduced code that makes it hard to create a new smart answer. Existing smart answers have not been affected.

If the `smart-answer` is not in the content store, then content_item instance variable is set to nil instead of the response from the content store.

The idea here is to return a hash after the request, such that no matter how many times it is executed, even if the slug/smart answer isn't defined on the content store the same type (i.e Hash) is returned.

In theory, the aforementioned is also regarded as idempotence

This PR moves the request for the content item to a class level method called fetch defined within a ContentItemRetriever Service.

Also inline with the existing code, a without_links_organisations class level method has also been defined to cater for the removal of organisations under the links key.

Tests have also been written to cover the aforementioned areas, including controller tests that ratify the assignment of navigation_helper and content_item instance variables.

[1] https://github.com/alphagov/smart-answers/issues/3136

## Expected changes
- Refactor the retrieval of content item from content store.
- Split assignment of navigation helpers and content item.
- Add tests to ratify the aforementioned changes
- Make it possible again to create a new smart answer

### DIFF
--- a/app/controllers/smart_answers_controller.rb
+++ b/app/controllers/smart_answers_controller.rb
@@ -116,24 +116,12 @@ private
   end
 
   def setup_navigation_helpers_and_content_item
-    @content_item = Services.content_store.content_item("/" + params[:id]).to_hash
-
-    # The GOV.UK analytics component[1] automatically sets `govuk:analytics:organisations`
-    # if there's a `organisations` key in the links. This will be sent to Google
-    # Analytics At the moment we want to avoid setting this because it will flood
-    # the analytics reports with (unexpected) data. We are currently working on
-    # a solution to this conundrum[2].
-    #
-    # [1] http://govuk-component-guide.herokuapp.com/components/analytics_meta_tags
-    # [2] https://trello.com/c/DkR63grd
-    if @content_item["links"]
-      @content_item["links"].delete("organisations")
-    end
-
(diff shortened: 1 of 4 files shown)


app/controllers/smart_answers_controller.rb
```diff
@@ -116,24 +116,12 @@ def set_expiry(duration = 30.minutes)
...
-    #
-    # [1] http://govuk-component-guide.herokuapp.com/components/analytics_meta_tags
-    # [2] https://trello.com/c/DkR63grd
-    if @content_item["links"]
-      @content_item["links"].delete("organisations")
-    end
-
-    @navigation_helpers = GovukNavigationHelpers::NavigationHelper.new(@content_item)
-  rescue GdsApi::HTTPNotFound, GdsApi::HTTPGone
     @navigation_helpers = nil
```
alecgibson: (You could maybe move this `nil` assignment closer to the `@content_item.present?` check? Using either the ternary operator or the return from the `if` block?)

ikennaokpala: Yeah.. I have moved it

PR #3153, opened 2017-07-25
Send update_type with put content

emmabeynon:
We are deprecating sending the `update_type` in publish requests made to the
Publishing API, and now expect clients to send `update_type` in PUT requests
instead.  We also upgrade `gds-api-adapters` to bring this app up to date with
the new expected behaviour, as well as `govuk_navigation_helpers` to resolve
some dependency issues.

https://trello.com/c/LnJlkb9e/987-5-deprecate-the-usage-of-updatetype-in-publish-for-publishing-api-and-update-usage-across-govuk

### DIFF
--- a/app/services/content_item_publisher.rb
+++ b/app/services/content_item_publisher.rb
@@ -3,11 +3,11 @@ class ContentItemPublisher
     flow_presenters.each do |smart_answer|
       start_page_content_item = StartPageContentItem.new(smart_answer)
       Services.publishing_api.put_content(start_page_content_item.content_id, start_page_content_item.payload)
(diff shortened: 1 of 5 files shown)


app/services/content_item_publisher.rb
```diff
@@ -125,6 +125,7 @@ def publish_transaction_via_publishing_api(base_path, publishing_app:, title:, c
     payload = {
       base_path: base_path,
       title: title,
+      update_type: 'major',
```
kevindew: It looks like the single quotes here are inconsistent with the rest of this file.

emmabeynon: I've changed these to double quotes now 👍

PR #3175, opened 2017-08-02
Azerbaijan marriage abroad - all opposite sex outcomes

hannahwhittaker:
https://trello.com/c/jYdeSZ1C/868-1-sept-azerbaijan-marriage-tool-opposite-sex-marriage-change-of-process-to-be-published-on-1-september-2017-content-change-reque

I've updated the 9 opposite sex outcomes. The law in Azerbaijan is changing on 1/9/17 so users will have to declare an affirmation rather than get an CNI.

Can this be deployed as late as possible on 31/8/17?

This doesn't affect the logic.


ikennaokpala:
Superseded by #3184

PR #3185, opened 2017-08-10
Raise an error if the Worldwide API does not return any countries

alecgibson:
We recently had an incident where the Worldwide API was not returning
any countries to Smart Answers.

As such, the country select questions presented users with empty
dropdowns. This case is irrecoverable by the user, so we should fail
loudly.

This change checks if we receive any locations from the API. If not,
then we raise an exception. This will achieve two things:

- The error will be logged to Errbit
- The user will be presented with a generic 500 error page, informing
  them that something has gone wrong at our end.

### DIFF
--- a/app/models/world_location.rb
+++ b/app/models/world_location.rb
@@ -13,13 +13,18 @@ class WorldLocation
 
   def self.all
     cache_fetch("all") do
-      world_locations = Services.worldwide_api.world_locations.with_subsequent_pages.map do |response|
-        location = response.to_hash
-        if valid_world_location_format?(location)
-          self.new(location)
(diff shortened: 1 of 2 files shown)


alecgibson:
[(1) Add report error to Errbit if Whitehall WorldLocation.all returns empty](https://trello.com/c/dhlG3bgW/709-1-add-report-error-to-errbit-if-whitehall-worldlocationall-returns-empty)

alecgibson:
[(2) Display error page for empty county list](https://trello.com/c/76YLJVPD/710-2-display-error-page-for-empty-county-list)

PR #3181, opened 2017-08-07
Benefit Cap Calculator: Replacing multiple choice YES/NO with Checkbox options

ikennaokpala:
[Trello card](https://trello.com/c/V84GqmbD/702-3-benefit-cap-calculator-turning-yes-no-into-checkboxes)

## Motivation

As part of the accessibility improvements, we have been ask to replace the Yes/No multiple choice with checkbox that contain the list of exempt benefits.

The intention here, is to assist devices like screen reader etc, to better recognise the checkbox values.

This PR content changes to the hint in the affected questions.

These changes have been requested via this document [1] by the content
design team.

[1] https://docs.google.com/document/d/1L7iCySnGFcTxTM8iyDJuYYRwSFT7nbNmwAOIgbZ5CZ0/edit#

## Factcheck

[Preview link](https://smart-answers-preview-pr-3181.herokuapp.com/benefit-cap-calculator/y/yes/no)
[GOVUK](https://gov.uk/benefit-cap-calculator/y/yes/no)

## Expected changes

- [Question 3](https://smart-answers-preview-pr-3181.herokuapp.com/benefit-cap-calculator/y/yes/no):
  - Replace Yes/No with list of benefits
  - Content update

### Before
screen shot 2017-08-07 at 13 29 06

### After

screen shot 2017-08-07 at 13 28 55

PR #3180, opened 2017-08-04
Calculate your child maintenance: Replacing multiple choice YES/NO with Checkbox options

ikennaokpala:
[Trello card](https://trello.com/c/PornSWgk/700-5-uk-benefits-abroad-turning-yes-no-into-checkboxes)

## Motivation

As part of the accessibility improvements, we have been asked to replace the Yes/No multiple choice with checkboxes that contain the list of state benefits.

The intention here, is to assist devices like screen reader etc, to better recognise the checkbox values.

This PR content changes to the title and hints in the affected questions.

The question title has also changes as instructed in [1].

These changes have been requested via this document [1] by the content
design team.

[1] https://docs.google.com/document/d/1L7iCySnGFcTxTM8iyDJuYYRwSFT7nbNmwAOIgbZ5CZ0/edit#

## Factcheck

[Preview link](https://smart-answers-preview-pr-3180.herokuapp.com/calculate-your-child-maintenance/y/pay/1_child)
[GOVUK](https://gov.uk/calculate-your-child-maintenance/y/pay/1_child)

## Expected changes

- [Question 2](https://smart-answers-preview-pr-3180.herokuapp.com/calculate-your-child-maintenance/y/pay/1_child):
  - Replace Yes/No with list of benefits
  - Content update

### Before

screen shot 2017-08-04 at 16 38 26

### After
screen shot 2017-08-04 at 16 38 13

PR #3179, opened 2017-08-04
Uk Benefits Abroad: Replacing multiple choice YES/NO with Checkbox options

ikennaokpala:
[Trello card](https://trello.com/c/PornSWgk/700-5-uk-benefits-abroad-turning-yes-no-into-checkboxes)

## Motivation

As part of the accessibility improvements, we have been ask to replace the Yes/No multiple choice with checkbox that contain the list of options/periods/benefits etc.

The intention here, is to assist devices like screen reader etc, to better recognise the checkbox values.

This PR content changes to the title and hints in the affected questions.

The question title has also changes as instructed in [1].

These changes have been requested via this document [1] by the content
design team.

[1] https://docs.google.com/document/d/1L7iCySnGFcTxTM8iyDJuYYRwSFT7nbNmwAOIgbZ5CZ0/edit#

## NB:

- Update this branch with some updates for Monday:  [uk-benefits-abroad/yes-no-to-checkboxes-updated](https://github.com/alphagov/smart-answers/tree/uk-benefits-abroad/yes-no-to-checkboxes-updated)

## Factcheck

[Preview link](https://smart-answers-preview-pr-3179.herokuapp.com/uk-benefits-abroad/y/already_abroad/child_benefit/austria/)
[GOVUK](https://gov.uk/uk-benefits-abroad/y/already_abroad/child_benefit/austria/)

## Expected changes

- [Question 13](https://smart-answers-preview-pr-3179.herokuapp.com/uk-benefits-abroad/y/going_abroad/child_benefit/austria): 
  - Replace Yes/No with list of benefits
  - Content update
- [Question 37](https://smart-answers-preview-pr-3179.herokuapp.com/uk-benefits-abroad/y/going_abroad/income_support/is_under_a_year_other/none/none): 
  - Replace Yes/No with dispute criteria
  - Content update
- [Question 33](https://smart-answers-preview-pr-3179.herokuapp.com/uk-benefits-abroad/y/going_abroad/income_support/is_under_a_year_other): 
  - Replace Yes/No with partner premium
  - Content update
- [Question 34](https://smart-answers-preview-pr-3179.herokuapp.com/uk-benefits-abroad/y/going_abroad/income_support/is_under_a_year_other/none): 
  - Replace Yes/No with impairments
  - Content update
- [Question 36](https://smart-answers-preview-pr-3179.herokuapp.com/uk-benefits-abroad/y/going_abroad/income_support/is_under_a_year_other/none/too_ill_to_work/no): 
[...]

PR #3187, opened 2017-08-14
Define and apply List Validator.

ikennaokpala:
[Trello card](https://trello.com/c/QtiKCxmd/718-dryup-reduce-duplication-after-accessibility-changes)

## Description 

Following the accessibility improvements, a few of repetitive predicate
methods have been defined and apply across the following smart answers changes:

- https://github.com/alphagov/smart-answers/pull/3181
- https://github.com/alphagov/smart-answers/pull/3180
- https://github.com/alphagov/smart-answers/pull/3179

This commit defines a list validator that reduces the duplication
introduced during the accessibility improvements.

The idea here is to check the supplied argument and its content that
exist in the list that serves as a source of truth.

if the type or contents don't match with the pre-defined list, false is
returned else true.

## Expected changes

- none (i.e flow or regression test artefacts have not been affected)

### DIFF
--- a/lib/list_validator.rb
+++ b/lib/list_validator.rb
@@ -1,0 +1,11 @@
+class ListValidator
+  def initialize(list)
+    @list = list
+  end
+
+  def all_valid?(elements)
+    elements.present? &&
+      elements.is_a?(Array) &&
+      elements.all? { |element| @list.include?(element) }
+  end
+end
--- a/lib/smart_answer/calculators/benefit_cap_calculator_configuration.rb
+++ b/lib/smart_answer/calculators/benefit_cap_calculator_configuration.rb
@@ -26,9 +26,7 @@ module SmartAnswer::Calculators
     end
 
     def exempted_benefits?
-      exempted_benefits.present? &&
-        exempted_benefits.is_a?(Array) &&
-        valid_exempted_benefits?
+      ListValidator.new(exempt_benefits.keys).all_valid?(exempted_benefits)
     end
 
     def questions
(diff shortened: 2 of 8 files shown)


lib/list_validator.rb
```diff
@@ -0,0 +1,11 @@
+class ListValidator
+  def initialize(list)
+    @list = list
+  end
+
+  def all_valid?(elements)
+    elements.present? &&
```
alecgibson: We could potentially switch:
`elements.present? && elements.is_a?(Array) && ...`
with:
`elements.is_a?(Array) && elements.any? && ...`

...if you wanted to be a bit clearer about why we're checking `.present?`, but I only mention it as something we'd discussed.

lib/list_validator.rb
```diff
@@ -0,0 +1,11 @@
+class ListValidator
+  def initialize(list)
+    @list = list
```
alecgibson: (It would be slightly more efficient to store `list` as a `Set`.)

ikennaokpala: I buy your idea.. will research it a bit.. howbeit this may not be handle huge lists

lib/list_validator.rb
ikennaokpala: I prefer using `present?`

lib/list_validator.rb
alecgibson: Yeah, it's a nitpick - I don't think we'll be handling anything huge, so fine to merge.

PR #3189, opened 2017-08-17
Avoid big numbers for money

mgrassotti:
## Trello
https://trello.com/c/y3bOwPKj/293-2-floatdomainerror-infinity-in-multiple-smart-answers

## Problem
We noticed some `Float::Infinity` errors were happening in production, leading to a `500 Error`. That’s probably due to some users trying to input very big numbers.

## Test path
https://www.gov.uk/am-i-getting-minimum-wage/y/current_payment/not_an_apprentice/111/23/2.0/12.0/2.4132314123431646e+273/1111111111111111122222222222222222222222222222222222222222222222222222222222222222222222222222222225555555555555555555555555555555555555555555555555555555555222222222222222222222211111111111111111133333333333333333333555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555444444444444444444444444444448888888888888888888888888888888888888888888888888888888888887777777777777778888888888888889999999999.0

## Solution 
Raise an `InvalidResponse` error if the value for money is too
big to be represented in a `Float` variable.


alecgibson:
Shouldn't the `infinite?` check go last? Will this pass if I feed in `"999,999,999,...,999"` ?

alecgibson:
Okay, so apparently in Ruby:
`"999,999,999".to_f # => 999.0`

Which means that `"999,999,<loads of numbers>,999".to_f.infinte? # => nil`

alecgibson:
An example of the failure I'm trying to point out: https://smart-answers-preview-pr-3189.herokuapp.com/am-i-getting-minimum-wage/y/current_payment/not_an_apprentice/111/23/2.0/12.0/2.4132314123431646e+273/111,1111111111111122222222222222222222222222222222222222222222222222222222222222222222222222222222225555555555555555555555555555555555555555555555555555555555222222222222222222222211111111111111111133333333333333333333555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555555444444444444444444444444444448888888888888888888888888888888888888888888888888888888888887777777777777778888888888888889999999999.0

mgrassotti:
@alecgibson I put it at the end because otherwise it would fall in the condition `raw_input.is_a?(Numeric)`. Why do you think it should go at the end? Thanks

alecgibson:
@mgrassotti have you clicked the example link above? It still breaks the site.

At the moment we have **two** paths to generating a `BigDecimal`:
1. Check if the input is `Numeric`, and if so, cast it
2. Do some logic to a string to try and turn it into a `Numeric`, and _then_ cast it

The above link breaks the code, because it follows the second path, where we don't check `infinite?`.

I think logically we should do all casting first, and then check `infinite?`, kind of like:

```
money_as_float = some_function_that_definitely_returns_a_number_or_raises(some_input_string)
raise if money_as_float.infinite?
```

mgrassotti:
@alecgibson sorry I understand now. I pushed an update

lib/smart_answer/money.rb
```diff
@@ -9,14 +9,11 @@ class Money
     attr_reader :value
 
     def initialize(raw_input)
-      if raw_input.is_a?(Numeric)
-        @value = BigDecimal.new(raw_input.to_s)
+      number, error_message = self.class.raw_input_to_number(raw_input)
```
alecgibson: I'm not sure I like this pattern of returning a tuple with an error message. Why not just `raise` directly from `raw_input_to_number`? It should hopefully tidy up your control logic a lot more.

lib/smart_answer/money.rb
```diff
@@ -32,6 +29,23 @@ def <=>(other)
...
     end
 
+    def self.raw_input_to_number(raw_input)  
+      if raw_input.is_a?(Numeric)
+        raw_input.to_f.infinite? ? 
+          [nil, "Sorry, that number is too big. Please try again."] :
+          [raw_input, nil]
+      else
+        raw_input = raw_input.to_s.delete(',').gsub(/\s/, '')
+        if raw_input.to_f.infinite?
```
alecgibson: As per my previous comment, you could avoid having to check `infinite?` multiple times if you just call it once when you _know_ you have a `Numeric` (probably outside this method).

mgrassotti:
@alecgibson check it now

lib/smart_answer/money.rb
```diff
@@ -32,8 +25,21 @@ def <=>(other)
...
+    def self.raw_input_to_number(raw_input)
+      raw_input = raw_input.is_a?(Numeric) ? raw_input : raw_input.to_s.delete(',').gsub(/\s/, '')
+      if raw_input.to_f.infinite?
+        raise InvalidResponse, "Sorry, that number is too big. Please try again.", caller
+      else
+        valid_raw_input(raw_input)
+      end
+    end
+
+    def self.valid_raw_input(raw_input)
```
alecgibson: Would be nice to keep the validation predicate separate from the validation method. So, I'd expect this to be two methods:

- `valid_raw_input?`, which returns a boolean
- `validate_raw_input`, which calls `valid_raw_input?` and will `raise` if it's invalid. You could also move the `infinite?` check into the validation, because it's part of what makes the input valid.

lib/smart_answer/money.rb
```diff
@@ -32,8 +25,21 @@ def <=>(other)
...
+      if raw_input.is_a?(Numeric) || raw_input.is_a?(Money) || raw_input =~ /\A *[0-9]+(\.[0-9]{1,2})? *\z/
+        raw_input
+      else
+        raise InvalidResponse, "Sorry, I couldn't understand that number. Please try again.", caller
```
alecgibson: (I know this isn't  part of your change, but while we're here, do you think "**I** couldn't understand that number" is a normal sort of error message to return? I don't think I'd expect a validation message to use the first person. Maybe more like "The entered value is not a valid number. Please try again."?)

lib/smart_answer/money.rb
```diff
@@ -32,8 +25,21 @@ def <=>(other)
       end
     end
 
-    def self.valid?(raw_input)
-      raw_input.is_a?(Numeric) || raw_input.is_a?(Money) || raw_input =~ /\A *[0-9]+(\.[0-9]{1,2})? *\z/
+    def self.raw_input_to_number(raw_input)
+      raw_input = raw_input.is_a?(Numeric) ? raw_input : raw_input.to_s.delete(',').gsub(/\s/, '')
+      if raw_input.to_f.infinite?
+        raise InvalidResponse, "Sorry, that number is too big. Please try again.", caller
```
alecgibson: I still find it a bit untidy how our validation logic is tangled up in the casting logic.

I basically expect these three separate steps to happen:

- Attempt to coerce the raw input into a number
- Validate the number (and raise if appropriate)
- Cast the number to a `BigDecimal`

I'd expect these three steps to be represented at a top-level function (probably `initialize`), and to each have a single line of code, calling to methods if required. This way you get proper separation of concerns.

lib/smart_answer/money.rb
```diff
@@ -9,15 +9,8 @@ class Money
...
-        @value = BigDecimal.new(raw_input.to_s)
-      else
-        raw_input = raw_input.to_s.delete(',').gsub(/\s/, '')
-        if !self.class.valid?(raw_input)
-          raise InvalidResponse, "Sorry, I couldn't understand that number. Please try again.", caller
-        end
-        @value = BigDecimal.new(raw_input.to_s)
-      end
+      number = self.class.raw_input_to_number(raw_input)
+      @value = BigDecimal.new(number.to_s)
```
alecgibson: Why do we call `.to_s`?

mgrassotti: @alecgibson that should be useless indeed

mgrassotti: @alecgibson indeed without the `.to_s` call it is returning `ArgumentError: can't omit precision for a Float.`. It probably guesses the precision needed if the argument is a String. That's a little bit weird. Anyway I would leave it as it is for moment, if it is ok for you.

lib/smart_answer/money.rb
mgrassotti: @alecgibson ok

lib/smart_answer/money.rb
mgrassotti: @alecgibson ok

lib/smart_answer/money.rb
mgrassotti: Ok.

lib/smart_answer/money.rb
alecgibson: Huh, weird. Yeah fine to leave then.

lib/smart_answer/money.rb
```diff
@@ -9,15 +9,9 @@ class Money
...
-        @value = BigDecimal.new(raw_input.to_s)
-      else
-        raw_input = raw_input.to_s.delete(',').gsub(/\s/, '')
-        if !self.class.valid?(raw_input)
-          raise InvalidResponse, "Sorry, I couldn't understand that number. Please try again.", caller
-        end
-        @value = BigDecimal.new(raw_input.to_s)
-      end
+      input = self.class.parse(raw_input)
+      number = self.class.valid?(input) && input
```
alecgibson: Why bother assigning `number`? Given that `valid?` raises, couldn't you just do something like:

```ruby
input = self.class.parse(raw_input)
self.class.validate(input)
@value = BigDecimal.new(input.to_s)
```

mgrassotti: 👍

PR #3190, opened 2017-08-17
Wrap post code errors in a `StandardError` class

alecgibson:
Due to some combination of our usage of an old fork of Airbrake, and a
forked version of Errbit, usage of `Airbrake#notify` does not appear
to work when called with a string or hash. These appear in Errbit as
empty hashes, which is obviously not ideal for error reporting.

This change ensures that we invoke `Airbrake#notify` with an actual
`StandardError` object, which does present the error correctly.


alecgibson:
[Investigate production error in Landlord Inmigration Check](https://trello.com/c/ddeLCkMZ/690-investigate-production-error-in-landlord-inmigration-check)

PR #3194, opened 2017-08-18
Tidy up Post Code lookup error handling

alecgibson:
When a postcode cannot be found in Imminence API, the endpoint
returns a `404` status. We should accept this error as standard control
logic, and raise a corresponding error within our own application. This
shows the user that the problem is to do with their input.

Any other error is an unexpected error, and should raise an exception
showing that the problem with lookup is at our end.

This change also removes the special logic for `GdsApi::BaseError`. We
should never get this error, and if we do, it’s truly exceptional and
the website should fail “noisily”. Removing this `rescue` block will
log the error directly to Errbit, and show the user a generic `500`
page, whilst also tidying up our own control logic.

Further, this change adds a `LoggedError`, which allows us to add logging
on top of our control logic.

## Acceptance Criteria

We currently are logging errors when non-existent post codes are looked up.
For example: 
https://errbit.publishing.service.gov.uk/apps/533c35ae0da1159384044f5f/problems/599a04c9657863558e8e3801

After deploying this change, we should no longer see errors of this kind, and
looking up - for example - W13 0FE should not log an error.

### DIFF
--- a/lib/smart_answer/calculators/landlord_immigration_check_calculator.rb
+++ b/lib/smart_answer/calculators/landlord_immigration_check_calculator.rb
@@ -15,25 +15,22 @@ module SmartAnswer::Calculators
     def areas_for_postcode
       response = Services.imminence_api.areas_for_postcode(postcode)
 
-      if response.dig("_response_info", "status") == "ok"
+      case response.dig("_response_info", "status")
+      when "ok"
         response["results"]
+      when 404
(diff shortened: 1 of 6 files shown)


alecgibson:
[Fix Production post code lookup error](https://trello.com/c/jWrYU1uu/721-fix-production-post-code-lookup-error)

alecgibson:
@ikennaokpala I've added acceptance criteria.

PR #3200, opened 2017-08-24
Marriage Abroad - One outcome per path for Cambodia

mgrassotti:
In order to simplify the maintenance of marriage abroad smart answers, we are going to create one outcome per path for Cambodia.
Trello: https://trello.com/c/GbsqRWXv/728-marriage-abroad-create-one-outcome-per-path-for-cambodia

## Factcheck
https://smart-answers-preview-pr-3200.herokuapp.com/marriage-abroad/y/cambodia


mgrassotti:
@ikennaokpala ok, I am using the partial "contact-method" in 

```
<%= render partial: 'contact_method.govspeak.erb',
           locals: { calculator: calculator } %>
```

is that wrong?

ikennaokpala:
@mgrassotti Partials like `_contact_method.govspeak.erb` are not required as part of the flattening.
See https://github.com/alphagov/smart-answers/pull/3045/files

mgrassotti:
@ikennaokpala yes I see, but as the "contact the embassy..." part is repeated in almost every partial, I was wondering if it could be good to leave that...

ikennaokpala:
@mgrassotti The long term goal is to remove/reduce dependency on partials like `_contact_method.govspeak.erb`  that contain view country logic and allow content designer freedom to edit. There partials like `_consular_fees_table_items.govspeak.erb` that don't contain country logic and are derived from rates data. 

Other partials like `_you_have_been_divorced_or_widowed.govspeak.erb` don't contain logic and hence can be ignore and may be removed by the content designer later.

mgrassotti:
@ikennaokpala 👍

PR #3201, opened 2017-08-31
Algeria marriage abroad

hannahwhittaker:
https://govuk.zendesk.com/agent/tickets/2223115

I've updated the 3 user in the UK, opposite sex outcomes to tell users to exchange their CNI at the embassy.

This doesn't affect the logic.


mgrassotti:
superseded by https://github.com/alphagov/smart-answers/pull/3203

PR #3204, opened 2017-09-04
Create one outcome per path for Nicaragua

mgrassotti:
In order to simplify the maintenance of marriage abroad smart answers, we are going to create one outcome per path for Nicaragua.

Trello: https://trello.com/c/BqtzqHmG/738-marriage-abroad-create-one-outcome-per-path-for-nicaragua


mgrassotti:
@ikennaokpala 👍 let me know if that is ok now

PR #3215, opened 2017-09-13
Move Smart Answers from Errbit to Sentry

mgrassotti:
Trello: https://trello.com/c/6BqK4u39/740-move-smart-answers-from-errbit-to-sentry

We need to migrate CPM to Sentry which is the new default.
There are 2 new pages in the docs to help with this:

- https://docs.publishing.service.gov.uk/manual/error-reporting.html
- https://docs.publishing.service.gov.uk/manual/upgrade-to-sentry.html

https://trello.com/c/qAicE11v


mgrassotti:
at the moment there is an error raised because the new class `GovukError` seems not to be found after adding all the gems.
I tried to manually require it as stated in the documentation 
```
require 'govuk_app_config/configure'
``` 
but it doesn't work either.

ikennaokpala:
@mgrassotti  regarding the issue with govuk_app_config. I noticed that the version installed was 0.1.0 and the latest which contain GovError was actually 0.2.0. 

But the latest version had a conflict with statsd-ruby 1.3.0, so this I have upgraded to 1.4.0 first, before upgrading govuk_app_config. 

This has inturned introduced more problem.. will continue looking at it.

mgrassotti:
@tijmenb 👍 thanks I think I addressed your comments

config/initializers/govuk_navigation_helpers.rb
```diff
@@ -1,4 +1,3 @@
 GovukNavigationHelpers.configure do |config|
-  config.error_handler = Airbrake
-  config.statsd = Services.statsd
+  config.error_handler = GovukError
```
tijmenb: You'll need `config.statsd = GovukStatsd` here like:

https://github.com/alphagov/collections/blob/master/config/initializers/govuk_navigation_helpers.rb#L1-L4

mgrassotti: 👍

tijmenb:
@mgrassotti looks like a few errors on CI, but otherwise looks good.

mgrassotti:
@tijmenb great, this is the only one error left:
```
Error:
SmartAnswersControllerTest#test_: GET /<slug> A/B testing pages under A/B test should show taxon breadcrumbs for the 'B' version. :
ActionView::Template::Error: undefined method `time' for GovukStatsd:Module
Did you mean?  timeout
    app/views/smart_answers/_related_items.html.erb:4:in `_app_views_smart_answers__related_items_html_erb__1451734698928877256_70125785253940'
    app/views/smart_answers/landing.html.erb:2:in `block in _app_views_smart_answers_landing_html_erb___360116743409834156_70125785130460'
    app/views/smart_answers/landing.html.erb:1:in `_app_views_smart_answers_landing_html_erb___360116743409834156_70125785130460'
    app/controllers/smart_answers_controller.rb:37:in `block (2 levels) in show'
    app/controllers/smart_answers_controller.rb:31:in `show'
    test/functional/smart_answers_controller_test.rb:280:in `block (5 levels) in <class:SmartAnswersControllerTest>'
    test/functional/smart_answers_controller_test.rb:279:in `block (4 levels) in <class:SmartAnswersControllerTest>'
    test/functional/smart_answers_controller_test.rb:291:in `instance_exec'
    test/functional/smart_answers_controller_test.rb:291:in `block in create_test_from_should_hash'

bin/rails test test/functional/smart_answers_controller_test.rb:278
```
do you have any idea about it?

mgrassotti:
@tijmenb sorry maybe I didn't addressed well a previous comment from yours. Now I used `config.statsd = GovukStatsd.client` in `config/initializers/govuk_navigation_helpers.rb`, seeing if it works.

tijmenb:
@mgrassotti sorry, my bad! I gave the wrong advice. `config.statsd = GovukStatsd.client` is correct.

lib/smart_answer/logged_error.rb
```diff
@@ -1,14 +1,10 @@
 module SmartAnswer
   class LoggedError < BaseStateTransitionError
+    attr_reader :log_message
```
ikennaokpala: is this line really required ?

mgrassotti:
@tijmenb no worries! Now tests should all be ok. Do you want to approve this PR? Many thanks

lib/smart_answer/logged_error.rb
mgrassotti: @ikennaokpala it seems it raises an Error otherwise:
```
Error:
FlowTest#test_: sequence of two questions a question raises a logged error should notify Sentry. :
NoMethodError: undefined method `log_message' for #<SmartAnswer::LoggedError: Sorry, that's not valid>
    lib/smart_answer/flow.rb:127:in `rescue in block in process'
    lib/smart_answer/flow.rb:122:in `block in process'
    lib/smart_answer/flow.rb:120:in `each'
    lib/smart_answer/flow.rb:120:in `inject'
    lib/smart_answer/flow.rb:120:in `process'
    test/unit/flow_test.rb:292:in `block (3 levels) in <class:FlowTest>'
    test/unit/flow_test.rb:303:in `instance_exec'
    test/unit/flow_test.rb:303:in `block in create_test_from_should_hash'

bin/rails test test/unit/flow_test.rb:290
```

mgrassotti:
@tijmenb sorry I am doing other changes, I'll let you know when it's definitive

mgrassotti:
@tijmenb it should be ok now

PR #3219, opened 2017-09-29
Use new DateRange class in existing flows

mgrassotti:
Trello: https://trello.com/c/ubcEk4vh/731-make-use-of-new-daterange-class-in-existing-flows

Use the new class `SmartAnswer::DateRange` in the existing flows, where possible, instead of manual calculations on date ranges.

### DIFF
--- a/lib/smart_answer/date_range.rb
+++ b/lib/smart_answer/date_range.rb
@@ -26,7 +26,8 @@ module SmartAnswer
     end
 
     def include?(date)
-      (ComparableDate.new(date.to_date) >= ComparableDate.new(@begins_on)) && (ComparableDate.new(date.to_date) <= ComparableDate.new(@ends_on))
+      (ComparableDate.new(date.to_date) >= ComparableDate.new(@begins_on)) &&
+        (ComparableDate.new(date.to_date) <= ComparableDate.new(@ends_on))
     end
 
     def intersection(other)
@@ -74,6 +75,35 @@ module SmartAnswer
         begins_on: [ComparableDate.new(ends_on), ComparableDate.new(other.ends_on)].min.date + 1,
         ends_on: [ComparableDate.new(begins_on), ComparableDate.new(other.begins_on)].max.date - 1
       )
+    end
+
+    def years
+      (begins_on.year..ends_on.year).to_a
+    end
+
+    def feb29th_date(year)
+      Date.new(year, 2, 29)
+    end
+
+    def leap_dates
+      years.inject([]) { |mem, year|
+        if Date.leap?(year) && include?(feb29th_date(year))
+          mem << feb29th_date(year)
+        end
+        mem
+      }
+    end
+
+    def leap?
+      leap_dates.any?
+    end
+
+    def weeks_later(weeks)
+      self + weeks * 7
+    end
+
+    def formatted
+      "#{formatted_date begins_on} to #{formatted_date ends_on}"
     end
   end
 end
--- a/lib/smart_answer/date_range.rb
+++ b/lib/smart_answer/date_range.rb
@@ -1,5 +1,7 @@
 module SmartAnswer
   class DateRange
+    include SmartAnswer::DateHelper
+
     EARLIEST_DATE = -Date::Infinity.new
     LATEST_DATE = Date::Infinity.new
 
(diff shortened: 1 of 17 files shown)


lib/smart_answer/date_range.rb
```diff
@@ -102,5 +108,13 @@ def leap_dates
     def leap?
       leap_dates.any?
     end
+
+    def weeks_later(weeks)
```
chao-xian: `date_range.increase_weeks 10` might make more sense?

test/unit/date_range_test.rb
```diff
@@ -95,6 +95,26 @@ class DateRangeTest < ActiveSupport::TestCase
...
         assert_equal range, @date_range.to_range
       end
+
+      should 'let to add days' do
+        new_range = @date_range + 10
+        assert_equal new_range.begins_on, @date_range.begins_on + 10
+        assert_equal new_range.ends_on, @date_range.ends_on + 10
+      end
+
+      should 'let to add weeks' do
```
chao-xian: `should 'allow increase by X weeks`?

lib/smart_answer/date_range.rb
```diff
@@ -102,5 +108,13 @@ def leap_dates
...
+      self + weeks * 7
+    end
+
+    def formatted
```
chao-xian: Would it make sense to call it `.to_s`?

`.formatted` to me implies that we can apply different formatting, and then I would expect to use `.to_formatted_s(:format)` on it.

lib/smart_answer/date_range.rb
```diff
@@ -75,5 +75,10 @@ def gap_between(other)
         ends_on: [ComparableDate.new(begins_on), ComparableDate.new(other.begins_on)].max.date - 1
       )
     end
+
+    def to_range
```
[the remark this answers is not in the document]
mgrassotti: @chao-xian 👍

lib/smart_answer/date_range.rb
```diff
@@ -38,7 +38,11 @@ def intersection(other)
     alias_method :&, :intersection
 
     def number_of_days
-      infinite? ? Float::INFINITY : (@ends_on - @begins_on).to_i + 1
+      available_days + 1
+    end
+
+    def available_days
```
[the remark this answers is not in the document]
mgrassotti: @chao-xian 👍

lib/smart_answer/date_range.rb
mgrassotti: @chao-xian 👍

test/unit/date_range_test.rb
mgrassotti: @chao-xian 👍

mgrassotti:
@chao-xian I reviewed your comments and updated the commits coherently

lib/smart_answer/date_range.rb
chao-xian: Having rethought this I recommend we rename this to `.weeks_after` because `later` always suggests in the future. But we are using it with minus X weeks to go backwards in a date range, which doesn't fit right with `later`. 

I think`.weeks_after(-10)` feels a bit more flexible and appropriate.

chao-xian:
@mgrassotti thanks for making your changes. Just one minor comment.

lib/smart_answer/date_range.rb
mgrassotti: @chao-xian what about `.shift_weeks(-10)` ?

chao-xian: This is just returning a result though, not setting a new value and that's what shift means to me. Ideally it would read as something like `Date range at +/- X weeks`

Mabe just swap the words in my suggestion so it'd be `date_range.after_weeks(10)` and equally `date_range.after_weeks(-10)` I think would work. What do you think of that?

mgrassotti: @chao-xian that is fine 👍

mgrassotti: @chao-xian but I would prefer `weeks_after`...

chao-xian: Sorry @mgrassotti I didn't realise that you'd addressed my comments about renaming the methods in [this commit](2e3fd972701ca9637903539064e3c6890f699f4d). 

I will approve this PR now but I think it would have been better to make that amendment in the commit where you introduced the new code.

Thanks for your work!

PR #3224, opened 2017-10-03
Consistently format money in ERB templates

mgrassotti:
Trello: https://trello.com/c/9Fu5uqGV/733-consistently-format-money-in-erb-templates

Use the `format_money` helper in ERB templates, instead of ```£<%= amount %>```.


chao-xian:
Hey @mgrassotti thanks for applying this improvement.

The commits on this PR will need re-working to keep each commit atomic: `Ensure your branch contains logical atomic commits before sending a pull request` -
 https://github.com/alphagov/smart-answers/blob/master/CONTRIBUTING.md

As it is, the first 2 commits break the build. 

You can either:

* squash into a single commit

or

* to achieve smaller commits apply changes to a single smart answer template with regression test artefact and checksum updates in each commit

The latter would be preferable as it would be easier to read and understand.

mgrassotti:
👍 Thanks, it should be done now

PR #3227, opened 2017-10-09
Use YearRange

mgrassotti:
In order to simplify the code, we are introducing the class `YearRange`
to deal with 1 year date ranges.


lib/smart_answer/date_range.rb
```diff
@@ -111,5 +114,54 @@ def weeks_after(weeks)
...
+    def whole_years_away
+      month_difference / 12
+    end
+
+    def whole_months_away
+      month_difference % 12
+    end
+
+    def month_difference
+      m = 0
```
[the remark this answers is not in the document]
mgrassotti: 👍

lib/smart_answer/date_range.rb
```diff
@@ -111,5 +114,54 @@ def weeks_after(weeks)
...
+      loop do
+        m += 1
+        break unless (begins_on >> m) <= ends_on
+      end
+      m - 1
+    end
+
+    def whole_days_away
+      closest_month_anniversary_before_to_date = (begins_on >> month_difference)
```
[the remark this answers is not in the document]
mgrassotti: @chao-xian I renamed it `begins_on_shifted_to_ends_on_month`. Let me know if it makes more sense.

mgrassotti:
@andrewgarner @chao-xian is this ok for you now?

lib/smart_answer/calculators/pay_leave_for_parents_calculator.rb
```diff
@@ -97,60 +97,36 @@ def earnings_employment(earnings_employment, work_employment)
...
-      saturday_before(due_date - 14.weeks)
-    end
-
-    def in_2013_2014_fin_year?(date)
-      (Date.new(2013, 05, 06)..Date.new(2014, 05, 05)).cover?(date)
-    end
-
-    def in_2014_2015_fin_year?(date)
-      (Date.new(2014, 05, 06)..Date.new(2015, 05, 05)).cover?(date)
+      maternity_leave_notice_date
```
andrewgarner: Maybe we could use an alias so that is more clear that these two methods are the same?

```rb
def maternity_leave_notice_date
  saturday_before(due_date - 14.weeks)
end

alias_method :paternity_leave_notice_date :maternity_leave_notice_date
```

mgrassotti: @andrewgarner Yes that was my first idea. But then I realised that the fact that these methods are calculated in the same way, doesn't mean they are the same. Indeed they are probably not, as the paternity leave and maternity leave could be calculated differently in the future, and I think that defining one them as an alias could be more confusing.

andrewgarner: I think an alias is a good design decision because it makes it very clear that they are currently the same. If someone in the future comes to change the implementation of `maternity_leave_notice_date` it will be clear to them that they are also changing the implementation of `paternity_leave_notice_date` and can consider if that is the right thing to do at the time.

With the implementation we have now, it might be less clear to future authors that changing the implementation of `maternity_leave_notice_date` has a side effect on `paternity_leave_notice_date`.

mgrassotti: @andrewgarner ok I agree, I updated it

PR #3234, opened 2017-10-17
Display money less than one pound in pence

h-lame:
For: https://trello.com/c/BOEQuDjb/257-formatting-of-numbers-in-benefit-cap-calculator

We tweak the `format_money` helper to show values less than a pound as pence values (e.g. `50p` not `£0.50`).

We also do some follow up from #3224 and make sure that views that were using `number_to_currency` now also use `format_money`.


lib/smart_answer/formatting_helper.rb
```diff
@@ -2,12 +2,18 @@ module SmartAnswer
...
       if show_in_pence?(amount)
         number_to_currency(amount * 100, precision: 0, unit: 'p', format: '%n%u')
       else
-        number_to_currency(amount, precision: ((amount == amount.round) ? 0 : 2))
+        precision =
+          if pounds_only || amount == amount.round
+            0
+          else
+            2
+          end
```
mgrassotti: @h-lame do you think a ternary operator would not be readable enough here?

h-lame: I think when the condition is more than a single clause it can get confusing to read (as you have to start thinking about operator precedence.  My preference was to add brackets, but rubocop didn't like it when I put brackets around the ternary so I exploded it out into a full if statement.

h-lame: Rejigged this after a discussion on slack.

PR #3246, opened 2017-10-30
Change rummager payload document type

suzannehamilton:
The start page for a smartanswer is a `transaction` document type in the content store. This should be consistent in the search index.

https://trello.com/c/BTZkG7CX/400-fix-documenttype-of-smart-answer-start-pages-in-search-index


alecgibson:
@suzannehamilton while we're here, do you happen to know if `format: smart-answer` matches Elasticsearch?

PR #3251, opened 2017-11-02
Improve arbitary input handling of CurrentQuestionHelper#default_for_date

alecgibson:
We recently received alerts in [Sentry][1] for 500 errors when receiving certain inputs for a date. These alerts arose from passing an array response to a date question. Our handling of arbitrary inputs to the `CurrentQuestionHelper#default_for_date` was no robust enough to deal with this, and attempted to call `.to_i` on an `Array`, which throws a `NoMethodError`.

This change makes the method more robust by direclty attempting to cast to `Integer`, and rescuing errors by returning `nil`. This change also adds a number of tests to cover various types of input to the  method.

[1]: https://sentry.io/govuk/app-smart-answers/issues/396675536/

### DIFF
--- a/app/helpers/current_question_helper.rb
+++ b/app/helpers/current_question_helper.rb
@@ -22,7 +22,10 @@ module CurrentQuestionHelper
   end
 
   def default_for_date(value)
-    value.blank? ? nil : value.to_i
+    integer = Integer(value)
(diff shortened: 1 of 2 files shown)


alecgibson:
[Investigate `[].to_i` failure](https://trello.com/c/kgoLBDu9/773-investigate-toi-failure)

PR #3256, opened 2017-11-07
Fix Maternity Smart Answer Error

theKHutDeveloper:
There's a problem with the calculator when the last SMP payment is after the last working day
of the month. This was spotted for June 2018, but we think it would happen whenever this
scenario arises.

Previously the code iterated the SMP payments according to the last working day of each month
but it did not tackle the scenario where the final SMP payment is outside of the last working
day of leave.

The new code iterates through the last working day of each month while the last working day of
the month is less than the final SMP pay date. This ensures that all payments are made. In this
scenario an extra month is added to which the last payment will be made.

### DIFF
--- a/lib/smart_answer/calculators/maternity_paternity_calculator.rb
+++ b/lib/smart_answer/calculators/maternity_paternity_calculator.rb
@@ -265,13 +265,18 @@ module SmartAnswer::Calculators
     end
 
     def paydates_last_working_day_of_the_month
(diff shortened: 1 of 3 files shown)


alecgibson:
[(5) Maternity smart answer outcome error](https://trello.com/c/rUhCsXAo/751-5-maternity-smart-answer-outcome-error)

PR #3258, opened 2017-11-09
Fix state pension calculator for leap year births

theKHutDeveloper:
This issue affects users born after 5th April 1976 on a leap year. These users will currently be eligible for their state pension on their 68 birthday (which will also be a leap year).

Currently users are given the correct entitlement date for their state pension but says their state pension age will be 67 years, 11 months, 30 days. This is because the calculator deducts a day for leap year births, but does not take into account the state pension date also being a leap year.

This fix updates the calculator to check whether the state pension date is also a leap year. If so no adjustment needs to be made to the state pension age.

### DIFF
--- a/test/unit/calculators/state_pension_age_calculator_test.rb
+++ b/test/unit/calculators/state_pension_age_calculator_test.rb
@@ -92,7 +92,7 @@ module SmartAnswer::Calculators
           should "someone born on #{test_case[:birth_date]} and getting their pension on #{test_case[:pension_date]} should have a pension age of #{test_case[:age]}" do
             @calculator.stubs(:dob).returns(test_case[:birth_date])
             @calculator.stubs(:state_pension_date).returns(test_case[:pension_date])
-            assert_equal @calculator.state_pension_age, test_case[:expected_age]
+            assert_equal test_case[:expected_age], @calculator.state_pension_age
           end
         end
       end
@@ -128,7 +128,7 @@ module SmartAnswer::Calculators
           should "someone born on #{test_case[:birth_date]} and getting their pension on #{test_case[:pension_date]} should have a pension age of #{test_case[:age]}" do
             @calculator.stubs(:dob).returns(test_case[:birth_date])
             @calculator.stubs(:state_pension_date).returns(test_case[:pension_date])
-            assert_equal @calculator.state_pension_age, test_case[:expected_age]
+            assert_equal test_case[:expected_age], @calculator.state_pension_age
           end
         end
       end
@@ -157,7 +157,7 @@ module SmartAnswer::Calculators
         end
 
         should "be true for a date that is the 29th of Feb" do
-          assert_equal true, @calculator.pension_on_feb_29?
+          assert @calculator.pension_on_feb_29?
         end
       end
 
@@ -169,7 +169,7 @@ module SmartAnswer::Calculators
         end
 
         should "be false for a date that is not the 29th of Feb" do
-          assert_equal false, @calculator.pension_on_feb_29?
+          refute @calculator.pension_on_feb_29?
         end
       end
     end
--- a/test/unit/calculators/state_pension_age_calculator_test.rb
+++ b/test/unit/calculators/state_pension_age_calculator_test.rb
@@ -61,14 +61,75 @@ module SmartAnswer::Calculators
         end
       end
 
-      context 'given a state_pension_date on the 29th of Feb when the pension age is 68' do
-        setup do
-          @calculator.stubs(:dob).returns(Date.parse("29 Feb 1980"))
-          @calculator.stubs(:state_pension_date).returns(Date.parse('29 Feb 2048'))
+      context 'given variable changes to state pension age between 66 - 67 years for births between 6 Apr 1960 - 5 Mar 1961' do
+        [
+          {
+            birth_date: Date.parse('6 Apr 1960'),
+            pension_date: Date.parse('6 May 2026'),
+            expected_age: '66 years, 1 month'
+          },
+          {
+            birth_date: Date.parse('20 May 1960'),
+            pension_date: Date.parse('20 Jul 2026'),
+            expected_age: '66 years, 2 months'
+          },
+          {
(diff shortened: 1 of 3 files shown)


theKHutDeveloper:
[(5) Check your state pension age calculator](https://trello.com/c/SUk2Bjsm/776-5-check-your-state-pension-age-calculator)

test/unit/calculators/state_pension_age_calculator_test.rb
```diff
@@ -60,6 +60,17 @@ class StatePensionAgeCalculatorTest < ActiveSupport::TestCase
           assert_equal @calculator.state_pension_age, '60 years'
         end
       end
+
+      context 'given a state_pension_date on the 29th of Feb when the pension age is 68' do
```
alecgibson: You could move this into the `context` above, because it's closely related.

test/unit/calculators/state_pension_age_calculator_test.rb
```diff
@@ -60,6 +60,17 @@ class StatePensionAgeCalculatorTest < ActiveSupport::TestCase
...
+        setup do
+          @calculator.stubs(:dob).returns(Date.parse("29 Feb 1980"))
+          @calculator.stubs(:state_pension_date).returns(Date.parse('29 Feb 2048'))
```
alecgibson: It seems very odd to me that in the test above this one, we return a date of 1 March 2040 (also a leap year), and it _also_ claims that state pension is exactly 60 years away. That's not right, is it?

theKHutDeveloper: I completely agree, anyone born in 1980 will not be eligible for their state pension at 60 years old

test/unit/calculators/state_pension_age_calculator_test.rb
theKHutDeveloper: Agreed

alecgibson:
Are we still passing the case where:
- Date of birth: 29th Feb 1980
- Pension date: 1st Mar 2040
- Pension age: 60 years exactly

Because that shouldn't pass. It should report a pension age of 60 years and 1 day.

theKHutDeveloper:
@alecgibson It does pass because the calculator deducts 1 day from that scenario

alecgibson:
@theKHutDeveloper the test shouldn't pass, though.

test/unit/calculators/state_pension_age_calculator_test.rb
```diff
@@ -60,6 +60,78 @@ class StatePensionAgeCalculatorTest < ActiveSupport::TestCase
...
+          {
+            birth_date: Date.parse('28 Feb 1961'),
+            pension_date: Date.parse('28 Jan 2028'),
+            expected_age: '66 years, 11 months'
+          },
+        ].each do |test_case|
+          should "someone born on #{test_case[:birth_date]} and getting their pension on #{test_case[:pension_date]} should have a pension age of #{test_case[:age]}" do
+            @calculator.stubs(:dob).returns(test_case[:birth_date])
+            @calculator.stubs(:state_pension_date).returns(test_case[:pension_date])
+            assert_equal @calculator.state_pension_age, test_case[:expected_age]
```
andrewgarner: We've got our arguments in the wrong order here. This will give us confusing failure messages.

It should be `assert_equal(expected, actual)` not `assert_equal(actual, expected)`

See https://apidock.com/ruby/Test/Unit/Assertions/assert_equal

test/unit/calculators/state_pension_age_calculator_test.rb
```diff
@@ -76,6 +148,32 @@ class StatePensionAgeCalculatorTest < ActiveSupport::TestCase
...
+    context '#pension_on_feb_29??' do
+      context 'pension due on 29th of Feb' do
+        setup do
+          @calculator = StatePensionAgeCalculator.new(
+            gender: "male", dob: Date.parse("29 Feb 1980"))
+          @calculator.stubs(:state_pension_date).returns(Date.parse('29 Feb 2048'))
+        end
+
+        should "be true for a date that is the 29th of Feb" do
+          assert_equal true, @calculator.pension_on_feb_29?
```
andrewgarner: We should use `assert` here which fails unless the supplied argument is a `true` value.

```ruby
assert @calculator.pension_on_feb_29?
```

See http://ruby-doc.org/stdlib-2.0.0/libdoc/minitest/rdoc/MiniTest/Assertions.html#method-i-assert

test/unit/calculators/state_pension_age_calculator_test.rb
```diff
@@ -76,6 +148,32 @@ class StatePensionAgeCalculatorTest < ActiveSupport::TestCase
...
+
+      context 'pension not due on 29th of Feb' do
+        setup do
+          @calculator = StatePensionAgeCalculator.new(
+            gender: "male", dob: Date.parse("28 Aug 1974"))
+          @calculator.stubs(:state_pension_date).returns(Date.parse('28 Aug 2041'))
+        end
+
+        should "be false for a date that is not the 29th of Feb" do
+          assert_equal false, @calculator.pension_on_feb_29?
```
andrewgarner: We should use `refute` here which fails if the supplied argument is a true value.

```ruby
refute @calculator.pension_on_feb_29?
```

See http://ruby-doc.org/stdlib-2.0.0/libdoc/minitest/rdoc/MiniTest/Assertions.html#method-i-refute

test/unit/calculators/state_pension_age_calculator_test.rb
theKHutDeveloper: Noted

test/unit/calculators/state_pension_age_calculator_test.rb
theKHutDeveloper: Noted

test/unit/calculators/state_pension_age_calculator_test.rb
theKHutDeveloper: Thanks for this, I didn't know about refute.

PR #3260, opened 2017-11-09
Use standard buildProject call in Jenkinsfile

h-lame:
Our job in jenkins is 'smartanswers', but our repo on github is
'smart-answers'.  The default buildProject used to not understand this
difference which is why we didn't use it.  However, it's now been updated
so that it does understand it and we no longer need our custom version.

We also need to:

1. configure an extra build param: RUN_REGRESSION_TESTS - used to control
   if we should run the (slow) regression tests or not
2. customise the test task so that it runs the default tests, but also
   will run the regression tests if we're on the master branch, or the
   RUN_REGRESSION_TESTS param is set.

Similar to alphagov/contacts-admin#280 and alphagov/licence-finder#119 - our goal here is to move smartanswers out of [this block in govuk-content-schemas Jenkinsfile](https://github.com/alphagov/govuk-content-schemas/blob/master/Jenkinsfile#L33-L38) because it uses the standard buildProject and will report back to github itself.


h-lame:
Note to self: this build passes, but it seems you can't call `properties` twice in a jenkins script and expect it to aggregate the parameters defined.  See the parameters for the passing build (https://ci.integration.publishing.service.gov.uk/job/smartanswers/job/use-standard-buildProject-in-jenkinsfile/2/parameters/) and compare it with the parameters for a passing build on another branch still using the custom jenkinsfile (https://ci.integration.publishing.service.gov.uk/job/smartanswers/job/master/3783/parameters/).  Our parameters don't include the RUN_REGRESSION_TESTS param, even though we called `properties` to define it.  Looks like it was overridden by the call in buildProject that adds the other params.  Suspect we'll have to change buildProject to add the ability to define extra params.

h-lame:
This PR now relies on alphagov/govuk-puppet#6773 which changes `buildProject` to allow providing extra jenkins build parameters.

h-lame:
The PR for govuk-puppet that this relies on has been merged, so this is good to review + merge.

PR #3262, opened 2017-11-10
Refactor calculators to use `YearRange`

alecgibson:
We have a few occurrences of different calculators relying on the concept of a year starting on a particular date (eg a Tax Year, which starts on 6th April).

This change refactors the existing `TaxYear` class into a more generic class, which will take a fixed day and month, and can generate the appropriate year range given a date, or a starting year.

Any instances that could be found in the calculators where they rely on some concept of these years has been refactored to use these new methods, which also removes "magic string" dates.


alecgibson:
[(3) Make use of new TaxYear class in existing flows](https://trello.com/c/pNH4EBs0/734-3-make-use-of-new-taxyear-class-in-existing-flows)

alecgibson:
@mgrassotti the concept of a Tax Year _always_ runs from 6th April. Why would it ever make sense to say `agricultural_holiday_year = TaxYear.new(begins_on: "1 October")` ?

I don't think I understand what coupling you're talking about? Everything is free to define its own year now.

mgrassotti:
@alecgibson sorry I don't understand why we need a `YearRangeWithFixedStartDate` class then. `YearRange` shouldn't be already able to initialised to a specific start date? About coupling I meant I would avoid to introduce methods name which refers to a specific calculator need inside the TaxYear or YearRange. But maybe I am missing something.

alecgibson:
@mgrassotti `YearRange` expects an exact start date. For example, `YearRange.new(Date.today)` gives me a 1 year range starting today, and ending a year from now.

The idea of a `TaxYear` is that it _always_ starts on the same date every year. Sometimes I want to know which tax year I am currently in, which the `TaxYear` class solves (eg `TaxYear.current` will tell me that I'm in a `YearRange` starting on 6th April 2017, and ending on 5th April 2018).

We _also_ have other years that do a similar thing (eg the agricultural holiday year), and if we were to write - for example - an `AgriculturalHolidayYear` class, it would look _exactly_ the same as the `TaxYear` class, just with a different `BEGINS_ON` date.

This commonality in the code is crying out for the class to be abstracted. We keep all the existing logic in `TaxYear`, but make it into something more generic: a year range who alway starts on a particular date, which is provided at initialization (much as you suggest doing with `TaxYear` - just with a name that doesn't tie the class to a particular date, which `TaxYear` does).

I may be missing something; please write some example code of what you would expect to see if you'd do this differently.

mgrassotti:
@alecgibson ok then, thanks for your exhaustive explanation. I would just use a constant for `AGRICULTURAL_HOLIDAY_YEAR_START` instead of a dedicated method:
``` 
def self.agricultural_holiday_year
  resetting_on("1 October")
end
``` 
But I let you the choice

alecgibson:
@mgrassotti a constant is all well and good, but how would you use it?

mgrassotti:
@alecgibson defining it inside `agricultural_holiday_entitlement_calculator.rb` and then calling `YearRange.resetting_on(AGRICULTURAL_HOLIDAY_YEAR_START)`. Is that ok?

alecgibson:
@mgrassotti I think that's a bit unnecessarily verbose? It's not particularly readable, and I'm not sure what advantages it gives us?

I see what you mean about wanting to isolate the concern of a calculator, which is arguable for the agricultural holiday, but things like a tax year are certainly a global construct, so I think a case can be made for `YearRange` knowing about these things.

I was actually quite tempted to make `resetting_on` a private class method, so that we wouldn't get re-runs of the [tax year bug][https://github.com/alphagov/smart-answers/pull/3263] where people are tempted to manually set the start date.

mgrassotti:
@alecgibson ok fine then 👍

PR #3263, opened 2017-11-13
Fix the tax year in the Pay Leave for Parents calculator

alecgibson:
The Pay Leave for Parents Calculator currently uses a tax year
that erroneously runs from 6th May, rather than 6th April. I cannot find
any context around this (old!) code, but since the [copy][1] that
references these methods talks about a tax year running from 6th April,
it is clear that the method should run from 6th April.

This change adds tests that would have caught this bug, and fixes the
year to run from 6th April.

[1]: https://github.com/alphagov/smart-answers/blob/736dc25024e40ae32e262148eb1a37081b8f9510/lib/smart_answer_flows/pay-leave-for-parents/outcomes/_additional_pat_pay.govspeak.erb#L7-L11


alecgibson:
[Wrong Financial Year start date in Pay Leave for Parents calculator](https://trello.com/c/GNBDGw0g/779-wrong-financial-year-start-date-in-pay-leave-for-parents-calculator)

alecgibson:
This was fixed in this refactor: https://github.com/alphagov/smart-answers/pull/3262

PR #3264, opened 2017-11-13
Tidy up `DateRange` ends on logic

alecgibson:
The `DateRange` class currently handles the case of having an end date
before the start date in a confusing manner.

If I have an end date before a start date, the `DateRange` will rewrite
my end date to be 1 day before my start date, which is _still_ before
the start date, but _not_ the date that I set.

For example:

```ruby
range = DateRange.new(
  begins_on: Date.parse("13 Nov 2017"),
  ends_on: Date.parse("1 Nov 2017")
)

range.ends_on # => 12th November 2017
```

As far as I can tell, this confusing behaviour exists solely to abuse
the subtraction that occurs in `non_inclusive_days`, which will in turn
generate the desired answer of `0` when calling `number_of_days`.

However - assuming the same `range` variable as above, this flow looks
like:

```ruby
range.non_inclusive_days # => -1
range.number_of_days     # => -1 + 1 == 0
```

This abuse of magical addition is a little obscure.

This change removes this extra `ends_on` logic on initialization, and
instead explicitly handles the case where the date range is not valid.

### DIFF
--- a/lib/smart_answer/date_range.rb
+++ b/lib/smart_answer/date_range.rb
@@ -25,7 +25,6 @@ module SmartAnswer
     def initialize(begins_on: nil, ends_on: nil)
       @begins_on = (begins_on || EARLIEST_DATE).to_date
       @ends_on = (ends_on || LATEST_DATE).to_date
-      @ends_on = [@begins_on - 1, @ends_on].max unless infinite?
     end
 
     def include?(date)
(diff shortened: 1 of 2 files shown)


alecgibson:
[Improve DateRange corner case handling](https://trello.com/c/YVhH1JFx/780-improve-daterange-corner-case-handling)

alecgibson:
We possibly need to rethink this class in general - it's still possible to have an end date before a start date; is that valid? We can't just raise in this case, because we'd need to think carefully about how to represent - for example - the intersection of two date ranges that don't overlap. We could potentially do that with an `EmptyDateRange` class which has the same interface as `DateRange`, but the overheads this adds may not be worth it.

PR #3259, opened 2017-11-09
Update adoption leave start date validation

alecgibson:
This change addresses the Adoption Calculator. We currently validate the adoption leave start date by ensuring that it is no more than 14 days before the adoption placement date.  It also adds a question to the Adoption calculator flow, asking if the child is being adopted from abroad.

This information can be used to update the copy for some questions, and to add validation for when the employee's adoption leave can start (which is different depending on if the child is being adopted from the UK or abroad).

We also need to validate that the leave start date is no later than one day after the placement start date.

This change is broken into several commits
- the first adds a series of basic unit tests to the Adoption Calculator flow, which was previously only covered by an integration test
- the second adds the validation for the adoption leave start date, and updates the accompanying copy
- the third updates the test artefacts and file checksums
- the fourth adds the new question
- the fifth updates the copy and tests to include the new question 

## Preview

[Current](https://www.gov.uk/maternity-paternity-calculator/y/adoption/maternity/2017-01-01/2017-12-01/yes/yes/yes/2017-12-03)
[New](https://smart-answers-preview-pr-3259.herokuapp.com/maternity-paternity-calculator/y/adoption/maternity/2017-01-01/2017-12-01/yes/yes/yes?utf8=%E2%9C%93&response%5Bday%5D=3&response%5Bmonth%5D=12&response%5Byear%5D=2017&next=1)

## Screenshots

adoption-validation

adoption-abroad

PR #3266, opened 2017-11-15
[WIP] Adoption from abroad validation

alecgibson:
This change builds on top of the changes made in the `adoption-leave-validation` branch. It adds a question to the Adoption calculator flow, asking if the child is being adopted from abroad.

This information can be used to update the copy for some questions, and to add validation for when the employee's adoption leave can start (which is different depending on if the child is being adopted from the UK or abroad).

## Screenshot

adoption-abroad

### DIFF
--- a/lib/smart_answer_flows/maternity-paternity-calculator/adoption_calculator_flow.rb
+++ b/lib/smart_answer_flows/maternity-paternity-calculator/adoption_calculator_flow.rb
@@ -51,7 +51,12 @@ module SmartAnswer
           end
 
           calculate :a_leave_earliest_start do
-            adoption_placement_date - 14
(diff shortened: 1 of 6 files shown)


alecgibson:
[Adoption leave calculator - new question needed for validation](https://trello.com/c/Ufnu2oKt/781-adoption-leave-calculator-new-question-needed-for-validation)

h-lame:
I'm going to close this PR and manually put the commits from this branch onto the branch for the other PR (#3259) even though they're not passing yet.  I'll deal with the problems in that other PR.  We need to do both these bits of work as one is not useful without the other so having two PRs gives us more things to do.

PR #3269, opened 2017-11-23
Improve DateRange negative case handling

mgrassotti:
Trello: https://trello.com/c/YVhH1JFx/780-3-improve-daterange-corner-case-handling

If the end date is initialised to a date before the begins date,
`DateRange` silently sets it to the day before the begins date. This was
probably done to make it return 0 number of days. This is a strange
behaviour and could lead to unexpected results.

Instead of automatically set the end date, this PR allows the end date to
be before the begins date and makes DateRange return the exact amount of
negative number of days. The `empty?` method is also set to return
true in a case of negative or zero number of days.

This avoids to break the other related calculations (e.g.
`gap_between`) and doesn’t make changes to the end date initially set.

### DIFF
--- a/lib/smart_answer/date_range.rb
+++ b/lib/smart_answer/date_range.rb
@@ -25,7 +25,6 @@ module SmartAnswer
     def initialize(begins_on: nil, ends_on: nil)
       @begins_on = (begins_on || EARLIEST_DATE).to_date
       @ends_on = (ends_on || LATEST_DATE).to_date
-      @ends_on = [@begins_on - 1, @ends_on].max unless infinite?
     end
 
     def include?(date)
@@ -72,7 +71,7 @@ module SmartAnswer
     end
 
     def empty?
-      number_of_days == 0
+      number_of_days <= 0
     end
 
     def begins_before?(other)
--- a/test/unit/date_range_test.rb
+++ b/test/unit/date_range_test.rb
@@ -253,6 +253,10 @@ module SmartAnswer
       should 'be empty' do
         assert @date_range.empty?
       end
+
+      should 'set the ends_on date to the date before the begins_on date' do
+        assert_equal Date.parse('2000-01-01'), @date_range.ends_on
+      end
     end
 
     context 'intersection of' do
--- a/test/unit/date_range_test.rb
+++ b/test/unit/date_range_test.rb
@@ -246,15 +246,15 @@ module SmartAnswer
         refute @date_range.include?(@date_range.ends_on)
       end
 
-      should 'contain zero days' do
-        assert_equal 0, @date_range.number_of_days
+      should 'contain a non positive amount days' do
+        refute @date_range.number_of_days.positive?
       end
 
       should 'be empty' do
         assert @date_range.empty?
       end
 
-      should 'set the ends_on date to the date before the begins_on date' do
+      should 'not modify the ends_on date' do
         assert_equal Date.parse('2000-01-01'), @date_range.ends_on
       end
     end


lib/smart_answer/date_range.rb
```diff
@@ -72,7 +71,7 @@ def infinite?
     end
 
     def empty?
```
andrewgarner: I'm not sure I agree with this logic. Days fall between the two dates so it is wrong to say the range is empty.

mgrassotti: @andrewgarner sorry I don't understand, which days fall between the 2 dates?

andrewgarner: I was thinking in regards to ends on being before begins on. However, if were to use [Ruby's Range](https://ruby-doc.org/core-2.2.0/Range.html) library then it would be correct to say to is empty.

```
>> require 'date'
true
>> today = Date.today
#<Date: 2017-11-27 ((2458085j,0s,0n),+0s,2299161j)>
>> october = Date.new(2017, 10, 1)
#<Date: 2017-10-01 ((2458028j,0s,0n),+0s,2299161j)>
>> (october..today).to_a
[
    [ 0] #<Date: 2017-10-01 ((2458028j,0s,0n),+0s,2299161j)>,
    [ 1] #<Date: 2017-10-02 ((2458029j,0s,0n),+0s,2299161j)>,
    [ 2] #<Date: 2017-10-03 ((2458030j,0s,0n),+0s,2299161j)>,
    [ 3] #<Date: 2017-10-04 ((2458031j,0s,0n),+0s,2299161j)>,
    # ...
    [57] #<Date: 2017-11-27 ((2458085j,0s,0n),+0s,2299161j)>
]
>> (today..october).to_a
[]
```

Perhaps we should delegate `empty?` to a range?

mgrassotti: @andrewgarner but, if I understood well, when you define the range as `(october..today)` you are defining a different range then `(today..october)`. The first is not empty while the lattest is empty. So it should be coherent also with our implementation, isn't it?

chao-xian:
@mgrassotti from conversation earlier with @andrewgarner to help clarify:


> Thinking was that in Ruby there is a `Range` type
> so `[1...10]` is a range of numbers.
> so if we're going to have a `DateRange` it should act like a specialist type of `Range`.
> so if you instantiate a new `DateRange`, internally it would hold those dates as a `Range` and you could delegate `empty?` to it.
> It feels like we should have a class that is more specific if we have rules about start date succeeding the end date etc

mgrassotti:
@chao-xian what about thinking about this in another PR? It feels to me like a refactoring that falls outside this PR objectives.

chao-xian:
@mgrassotti I think we should have a test that covers the situation described in the Trello card: 
> If you try to set an end date before the start date, it sets your end date to the day before your start date, which is _still_ before your start date, but not the date you tried to set.

pmanrubia:
Would it be possible to add a few test to exercise all the edge cases and make them pass (first commit), then in upcoming commits refactor it so it is easier to read?

mgrassotti:
@chao-xian done 👍

mgrassotti:
@pmanrubia I tried to do that, let me know if that could be ok. Thanks

lib/smart_answer/date_range.rb
```diff
@@ -72,7 +71,7 @@ def infinite?
...
-      number_of_days == 0
+      number_of_days <= 0
```
alecgibson: How about `number_of_days.positive?` ?

mgrassotti: 👍

mgrassotti: @alecgibson or did you mean `!number_of_days.positive?`?

PR #3270, opened 2017-11-27
Consistently format dates in ERB

mgrassotti:
Trello: https://trello.com/c/cVp6ntVj/736-3-consistently-format-dates-in-erb

- Apply the default standard format ”d Month YYYY” for dates across all
smart-answers, where not differently specified.
- Define all formats in `Date::DATE_FORMATS`
- Uniform the date formatting across all smart answers
- Update tests, artefacts and checksums


config/initializers/date_formats.rb
```diff
@@ -0,0 +1,6 @@
+#Initialise default Gov UK date formats
```
andrewgarner: It might be worth referencing the style guide in the comment too for some additional context. Also we should say "GOV.UK" :)

lib/smart_answer/formatting_helper.rb
```diff
@@ -16,9 +16,8 @@ def format_salary(salary)
       number_to_currency(salary.amount, precision: 0) + " per " + salary.period
     end
 
-    def format_date(date)
-      return nil unless date
-      date.strftime('%e %B %Y')
+    def format_date(date, date_format = :govuk)
```
andrewgarner: I'm not sure we really need this helper.

We have configured `Date::DATE_FORMATS[:govuk]` so that we are not duplicating our format strings.

`to_s(format = :default)` is essentially the same as `format_date(date, date_format = :govuk)`

By creating a `format_date` method we've encapsulated that the default date format should be `:govuk`, but I'm not sure that really gives us any value because we still have to remember to use the `format_date` method which is no different to having to remember to use `to_s(:govuk)`.

I'd favour using `to_s(:govuk)` over the extra layer of encapsulation.

test/artefacts/maternity-paternity-calculator/adoption/maternity/2018-04-05/2018-04-06/yes/no/yes/2018-04-06/2018-01-01/2017-11-01/every_2_weeks/3000.0/4/usual_paydates/2018-01-01.txt
```diff
@@ -6,8 +6,8 @@ The employee is entitled to Statutory Adoption Leave.
 
 |Leave                         | Key dates                    |
 |------------------------------|------------------------------|
-|Start                         |  6 April 2018          |
-|End                           |  4 April 2019            |
+|Start                         | 6 April 2018          |
```
andrewgarner: In this file we've got two approached to markdown tables:

formatted (broken)

```
|Leave                         | Key dates                    |
|------------------------------|------------------------------|
|Start                         | 6 April 2018          |
|End                           | 4 April 2019            |
|Latest date to [claim leave](/employers-paternity-pay-leave/)     |   |
```

and inline

```
Date | SAP amount
-|-
9 April 2018|£337.50
23 April 2018|£675.00
```

Maybe we should fix the formatted approach here, to keep the PR focused, but maybe consider using the inline syntax so we don't litter PRs with whitespace changes?

```
| Leave                                                         | Key dates    |
| ------------------------------------------------------------- | ------------ |
| Start                                                         | 6 April 2018 |
| End                                                           | 4 April 2019 |
| Latest date to [claim leave](/employers-paternity-pay-leave/) |              |
```

See https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#tables

lib/smart_answer/formatting_helper.rb
mgrassotti: @andrewgarner yes you are right, but I tried the other approach and if we use `to_s(:govuk)` in the code, then we should manage somewhere else the `nil` case:
```
date = nil
date.to_s(:govuk)
=>  error: 1 parameter instead of 0 given to `to_s` method 
```
Apart from this I feel that `format_date` is easier to write and remember than `to_s(:govuk)`.
But let me know if you are still for the latter approach. Thanks.

andrewgarner: Hmm, I feel like that is telling us we should be handling the `nil` case correctly elsewhere though.

test/artefacts/maternity-paternity-calculator/adoption/maternity/2018-04-05/2018-04-06/yes/no/yes/2018-04-06/2018-01-01/2017-11-01/every_2_weeks/3000.0/4/usual_paydates/2018-01-01.txt
mgrassotti: @andrewgarner regarding date formatting, the PR should be ok now. What about fixing tables in another PR? I am not sure if you are proposing that too. Thanks

config/initializers/date_formats.rb
```diff
@@ -0,0 +1,14 @@
+# Initialise default GOV.UK date formats
+# https://www.gov.uk/guidance/style-guide/a-to-z-of-gov-uk-style#dates
+Date::DATE_FORMATS[:govuk] = "%-e %B %Y"
+Date::DATE_FORMATS[:day_month] = "%-e %b %Y"
+Date::DATE_FORMATS[:weekday_name] = "%A, %-e %B %Y"
+Date::DATE_FORMATS[:short] = "%-e %b %Y"
+Date::DATE_FORMATS[:db] = "%Y-%m-%d"
+
+# Manage nil case
+class NilClass
```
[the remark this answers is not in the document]
mgrassotti: @chao-xian I tried to do that for each single case, but then it turned out changing every calculator potentially producing a nil date. I think this change would have a smaller impact on the code, also it doesn't change much the `to_s` method of the Nil class, it just let it being more flexible. What do you think?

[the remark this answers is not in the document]
chao-xian: Monkey patching is not an option though, and definitely not recommend in GOV.UK (there was a discussion on this a few months ago in a dev meetup). 

It's ok to be explicit in each calculator. We should try to deal with `nil` before we try to format it really.

mgrassotti:
@chao-xian I removed the patching. Now the tests are passing, it seems quite odd though as they were not working before... I'll do other checks

PR #3276, opened 2017-12-04
Upgrade to Ruby 2.4.2 & Bundle update

danielburnley:


tijmenb:
@pmanrubia to review with a content designer to make sure that the rounding changes are correct.

chao-xian:
[(1) Check that rounding up from down of pennies is ok post Ruby upgrade](https://trello.com/c/pldzApVg/791-1-check-that-rounding-up-from-down-of-pennies-is-ok-post-ruby-upgrade)

chao-xian:
Update: the figures affected by the rounding changes have been confirmed as still correct by the department,

test/data/am-i-getting-minimum-wage-files.yml
```diff
@@ -21,7 +21,7 @@ lib/smart_answer_flows/am-i-getting-minimum-wage/questions/how_often_did_you_get
 lib/smart_answer_flows/am-i-getting-minimum-wage/questions/how_often_do_you_get_paid.govspeak.erb: 7c8cfebb76f688bfb03c3625cd413a24
 lib/smart_answer_flows/am-i-getting-minimum-wage/questions/how_old_are_you.govspeak.erb: 9d71f5503c82bb150961f8537453cce5
 lib/smart_answer_flows/am-i-getting-minimum-wage/questions/how_old_were_you.govspeak.erb: bfd909f04f6d523614226a0f9b8e7ea5
-lib/smart_answer_flows/am-i-getting-minimum-wage/questions/is_provided_with_accommodation.govspeak.erb: 083f1e522ae617d0aacf6851479ef28c
+lib/smart_answer_flows/am-i-getting-minimum-wage/questions/is_provided_with_accommodation.govspeak.erb: '083f1e522ae617d0aacf6851479ef28c'
```
suzannehamilton: Did the checksum update script add these quote marks? Is it anything to do with the Ruby or dependency updates?

I'm just wondering if we can remove the change to this line, or if we should prepare for more churn in these files.

suzannehamilton: I've reproduced this issue. We're not sure what's causing it, but it doesn't affect the tests so I'm going to approve this PR.

(All the affected checksums start with a zero, so our best guess is that it's something to do with avoiding [octal parsing](http://girders.org/blog/2009/08/13/ruby-yaml-loads-leading-zero-numbers-as-octal/). But some checksums that start with a zero haven't been updated, which is odd.)

This will probably add more churn, but the checksums already have a lot of churn and we'd like to improve them (or get rid of them) at some point anyway.

PR #3255, opened 2017-11-07
Retire student finance forms

mgrassotti:
Trello: https://trello.com/c/EnvfuMBh/754-3-retire-student-finance-forms-ie-firebreak-task
Retire **Student Finance Forms** Smart Answer and change its format to a _Simple Smart Answer_.

After deleting its files from smart-answers ([guide](https://github.com/alphagov/smart-answers/blob/master/doc/retiring-a-smart-answer.md)), the following steps have to be followed to apply the changes to the other applications and on **staging/production environments**:

**Retire the smart-answer via rake task**
- go to Jenkins: https://deploy.integration.publishing.service.gov.uk/job/run-rake-task/build?delay=0sec
- set & run:
  ``` 
  TARGET_APPLICATION = smartanswers
  MACHINE_CLASS = calculators_frontend
  RAKE_TASK = <rake task name only (no bundle exec rake) …. > 
  ```
  ```
  (bundle exec rake) retire:unpublish[67764435-e8ed-4700-a657-2e0432cb1f5b]
  (bundle exec rake) retire:change_owning_application[/student-finance-forms,publisher]
  ```
Verify success on Publishing API with:
  - SSH to backend: ```ssh publishing-api-1.backend.(integration|staging|production)```
  - rails console: ```govuk_app_console publishing-api```
  - check publishing_app: "publisher" (no more smart-answers) with `PathReservation.find_by(base_path: '/student-finance-forms')`

**Publish the simple smart answer document with the format manually**
  - Manually archive existing Artefact on publisher storage.
    - SSH to backend: ```ssh backend-1.backend.(integration|staging|production)```
    - rails console: ```govuk_app_console publisher```
    ``` 
    artefact = Artefact.find_by(slug: "student-finance-forms")
    artefact.state = "archived"
    artefact.save!
    artefact.slug = "archived-student-finance-forms"
    artefact.save!
    ```
  - Publish new simple smart answer for student-finance-forms from publisher
  - Update the retirement log ([link](https://github.com/alphagov/smart-answers/blob/master/doc/retirement-log.md))

** Remove the existing backend route
  - SSH to router backend ```ssh publishing-api-1.backend.(integration|staging|production)```
  - Rails console ```govuk_app_console router-api```
  - Delete the route
    ```
    route = Route.find_by(incoming_path: "/student-finance-forms/y")
    route.soft_delete
    RouterReloader.reload
    ```

This was successfully tested on integration.

PR #3279, opened 2017-12-05
Use the unpublish type of vanish for some retiring

cbaines:
When retiring a smart answer, and replacing it with a simple smart
answer, use an unpublishing type of "vanish" for unpublishing the
smart answer.

Previously, unpublishing the smart answer with a type of "gone" would
leave the route in place for the smart answer, and being a more
specific route than the simple smart answer, this would take
precedence.

Until recently, unpublishing with a type of vanish wouldn't have
removed the route from the router, but this should now happen.

### Extra background

I started looking at this when I was on 2ndline, as there was a problem with the display of the Student Finance Forms simple smart answer on the draft stack. As described in https://github.com/alphagov/smart-answers/pull/3255 , the route in the router for the smart answer had been manually deleted from the live router, but still existed on the draft router, which meant that parts of the simple smart answer couldn't be seen.

There is a somewhat unwritten and poorly followed architectural principle with the publishing platform, that the only app to change the state of the content stores should be the Publishing API, and the only app to change the state in the Router/Router API should be the Content Store (with data from the Publishing API). I think this is a good direction to move in, as it centralises and simplifies the state, holding it all in the Publishing API, and sending it out to where its needed.

Following on from this, manually deleting routes from the Router API isn't ideal. Even more unfortunate was that at the time, this was most probably the best practical approach, as up until recently, there was no way of deleting routes from the Router API, from the Publishing API.

However, this is something that should now be supported, as the Content Store should now delete routes from the Router API, when the corresponding content is removed [1]. The Publishing API removes content from the Content Store when you unpublish with a type of "vanish", which brings me back to this change. To replace a smart answer with a simple smart answer, if the unpublishing type of "vanish" is used, this should work without requiring any manual intervention in the Router API.

1: https://github.com/alphagov/content-store/pull/329

### DIFF
--- a/app/services/content_item_publisher.rb
+++ b/app/services/content_item_publisher.rb
@@ -21,6 +21,12 @@ class ContentItemPublisher
     )
   end
 
+  def unpublish_with_vanish(content_id)
+    raise "Content id has not been supplied" unless content_id.present?
+
+    Services.publishing_api.unpublish(content_id, type: "vanish")
+  end
+
   def publish_redirect(path, destination)
     raise "The destination or path isn't defined" unless path.present? && destination.present?
 
--- a/doc/retiring-a-smart-answer.md
+++ b/doc/retiring-a-smart-answer.md
@@ -202,7 +202,7 @@ In this document is prescribed the steps that need to be taken:
     RAKE_TASK = <rake task name only (no bundle exec rake) …. > 
     ```
     ```
-    (bundle exec rake) retire:unpublish[67764435-e8ed-4700-a657-2e0432cb1f5b]
+    (bundle exec rake) retire:unpublish_with_vanish[67764435-e8ed-4700-a657-2e0432cb1f5b]
     (bundle exec rake) retire:change_owning_application[/student-finance-forms,publisher]
     ```
   Verify success on Publishing API with:
--- a/lib/tasks/retire.rake
+++ b/lib/tasks/retire.rake
@@ -18,6 +18,13 @@ namespace :retire do
     ContentItemPublisher.new.unpublish(args.content_id)
   end
 
+  desc "Unpublish a smart answer from publishing-api, with an unpublish type of vanish"
+  task :unpublish_with_vanish, [:content_id] => :environment do |_, args|
+    raise "Missing content-id parameter" unless args.content_id
+
+    ContentItemPublisher.new.unpublish_with_vanish(args.content_id)
+  end
+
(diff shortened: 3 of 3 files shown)


cbaines:
Thanks for reviewing Mat :)

> If vanishing stuff is a common thing to do can we move this rake task to the publishing API itself?

Unpublishing things is common, but the common way to do this is via a publishing app, not as a developer task. I see the Smart Answers use case as a bit of an exception.

> I think we could do with better documentation around unpublishing as I don't think everyone knows what is good and bad practise.

I'm quite open to trying to write some, but I'm not quite sure what the areas of uncertainty are?

> We have a manual page on manually removing stuff from the router so if this is no longer necessary we should get rid of/update it.

Good spot. In my opinion, it shouldn't be necessary for normal operation, but there is at least a couple of incidents that I've been involved in where it was helpful to make route changes manually in the router, before putting a more permanent fix in place. I've put up a Pull Request [1] to try and clarify when it is appropriate to make manual changes, and when not to.

1: https://github.com/alphagov/govuk-developer-docs/pull/640

MatMoore:
> I'm quite open to trying to write some, but I'm not quite sure what the areas of uncertainty are?

Cool 👍 Let me know if I can help at all.

I was just thinking of this:

> the only app to change the state of the content stores should be the Publishing API, and the only app to change the state in the Router/Router API should be the Content Store

Is this spelled out anywhere? It's what I'd generally expect but I've also manually changed data in the router before when working on support tickets, and if this is no longer required we definitely shouldn't be doing it. I don't think anyone should break this rule unless it's part of a temporary measure to fix something during an incident.

It seems like there are really a lot of scenarios involving unpublishing. It's not always immediately obvious what to do when unpublishing stuff as part of development work, or when I'm providing support to publishers who are trying to unpublish stuff themselves. I have enough context to muddle through it, but I wouldn't say I have a good understanding of how to handle this kind of thing.

Maybe we could write something that explains:

- what is unpublishing, and why do we do it
- how to reverse an unpublishing
- what kinds of things are hard deleted or appear that way to publishers
- what content items should never be unpublished (if anything)
- why and how we (usually) preserve URLs when content is unpublished
- when we make exceptions to that rule
- what each unpublishing type means and when to use it

(assuming this doesn't already exist somewhere)

cbaines:
> Can this be merged now?

Yeah, sure, I've just tested it locally, and it worked.

I've still got Mat's questions on my TODO list, so I might get around to doing something eventually...

I'll merge this now.

PR #3280, opened 2017-12-06
ADR for rummager integration via Publishing API

dwhenry:
https://trello.com/c/tSPsLjz3/489-make-smartanswers-indexable

### DIFF
--- a/doc/arch/adr-002-rummager-integration-via-publishing-api.md
+++ b/doc/arch/adr-002-rummager-integration-via-publishing-api.md
@@ -1,0 +1,45 @@
+# ADR 2: Rummager integration via Publishing API
+
+## Context
+
+We are currently moving to use the publishing API as the source of truth for data in our search index - we had each application integrate directly with rummager - this will help standardise how data is inserted into our index and simplify the reindexing process.
+
+Smart answers currently sends a single document to rummager per a smart answer which is indexed under the smart page URL. This document contains all the text from all the nodes (questions and outcomes) associated with the smart answer.
+
+In addition to this smart answers sends two documents to the publishing API for each smart answer:
+
+1. A document representing the start page, this document only contains the text present on the start page. This has a document type of `transaction`
+1. A document representing the flow, this has the URL for the flow base page and does not contain any content. This has a document type of `smart-answer`
+
+## Options
+
+In order to move to indexing via rummager we will need to change what data is sent to the publishing API, the following options have been considered:
+
+1. Add the content for each node to a new array item on the document schema at `details -> nodes` for the start page.
+
+  This has the negative side effect of adding the new item to all `transaction` dcument types - this document type is also used by the `publisher` application. It has also been suggested that "something feels wrong about adding context-specific data to a schema because search happens to use it".
+
+1. Change the document type for the start page then add the new array item to the new content schema.
+
+  This would avoid the issue posed by multiple applications using the same document type, however this is potentially a large piece of work.
+
+1. Add the content to any existing document schema item at `hidden_indeaxable_content` for the flow base page.
+
+  This would increase the system complexity as it would require rummager to merge the two records from the publishing API into a single item in the rummager index, this functionality is not currently supported and would require a reworking of the queue based processing.
+
+  This raises the question about the `hidden_indexable_content` field which is something that we want to remove in the future as it is considered a hack which is primarily used by `specialist_publisher` to store PDF content as we are not currently able to extract it from the PDF documents.
+
+1. Add the content to any existing document schema item at `hidden_indeaxable_content` for the start page.
+
+  As stated above we want to remove this field long term.
+
+  This is consider the easiest option to implement.
+
+
+## Decision
+
+
+## Status
+
+
+## Consequences


tijmenb:
I don't think `hidden_indexable_content` is a hack, or at least not a bad one. I'm pretty sure it's reasonable for pages to have extra metadata only used for search purposes like that, demonstrated by this issue. I'm for Option No. 4 all the way.

doc/arch/adr-002-rummager-integration-via-publishing-api.md
```diff
@@ -0,0 +1,45 @@
...
+
+  This would increase the system complexity as it would require rummager to merge the two records from the publishing API into a single item in the rummager index, this functionality is not currently supported and would require a reworking of the queue based processing.
+
+  This raises the question about the `hidden_indexable_content` field which is something that we want to remove in the future as it is considered a hack which is primarily used by `specialist_publisher` to store PDF content as we are not currently able to extract it from the PDF documents.
+
+1. Add the content to any existing document schema item at `hidden_indeaxable_content` for the start page.
+
+  As stated above we want to remove this field long term.
+
+  This is consider the easiest option to implement.
```
MatMoore: I agree with this, it seems like the most pragmatic option for the short term out of the ones listed. However I feel that `hidden_indexable_content` doesn't go far enough to separate data from presentation, so this just clutters up the API with fields that cannot be reused for anything else.

I think it's a bad idea to couple individual fields in the content store/API to rummager fields. The stuff that smart answers would call `hidden_indexable_content` feels different from the content specialist publisher considers `hidden_indexable_content`, and both are very different from the real `indexable_content`, which is content that is immediately available on the page without clicking through to stuff. It's also possible that the way we represent this information in rummager could change in the future in order to better balance the influences on search results.

doc/arch/adr-002-rummager-integration-via-publishing-api.md
```diff
@@ -0,0 +1,45 @@
...
+
+
```
MatMoore: Another alternative approach could be to strip this down to a set of curated keywords or search terms, which is what I originally suggested to @brenetic for https://github.com/alphagov/govuk-content-schemas/pull/672

This is something that would be more generally useful, as knowing keywords that the content is supposed to target enable us to monitor internal and external search performance for those queries.

doc/arch/adr-002-rummager-integration-via-publishing-api.md
```diff
@@ -0,0 +1,45 @@
...
+1. Change the document type for the start page then add the new array item to the new content schema.
+
+  This would avoid the issue posed by multiple applications using the same document type, however this is potentially a large piece of work.
+
+1. Add the content to any existing document schema item at `hidden_indeaxable_content` for the flow base page.
+
+  This would increase the system complexity as it would require rummager to merge the two records from the publishing API into a single item in the rummager index, this functionality is not currently supported and would require a reworking of the queue based processing.
+
+  This raises the question about the `hidden_indexable_content` field which is something that we want to remove in the future as it is considered a hack which is primarily used by `specialist_publisher` to store PDF content as we are not currently able to extract it from the PDF documents.
```
MatMoore: It's a hack in the sense that users have to copy and paste PDF contents into a text box, instead of rummager properly processing attachments (with something like [Tika](https://tika.apache.org/))

doc/arch/adr-002-rummager-integration-via-publishing-api.md
```diff
@@ -0,0 +1,45 @@
...
+1. A document representing the start page, this document only contains the text present on the start page. This has a document type of `transaction`
+1. A document representing the flow, this has the URL for the flow base page and does not contain any content. This has a document type of `smart-answer`
+
+## Options
+
+In order to move to indexing via rummager we will need to change what data is sent to the publishing API, the following options have been considered:
+
+1. Add the content for each node to a new array item on the document schema at `details -> nodes` for the start page.
+
+  This has the negative side effect of adding the new item to all `transaction` dcument types - this document type is also used by the `publisher` application. It has also been suggested that "something feels wrong about adding context-specific data to a schema because search happens to use it".
```
MatMoore: Another idea is to introduce a different schema instead of a different document type.

Schemas and document types are independent of each other, so you could capture the fact that these pages are functionally start pages, but there is smart answers specific data associated with them.

doc/arch/adr-002-rummager-integration-via-publishing-api.md
```diff
@@ -0,0 +1,45 @@
...
+## Context
+
+We are currently moving to use the publishing API as the source of truth for data in our search index - we had each application integrate directly with rummager - this will help standardise how data is inserted into our index and simplify the reindexing process.
+
+Smart answers currently sends a single document to rummager per a smart answer which is indexed under the smart page URL. This document contains all the text from all the nodes (questions and outcomes) associated with the smart answer.
+
+In addition to this smart answers sends two documents to the publishing API for each smart answer:
+
+1. A document representing the start page, this document only contains the text present on the start page. This has a document type of `transaction`
+1. A document representing the flow, this has the URL for the flow base page and does not contain any content. This has a document type of `smart-answer`
```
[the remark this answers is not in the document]
tijmenb: The `flow` content item represents the actual smart answer. It takes care of the routes (the `start` item points to a different frontend app). I think it makes sense for it to be entirely separate from the start page.

doc/arch/adr-002-rummager-integration-via-publishing-api.md
tijmenb: I think it's a lot better to create a field that explicitly contains the extra stuff for rummager than to try to put this content into a schema structure that can't actually be used. I understand the desire not to couple, but I think it's probably unavoidable for publishing apps to send search-specific data to publishing-api. The alternative is to clutter the system with schema / schema changes that secretly are rummager-only, but look generic (https://github.com/alphagov/govuk-content-schemas/pull/677 is another example).

doc/arch/adr-002-rummager-integration-via-publishing-api.md
tijmenb: Hmm, I missed this comment before https://github.com/alphagov/smart-answers/pull/3280#discussion_r155308061. This seems really close to `hidden_indexable_content` (so I like it), how does it differ?

pmanrubia:
I don't have the whole picture of the architecture around search, but with 
my limited experience, I would feel comfortable having indexable metadata 
attached to a content item so it is indexed by search. 
This seems like a feature to me. 

I am not sure either, why PDF content is that different to all the nodes of a 
smart-answer's flow. To me, all looks like part of the same feature: 
`indexable content that is not presented to the user`. 

I think @tijmen [was spot on in this comment][1]. 

I am also quite aligned with the PR from @brenetic. The name describes quite well its purpose.

[1]: https://github.com/alphagov/smart-answers/pull/3280#discussion_r155308061
[2]: https://github.com/alphagov/govuk-content-schemas/pull/672/files

pmanrubia:
@dwhenry @tijmenb @MatMoore:
Not sure if I am the right person to make a decision here; this ADR could affect other apps and has architectural implications.
Shouldn't this live in Search or govuk-schemas? Or even a proper RFC?

MatMoore:
@dwhenry can you update this PR with the approach we've taken? (assuming we still want the ADR to live in this repo)

tijmen:
My name is indeed Tijmen and I do work on ADRs (adiabatic demagnetization refrigerators) but I'm pretty sure you've got the wrong guy :)

pmanrubia:
@dwhenry: happy to approve the ADR if you rework it with the approach that has been decided.

dwhenry:
This was super helpful for the discussion and decision, but ultimately I don't think we need to keep this longer term.

PR #3283, opened 2017-12-12
Send flow content to publishing API

dwhenry:
This content has been added to the start page content item as that allows us to
index all the content fropm the flows back to the start page

https://trello.com/c/tSPsLjz3/489-make-smartanswers-indexable

### DIFF
--- a/app/presenters/flow_registration_presenter.rb
+++ b/app/presenters/flow_registration_presenter.rb
@@ -57,6 +57,28 @@ class FlowRegistrationPresenter
     end
   end
 
+  def flows_content
+    content = @flow.nodes.flat_map do |node|
+      case node
+      when SmartAnswer::Question::Base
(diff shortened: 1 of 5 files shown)


app/presenters/flow_registration_presenter.rb
```diff
@@ -57,6 +57,28 @@ def method_missing(method, *_args, &_block)
...
+      when SmartAnswer::Outcome
+        pres = OutcomePresenter.new(node, nil, helpers: [MethodMissingHelper])
+        [pres.title, pres.body]
+      end
+    end
+
+    content
+      .compact
+      .map do |html|
+        HTMLEntities.new
```
MatMoore: Having to strip html out here doesn't seem great, but I don't have a better suggestion, and this'll work for search.

dwhenry: I agree, but think that makes more sense than sending this as html.

PR #3284, opened 2017-12-13
Remove rummager integration

dwhenry:
https://trello.com/c/6YLtQotp/517-remove-rummager-code-from-smart-answers


lib/tasks/rummager.rake
```diff
@@ -1,7 +0,0 @@
-namespace :rummager do
-  desc "Indexes all smart answers in Rummager"
-  task index: :environment do
```
suzannehamilton: This task is mentioned in the readme, so we should update that too.

PR #3295, opened 2017-12-15
[WIP] Link to partial templates on GitHub

chao-xian:
## Reason

This is to progress a spike that was done in [this branch](https://github.com/alphagov/smart-answers/tree/link-to-partial-templates-on-github).

To help content designers to easily find the partials that contain the Govspeak and content that they need to edit. Smart Answers has a lot of templates and uses partials in a large number of them. When content designers want to edit content that is hidden from them in partials, they don't and shouldn't need to have knowledge of the structure of the codebase to find these partials.

This PR adds markup around the relevant partials with the path to the files in Github. Working in conjunction with the GOV.UK Chrome Toolkit extension. A PR for that adds a link to activate highlighting around the marked up partials for a page, and a link to Github.

We will only enable this feature on integration, as we do *not* want to inject debug markup into production. This will be toggled by an environment variable.

## Depends on

- [x] https://github.com/alphagov/govuk-puppet/pull/6973

## After
 
image


chao-xian:
[(8) Provide a way for Content Designers to navigate from content to *partial* templates on GitHub](https://trello.com/c/jeyLKQLQ/785-8-provide-a-way-for-content-designers-to-navigate-from-content-to-partial-templates-on-github)

config/initializers/debug_partial_template_paths.rb
```diff
@@ -0,0 +1,11 @@
+if Rails.env.production?
```
pmanrubia: Why do we need to check if the environment is production? Shouldn't it be enough with the environment variable.

chao-xian: This is so that we can set the config value to true by default on non-production environments.

chao-xian: @pmanrubia addressed in https://github.com/alphagov/smart-answers/pull/3295/commits/b04513ed4e9392b294f8911f8c95225a2a96a548

pmanrubia:
Thank you for your great work on this 👍

pmanrubia:
It's been quite a long time since this PR was implemented, I no longer feel comfortable with merging it. I would recommend closing it so it can be re-opened in the future if the user need is still there.

cc/ @chao-xian

PR #3341, opened 2018-01-17
Update continuous integration docs

suzannehamilton:
Rewrite the CI docs to explain the current configuration, which uses Jenkins pipeline builds and a `Jenkinsfile` rather than separate builds for master, branches and regression tests.

### DIFF
--- a/doc/deploying.md
+++ b/doc/deploying.md
@@ -2,17 +2,15 @@
 
 ## Integration
 
-This happens automatically as a post build action when the [govuk_smart_answers](https://ci.dev.publishing.service.gov.uk/job/govuk_smart_answers/) build passes on CI.
+This happens automatically as a post build action when the [govuk_smart_answers master build](https://ci.integration.publishing.service.gov.uk/job/smartanswers/job/master/) passes on CI.
 
 ### Manually
 
-If you have sufficient permissions, you can use the [Integration Deploy App](https://deploy.integration.publishing.service.gov.uk/job/Deploy_App/). Click the "Build with Parameters" link and enter "release" as the "TAG".
-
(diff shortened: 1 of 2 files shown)


doc/deploying.md
```diff
@@ -2,17 +2,17 @@
...
 
-This happens automatically as a post build action when the [govuk_smart_answers](https://ci.dev.publishing.service.gov.uk/job/govuk_smart_answers/) build passes on CI.
+This happens automatically as a post build action when the [govuk_smart_answers master build](https://ci.integration.publishing.service.gov.uk/job/smartanswers/job/master/) passes on CI.
 
 ### Manually
 
 If you have sufficient permissions, you can use the [Integration Deploy App](https://deploy.integration.publishing.service.gov.uk/job/Deploy_App/). Click the "Build with Parameters" link and enter "release" as the "TAG".
 
-If you do not have sufficient permissions, you can either kick the [main CI build](continuous-integration.md#main) in Jenkins or commit & push something to the `master` branch.
+If you do not have sufficient permissions, you can either kick the [master CI build](continuous-integration.md#master) in Jenkins or commit & push something to the `master` branch.
```
h-lame: I think we should stop suggesting that people make arbitrary commits to `master` just to trigger a build.  That doesn't seem like a good idea.

suzannehamilton: I did not read the whole of that sentence. 🤦‍♀️ 

Maybe we should remove this paragraph. Does anyone have permission to see CI-Jenkins but not deploy-Jenkins, and still needs to run deployments? Also, this was probably written before the `master` build was much faster because it didn't include the whole regression suite.

h-lame: Good point, these days I can't think of anyone who would be allowed to commit, but not see deploy or ci.  Delete!

suzannehamilton: Done. 👍

PR #3346, opened 2018-01-22
Cache the holiday data file in dev and test

dwhenry:
This shaves about 0.2sec off the load time, but more 
importantly it avoids this process timing out which 
we have seen multiple issues of when running tests/rake
tasks on the dev VM.

### DIFF
--- a/lib/working_days.rb
+++ b/lib/working_days.rb
@@ -37,15 +37,39 @@ private
     self.class.workday?(date)
   end
 
-  def self.bank_holidays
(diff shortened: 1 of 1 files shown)


cbaines:
How about removing the initializer that calls this code [1]? I'm assuming that would mean that it would be loaded lazily when needed.

1: https://github.com/alphagov/smart-answers/blob/3e784b05ad09d91af200b4dacb475f56777956a0/config/initializers/preload_working_days.rb

PR #3357, opened 2018-01-29
Upgrade govuk_app_config to 1.2.1

boffbowsh:
Including config changes for logstasher.

The bump to 0.2.0 enabled debug logging which is sending a lot of data to logit.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -60,12 +60,7 @@ Rails.application.configure do
 
   config.action_controller.asset_host = ENV['GOVUK_ASSET_HOST']
 
-  if ENV['RUNNING_ON_HEROKU'].blank?
-    # Enable JSON-style logging
-    config.logstasher.enabled = true
-    config.logstasher.logger = Logger.new("#{Rails.root}/log/#{Rails.env}.json.log")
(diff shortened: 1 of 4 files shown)


cbaines:
It's probably a good idea to remove https://github.com/alphagov/smart-answers/blob/master/config/initializers/logstasher.rb at the same time, as I believe this duplicates what the newer versions of govuk-app-config do.

boffbowsh:
Good catch, thanks 👍

config/environments/production.rb
```diff
@@ -60,12 +60,7 @@
 
   config.action_controller.asset_host = ENV['GOVUK_ASSET_HOST']
 
-  if ENV['RUNNING_ON_HEROKU'].blank?
-    # Enable JSON-style logging
-    config.logstasher.enabled = true
-    config.logstasher.logger = Logger.new("#{Rails.root}/log/#{Rails.env}.json.log")
-    config.logstasher.suppress_app_logs = true
-  else
+  if ENV['RUNNING_ON_HEROKU']
```
tijmenb: I think we might be able to skip this, because `govuk_app_config` will log to STDOUT by default?

PR #3370, opened 2018-02-07
Change the output of the artefact check test

h-lame:
The previous version suggested that you:

1. regenerate the checksums
2. run the test again
3. commit all the changes

However, as the test does a `git status` and checks for empty output you
actually have to commit the changes before you re-run the test, otherwise
the changes will still appear in the `git status` output.  So we change the
output of the test to suggest that you:

1. regenerate the checksums
2. commit all the changes
3. run the test again

### DIFF
--- a/test/regression/smart_answers_regression_test.rb
+++ b/test/regression/smart_answers_regression_test.rb
@@ -145,7 +145,7 @@ class SmartAnswersRegressionTest < ActionController::TestCase
 
       should "#{RUN_ME_LAST} and generate the same set of output files" do
         diff_output = `git status --short -- #{smart_answer_helper.path_to_outputs_for_flow}`
-        assert diff_output.blank?, "Changes in outcome page artefacts have been detected:\n#{diff_output}\nIf these changes are expected then re-generate the checksums, re-run the regression test, and commit all the changes."
+        assert diff_output.blank?, "Changes in outcome page artefacts have been detected:\n#{diff_output}\nIf these changes are expected then re-generate the checksums, commit all the changes, and re-run the regression test."
       end
     end
   end


h-lame:
Not 100% sure that this change reflects the intention of the test properly as it doesn't seem great that you have to commit a change before the test will pass.

chrisroos:
Yo @h-lame. I just tried this locally to try to remind myself of the intention. I modified an ERB template in the check-uk-visa smart-answer and ran the regression tests:

```
$ RUN_REGRESSION_TESTS=check-uk-visa \
  ruby test/regression/smart_answers_regression_test.rb

Failure:
SmartAnswersRegressionTest#test_: Smart Answer: check-uk-visa should have up to date checksum data.  [test/regression/smart_answers_regression_test.rb:92]:
Expected /Users/chrisroos/Code/alphagov/smart-answers/test/data/check-uk-visa-files.yml to exist and to contain up to date data. Use the checksums:add_files[check-uk-visa] rake task to update it.
Expected: false
  Actual: true

# Update the checksums as prompted
$ bundle exec rake checksums:add_files[check-uk-visa]

# Re-run the regression tests
$ RUN_REGRESSION_TESTS=check-uk-visa \
  ruby test/regression/smart_answers_regression_test.rb

Failure:
SmartAnswersRegressionTest#test_: Smart Answer: check-uk-visa should zzzzzzzzzzz run me last and generate the same set of output files.  [test/regression/smart_answers_regression_test.rb:148]:
Changes in outcome page artefacts have been detected:
 M test/artefacts/check-uk-visa/check-uk-visa.txt

If these changes are expected then re-generate the checksums, re-run the regression test, and commit all the changes.
```

It makes me wonder whether the message you're modifying in this branch was written before we added the test to verify that the checksums were up to date (i.e. the failure I saw before I was able to see the message you're modifying). And I wonder whether the order of the steps in the message is still correct but that it's confusing because you've probably already updated the checksums (step 1), and run the tests (step 2) by the time you see this message. Maybe the message could read, "If these changes are expected then ensure the checksums are up to date and commit the changes"?

I don't think I've done a very good job of explaining what I mean so please follow-up if you think I might be able to help any further.

h-lame:
@chrisroos / @floehopper thanks for taking a look.  Sorry for dragging you back to smart-answers.

I am pretty sure I did update the checksums first, but as far as I can tell the test won't pass until the changes are committed because of how it's implemented.  @chrisroos what did you have to do to make the test pass?  I am fairly sure you have to commit before the test will pass, but if that's not the case then I can abandon this PR.

chrisroos:
You're quite correct that you have to commit the updated artefacts in order to get the tests to pass, @h-lame. Apologies for any confusion.

I think the workflow we were using was something like:

1. Make a change to a smart answer
2. Run the tests
    * Update the checksum data if the tests indicate that they're out of date
    * Re-run the tests
3. Use the difference in the regression test artefacts to ensure a) nothing has changed, or b) they've changed as expected
4. Commit any changes from step 3.
5. Re-run the tests to ensure that they all pass after committing the changes in step 4.

Having typed that out, I now understand your change to the error message and so am approving this PR. 

Sorry it's taken so long to review such a small change!

h-lame:
Thanks @chrisroos !

PR #3425, opened 2018-03-15
Changing pensions

elenatanasoiu:
Changes requested can be found here: https://docs.google.com/document/d/1QZQrYyBEytxaU0npBjit3-nUJsKrhbKn34GZICTgSNw/edit

For: https://trello.com/c/SsZ8u5xF/110-2-state-pension-age-government-proposals

This PR does two things:

- moves 2 lines in the `not_yet_reached_sp_age.govspeak.erb` file that are under the
"Pension Credit" section to above the "Pension Credit section", as they are not actually
 related to that topic.
- adds a condition to notify only the people born between 06/04/1970 and 05/04/1978
that their pension age is going to increase

# Before
screen shot 2018-03-15 at 14 25 19

# After
screen shot 2018-03-15 at 14 25 27

### DIFF
--- a/test/unit/calculators/state_pension_age_calculator_test.rb
+++ b/test/unit/calculators/state_pension_age_calculator_test.rb
@@ -343,5 +343,22 @@ module SmartAnswer::Calculators
         end
       end
     end
+
+    context "#rolling_increase_period?" do
+      should "be true for a date that is between 06/04/1970 and 05/04/1978" do
+        @calculator = StatePensionAgeCalculator.new(dob: Date.parse("6 April 1970"))
+        assert_equal true, @calculator.rolling_increase_period?
+      end
+
+      should "be false for a date that is before 06/04/1970" do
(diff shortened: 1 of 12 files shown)


test/unit/calculators/state_pension_age_calculator_test.rb
```diff
@@ -343,5 +343,25 @@ class StatePensionAgeCalculatorTest < ActiveSupport::TestCase
         end
       end
     end
+  
+    context "#rolling_increase_period?" do
+      should "be true for a date that is between 06/04/1970 and 05/04/1978" do
+        @calculator = StatePensionAgeCalculator.new(
+          gender: "male", dob: Date.parse("6 April 1970"))
```
suzannehamilton: Does gender affect this calculation? It looks a bit misleading if it doesn't affect the outcome, so maybe we could we skip it? (Unless that causes an error.)

PR #3439, opened 2018-03-20
Remove breadcrumbs from pages

tijmenb:
The breadcrumbs were introduced in https://github.com/alphagov/smart-answers/pull/2805 (by me).

At this time this app also rendered the start pages like https://www.gov.uk/calculate-agricultural-holiday-entitlement. Those have since moved to government-frontend.

This PR removes the breadcrumbs from the pages because they don't add any information for the user. Smart answer pages are kind of transaction-y pages that don't need navigation elements because people need to focus on answering the questions.

Also removes the sidebar from the "landing page", but that's only visible in development. 

https://trello.com/c/4ffi5zAj


Gemfile
```diff
@@ -28,7 +28,6 @@ gem 'tilt', '2.0.8'
 gem 'uglifier'
 gem 'uk_postcode', '~> 2.1.2'
 gem 'rails_stdout_logging'
-gem 'govuk_navigation_helpers', '~> 6.3.0'
```
suzannehamilton: 👍

PR #3460, opened 2018-04-03
Added further options for tourists from outside EEA

gpclaridge:
I've added further options for tourists from non-EEA countries wishing to enter the UK for a short stay. This is a short-term fix: the text is doing too much heavy lifting at present, so I've raised a ticket to make changes to the flow. But at least the info we're giving is correct.

Here's an example of an outcome that'll change once these changes go live: https://www.gov.uk/check-uk-visa/y/russia/tourism


tlwr:
I'm closing this PR as it has been open for a considerable amount of time and we're trying to cut down the number of orphaned PRs in alphagov.

PR #3477, opened 2018-04-13
Fix date range in maternity/paternity calculator

rubenarakelyan:
In cases where the leave date was after the pay date in the same month, the pay was being calculated from the pay date rather than the leave date.

Trello: https://trello.com/c/GN11p1I7/61-urgent-maternity-adoption-and-paternity-calculator-for-employers


lib/smart_answer/calculators/maternity_paternity_calculator.rb
```diff
@@ -226,11 +226,14 @@ def total_statutory_pay
     def paydates_and_pay
       paydates = send(:"paydates_#{pay_method}")
       [].tap do |ary|
+        last_paydate = pay_start_date
         paydates.each_with_index do |paydate, index|
```
bevanloon: do we need `each_with_index` anymore? - it looks like we're no longer using `index`

rubenarakelyan: Nope, removed

PR #3497, opened 2018-04-26
Rubyify flattening script

steventux:
https://trello.com/c/JV9rjSnw/109-look-at-making-the-flattening-script-a-more-long-term-solution-2

Based on @cbaines 's script in https://github.com/alphagov/smart-answers/pull/3131 make flattening country outcomes easier with a rake task.

Tested with recent output for Ecuador generated by original bash script, produces the same output according to git.

### DIFF
--- a/lib/marriage_abroad_outcome_flattener.rb
+++ b/lib/marriage_abroad_outcome_flattener.rb
@@ -1,0 +1,90 @@
+class MarriageAbroadOutcomeFlattener
+  COUNTRIES_DIR = "lib/smart_answer_flows/marriage-abroad/outcomes/countries".freeze
+
+  FEE_TABLE_PARTIAL = <<~FEE_TABLE.freeze
+    <%= render partial: 'consular_fees_table_items.govspeak.erb',
+        collection: calculator.services,
+        as: :service,
+        locals: { calculator: calculator } %>
+  FEE_TABLE
+
+  HOW_TO_PAY_PARTIAL = <<~HOW_TO_PAY.freeze
+    <%= render partial: 'how_to_pay.govspeak.erb', locals: {calculator: calculator} %>
+  HOW_TO_PAY
+
+  TITLE_PARTIAL_CONTENT = <<~TITLE.freeze
+    <% if calculator.partner_is_same_sex? %>
+      Civil partnership in COUNTRY
+    <% else %>
+      Marriage in COUNTRY
+    <% end %>
+  TITLE
+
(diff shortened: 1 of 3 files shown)


lib/marriage_abroad_outcome_flattener.rb
```diff
@@ -0,0 +1,90 @@
...
+  TITLE
+
+  PAYMENT_RE = /^\^?You can( only)? pay by.*?\n.*?^$/mi
+  SERVICE_FEE_RE = /^Service \| Fee\n.*?^$/mi
+
+  attr_reader :country, :logger
+
+  def initialize(country, logger = STDOUT)
+    @country = country
+    @logger = logger
```
kevindew: Are you anticipating any other loggers than STDOUT? I notice the Rails one has `puts` as a private method and that seems to be only method we call on the logger.

lib/marriage_abroad_outcome_flattener.rb
```diff
@@ -0,0 +1,90 @@
...
+  # Strips the first two lines of the file (the title).
+  # Replaces payment instructions with a shared partial erb snippet.
+  # Replaces service fee with a fees table shared partial erb snippet.
+  def insert_payment_partials(filename)
+    lines = IO.readlines(filename)
+    text = lines[2..-1].join
+    text = text.gsub(PAYMENT_RE, HOW_TO_PAY_PARTIAL)
+    text = text.gsub(SERVICE_FEE_RE, FEE_TABLE_PARTIAL)
+
+    File.open(filename, "w") { |f| f.write(text) }
```
kevindew: Could we use `File.write(filename, text)` ?

lib/marriage_abroad_outcome_flattener.rb
```diff
@@ -0,0 +1,90 @@
...
+    remove_old_country_partials
+    copy_test_artefacts
+    rename_and_add_partials_to_country_files
+    create_title_partial
+  end
+
+private
+
+  def rename_and_add_partials_to_country_files
+    Dir["#{country_partials_dir}/**/*.txt"].each do |src_filename|
```
kevindew: Are the filenames pretty rigid here? It looks like this would all work good so long as the file was named `same_sex.txt` or `opposite_sex.txt`but looks like it could go a bit wacky (unless I've missed something) if the glob matched something else.

steventux:
Thanks @kevindew, very helpful. I think I've covered your comments now.

PR #3508, opened 2018-05-02
Strip PII from URLs

tijmenb:
This was accidentally removed in https://github.com/alphagov/smart-answers/pull/3506/files.

https://trello.com/c/2w1AKyuW

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -7,7 +7,10 @@
     <%= yield :head %>
 
     <% if @content_item %>
-      <%= render "govuk_publishing_components/components/meta_tags", content_item: @content_item %>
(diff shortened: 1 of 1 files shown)


vanitabarrett:
Taken a closer look at the component itself - I feel like https://github.com/alphagov/govuk_publishing_components/commit/a30650124bd2d859b4cb56f1255d864041b2a22d (and the corresponding methods for postcode) should have covered this scenario?

It seems to suggest that even if `strip_dates_pii` isn't passed, if the document is a smart_answer or search then this should apply by default?

tijmenb:
Found it: `@content_item` in smart answers is actually not the content item for the smart answer (https://www.gov.uk/api/content/marriage-abroad/y), but the start page (https://www.gov.uk/api/content/marriage-abroad).

PR #3536, opened 2018-05-30
Ethiopia marriage abroad

hannahwhittaker:
https://trello.com/c/GQNaPVoW/1613-getting-married-abroad-ethiopia-opposite-sex-text-full-re-write

I've updated the 9 opposite sex outcomes to remove information about getting a CNI as one is no longer required to get married in Ethiopia.


thomasleese:
Oops pressed the wrong button!

PR #3553, opened 2018-06-12
Update "Register a Birth Abroad" instructions for Venezuela

barrucadu:
[Trello card](https://trello.com/c/uGTG0j8A/224-update-register-a-birth-abroad-venezuela)

### DIFF
--- a/lib/smart_answer_flows/register-a-birth/outcomes/oru_result.govspeak.erb
+++ b/lib/smart_answer_flows/register-a-birth/outcomes/oru_result.govspeak.erb
@@ -215,13 +215,20 @@
 
   <% if calculator.country_of_birth == 'venezuela' && calculator.same_country? %>
 
-    ##4. Go to the British embassy
+    ##4. Send your registration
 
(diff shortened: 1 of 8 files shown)


lib/smart_answer_flows/register-a-birth/outcomes/oru_result.govspeak.erb
```diff
@@ -215,13 +215,20 @@
 
   <% if calculator.country_of_birth == 'venezuela' && calculator.same_country? %>
```
thomasleese: Does this true case of the `if` now do the same as the `else`? Can we get rid of the `if` statement completely?

barrucadu: `same_country?` and `in_the_uk?` are mutually exclusive, so yes it is.

maxgds:
Hi @barrucadu this change has introduced a smokey failure, please could you update the smokey repo in line with this?
https://deploy.staging.publishing.service.gov.uk/job/Smokey/9534/console

PR #3126, opened 2017-07-05
Store Start Pages independently of Flow content

chrisroos:
This is an alternative approach to PR #3122.

This PR splits each Smart Answer into a Start Page and a Flow in terms of the Publishing API/Content Store. It publishes the content of the Start Page as a `transaction` and defines the `rendering_app` to be `frontend`. I believe this achieves the main goal of being able to render the Smart Answer Start Pages (along with most/all other Start Pages) through the Frontend app.

In contrast to PR #3122, this branch continues to set the `publishing_app` of the Start Pages as `smartanswers`. This avoids us having to forcefully reserve routes in order to change the `publishing_app` to `Publisher` each time we deploy Smart Answers. That change can happen separately and should probably be a one-off task that's executed within the Publisher environment (something like get Start Page content from Publishing API, add corresponding Transaction artefact to Publisher, reserve the route for Publisher, publish the content through Publisher).

I've deployed this branch to Integration and visually checked each of the Smart Answers using the [links in this gist](https://gist.github.com/chrisroos/a9a0fb6c4c27cf7b472464edf349d66b). Aside from the cosmetic change to the "Start now" button, everything else appears to be identical. The only difference is with the [maternity-paternity-calculator](https://www.gov.uk/maternity-paternity-calculator). The maternity-paternity-calculator start page now includes "Before you start" below the start button instead of "What you need".

I've highlighted the differences mentioned above in the screenshots below.

## maternity-paternity-calculator

smart-answers-maternity-paternity-calculator-start-page

screenshot 2017-07-06 10 53 29

PR #3558, opened 2018-06-19
[WIP] Use flow content item for rendering meta tags

thomasleese:
Currently we use the content item of the start page, not the actual smart answer itself, when rendering the meta tags of a smart answers.

This changes it to use the flow content item instead, but I'm not 100% clear if there are will be any side affects to doing this and whether it was intentional to use the start page content item.

It appears that https://github.com/alphagov/smart-answers/pull/3126 introduced this distinction in the first place, and since then the analysts have reported that the `govuk:format` for smart answers is `transaction` not `smart_answer`.

[Trello Card](https://trello.com/c/vhfMGngX/239-investigate-and-correct-smart-answer-google-analytics-tracking)


app/services/flow_content_item_retriever.rb
```diff
@@ -1,8 +1,7 @@
-class ContentItemRetriever
+class FlowContentItemRetriever
   def self.fetch(slug)
-    Services.content_store.content_item("/#{slug}")
+    Services.content_store.content_item("/#{slug}/y")
```
kevindew: I wonder if we have a path we can use here rather than having to create the path ourselves (and thus hardcode the /y)

thomasleese: Although we do have access to the path, I'm not sure about if we'll hit issues around whether the format is set: https://github.com/alphagov/smart-answers/blob/master/config/routes.rb#L12

What do you think about instead providing a method which turns a slug into this path and then we can use it here as well:

https://github.com/alphagov/smart-answers/blob/f3790fb4650283827d1a0ab66bb5b21d85dabcc1/app/presenters/flow_content_item.rb#L38-L40

binaryberry:
Closing as we've gone with a different approach instead - adding "publishing-app" as a custom dimension in GA to help differentiate transactions from smart-answers.

PR #3636, opened 2018-08-09
Fix leave for parents pay rates

elliotcm:
https://trello.com/c/LHDD9g5m/373-pay-leave-for-parents-calculator-change-to-earnings-limit


thomasleese:
I've renamed the PR so the seal ignores it.

elliotcm:
OK.  We should probably add labels into that check.

PR #3648, opened 2018-08-17
Revert "Bump mocha from 1.4.0 to 1.7.0"

thomasleese:
Reverts alphagov/smart-answers#3637

As per https://github.com/alphagov/smart-answers/pull/3633 "Regression tests have been failing on master since this was merged. I had a quick look, but couldn't figure out how to fix it. Reverting for now to unblock deployments."

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -47,7 +47,7 @@ group :test do
   gem 'ci_reporter'
   gem 'minitest', '~> 5.11'
   gem 'minitest-focus', '~> 1.1', '>= 1.1.2'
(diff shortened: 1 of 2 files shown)


binaryberry:
cc @cbaines

cbaines:
> cc @cbaines

Thanks for looking at this. I've got one idea how we can avoid this in the future [1]. I also looked at speeding up the regression tests, but unfortunately no luck there yet...

1: https://github.com/alphagov/smart-answers/pull/3651

PR #3670, opened 2018-09-07
Update maternity/paternity calculator rounding

elliotcm:
Two things here, as requested by HMRC:
1. Only round up calculation results when we get to the monthly Statutory Maternity Pay (SMP) number
2. Show the full Average Weekly Earnings (AWE) number up to 7 decimal places

Also seems like the expected test results file wasn't updated for a long time for this calculator, so a bunch of those updates are in here too.

https://trello.com/c/uH75vcF5/388-maternity-paternity-rounding-up-issue


elliotcm:
Preview approved by Sandie at HMRC.

PR #3693, opened 2018-09-24
Get a simply - typo?

erikeide:
I dunno


issyl0:
Thanks for this. It looks like we've since fixed it elsewhere. 😄

PR #3704, opened 2018-09-28
Remove checksums

barrucadu:
I don't think the checksums provide any value:

* They tell us if some code changes, but that's what code review is
  for.

* They tell us if some content changes, but that's what content
  review and the regression tests are for.

When was the last time you saw a change to the checksums and thought
"oh, that was unexpected, I'll go double-check that file", rather than
just assuming that the new checksums are correct and committing the
diff?


deborahchua:
Yeh I also agree, I don't feel any more reassured after running them and feels like an unnecessary additional step.

barrucadu:
Something seems wrong with this, the tests are now taking about 11 hours to run.  I might look into it further later in the week.

PR #3719, opened 2018-10-05
Resolve linting issues

bilbof:
Similar to alphagov/feedback#522, this resolves any linting issues in this repository raised by govuk-lint-ruby. Reviewing by commit is probably the easiest way to look through the changes.

Please let me know if the new checksums are not required. It sounded like those were needed in order for the tests to pass.


lib/graph_presenter.rb
```diff
@@ -75,8 +75,12 @@ def word_wrap(text, line_width = 40)
   end
 
   module MethodMissingHelper
+    def respond_to_missing?(*_args)
+      true
+    end
```
cbaines: I had some problems when I tried something similar in Whitehall, although I understand very little about this aspect of Ruby... might be worth double checking the code where this has an effect.

lib/method_missing_object.rb
```diff
@@ -6,9 +6,13 @@ def initialize(method, parent_method = nil, blank_to_s = false, overrides = {})
     @overrides = overrides
   end
 
+  def respond_to_missing?(*_args)
+    true
+  end
```
cbaines: Same comment as here https://github.com/alphagov/smart-answers/pull/3719/files#r223651837

lib/smart_answer/calculators/register_a_birth_calculator.rb
```diff
@@ -38,7 +38,7 @@ def paternity_declaration?
     end
 
     def before_july_2006?
-      Date.new(2006, 07, 01) > childs_date_of_birth
+      Date.new(2006, 0o7, 0o1) > childs_date_of_birth
```
cbaines: I guess this is technically right (which is why Rubocop makes this change automatically), but I'd suggest changing these numbers in to decimal instead.

lib/smart_answer_flows/plan-adoption-leave.rb
```diff
@@ -29,15 +29,16 @@ def define
...
           response
         end
 
         calculate :calculator do
           Calculators::PlanAdoptionLeave.new(
             match_date: match_date,
             arrival_date: arrival_date,
-            start_date: start_date)
+            start_date: start_date
+)
```
cbaines: I'm guessing this is a automatic fix. I sometimes 'correct' this, as while Rubocop doesn't have an issue with it, it still seems a bit odd to me. But I don't really mind either way.

config/initializers/preload_working_days.rb
```diff
@@ -3,5 +3,5 @@
 # Don't preload the bank_holidays in test, so that we can stub out the calls in test setup methods.
 unless Rails.env.test?
   # This loads the bank holiday dates from https://www.gov.uk/bank-holidays.json
-  WorkingDays.bank_holidays
+  WorkingDays.send(:bank_holidays)
```
barrucadu: Why is this change necessary?

bilbof: As part of [this commit](https://github.com/alphagov/smart-answers/pull/3719/commits/d7ad47aa9d78cc492054b4d0f1ec3910fa56590a) I made the `self.bank_holidays` method private. It looks like the intent was for `bank_holidays` to be a private method, but as a singleton method it was still accessible. Since the method is now private I needed to use `.send` here so that I could bypass the private check.

Let me know if it would be better to make the method public. I can also undo this change, and add in a `rubocop:disable` comment.

bilbof:
Closing for now. I have a WIP locally that I will come back to.

PR #3732, opened 2018-10-24
[DO NOT MERGE] Add content to statutory sick pay start page 

binaryberry:
The calculator doesn't allow users to add more than one linked period of
sickness, so we want to direct users to content that explains how to
manually work out statutory sick pay.

This also fixes two negative contractions: don't and can't.

*Deployment*: this will require a rake task to be run when deployed, to publish the start page to the publishing-api.

[Trello](https://trello.com/b/16mTLUnP/platform-health-doing)


binaryberry:
Indeed @barrucadu ! Amended now.

PR #3795, opened 2018-12-17
Log the backtrace for timeout errors

cbaines:
Currently it's hard to tell where exactly the timeout errors are
coming from. So log the backtrace when this happens, as this is the
least worst way of doing this that I could find.


cbaines:
I think I've found a better way https://github.com/alphagov/smart-answers/pull/3796

PR #3801, opened 2018-12-18
Trello-386: workaround 500's on smartanswers

ronocg:
Since migrating the content-store to AWS, we are seeing timeouts on
some requests on smartanswers when it makes calls to retrieve content items.
The latency has increased since the migration so we need to temporarily
increase the timeout time for the calls to the content-store.
The default timeout is 4 seconds we are doubling it to 8 seconds until
the frontends are migrated to AWS.

More info here: https://github.com/alphagov/gds-api-adapters/blob/a83d405df57f84fbf7080903a485d562a1350543/README.md#setting-the-timeout

Here the timeout exceptions are being rescued as 503's: https://github.com/alphagov/smart-answers/blob/09c7f542e758c307f282dba349ee620a0ffd3cc0/app/controllers/application_controller.rb

Pair: @ronocg <conor.glynn@digital.cabinet-office.gov.uk>
 @karlbaker02 <karl.baker@digital.cabinet-office.gov.uk>

### DIFF
--- a/lib/services.rb
+++ b/lib/services.rb
@@ -22,7 +22,8 @@ module Services
 
   def self.content_store
     @content_store ||= GdsApi::ContentStore.new(
-      Plek.new.find('content-store')
+      Plek.new.find('content-store'),
+      timeout: 8
     )
   end
 


lib/services.rb
```diff
@@ -22,7 +22,8 @@ def self.worldwide_api
 
   def self.content_store
     @content_store ||= GdsApi::ContentStore.new(
-      Plek.new.find('content-store')
+      Plek.new.find('content-store'),
+      timeout: 8
```
thomasleese: It seems that this is likely to affect a lot of the frontend apps until they're all running in AWS. I've certainly seen it on government-frontend and finder-frontend as well as smart-answers.

I wonder if it's worth increasing the default timeout from 4 to 8 on [`GdsApi::ContentStore`](https://github.com/alphagov/gds-api-adapters/blob/master/lib/gds_api/content_store.rb#L6) to allow more apps to take advantage of this? As long as we remember to decrease it again once the frontends have migrated.

karlbaker02: If we're going to do this then would it make more sense to use an environment variable and override it per app? If we make the change in `GdsApi::ContentStore` this would involve two releases (one to increase the timeout, another to revert the timeout), plus two releases for each of the apps consuming the gem. 

There could also be some inconsistencies if we have any apps already migrated to AWS using the extended timeout, which is not the change we're looking to make here.

ronocg:
Closing this, we will make the change on gds-api-adapters

PR #3810, opened 2018-12-21
Fixes tests and makes variables more dynamic

jonodrew:
This commit changes the way variables are passed through each context block. We’ve done this because they were previously quite brittle and had a number of hard-coded values that would have to be changed every leap year. Now the number of days in a year is dynamically calculated by the number_of_days method of the YearRange class.

Co-authored-by: Rosa Fox <rosa.fox@digital.cabinet-office.gov.uk>

### DIFF
--- a/test/integration/calculators/part_year_profit_tax_credits_calculator_test.rb
+++ b/test/integration/calculators/part_year_profit_tax_credits_calculator_test.rb
@@ -6,42 +6,43 @@ module SmartAnswer
       setup do
         @calculator = Calculators::PartYearProfitTaxCreditsCalculator.new
 
-        @calculator.tax_credits_award_ends_on  = Date.parse('2015-08-01')
+        @calculator.tax_credits_award_ends_on  = Date.parse('2018-08-01')
         @calculator.accounts_end_month_and_day = Date.parse('0000-04-05')
         @calculator.taxable_profit             = SmartAnswer::Money.new(10_000)
+
+        @expected_accounting_year = YearRange.new(
+          begins_on: Date.parse('2018-04-06')
+          )
+        @expected_profit_per_day = (
+          SmartAnswer::Money.new(10_000) / @expected_accounting_year.number_of_days
+          ).floor(2)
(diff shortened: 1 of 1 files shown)


test/integration/calculators/part_year_profit_tax_credits_calculator_test.rb
```diff
@@ -6,42 +6,43 @@ class PartYearProfitTaxCreditsCalculatorTest < ActiveSupport::TestCase
       setup do
         @calculator = Calculators::PartYearProfitTaxCreditsCalculator.new
 
-        @calculator.tax_credits_award_ends_on  = Date.parse('2015-08-01')
+        @calculator.tax_credits_award_ends_on  = Date.parse('2018-08-01')
```
thomasleese: I'm not quite sure why the year has changed in this test? do we need to update the `context` labels as well?

jonodrew: Ah shoot, you're right. This was supposed to be generic changes, but could update it for the most recent period as well

jonodrew: Or I could change it back so it fits the old format and leave it to someone else to update the next time we change this?

thomasleese: Could we leave this one as 2015 and create another test for 2018? Or will that just make it worse if we're copying and pasting a lot of tests?

jonodrew: I don't think that's a good idea. This is just to improve the tests so there's fewer places to change the numbers when the tests change, and fewer values hard-coded.

jonodrew: @thomasleese what do you think?

thomasleese: @jonodrew I think we can close this one for now and Platform Health will have a think about how best to do it next quarter.

PR #3819, opened 2019-01-02
Increase the Content Store cache time

cbaines:
5 minutes reduces the failures by a lot, but there are still more
failed requests than I'd like.

Upping this to 30 minutes should avoid even more failures, and manual
cache clearing can be done if changes need to be made quickly.


cbaines:
> Might this become a source of tickets about delayed updates? I wonder whether it would be better to focus on finding the underlying problem, as it does seem to be effecting all of the apps.

Smart Answers is a bit of a special case, as there isn't a corresponding publishing app. The changes to the content in the content store are made infrequently by developers, and then deployed as code changes before the content store is updated.

PR #3837, opened 2019-01-14
Flatten Austria 18 outcomes

edwardkerry:
Austria has legalised same-sex marriage so this has been flattened for content changes.

(Trello card)[https://trello.com/c/irPNRnj8/709-marriage-abroad-flatten-austria-to-18-outcomes]


test/artefacts/marriage-abroad/austria/ceremony_country/partner_british/same_sex.txt
```diff
@@ -1,4 +1,4 @@
-Civil partnership in Austria
+Same-sex marriage in Austria
```
barrucadu: You could update the erb file the flattening script produced to avoid this content change.  I imagine this will change at some point anyway, but it means you don't need to get a content designer to approve this now.

edwardkerry: Thanks @barrucadu I've made the change to remove the test artefacts.

PR #3867, opened 2019-02-01
Removed reference to civil partnership

hannahwhittaker:
https://trello.com/c/USToYLTT/1999-marriage-abroad-change-to-cni-content-for-finland
 
I've updated the title file for Finland marriage abroad to remove reference to civil partnership as it's no longer possible to get one.

I've also removed an inaccurate definition of 'legalisation'.


elliotcm:
Branch updated using `bin/prep_for_pull_request.sh marriage-abroad`.

PR #3871, opened 2019-02-06
Update plek service static uri

AlanGabbianelli:
Assets CDN has been added to integration, where before we used origin
directly. It also seem that external access to origin has been blocked
because review links now return `Application error` from Heroku.

The edited url goes through the CDN and is publicly accessible, so it
should solve the problem.


AlanGabbianelli:
Yeah, I can do that later @thomasleese 😉

PR #3882, opened 2019-02-12
Report a lost/stolen passport abroad (deploy on 26 Feb)

hannahwhittaker:
https://trello.com/c/LXA2O1YN/2006-26-feb-changes-to-cancelling-a-lost-stolen-passport-overseas

I've made 2 changes:

- updated the first question to ask where the user is rather than where the passport was lost
- updated the content for the abroad outcome as it's now the same for all overseas countries


hannahwhittaker:
In addition to the content changes, it also needs a logic change to remove the question 'In which country': https://www.gov.uk/report-a-lost-or-stolen-passport/y/abroad

The content changes are currently with HMPO to fact check.

PR #3899, opened 2019-02-21
Rename "Macedonia" to "North Macedonia"

AlanGabbianelli:
HM Government's FCO has officially recognised the change of Macedonia's
name to North Macedonia: 
https://www.gov.uk/government/news/foreign-secretary-statement-on-the-republic-of-north-macedonia

Trello card: https://trello.com/c/ySMq2bcm/786-slug-and-tag-changes-country-name-change-macedonia-to-north-macedonia

Co-Authored-By: Debrah Chua <deborah.chua@digital.cabinet-office.gov.uk>

### DIFF
--- a/lib/data/prisoner_packs.yml
+++ b/lib/data/prisoner_packs.yml
@@ -303,8 +303,6 @@
   translators: /government/publications/british-embassy-luxembourg-translators
   prison: /government/publications/british-embassy-luxembourg-prisoner-pack
 - slug: macao
-- slug: macedonia
-  pdf: /government/publications/macedonia-prisoner-pack
 - slug: madagascar
   lawyer: /government/publications/madagascar-list-of-lawyers
 - slug: malawi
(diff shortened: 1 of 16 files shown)


lib/data/prisoner_packs.yml
```diff
@@ -368,6 +366,8 @@
   pdf: /government/publications/nigeria-prisoner-pack
   lawyer: /government/publications/nigeria-list-of-lawyers
 - slug: north-korea
+- slug: north-macedonia
+  pdf: /government/publications/macedonia-prisoner-pack
```
edwardkerry: Looks like `government/publications/north-macedonia-prisoner-pack` is already up?

lib/data/translators.yml
```diff
@@ -53,6 +52,7 @@ namibia: /government/publications/namibia-list-of-lawyers
 nepal: /government/publications/nepal-list-of-lawyers
 netherlands: /government/publications/netherlands-list-of-lawyers
 nicaragua: /government/publications/nicaragua-list-of-lawyers
+north-macedonia: /government/publications/macedonia-list-of-lawyers
```
edwardkerry: does this also need updating? it redirects to
`https://www.gov.uk/government/publications/north-macedonia-lawyers`

lib/data/prisoner_packs.yml
AlanGabbianelli: @deborahchua and I think that content designers already made those changes, so we'll update them here. 👍

lib/data/translators.yml
AlanGabbianelli: Yup, same as above. 😄

PR #3905, opened 2019-02-27
Simplified info about the Indonesian affirmation/affidavit process

hannahwhittaker:
https://trello.com/c/hw1lEj6R/2060-getting-married-abroad-indonesia

I've updated the 9 opposite sex outcomes to replace out-of-date info with the new process.


edwardkerry:
LGTM, squashed 9 commits into 1

PR #3909, opened 2019-02-28
[Do not merge] Update smart answer start page

strawbridge:
Telling users about forthcoming changes to VAT records and returns. They need to be able to act on these now.


issyl0:
This PR is from February and has conflicts. Closing.

PR #3912, opened 2019-03-01
Flatten marriage abroad outcomes for North Macedonia

elliotcm:
https://trello.com/c/JANhScNn/814-flatten-north-macedonia-to-18-outcomes


test/integration/smart_answer_flows/marriage_abroad_test.rb
```diff
@@ -431,7 +431,7 @@ def self.translations
       add_response 'opposite_sex'
     end
     should "go to outcome_opposite_sex_marriage_in_consular_cni_countries_when_residing_in_third_country" do
```
bevanloon: is the description here still accurate given the assertion on 434? not that I have a better description, just wondering if the descriptions captures what we are actually testing really.

elliotcm: Good call.  Since it's no longer special the tests don't really need to stay, so I've removed them.

elliotcm:
https://trello.com/c/JANhScNn/814-flatten-north-macedonia-to-18-outcomes#comment-5c795130b5ce65604d085c16

PR #3954, opened 2019-04-03
Implement new EEA travel rules

elliotcm:
As per published guidance, these are the rules for travel to the UK from an EEA country if the UK is outside the EU at the time.

https://www.gov.uk/guidance/european-temporary-leave-to-remain-in-the-uk#who-will-need-to-apply

All EEA citizens can stay in the UK for up to and including 3 months without a visa.  Beyond that they will need a European Temporary Leave to Remain.

Republic of Ireland citizens do not need a visa even if staying in the UK longer than 3 months.

https://trello.com/c/VIhHyhO6/913-for-12-april-only-if-no-deal-check-if-you-need-a-visa-content-requests

### DIFF
--- a/lib/smart_answer/calculators/uk_visa_calculator.rb
+++ b/lib/smart_answer/calculators/uk_visa_calculator.rb
@@ -41,6 +41,10 @@ module SmartAnswer::Calculators
       @passport_country == 'syria'
     end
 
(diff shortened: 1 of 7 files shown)


elliotcm:
Please do not delete this branch in case it is needed later.

PR #3993, opened 2019-05-15
Fix lower earnings limits on adoption pay

elliotcm:
The legislation says that
> (2)(d)that [their] normal weekly earnings [lower earnings limit] for the period of 8 weeks ending with the relevant week are not less than the lower earnings limit in force under section 5(1)(a) at the end of the relevant week; and
> ...
> (3)The references in subsection (2)(b) and (d) above to the relevant week are to the week in which the person is notified that he has been matched with the child for the purposes of adoption.

In the case of birth this "relevant week" is 16 weeks before the due date.  For adoption it's the match week, analogous to the due week in most cases.

As such we should be using the end of the match week as the point to check the rates file.

https://trello.com/c/SViOJG0s/991-employers-maternity-calculator-change-calculation-in-adoption-paternity-branch

### DIFF
--- a/lib/smart_answer/calculators/adoption_pay_calculator.rb
+++ b/lib/smart_answer/calculators/adoption_pay_calculator.rb
@@ -13,7 +13,11 @@ module SmartAnswer::Calculators
     end
 
     def lower_earning_limit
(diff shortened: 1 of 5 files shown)


bevanloon:
.

PR #4001, opened 2019-05-20
Ignore definite article "the" from country names when alphabetising dropdown

edwardkerry:
The Worldwide Locations API returns a specific order of country names to display those
starting with `The` in the most convenient position for the user, for example
`The Gambia` appears under `G` rather than `T`. However, the CountrySelect may add some additional countries and then re-order the list.

This PR ensures that SmartAnswers continues to display the user-friendly order by ignoring the definite article `The` at the beginning of any country names when sorting.

[Trello](https://trello.com/c/X2u3t3br/1011-check-if-you-need-a-uk-visa-update-the-gambia-in-dropdown)

### DIFF
--- a/lib/smart_answer/question/country_select.rb
+++ b/lib/smart_answer/question/country_select.rb
@@ -40,7 +40,7 @@ module SmartAnswer
           countries = countries.reject { |c| @exclude_countries.include?(c.slug) }
         end
         if @additional_countries
-          countries = (countries + @additional_countries).sort { |a, b| a.name <=> b.name }
+          countries = (countries + @additional_countries).sort { |a, b| a.name.sub(/^The /i, '') <=> b.name }
         end
         countries
       end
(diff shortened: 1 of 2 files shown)


lib/smart_answer/question/country_select.rb
```diff
@@ -40,7 +40,7 @@ def load_countries
           countries = countries.reject { |c| @exclude_countries.include?(c.slug) }
         end
         if @additional_countries
-          countries = (countries + @additional_countries).sort { |a, b| a.name <=> b.name }
+          countries = (countries + @additional_countries).sort { |a, b| a.name.sub(/^The/, '') <=> b.name }
```
MuriloDalRi: Does it matter if this is case sensitive? And do we care about whitespaces?

edwardkerry: I don't think the case matters, as we are looking only at the beginning of the string. That said, it won't hurt to take care of it just in case. 👍

PR #4076, opened 2019-07-22
Update national minimum wage flow

edwardkerry:
This PR changes the way the Minimum Wage Calculator works for both employee and employer branch.

It removes the option to run the calculator against a specific year, but instead only deals with the current or previous year. If the user selects the `past` branch, the calculator date is now set to April 1st 2018 to access minimum wage dates for last year.

It removes a question asking if the worker works overtime, and a sub-question asking how many hours. As this information is no longer used, overtime information has also been removed from the calculator and all tests.

It adds a question asking if the worker is charged for requirements needed for their job.

It adds a question asking if the worker works hours outside their shift, and a sub-question asking if they are paid for these.

It also updates copy across several questions and outcomes.

[Trello](https://trello.com/c/iei6H1pB/1121-changes-to-national-minimum-wage-smart-answers)


lib/smart_answer_flows/am-i-getting-minimum-wage/questions/current_paid_for_work_outside_shift.govspeak.erb
```diff
@@ -0,0 +1,8 @@
+<% content_for :title do %>
+  Does your employer pay you for any additional time you work outside of your shift?
+<% end %>
+
+<% options(
+  "yes": "Yes",
+  "no": "No"
+) %>
```
cbaines: You're missing a newline at the end of this file (and some other files as well).

PR #4092, opened 2019-08-05
 Add new maternity leave question

AlanGabbianelli:
This is to give a more accurate outcome.

From Content team:
```
An earlier version of the smart answer assumed you weren't eligible for
Statutory Maternity Leave if you weren't eligible for Statutory
Maternity Pay - that's true in 99% of cases but not all.

We fudged the outcome by adding 'The employee is entitled to Statutory
Maternity Leave if... they have an employment contract with you,
regardless of when it started', but it's better to have proper routing
and make the outcomes easier to understand.
```

Trello card: https://trello.com/c/QUboSszZ/1159-employers-maternity-calculator-new-question-in-maternity-branch

### DIFF
--- a/lib/smart_answer_flows/maternity-paternity-calculator/maternity_calculator_flow.rb
+++ b/lib/smart_answer_flows/maternity-paternity-calculator/maternity_calculator_flow.rb
@@ -63,7 +63,7 @@ module SmartAnswer
           end
         end
 
-        # QM3
+        ## QM3
         multiple_choice :did_the_employee_work_for_you_between? do
           option :yes
           option :no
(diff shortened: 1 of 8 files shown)


lib/smart_answer_flows/maternity-paternity-calculator/maternity_calculator_flow.rb
```diff
@@ -78,19 +78,38 @@ def define
...
 
         ## QM4
+        multiple_choice :does_the_employee_work_for_you_now? do
+          option :yes
+          option :no
+
+          save_input_as :has_employment_contract_now
+
+          next_node do |response|
+            case response
```
edwardkerry: Does this need to be a case statement, if both responses lead to the same outcome? Or is it more for documentation?

AlanGabbianelli: You're right, it doesn't. I was originally following the structure of the other questions but it doesn't make sense in this case.

PR #4105, opened 2019-08-22
Corrected consular appointments link

hannahwhittaker:
Updated the 9 opposite sex outcomes to update the link to the consular appointments service.


hannahwhittaker:
Thanks @cbaines! Can you hold off on deploying? I'm just double checking with the embassy that they're happy.

cbaines:
> Thanks @cbaines! Can you hold off on deploying? I'm just double checking with the embassy that they're happy.

Sure.

PR #4131, opened 2019-09-18
Upgrade components gem

andysellick:
Upgrade smart answers to the new components gem. Not much needed doing, although I've had to apply a workaround to get `govuk-link` styles into most of the text, by wrapping the text in a new class and giving that the required link styles.

There are a lot of elements in this application that would benefit from components but this is not a small change so I'm attempting it in a [separate PR](https://github.com/alphagov/smart-answers/pull/4132).

### DIFF
--- a/app/assets/stylesheets/smart_answers.scss
+++ b/app/assets/stylesheets/smart_answers.scss
@@ -41,6 +41,10 @@ $is-ie: false !default;
     min-height: 0px;
   }
 
+  .descriptive__text a:focus {
+    @include govuk-focused-text;
+  }
+
   .previous-answers-top {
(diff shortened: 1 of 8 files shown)


app/assets/stylesheets/smart_answers.scss
```diff
@@ -41,6 +41,14 @@ $is-ie: false !default;
     min-height: 0px;
   }
 
+  .descriptive__text a:focus {
```
injms: Could be worth using `@include govuk-focused-text;` here

andysellick: Ah, of course. Good shout, thanks. Are you happy to approve?

PR #4132, opened 2019-09-18
Use govuk components in smart answers

andysellick:
## What
Updates smart answers to use GOVUK components for form elements. This includes using component labels/legends for questions instead of headings, and using built in error messages for form components instead of a custom error message.

This PR doesn't include a total removal of all govuk_frontend_toolkit stuff, as it's big enough already, but the user should have a much improved experience.

There are some things about this change that I'm not fully happy with and hoping that someone will point out a better way of doing them - mainly that there's now a lot of logic in the view, and it shouldn't be. I've tried to move it but getting the chosen value of form elements caused problems (e.g. accessing URL params from a presenter).

Test URL: https://smart-answers-preview-pr-4132.herokuapp.com/

The 'all smart answer questions' smart answer doesn't appear to be in the above deployment, so I'd recommend checking out this branch and running it locally to see it.

## Visual changes

All of these screenshots are from the 'All smart answer questions' test smart answer that's built into the app but not publicly visible, so some of the data/error messages are test content.

Before on the left, after on the right.

### Start page

No visual changes.

---

### Checkbox question

smartanswers1

(I've just noticed that the hint isn't correct in this screenshot, and corrected it)

There doesn't appear to be an error state for this question.

---

### Country select question

_smartanswers2

_smartanswers3

---

### Date select question

_smartanswers4
_smartanswers5

---

### Money question

Due to the way that components are laid out I can't reproduce the layout of the original question exactly, so I've reworded the hint text to be more descriptive (it's hardcoded). I tried to get a content person to help with this but no one stepped up. Happy to change the wording if anyone can think of something better.

You'll note also that the input component doesn't have the black outline that it should. This is being overridden by template, and I haven't bothered trying to override it. Worth doing?

_smartanswers6
_smartanswers7

---

### Multiple choice question

_smartanswers8
_smartanswers9

---

### Postcode question

_smartanswers10

_smartanswers11

---

### Multiple choice (yes/no) question

_smartanswers12

_smartanswers13

---

### Salary question

Another one that I had to make changes to based on how components are laid out. In the old version, `£` was the label for the input and `per` was the label for the select. So while this change may not be perfect, I'm confident it's more accessible than before.

Note that the application provides no error state for the select element in this question.

_smartanswers14

_smartanswers15

---

### Value question

There's a variation on this where text is added after the input, which I couldn't find a good example of. I've moved it to be the hint text for the input component, but we may need to revisit this.

_smartanswers16

_smartanswers17

### DIFF
--- a/app/presenters/date_question_presenter.rb
+++ b/app/presenters/date_question_presenter.rb
@@ -1,4 +1,5 @@
 class DateQuestionPresenter < QuestionPresenter
+  include CurrentQuestionHelper
   delegate %i[default_day default_month default_year] => :@node
 
   def response_label(value)
@@ -9,6 +10,37 @@ class DateQuestionPresenter < QuestionPresenter
     end
   end
 
+  def error_id
+    "error_id" if error
+  end
+
+  def days_options
+    days = Array(1..31).map { |number|
+      format_date(number, :day)
+    }
+    days.unshift(text: "", value: "")
+  end
+
+  def months_options
+    months = Array(1..12).map { |number|
+      format_date(number, :month)
+    }
+    months.unshift(text: "", value: "")
+  end
+
+  def years_options
+    years = Array(start_date.year..end_date.year).map { |number|
+      format_date(number, :year)
+    }
+    years.unshift(text: "", value: "")
+  end
+
+private
+
+  def only_display_day_and_month?(value)
+    value.year.zero?
+  end
+
   def start_date
     @node.range == false ? 1.year.ago : @node.range.begin
   end
@@ -17,9 +49,11 @@ class DateQuestionPresenter < QuestionPresenter
     @node.range == false ? 3.years.from_now : @node.range.end
   end
 
-private
-
-  def only_display_day_and_month?(value)
-    value.year.zero?
+  def format_date(number, type)
+    {
+      text: type.eql?(:month) ? Date::MONTHNAMES[number] : number,
+      value: number,
+      selected: default_for_date(prefill_value_for(self, type)) == number,
+    }
   end
 end
--- a/app/views/smart_answers/inputs/_date_question.html.erb
+++ b/app/views/smart_answers/inputs/_date_question.html.erb
@@ -1,17 +1,45 @@
-<fieldset>
-  <% unless question.default_day %>
-    <label for="response_day">Day
-      <%= select_day default_for_date(prefill_value_for(question, :day)), prefix: "response", :prompt => '' %>
-    </label>
-  <% end %>
-  <% unless question.default_month %>
-    <label for="response_month">Month
-      <%= select_month default_for_date(prefill_value_for(question, :month)), prefix: "response", prompt: '' %>
-    </label>
-  <% end %>
-  <% unless question.default_year %>
-    <label for="response_year">Year
-      <%= select_year default_for_date(prefill_value_for(question, :year)), prefix: "response", prompt: '', :start_year => question.start_date.year, :end_year => question.end_date.year %>
-    </label>
-  <% end %>
-</fieldset>
+<% form_contents = capture do %>
+  <div class="govuk-grid-row date-wrapper">
+    <% unless question.default_day %>
+      <div class="govuk-grid-column-one-third date-wrapper__col">
+        <%= render "govuk_publishing_components/components/select", {
+          id: "response_day",
+          name: "response[day]",
+          label: "Day",
+          error_id: question.error_id,
+          options: question.days_options
+        } %>
+      </div>
+    <% end %>
+    <% unless question.default_month %>
+      <div class="govuk-grid-column-one-third date-wrapper__col">
+        <%= render "govuk_publishing_components/components/select", {
+          id: "response_month",
+          name: "response[month]",
+          label: "Month",
+          error_id: question.error_id,
+          options: question.months_options
+        } %>
+      </div>
+    <% end %>
+    <% unless question.default_year %>
+      <div class="govuk-grid-column-one-third date-wrapper__col">
+        <%= render "govuk_publishing_components/components/select", {
+          id: "response_year",
+          name: "response[year]",
+          label: "Year",
+          error_id: question.error_id,
+          options: question.years_options
+        } %>
+      </div>
+    <% end %>
+  </div>
+<% end %>
+
+<%= render "govuk_publishing_components/components/fieldset", {
+  legend_text: question.title,
+  heading_size: 'm',
+  error_message: question.error,
+  error_id: question.error_id,
+  text: form_contents
+} %>
--- a/app/views/smart_answers/question.html.erb
+++ b/app/views/smart_answers/question.html.erb
@@ -14,7 +14,7 @@
 
       <div class="question" data-debug-template-path="<%= question.relative_erb_template_path %>">
         <% t = question.partial_template_name %>
-        <% if t != 'multiple_choice_question' && t != 'checkbox_question' && t != 'country_select_question' %>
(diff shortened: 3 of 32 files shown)


vanitabarrett:
I think this will help you move some of the additional logic from the view into the relevant presenter... 🤞  The issue is that `params` isn't automatically available within the presenters - it needs to be passed to them when they are initialised.

We can pass params through in `flow_presenter.rb`, which initialises each presenter

**Before:**
```
@node_presenters[node.name] ||= presenter_class.new(node, current_state)
```

**After:**
```
@node_presenters[node.name] ||= presenter_class.new(node, current_state, {}, params)
```

Then in `question_presenter.rb`, we can expand the initialize function to expect this new argument:

**Before:**
```
def initialize(node, state = nil, options = {})
   ...etc
end
```

**After:**
```
// I don't want to break existing calls to initialize(), so we'll make params the last argument expected by the function
def initialize(node, state = nil, options = {}, params = {})
    @params = params
   ....etc
end

def params
    @params
end
```

You should then be able to move some of that logic that calls `prefill_value_is?()` into the relevant presenter, as I think this was the function that was complaining about `params` not being available.

test/functional/smart_answers_controller_test.rb
```diff
@@ -116,7 +116,7 @@ def teardown
 
     should "display first question after starting" do
       get :show, params: { id: "smart-answers-controller-sample", started: "y" }
-      assert_select ".step.current [data-test=question]", /Do you like chocolate\?/
+      assert_select ".step.current .govuk-fieldset__legend", /Do you like chocolate\?/
```
vanitabarrett: Do we know if the `.step.current` class + associated styling are still needed? I tried removing it and it looks ok, but not sure if I'm missing a situation where it's needed as it also looks ok on the live site without it.

vanitabarrett:
Spacing between date fields on smaller screens doesn't look quite right: 

Screen Shot 2019-09-27 at 14 36 05

app/views/smart_answers/inputs/_date_question.html.erb
```diff
@@ -1,17 +1,80 @@
...
-    </label>
-  <% end %>
-  <% unless question.default_year %>
-    <label for="response_year">Year
-      <%= select_year default_for_date(prefill_value_for(question, :year)), prefix: "response", prompt: '', :start_year => question.start_date.year, :end_year => question.end_date.year %>
-    </label>
-  <% end %>
-</fieldset>
+<%
+  if question.error
```
vanitabarrett: This feels like something which could be a method in a presenter

vanitabarrett:
Not sure if this is your change, but I noticed the Start Again button takes up the full width because it's `display: block`:

Screen Shot 2019-09-27 at 14 47 07

app/views/smart_answers/question.html.erb
```diff
@@ -13,27 +13,14 @@
...
           <div class="<%= question.error && 'error' %>">
-            <% if question.error %>
-              <p class="error-message" id="current-error" role="alert">
-                <%= question.error %>
-              </p>
-            <% end %>
-
             <%= render partial: "smart_answers/inputs/#{question.partial_template_name}", locals: { question: question } %>
 
             <%= question.post_body %>
```
vanitabarrett: There's a line near the end of this file that renders a green button - this needs replacing with the button component, I think. It renders like this on https://smart-answers-preview-pr-4132.herokuapp.com/vat-payment-deadlines/y/2020-05-31 : 

Screen Shot 2019-09-27 at 14 50 01

vanitabarrett:
I didn't even know these pages existed... but the /visualise pages seem to be broken. Technically, they're pretty broken on live already as they're so large they overflow the page, but the flow diagram does render. On the preview, it seems to be completely collapsing so you can't see the links between anything. 

**On live (zoomed out!):**
Screen Shot 2019-09-27 at 15 03 25

**Heroku preview:**
Screen Shot 2019-09-27 at 15 03 11

andysellick:
@vanitabarrett thanks for the review, some great stuff there. Not sure what's going on with the visualise pages as they're broken on my local master as well (i.e. they're all collapsed on top of themselves as shown in your heroku screenshot). I'd tentatively suggest that wasn't me? 😄

test/functional/smart_answers_controller_test.rb
andysellick: We should be using govuk-frontend for all the layout stuff so yes, this should definitely be looked at. I decided that this PR was big enough already without trying to take out govuk_frontend_toolkit at the same time so I'm going to address that in a followup PR.

app/views/smart_answers/question.html.erb
andysellick: Fixed, good spot, thanks!

andysellick:
@vanitabarrett I think I've addressed all your comments... thanks again for the view/presenter stuff, I've been able to move all the logic out of the view now. 

I haven't addressed the wide 'start again' link as I'll do that in a separate PR, and the flow visualisation stuff seems to be broken on master, so I'm not sure what to do about that. Maybe it doesn't work locally?

vanitabarrett:
@andysellick Yeah, the visualize stuff looks fine on integration so I guess it must depend on something which isn't available locally/on Heroku... I'll re-review the other changes 🙂

chao-xian:
@andysellick I've spoken to Xander about getting a Content Designer to check the £ changes.

app/presenters/date_question_presenter.rb
```diff
@@ -9,17 +10,54 @@ def response_label(value)
...
+        value: number,
+        selected: default_for_date(prefill_value_for(self, :day)) == number,
+      }
+    }
+    days.unshift(text: "", value: "")
+  end
+
+  def months_options
+    months = Array(1..12).map { |number|
+      {
```
vanitabarrett: I wonder if we can simplify some of this logic into one method, as there's a lot of repetition for each days, months, years method. Something like this, which can be called by each method?

```
def format_date(number, type)
    {
      text: type.eql?(:month) ? Date::MONTHNAMES[number] : number,
      value: number,
      selected: default_for_date(prefill_value_for(self, type)) == number,
    }
  end
```

vanitabarrett:
@andysellick Just remembered what I was meant to say in a comment above, sorry! Quite a few of the commit messages are now out of date (talking about needing to refactor/move logic which has now been done). Could you please update them?

app/presenters/date_question_presenter.rb
```diff
@@ -9,17 +10,54 @@ def response_label(value)
     end
   end
 
-  def start_date
-    @node.range == false ? 1.year.ago : @node.range.begin
+  def error_id
+    "error_id" if error
```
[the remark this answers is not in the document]
andysellick: Yes, it's intentional. There are up to three selects on this page and in the event of an error smart answers only provides one error message, so we need to pass the id of that error to each select component for the `aria-describedby` attribute.

Ordinarily, yes, each select would have it's own error and we'd just pass them to the component and let it generate it's own error ID.

app/presenters/date_question_presenter.rb
andysellick: Nice. Thanks for that. Have implemented.

andysellick: Have also fixed the commit messages, I had cleaned up some of them, but obviously had missed a couple.

andysellick:
@vanitabarrett I think I've addressed your changes, are you happy for me to merge?

@chao-xian have spoken with a content designer and tweaked the hint text for some of the questions.

vanitabarrett:
@andysellick Yep, looks good to me!

PR #4150, opened 2019-10-03
Fix question layouts

andysellick:
Recent changes to use the components for smart answers introduced a few question readability problems, specifically to do with the layout of questions - body text was appearing in the wrong place, hint text was sometimes not appearing.

Example of the problem (from [this smart answer](https://www.gov.uk/landlord-immigration-check/y/E1%208QS/yes/yes/eea)):

Screen Shot 2019-10-03 at 10 44 01

Solution:

Screen Shot 2019-10-03 at 10 43 36

Another problem was that suffixes (text immediately to the right of some value questions, e.g. "days") didn't make sense in the new layout, or repeated what was already said in the question.

This PR solves these problems by doing the following layout changes:

- where components have a description field as well as a hint, body text is used for description and so included after the question label or legend (multiple choice question, date question)
- where components do not have a description field, body text, hint text and suffix text are combined into a single element and used for the hint text (value questions)

This PR also includes some content changes to questions, where the above measures did not help with the text layout of the questions. These content changes are from [this content spreadsheet](https://docs.google.com/spreadsheets/d/1l-FfOM2ZqDRUYtEBuqKz1n77py6yTnQoV8pGip2vhOg/edit#gid=2072014057) and only those needed and marked as okay to do have been implemented.

This PR also fixes a bug with date questions, where if the year range started with a higher number (e.g. a range between 2019 and 1900) the range would return 0. This is fixed by using the smallest number as the start of the range (this happened in 'Calculate your married couples allowance' on the age of the husband question).

This PR also updates the 'all smart answer questions' test question to have more data in it, to improve the reliability of it when testing.

### DIFF
--- a/app/assets/stylesheets/smart_answers.scss
+++ b/app/assets/stylesheets/smart_answers.scss
@@ -254,6 +254,10 @@ $is-ie: false !default;
         margin-top: 20px;
       }
     }
+
+    li {
+      margin-bottom: govuk-spacing(1);
+    }
   }
 
   .current-question {
--- a/app/presenters/date_question_presenter.rb
+++ b/app/presenters/date_question_presenter.rb
@@ -29,7 +29,10 @@ class DateQuestionPresenter < QuestionPresenter
   end
 
   def years_options
-    years = Array(start_date.year..end_date.year).map { |number|
+    smallest = [start_date.year, end_date.year].min
+    biggest = [start_date.year, end_date.year].max
+
+    years = Array(smallest..biggest).map { |number|
       format_date(number, :year)
     }
     years.unshift(text: "", value: "")
--- a/app/presenters/money_question_presenter.rb
+++ b/app/presenters/money_question_presenter.rb
@@ -4,4 +4,9 @@ class MoneyQuestionPresenter < QuestionPresenter
   def response_label(value)
     format_money(SmartAnswer::Money.new(value))
   end
+
+  def hint_text
+    text = [body, "in £", suffix_label].reject(&:blank?).compact.join(", ")
+    ActionView::Base.full_sanitizer.sanitize(text)
+  end
 end
--- a/app/presenters/value_question_presenter.rb
+++ b/app/presenters/value_question_presenter.rb
@@ -6,6 +6,7 @@ class ValueQuestionPresenter < QuestionPresenter
   end
 
   def hint_text
-    suffix_label.present? ? suffix_label : label
+    text = [body, hint, suffix_label].reject(&:blank?).compact.join(", ")
+    ActionView::Base.full_sanitizer.sanitize(text)
   end
 end
--- a/app/views/smart_answers/inputs/_checkbox_question.html.erb
+++ b/app/views/smart_answers/inputs/_checkbox_question.html.erb
@@ -1,6 +1,7 @@
 <%= render "govuk_publishing_components/components/checkboxes", {
   name: "response[]",
   heading: question.title,
+  description: question.body,
   id: "response",
   error: question.error,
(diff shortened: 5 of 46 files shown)


injms:
I'm seeing a missing line of text on the maternity/paternity/adoption checker:

Preview app:
Screen Shot 2019-10-03 at 12 04 02

Live:
Screen Shot 2019-10-03 at 12 04 40

injms:
This input is feels strangely wide - do you think it could be made less wide?

Page found on: https://smart-answers-preview-pr-4150.herokuapp.com/am-i-getting-minimum-wage/y/current_payment/not_an_apprentice/34/31

Currently:
Screen Shot 2019-10-03 at 12 12 52

Maybe:
Screen Shot 2019-10-03 at 12 14 02

injms:
I think the lists need extra spacing between each list item to mirror `govuk-list govuk-list--bullet` styling.

Extreme example from https://smart-answers-preview-pr-4150.herokuapp.com/landlord-immigration-check/y/GU7%202LN/yes/yes/somewhere-else/no/no/no/no/no/no 

Without `govuk-list govuk-list--bullet`:
Screen Shot 2019-10-03 at 12 22 59


With `govuk-list govuk-list--bullet`:
Screen Shot 2019-10-03 at 12 23 13

injms:
I think the lists need extra spacing between each list item to mirror `govuk-list govuk-list--bullet` styling.

Extreme example from https://smart-answers-preview-pr-4150.herokuapp.com/landlord-immigration-check/y/SW1A%202AA/yes/yes/somewhere-else/no/no/no/no/no/no 

Without `govuk-list govuk-list--bullet`:
Screen Shot 2019-10-03 at 12 22 59


With `govuk-list govuk-list--bullet`:
Screen Shot 2019-10-03 at 12 23 13

selfthinker:
I gave feedback to Andy elsewhere which he has incorporated.

Just to explain the spreadsheet he linked in the description above...
When I was working on Smart Answer 3 years ago, I planned to implement the standard patterns. And when I tested that I discovered that there are a lot of content issues preventing this. The spreadsheet lists all of those issues (from 3 years ago):
* The “suffix_label” tab lists questions that have something that is to the right-hand side of a label. That would cause trouble because we don't have such a pattern. The content designers back then agreed that it's best to remove all of them and, if necessary, incorporate their meaning into the question instead. This PR removes them.
* The “label” tab lists questions which have a label. Because most input fields in smart answers don't actually have a label but the actual question should be one (or a legend), that would cause trouble because then you'd have two labels. The content designers back then agreed that it's best to just remove those labels as all of them just repeat what's already in the question/title. I believe this PR also removes those.
* The “body” tab lists questions which have text between the title and the input. If the title became the label or legend, that would cause trouble because we have no pattern that has extra meaningful text after the label. Because that extra "body" text is usually important to the question and can be really long, it is very difficult to solve this so it still makes semantic sense. This PR fixes some of that.

There are still issues left with some of the "body" texts. This PR is definitely an improvement and to solve these issues is not in the remit of this PR. But I'd like to point this out just so it isn't forgotten.

One example of an issue that has now been created: [This question](https://smart-answers-preview-pr-4150.herokuapp.com/landlord-immigration-check/y/E1%208QS/yes/yes/eea) has the question in the legend and the extra text underneath before the input. A screen reader users would go to the first radio button and get the legend read out in addition to the label but the legend doesn’t make sense without the additional text.

There are probably lots of similar issues left. (Best go through the spreadsheet to find out how many.) It's not a blocker as there are still ways for screen reader users to find the content. But ideally it should be on someone's radar to fix them.

PR #4171, opened 2019-10-16
Add rspec-rails gem + config & "default" rake task

AlanGabbianelli:
This is the basis to start migrating from minitest to rspec.

The `rspec-rails` gem has been added, plus the related configuration in
the `spec_helper.rb` file.

In the Rakefile, the "default" rake task has been overridden so that it
now runs both minitest and rspec tests. This allow us to use:
- `rake test` to only run minitest tests
- `rake spec` or `rspec` to only run rspec tests
- `rake` to run both minitest and rspec tests


cbaines:
I'd have expected most of the linting issues to have been fixed by https://github.com/alphagov/smart-answers/pull/4139

We don't normally lint the bin directory, which could be why you're finding issues there.

Rakefile
```diff
@@ -5,3 +5,13 @@
...
 
 SmartAnswers::Application.load_tasks
+
+# Delete the current "default" rake task and redefine it. This allow us to use:
+# - `rake test` to only run minitest tests
+# - `rake spec` or `rspec` to only run rspec tests
+# - `rake` to run both minitest and rspec tests
+Rake::Task["default"].clear
+task :default do
+  system("bundle exec rake spec")
```
cbaines: How does this handle failures, will the default task exit with a non 0 exit code if either of the sets of tests fail?

PR #4178, opened 2019-10-17
Rescue 403 errors from content-store

leenagupte:
Trello: https://trello.com/c/BFSkOY22
Follows on from: alphagov/collections#1324

What's changed and why?
At the moment smart-answers are throwing a 500, when it receives an error code it doesn't recognise
from content-store. That means that users are shown a "Problem has occurred" page.

We need to handle this error properly so that in the case of a 403, the user will be properly routed
to signon and asked to login if there is any form of access limiting on the content item (e.g. hosted on draft
stack, access limited to organisation etc)


ChrisBAshton:
Works (in collaboration with https://github.com/alphagov/frontend/pull/2001)

PR #4189, opened 2019-10-24
Fix styles of headings and lists in body text

andysellick:
- styling was missing for headings and lists in result and intro pages for smart answers
- adds govuk-frontend mixins to provide consistent font sizes etc. as content is rendered in govspeak and can't have specific classes applied
- also adds a fix for 'quote' elements (they look like blockquotes but they're not) as this change originally broke them a little

Before:

Screenshot 2019-10-24 at 15 46 55

After:

Screenshot 2019-10-24 at 15 47 14

### DIFF
--- a/app/assets/stylesheets/smart_answers.scss
+++ b/app/assets/stylesheets/smart_answers.scss
@@ -11,6 +11,29 @@ $is-ie: false !default;
   .outcome {
     min-height: 0; //overrides a much higher min height from static
   }
+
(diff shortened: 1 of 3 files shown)


andysellick:
@injms thanks. Yes, it is. Interestingly, despite having things like this that look like blockquotes, as far as I can tell we don't have any actual blockquotes on GOV.UK - apart from in govspeak (which is probably why there's no blockquote component).

PR #4204, opened 2019-11-06
Remove rspec tests

MahmudH:
The tests in minitest were removed to be replaced by rspec. Even though this is a good idea, we believe that this is a separate piece of work.


spec/spec_helper.rb
```diff
@@ -1,24 +0,0 @@
-ENV["RAILS_ENV"] ||= "test"
```
edwardkerry: Thinking about this we should probably leave the overall spec folder and spec_helper in place, and just remove the test file itself, as Rspec was setup in a different PR

MahmudH: I've added it back in 👍

PR #4211, opened 2019-11-08
Holiday entitlement calculator update

edwardkerry:
This PR is a major update for the holiday entitlement calculator. It updates the logic for all calculations in line with policy guidance, with a greater level of accuracy for leap years. It also introduces two new scenarios for people working annualised and irregular hours.

Individual commits have details on specific changes and calculations.

[Trello card](https://trello.com/c/DOulB3Hd/1297-epic-republish-holiday-entitlement-calculator-with-new-logic-%F0%9F%8D%90)


issyl0:
Part of me thinks it might be nice if the commits at the start were worded such that we could tell which path it was, like the later ones do?

So the first one:

```
HolidayEntitlement: Refactor logic and tests for full days calculations
```

could change to

```
Refactor logic and tests for days worked per week
```

and the following _n_ commits related to days worked per week could be squashed, and so on?

PR #4217, opened 2019-11-13
Use in-memory ActiveSupport cache, not lrucache

barrucadu:
lrucache hasn't been updated since 2012.  It depends on PriorityQueue,
which hasn't been updated since 2009, and its website domain has since
expired, been snapped up, and parked.  It's not good to use such old
dependencies, especially where the original sources have been lost.

This commit replaces the use of lrucache with the ActiveSupport cache.
This is much better because it's part of Rails, so it will work for as
long as we support smart-answers.


barrucadu:
@benthorner  I think there's still value in having a fallback cache which serves stale content if the regular cache can't be updated - the logic of "we'd like to expire things after a day but are happy to wait up to a week if the alternative is serving errors" seems sensible to me.

PR #4238, opened 2019-12-05
StudentFinanceCalculator: Remove 2018 dates, add 2020 dates and adhere to style guide

issyl0:
https://trello.com/c/qhBsbXHr/1608-add-student-finance-rates-for-2020-2021

PR #4243, opened 2019-12-06
StudentFinanceCalculator: Remove 2018 dates, add 2020 dates and adhere to the style guide

issyl0:
- Originally raised as #4238 and then reverted in #4242 as I was too quick for departmental fact-check. Raising this again for a new Heroku preview link.


issyl0:
From @OrcaTom: "This has been approved by the department this morning so can be deployed."

PR #4282, opened 2020-01-15
[DO NOT MERGE BEFORE alphagov/govuk-puppet#10054] Access Worldwide API via whitehall-frontend

whi-tw:
To simplify the migration of Whitehall to AWS, calls to the Worldwide API should be directed to whitehall-frontend, not whitehall-admin.

### DIFF
--- a/lib/services.rb
+++ b/lib/services.rb
@@ -1,8 +1,8 @@
 require "statsd"
+require "gds_api"
 require "gds_api/publishing_api_v2"
 require "gds_api/content_store"
 require "gds_api/imminence"
-require "gds_api/worldwide"
 
 module Services
   def self.publishing_api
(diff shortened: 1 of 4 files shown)


lib/services.rb
```diff
@@ -17,7 +17,7 @@ def self.imminence_api
   end
 
   def self.worldwide_api
-    @worldwide_api ||= GdsApi::Worldwide.new(Plek.new.find("whitehall-admin"))
+    @worldwide_api ||= GdsApi::Worldwide.new(Plek.new.find("whitehall-frontend"))
```
thomasleese:
```suggestion
    @worldwide_api ||= GdsApi.worldwide
```

Since https://github.com/alphagov/gds-api-adapters/pull/980 we can rely on that being configured to talk to whitehall-frontend and it makes it easier to change it again in the future across all our apps.

PR #4286, opened 2020-01-17
Self Assessment penalty calculator change

callumknights:
### Self Assessment penalty calculator change

Changed content as well as changed the daily interest rate from
2.75% to 3.25%, changed relevant rate change dates as well.

https://trello.com/c/sVVglBFQ/1685-2-self-assessment-penalty-calculator-content-requests.

### DIFF
--- a/lib/smart_answer/calculators/self_assessment_penalties.rb
+++ b/lib/smart_answer/calculators/self_assessment_penalties.rb
@@ -15,6 +15,7 @@ module SmartAnswer::Calculators
         "2015-16": ONLINE_FILING_DEADLINE_YEAR.starting_in(2017).begins_on,
         "2016-17": ONLINE_FILING_DEADLINE_YEAR.starting_in(2018).begins_on,
         "2017-18": ONLINE_FILING_DEADLINE_YEAR.starting_in(2019).begins_on,
+        "2018-19": ONLINE_FILING_DEADLINE_YEAR.starting_in(2020).begins_on,
       },
       offline_filing_deadline: {
         "2012-13": OFFLINE_FILING_DEADLINE_YEAR.starting_in(2013).begins_on,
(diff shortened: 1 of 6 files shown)


lib/smart_answer/calculators/self_assessment_penalties.rb
```diff
@@ -186,12 +193,12 @@ def calculate_interest_for_date(date)
...
-      rate_change_date = Date.new(2016, 8, 23)
+      # Rate increased to 3.25% on 21 August 2018
+      rate_change_date = Date.new(2018, 8, 21)
       if date < rate_change_date
         0.03 / 365.0
       else
-        0.0275 / 365.0
+        0.0325 / 365.0
       end
     end
```
AlanGabbianelli: Not sure about what the policy says, but should we have three options here?
before 23-08-2016 -> 0.03 / 365.0
between 23-08-2016 and 21-08-2018 -> 0.0275 / 365.0
after 21-08-2018 -> 0.0325 / 365.0

callumknights: According to content apparently not, I expect there is also a bit of "is that really worth it" at play here. Considering the interest rate has been wrong since 2018 and no one noticed till now, and it's only an estimate and not used for any real calculations anyway. On the other hand it might just be a HMRC quirk and thats the way it is 🤷‍♂

lib/smart_answer_flows/estimate-self-assessment-penalties/questions/which_year.govspeak.erb
```diff
@@ -8,5 +8,6 @@
   "2014-15": "6 April 2014 to 5 April 2015",
   "2015-16": "6 April 2015 to 5 April 2016",
   "2016-17": "6 April 2016 to 5 April 2017",
-  "2017-18": "6 April 2017 to 5 April 2018"
+  "2017-18": "6 April 2017 to 5 April 2018",
+  "2018-19": "6 April 2018 to 5 April 2019"
```
AlanGabbianelli:
```suggestion
  "2018-19": "6 April 2018 to 5 April 2019",
```

PR #4346, opened 2020-03-06
Update smart answer outcome for intestacy

bilbof:
Requested at https://govuk.zendesk.com/agent/tickets/3935092


barrucadu:
There'll need to be some change to the flow logic to send them to a question with the updated amount, somewhere here: https://github.com/alphagov/smart-answers/blob/master/lib/smart_answer_flows/inherits-someone-dies-without-will.rb#L42

Or you might be able to get away with changing how the `estate_over_250000` question (and answer) is displayed if the region is `england-and-wales`

bilbof:
Thanks Michael!

PR #4360, opened 2020-03-16
Fix minimum wage calculator rounding issues

callumknights:
It was discovered as part of the uprating work that the minimum wage calculator
could give false positive results due to it needing to round the minimum wage of
a given user to 2 decimal places to finish it's calculation. This meant in some
cases the total pay a person recieved in a given time period meant they were
earning slightly under mininum wage (maximum 0.5 pence per hour under).

To rectify this I changed the calculator to use the minimum entitlement for
a given number of hours as it's basis for deciding whether a user is being
above or below the minimum wage. This is more suitable as the total pay a user
has recieved is given to 2 decimal places in the first place, and the minimum
wage in the `rates.yml`, which minimum entitlement is based off is also given
to 2 decimal places. This avoids any unneccessary rounding and makes the
calculator more accurate in cases where the employee is earning an amount very
close to minimum wage.

I have also added some test cases to the flow testing, just to cover some extra
scenarios using accurate numbers given to me by content.

Trello card > https://trello.com/c/k8xvMhKs/1813-minimum-wage-calculator-bug

### DIFF
--- a/lib/smart_answer/calculators/minimum_wage_calculator.rb
+++ b/lib/smart_answer/calculators/minimum_wage_calculator.rb
@@ -108,7 +108,11 @@ module SmartAnswer::Calculators
     end
 
     def minimum_wage_or_above?
-      minimum_hourly_rate <= total_hourly_rate
+      minimum_entitlement <= total_pay
+    end
+
+    def minimum_entitlement
+      historical_entitlement
     end
 
     def accommodation_adjustment(charge, number_of_nights)
--- a/test/integration/smart_answer_flows/am_i_getting_minimum_wage_test.rb
+++ b/test/integration/smart_answer_flows/am_i_getting_minimum_wage_test.rb
@@ -1163,5 +1163,165 @@ class AmIGettingMinimumWageTest < ActiveSupport::TestCase
         assert_equal 3.90, current_state.calculator.minimum_hourly_rate
       end
     end
+    context "18 years old, first year apprentice, unpaid working time" do
+      should "earn above minmum wage" do
+        assert_current_node :what_would_you_like_to_check?
+        add_response "current_payment"
+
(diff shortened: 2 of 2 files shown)


lib/smart_answer/calculators/minimum_wage_calculator.rb
```diff
@@ -108,7 +108,11 @@ def historical_adjustment
     end
 
     def minimum_wage_or_above?
-      minimum_hourly_rate <= total_hourly_rate
+      minimum_entitlement <= total_pay
```
thomasleese:
```suggestion
      historical_entitlement <= total_pay
```

It doesn't feel like we need the `minimum_entitlement` method since it just returns the value of a different method.

callumknights: I had this discussion with myself at length, I only added that cause `historical_entitlement` is used only in the context of checking historical wages. I just felt `minimum_entitlement` was clearer when reading the code, I'm happy to be proven wrong though.

thomasleese: That's reasonable. Rather than create a new method, you can use [`alias`](https://apidock.com/ruby/Module/alias_method) which is slightly nicer!

```rb
alias_method :historical_entitlement, :minimum_entitlement
```

callumknights: Ooh I had absolutely no idea this existed! Thanks Tom!

callumknights: For some reason `alias_method` wouldn't work in this instance, seemed to have a scoping issue as that call eventually led to a private method which was erroring.

OrcaTom:
Tested and looks good from a content perspective.

PR #4372, opened 2020-03-20
Add callout to Statutory Sick Pay calculator start page

ChrisBAshton:
As requested: https://govuk.zendesk.com/agent/tickets/3955481

PR #4374, opened 2020-03-20
Revert content type to remove UTF information

ChrisBAshton:
This was added in ba67c415410e8cec4f412e80f3fefd06d87c09d5
automatically as part of Rails 6 upgrade. However, it caused
our publishing API rake tasks to break:

> GdsApi::HTTPUnprocessableEntity: URL: https://publishing-api.integration.govuk-internal.digital/v2/content/05d5412d-455b-485e-a570-020c9176a46e
> Response body:
> {"error":{"code":422,"message":"Details there must be at least one content type of (text/html, text/govspeak)","fields":{"details":["there must be at least one content type of (text/html, text/govspeak)"]}}}

Ideally we should add a test in future which validates against
the schema.

This is needed ASAP so we can publish https://github.com/alphagov/smart-answers/pull/4372 today.

### DIFF
--- a/app/presenters/start_page_content_item.rb
+++ b/app/presenters/start_page_content_item.rb
@@ -16,12 +16,12 @@ class StartPageContentItem
         introductory_paragraph: [
           {
             content: flow_presenter.start_page_body,
-            content_type: "text/govspeak; charset=utf-8",
+            content_type: "text/govspeak",
           },
         ],
         more_information: [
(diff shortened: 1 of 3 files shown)


ChrisBAshton:
Have just built this branch to integration and the rake task now works: https://deploy.integration.publishing.service.gov.uk/job/run-rake-task/131069/console

PR #4394, opened 2020-03-27
Fix question body formatting

andysellick:
It's possible to pass govspeak formatting to to question body but this is currently not being rendered properly. This PR fixes that for the following question types:

- checkbox question
- date question
- multiple choice question

Before:

Screenshot 2020-03-27 at 11 55 15

After:

Screenshot 2020-03-27 at 11 55 23


andysellick:
@alex-ju yes, had spotted that might need doing, was waiting for confirmation but I think I'll go ahead anyway 👍

PR #4414, opened 2020-04-06
Remove call out

NikkiAlbon:
Remove %The service to register an overseas death with the UK authorities is currently suspended. This page will be updated when it has resumed.% 
Because it draws the user's eye down to that warning. Users may not realise they can use the tool to find out how to register a death in the UK. The majority of users will be registering a death in the UK.
The General Register Office asked for it to be removed. FCO - who had asked for it to be added - agreed that it could be removed


NikkiAlbon:
@issyl0 I've created a Trello card and moved it to the 'To deploy column'.

NikkiAlbon:
This is not live on the site. This can now be deployed.

PR #4418, opened 2020-04-08
Add initial flow for coronavirus business support

benthorner:
https://trello.com/c/LoPI3DJc/31-investigate-viability-of-using-smart-answers

This adds a draft smart answer flow for supporting businesses during
the Coronavirus pandemic. Content is placeholder only, and we intend
to iterate this and add tests once the initial code is merged. Using
a 'draft' status means this will only be visible in development [1].
    
Since the purpose of this smart answer is to filter from a number of
potential results, we've chosen to use a single 'results' outcome page,
and a Calculator class [2] to determine which of the results to display
to the user; each result has its own bespoke criteria to display.
    
[1]: https://github.com/alphagov/smart-answers/blob/master/doc/smart-answers/flow-definition.md#metadata
[2]: https://github.com/alphagov/smart-answers/blob/master/doc/smart-answer-flow-development/refactoring.md

### DIFF
--- a/lib/smart_answer_flows/coronavirus-business-support.rb
+++ b/lib/smart_answer_flows/coronavirus-business-support.rb
@@ -1,0 +1,133 @@
+module SmartAnswer
+  class CoronavirusBusinessSupportFlow < Flow
+    def define
+      start_page_content_id "89edffd2-3046-40bd-810c-cc1a13c05b6a"
+      flow_content_id "1f589327-a6b3-4b5c-aea0-7a2752e2eddf"
+      name "coronavirus-business-support"
+      status :draft
+
+      # Q1
+      multiple_choice :business_based? do
+        option :england
+        option :scotland
+        option :wales
+        option :northern_ireland
+
+        on_response do |response|
+          self.calculator = Calculators::CoronavirusBusinessSupportCalculator.new
+          calculator.business_based = response
+        end
+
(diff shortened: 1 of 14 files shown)


lib/smart_answer_flows/coronavirus-business-support.rb
```diff
@@ -0,0 +1,133 @@
+module SmartAnswer
+  class CoronavirusBusinessSupportFlow < Flow
+    def define
+      start_page_content_id "89edffd2-3046-40bd-810c-cc1a13c05b6a"
+      flow_content_id "1f589327-a6b3-4b5c-aea0-7a2752e2eddf"
```
theseanything, line 5: Curious where do these come from?

benthorner: Turns out we need to make our own, as per https://github.com/alphagov/smart-answers/blob/master/doc/smart-answer-flow-development/content-ids.md

chao-xian: @theseanything Probably here https://github.com/alphagov/smart-answers/blob/master/doc/smart-answer-flow-development/content-ids.md

chao-xian: @benthorner snap

PR #4429, opened 2020-04-09
Allow sending draft Smart Answers to Publishing API and clean up related tasks

kevindew:
Trello: https://trello.com/c/qUyNDb1n/39-live-preview-of-smart-answer

First off, apologies that this PR is a bit meandering, whilst the focus was on resolving the Publishing API draft functionality there were various interconnected aspects that were in a muddle and I resolved them at the same time. This may make reviewing this a bit of a bigger job than it needs be - if that's the case or I've made contentious changes we can break it down.

Key things this does:

- Stops the smart-answers app deciding on whether to render smart-answers based on their status. the application will render them all regardless and it's up to whether the item is in the content store as to whether requests make it to the app (this allowed simplifying some rather confusing constant manipulation)
- It changes the ContentItemPublisher#publish method to send draft items to Publishing API as drafts and only publish those with a status of published. I renamed the class and method to ContentItemSyncer#sync to reflect it's wider purpose - I'm not super happy with this naming and happy to discuss alternatives
- It removes a bunch of additional publish themed rake tasks that allowed smart answers to create document types that mainstream publisher produces. Instead publishers should create documents there, I've updated the documentation to reflect this
- I've extracted the unpublishing code out of ContentItemPublisher and moved those tasks from the retire rake namespace to be within the publishing_api one


test/unit/services/content_item_publisher_test.rb
```diff
@@ -1,25 +1,51 @@
...
+  context "#sync" do
+    setup do
+      start_page_content_id = SecureRandom.uuid
+      flow_content_id = SecureRandom.uuid
+      @start_page_draft_request = stub_publishing_api_put_content(start_page_content_id, {})
+      @start_page_publishing_request = stub_publishing_api_publish(start_page_content_id, {})
+      @flow_draft_request = stub_publishing_api_put_content(flow_content_id, {})
+      @flow_publishing_request = stub_publishing_api_publish(flow_content_id, {})
+      @flow = stub("flow",
+                   name: "bridge-of-death",
```
benthorner: What inspired this?

test/unit/tasks/retire_test.rb
```diff
@@ -95,10 +112,14 @@ class RetireSmartAnswerRakeTest < ActiveSupport::TestCase
       assert_equal "Missing publishing_app parameter", exception.message
     end
 
-    should "invoke reserve_path_for_publishing_app method on ContentItemPublisher" do
-      ContentItemPublisher.any_instance.expects(:reserve_path_for_publishing_app).with("/base-path", "a-publisher").once
+    should "sends a path reservation to the Publishing API" do
```
benthorner:
```suggestion
    should "send a path reservation to the Publishing API" do
```

lib/tasks/retire.rake
```diff
@@ -1,36 +1,35 @@
...
     raise "Missing destination parameter" if args.destination.blank?
 
-    content_item_publisher = ContentItemPublisher.new
-
-    content_item_publisher.unpublish_with_redirect(
-      args.content_id, args.base_path, args.destination
-    )
+    Services.publishing_api.unpublish(args.content_id,
+                                      type: "redirect",
+                                      alternative_path: args.destination)
```
benthorner: Do we need to preserve this behaviour?

https://github.com/alphagov/smart-answers/pull/4429/commits/b66db24a7607c91f7c3c225634cefabff625c928#diff-8cef3e0c497f0b3bd493ed82777c86d2L35

doc/smart-answer-flow-development/retiring-a-smart-answer.md
```diff
@@ -1,69 +1,95 @@
...
+examples of common files:
+
+- Flow class files
+  - lib/smart_answer_flows/<\smart-answer-name>.rb
+- ERB templates directory
+  - lib/smart_answer_flows/<\smart-answer-name>
+- YAML files
+  - lib/data/rates/<\smart-answer-name>.yml
+  - lib/data/<\smart-answer-name>.yml
+- Tests for Calculators, rates, data query and other ruby files
```
benthorner, line 20: Minor: we shouldn't need to specify these, since they will (hopefully) fail.

README.md
```diff
@@ -76,7 +76,7 @@ This is a Ruby on Rails application that contains:
 
 ### Changes to the landing page
```
benthorner, line 77: Minor: I found this confusing when I read it initially. Could we rename it to "Changes to a Smart Answer"? The process is the same for all parts of the flow.

app/presenters/flow_registration_presenter.rb
```diff
@@ -88,10 +88,6 @@ def flows_content
       end
   end
 
-  def state
```
benthorner, line 87: I wonder if it was related to this: https://github.com/alphagov/smart-answers/blob/master/doc/smart-answer-flow-development/archiving.md.

Perhaps we can remove this doc as well (and the link to it in the README)? Looks like it's pretty out-of-date.

README.md
benthorner: Less minor: having this here did slow me down when writing/merging the initial smart answer, since it's weirdly separate to the "Smart Answer Flow Development" section in the README. 

Could we make it it's own section and link/step instead? Happy for this to be done later.

app/services/content_item_publisher.rb
```diff
@@ -1,13 +1,13 @@
 class ContentItemPublisher
-  def publish(flow_presenters)
+  def sync(flow_presenters)
     flow_presenters.each do |smart_answer|
       start_page_content_item = StartPageContentItem.new(smart_answer)
       Services.publishing_api.put_content(start_page_content_item.content_id, start_page_content_item.payload)
-      Services.publishing_api.publish(start_page_content_item.content_id)
+      Services.publishing_api.publish(start_page_content_item.content_id) if smart_answer.publish?
```
theseanything: Is this okay when a flow goes from published back to draft? Wasn't sure how Publishing API deals with that

lib/tasks/retire.rake
kevindew: Oh shoot! I've probably messed that up - good catch.

test/unit/services/content_item_publisher_test.rb
kevindew: Haha it was the existing the test data unfortunately not my imagination.

doc/smart-answer-flow-development/retiring-a-smart-answer.md
kevindew: Sure - I can drop that, it's a copy and paste from the existing doc

app/presenters/flow_registration_presenter.rb
kevindew: Sure - I'll see about doing that.

app/services/content_item_publisher.rb
kevindew: No it doesn't handle that situation well at all unfortunately. Publishing API would have a new draft of the items and the previous one would be published so you'd be served the Smart Answer still.

If you go from published back to draft you'd have to follow the retiring steps first, then this would be fine.

README.md
kevindew: Ah darn all I did was fix the rake task name on that line - I've not even read it 🙈. Your points sound good though. Is this something you wouldn't mind picking up? I think you've probably got a clearer idea than me what it should say.

lib/smart_answer/flow.rb
```diff
@@ -14,6 +14,7 @@ def self.build
 
     def initialize(&block)
       @nodes = []
+      status(:draft)
```
theseanything, line 17: If the default is draft? Do we need to add a `:published` flag to all existing smart answers?

kevindew: They already all have this

app/services/content_item_publisher.rb
theseanything: Ah righto. The naming `sync` maybe a little confusing against our publishing model. Probably something we can address in a future PR along with some more docs about the draft vs published states.

kevindew: Yeah - I'm not very happy about the naming. I felt that `#publish` was wrong but `#sync` isn't as sophisticated as you may expect.

app/presenters/flow_registration_presenter.rb
kevindew: I'll plan to do that after this PR actually since it's not interconnected.

kevindew:
Thanks @benthorner and @theseanything for the reviews.

@benthorner I've actioned your comments now if you'd like to take another look.

lib/tasks/retire.rake
benthorner: I'm still wondering: does this conflict with the instructions here: https://github.com/alphagov/smart-answers/pull/4429/files#diff-5450b7bbb07ec3d9169cb6e9fa889847R35? It looks like this is something we've added, compared to [the original](https://github.com/alphagov/smart-answers/pull/4429/files#diff-5450b7bbb07ec3d9169cb6e9fa889847L33), which only suggested running the task once.

kevindew: With the previous approach you should still have done it for both, however it's not clearly signposted: https://github.com/alphagov/smart-answers/pull/4429/files#diff-5450b7bbb07ec3d9169cb6e9fa889847L5-L6 I imagine it's often been omitted as people wouldn't anticipate visits to `/<smart-answer>/y`

Even with a prefix on the `/<smart-answer>` any visits to /<smart-answer>/y would go to that route as prefix only applies when there aren't routes with more matching segments.

benthorner: > prefix only applies when there aren't routes with more matching segments

That's good to know.

benthorner:
@kevindew this is good to go for me.

PR #4436, opened 2020-04-09
Add CLBILS scheme to the Coronavirus Business Support results

theseanything:
This is an another scheme to be shown to business with more than 45m and less than 500m turnover.

### DIFF
--- a/lib/smart_answer_flows/coronavirus-business-support/outcomes/results.govspeak.erb
+++ b/lib/smart_answer_flows/coronavirus-business-support/outcomes/results.govspeak.erb
@@ -123,5 +123,15 @@
     $CTA
   <% end %>
 
+  <% if calculator.show?(:large_business_loan_scheme) %>
+    $CTA
+    **Coronavirus Large Business Interruption Loan Scheme**
+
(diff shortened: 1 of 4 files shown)


lib/smart_answer_flows/coronavirus-business-support/outcomes/results.govspeak.erb
```diff
@@ -122,4 +122,12 @@
     [Find out more](https://www.google.com)
     $CTA
   <% end %>
+
+  <% if calculator.show?(:large_business_loan_scheme) %>
+    $CTA
+    **[Coronavirus Large Business Interruption Loan Scheme](https://www.gov.uk/guidance/apply-for-the-coronavirus-large-business-interruption-loan-scheme)**
```
kevindew, line 128: Ah do we have a preference for using absolute URLs? I'll update for consistency

theseanything: Sorry no preference that was just me copy and pasting. Should probably use relative, but that won't work in the Heroku app.

PR #4443, opened 2020-04-09
Remove integration links

kevindew:
These links are for the old host name and they 404 when switched to the
www.integration hostname


doc/debugging/viewing-templates-as-govspeak.md
```diff
@@ -11,10 +11,3 @@ The Govspeak version of the pages can be accessed by appending `.txt` to URLs.
 * [Marriage abroad landing page](http://smartanswers.dev.gov.uk/marriage-abroad.txt)
 * [Marriage abroad question page](http://smart-answers.dev.gov.uk/marriage-abroad/y.txt)
 * [Marriage abroad outcome page](http://smartanswers.dev.gov.uk/marriage-abroad/y/afghanistan/uk/partner_other/opposite_sex.txt)
-
-## In Integration environment
-
-**It's not currently possible to access the Govspeak version of landing pages in the Integration environment.** There's a [Trello card about this inconsistency](https://trello.com/c/qd5C7qDn/165-allow-landing-pages-to-be-rendered-as-govspeak-in-integration).
-
-* [Marriage abroad question page](https://www-origin.integration.publishing.service.gov.uk/marriage-abroad/y.txt)
-* [Marriage abroad outcome page](https://www-origin.integration.publishing.service.gov.uk/marriage-abroad/y/afghanistan/uk/partner_other/opposite_sex.txt)
```
cbaines, line 20: This link works for me. I'm not sure what's up with the link above, but I'm not quite sure why they're being removed?

kevindew: Oh so it does - sorry about that. I'm guessing the first one fails because the route in the router is /marriage-abroad/y and the prefix routing for that will only catch later segments, for example: `/marriage-abroad/y/other.anything`.

As for the why, I had thought it was a no-brainer because they both didn't work even with change of hostname however I didn't realise the second would work with a change of hostname. In a wider context though, as part of our work using Smart Answers we are finding documentation that is out of date so where applicable I'm getting rid of it to reduce ongoing maintenance. 

Do you know if this section provides dev with particular value? I'd still be keen to remove this with an amended commit message as I think a sufficient amount of the functionality doesn't work in an integration environment for it to not be particularly helpful.

cbaines: I think removing this section is OK, but for consistency I'd also remove the reference to Integration "It is enabled by default in the Integration environment" further up in this file.

kevindew: Thanks - I've made that change now.

PR #4447, opened 2020-04-14
Fix spacing for previous questions

benthorner:
This is consistent with the margin/class used on the results page,
and is particularly helpful on mobile.

### Before

Screenshot 2020-04-14 at 11 03 21

### After

Screenshot 2020-04-14 at 11 03 59

### DIFF
--- a/app/assets/stylesheets/smart_answers.scss
+++ b/app/assets/stylesheets/smart_answers.scss
@@ -80,7 +80,3 @@ $is-ie: false !default;
   @include govuk-font($size: 16);
   color: $govuk-secondary-text-colour;
 }
-
-.app-o-spacing {
-  margin-bottom: govuk-spacing(6);
-}
--- a/app/views/smart_answers/landing.html.erb
+++ b/app/views/smart_answers/landing.html.erb
@@ -13,7 +13,7 @@
     } %>
 
     <article role="article" data-debug-template-path="<%= start_node.relative_erb_template_path %>">
-      <section class="app-o-spacing">
+      <section class="govuk-!-margin-bottom-6">
         <%= render "govuk_publishing_components/components/govspeak", {
         } do %>
           <%= start_node.body %>
--- a/app/views/smart_answers/landing.html.erb
+++ b/app/views/smart_answers/landing.html.erb
@@ -23,7 +23,8 @@
           text: start_node.start_button_text,
           href: smart_answer_path(@name.to_s, started: "y"),
           rel: "nofollow",
-          start: true
+          start: true,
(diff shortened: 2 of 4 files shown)


benthorner:
Good question @kevindew. I was trying to be conssitent with the other pages. Turns out `govuk-spacing(6)` is _slightly_ larger than `margin-bottom-6`. I've added another couple of commits to replace the dodgy style, and with the following additional diff.

### Before
Screenshot 2020-04-14 at 12 01 57

### After
Screenshot 2020-04-14 at 12 02 17

app/assets/stylesheets/smart_answers.scss
```diff
@@ -80,7 +80,3 @@ $is-ie: false !default;
   @include govuk-font($size: 16);
   color: $govuk-secondary-text-colour;
 }
-
-.app-o-spacing {
-  margin-bottom: govuk-spacing(6);
```
andysellick, line 85: We had a discussion amongst the FE devs on GOV.UK about these classes relatively recently.  The `govuk-!-margin-something` classes are override classes - they contain `!important` and are meant for situations where a style is already present but can't be changed, only overridden. The decision made was that override classes like this should be used when dealing with legacy code where this kind of problem arises. For any other use, we should be writing new classes that use the `govuk-spacing` scale.

@alex-ju

benthorner: Oh no, we've been caught 😸. I'm open to a more conventional approach on this. The problem for me is that Smart Answers styles tend to be on the inline side of things, and I didn't want to invent classes for each of the places I was substituting this. There was also some existing uses of '!', so we were only making an existing problem worse 🔥.

Ways I can think to address this:

- Invent the classes so we can style the elements
- Use some other inline class that does the same thing

I'm pretty sure we don't have an alternative for inline stuff. Do you have any other suggestions, or should we make some `app-` classes?

andysellick: 😁 

For this, I wouldn't worry. And you're right, there was some inconsistency here, which is probably worse. For balance, you could make the argument that writing a new class to do something that an existing class already does only adds more CSS weight (although anything involving `!important` should only be used as a last resort).

So I'm happy to leave this change as it is, although I wasn't sure what you meant when you talked about 'inline stuff'?

benthorner: Thanks for taking another look @andysellick. We'll leave this as-is then. Turns out I was wrong about the 'inline (class/style) stuff' - I'm the main culprit here!

PR #4449, opened 2020-04-14
Add further tests for Business Coronavirus Support Checker

chao-xian:
A couple more flows needed testing in the checker.

Also replaces the Calculator integration test with a unit test.

Trello https://trello.com/c/l2QKcqWf


benthorner:
The tests themselves look OK. I'm struggling to understand the overall integration test, though: it's hard to work out why these particular tests are there, and why a new one is being added. The calculator test also looks more like a unit test, since it's testing the calculator class in isolation.

I'd find the calculator integration tests easier to review if each context described itself in a way that made the contexts comparable. For example, if we're testing individual scheme outcomes, then that could be the context (and it doesn't really matter what route we took to get there).

We should probably rethink our integration tests, in general, though - I was assuming they were all in one file. There seems to be some precedent for this: [see here, for example](https://github.com/alphagov/smart-answers/blob/master/test/integration/smart_answer_flows/am_i_getting_minimum_wage_test.rb#L66).

For this PR, the [change in the finder test](https://github.com/alphagov/smart-answers/pull/4449/files#diff-c98383ab813286a0602c6d49e67a37dcR43) seems OK. I'm not really sure why we're making the change in the calculator test, though. Can we add some explanation for that one?

chao-xian:
@benthorner thanks for your feedback - I'll re-arrange the contexts as you suggest.

The approach to the tests for the calculator was as I understood from this [guidance in the docs](https://github.com/alphagov/smart-answers/blob/master/doc/smart-answers-app-development/testing.md#integration-tests) but I may have mis-interpreted.

I agree the Calculator integration test isn't needed at all if the unit tests covers the same ground for the calculator, but I added it as part of the integration tests in lieu of there being a unit test for it. I was going to add the unit test as a separate PR but I can rework this PR to swap the integration test for the calculator.

test/integration/smart_answer_flows/business_coronavirus_support_finder_test.rb
```diff
@@ -39,4 +39,29 @@ class BusinessCoronavirusSupportFinderFlowTest < ActiveSupport::TestCase
       end
     end
   end
+
```
benthorner, line 42: Somewhat related: could we merge the contexts in the existing test in this file, so the two tests are structured consistently?

test/unit/calculators/business_coronavirus_support_calculator_test.rb
```diff
@@ -0,0 +1,304 @@
...
+          @calculator.business_rates = "yes"
+          @calculator.non_domestic_property = "over_15k"
+          @calculator.sectors = %w[retail hospitality leisure]
+        end
+
+        should "return true when criteria met" do
+          assert_equal true, @calculator.show?(:business_rates)
+        end
+
+        context "criteria not met" do
```
benthorner, line 109: Minor: I don't think we need this context.

test/unit/calculators/business_coronavirus_support_calculator_test.rb
```diff
@@ -0,0 +1,304 @@
...
+          @calculator.business_rates = "yes"
+          @calculator.non_domestic_property = "over_15k"
+          @calculator.sectors = %w[retail hospitality leisure]
+        end
+
+        should "return true when criteria met" do
+          assert_equal true, @calculator.show?(:grant_funding)
+        end
+
+        context "criteria not met" do
```
benthorner, line 139: Or this one (context not needed).

chao-xian:
Thanks @benthorner! I'll make your suggestions in another PR, as the tests should go in for now.

kevindew:
Thanks @chao-xian - I don't mind taking on some of these comments while modifying most of the questions (I think I'm going to touch a lot of them anyway)

chao-xian:
@kevindew only if it's easier that way, otherwise I was going to wait until you made your changes first. But I'm not feeling precious about it 👍

PR #4454, opened 2020-04-15
Move smart answer configuration out of lib/data

kevindew:
Since this is configuration information it feels somewhat out of place
within the lib directory of this project where you would anticipate
finding code.

I chose to rename the directory from data to smart_answers so that this
information is more clearly config for smart answers when looking at the
directory. The name "data" felt rather ubiquitous. The naming was a pain
though as the only clear convention is that these files are yaml.

Moving this resolves a deprecation warning where Zeitwerk tries to load
a Ruby constant of Data. This could probably have been resolved by
removing this path from being eager loaded.


kevindew:
> Nice one! I was about to just approve, and then had a couple of thoughts:
> 
> * Could we bring [this thing (for marriage abroad)](https://github.com/alphagov/smart-answers/blob/master/config/outcome_flattener.yml) into this overall change?

Yup, I think I can sort that out relatively painlessly will have a look

> * How about putting the config with each smart answer flow?
> 
> The latter is more a general pain point that the definition of a "smart answer" is split between the "flow" and a couple of other places (calculators, random helpers, YAML files). It would at least be more convenient, and arguably less surprising, if they were together, in one place.

This seems a wider job, I definitely agree it's a pain point. I find it strange that calculators are somewhere totally different and I continuously manage to open the start page govspeak expecting the flow definition. These data files seem to be in a muddle already with many being shared across multiple answers rather than logically making sense grouped with a flow.

So I'd rather defer that for now.

kevindew:
@benthorner I was able to remove the output flattener config if you don't mind taking another look.

kevindew:
> Thanks for checking in @kevindew. I didn't realise the flattener was such a niche thing. Given it's only got that one use case, perhaps we could just remove it, but that would be pushing the scope here!

Oh I know - it's a grim thing. It doesn't seem to be a straight forward thing to rip out as it seems to be something that was added back for Platform Health after initially removed so I'm hesitant to kill it again without a good understanding of how/why it's used. All looks pretty hacky though.

PR #4457, opened 2020-04-16
Next round of Coronavirus Business Support amends

kevindew:
This represents all of the changes requested for lunchtime of April 16th.

Highlights are:

- we now have two different flows where a user may or may not be shown sector information
- self employed question is reinstated though we're not super happy it makes sense in the flow or context
- non-domestic property is now phrased like an actual question 🎉


lib/smart_answer_flows/business-coronavirus-support-finder.rb
```diff
@@ -83,7 +97,7 @@ def define
         end
       end
 
-      # Q6
+      # Q7
```
benthorner, line 103: Minor: does these numbers actually help anyone? Since some questions may be conditional in general, I'm struggling to understand why these are here.

test/unit/calculators/business_coronavirus_support_calculator_test.rb
```diff
@@ -146,13 +155,13 @@ class BusinessCoronavirusSupportFinderCalculatorTest < ActiveSupport::TestCase
...
 
-        should "return false when no non domestic property" do
+        should "return false when no non-domestic property" do
           @calculator.non_domestic_property = "none"
           assert_not @calculator.show?(:nursery_support)
         end
 
         should "return false when not supported business sector" do
-          @calculator.sectors = %w[retail hospitality leisure]
+          @calculator.sectors = %w[retail_hospitality leisure]
```
benthorner, line 164:
```suggestion
          @calculator.sectors = %w[retail_hospitality_or_leisure]
```

benthorner: These loose option strings are definitely a pain point of this app!

lib/smart_answer_flows/business-coronavirus-support-finder.rb
kevindew: No, just conventional for smart answers. I'm not a fan of them either just continued the convention. I'll nuke them.

test/unit/calculators/business_coronavirus_support_calculator_test.rb
kevindew: Argh good catch. Yes, I find the fragility of the strings annoying and concerning. I did have a look and see if we could maybe use the enum functionality from Rails so they're more restricted but it turns out it's baked into ActiveRecord rather than ActiveModel so hard to use.

PR #4468, opened 2020-04-17
Remove Services module (and venture into the Rabbit hole)

kevindew:
This is a bit of meandering PR of me scratching an itch - sorry! hopefully it makes a few things cleaner and simpler

This removes a slightly odd concept, the fake content store, that was used for marriage abroad flattening and replaces it with calls to the live GOV.UK Content Store. Once the fake content store was gone, I then took to removing the Services module. This was rather unnecessary as it was calling the short version of GdsApi adapters already. Then I took to removing the stubbing on the existing Services module to replacing the code mocks with API mocks.


test/unit/world_location_test.rb
```diff
@@ -28,68 +31,56 @@ class WorldLocationTest < ActiveSupport::TestCase
     end
 
     should "filter out any results that don't have a slug" do
-      loc1 = world_location_details_for_slug("location-1")
-      loc2 = world_location_details_for_slug("location-2")
-      loc2["details"]["slug"] = nil
```
benthorner, line 36: I notice we're not handling this case anymore. Do we assume this can't be nil?

test/unit/world_location_test.rb
```diff
@@ -16,6 +16,9 @@ class WorldLocationTest < ActiveSupport::TestCase
       stub_worldwide_api_has_locations(@location_slugs)
 
       results = WorldLocation.all
+      assert_requested :get,
```
benthorner, line 19: It's unclear how the test was achieving it's aim before this came along!

lib/marriage_abroad_outcome_flattener.rb
```diff
@@ -22,11 +20,6 @@ def flatten
 
   attr_reader :country, :logger
 
-  def stubs_etc
```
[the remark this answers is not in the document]
kevindew: Ha oh dear - all this talk of removing the fake content store and I forgot that. Thanks good catch 👍

test/unit/world_location_test.rb
kevindew: The code under test doesn't make any special case for nil (uses present?) so I felt it didn't need to exercise both conditions.

Though I don't really want to pick too much at this as I think there's about 3 other conditionals that are part of this which aren't tested 🙈

Issue #3707, opened 2018-10-01
Graphviz representation of a smart answer flow returns 503


jennyd:
The interactive visualisations are working fine, but the Graphviz format (documented [here](https://github.com/alphagov/smart-answers/blob/f1176f991ee188294184ee9929dea5a5c93860bc/doc/debugging/visualising-flows.md)) returns 503 on all the ones I've tried.

I also can't find any tests for `GraphvizPresenter` to see what the expected output should look like.

kevindew:
Hi @jennyd - sorry you didn't hear back from anyone for a year and a half (I'm wondering if this was before or after you left GDS). The good news is this now works again and has a test.

jennyd:
Thanks @kevindew! It was after I left GDS - I can't remember exactly how I came across these visualisations being broken, but glad they're working now 🎉  They're a useful example to be able to point to of making a service more transparent. We should do something like this for forms on [MOJ's Form Builder platform](https://github.com/ministryofjustice/form-builder)...

kevindew:
Oh cool project. I'm going to have a nosey.

PR #4469, opened 2020-04-17
Fix the visualise system

kevindew:
This fixes both the JS and Graphviz abilities to see visualisation of Smart Answers. The JS version wasn't working on non-prod environments because of the content security policy and the Graphviz version was raising a template error.

I also corrected the show landscape / portrait button as this was set the wrong way round. 

Fixes https://github.com/alphagov/smart-answers/issues/3707

### DIFF
--- a/app/controllers/smart_answers_controller.rb
+++ b/app/controllers/smart_answers_controller.rb
@@ -52,7 +52,8 @@ class SmartAnswersController < ApplicationController
       end
 
       format.gv do
-        render text: GraphvizPresenter.new(@smart_answer).to_gv
+        render plain: GraphvizPresenter.new(@smart_answer).to_gv,
+               content_type: "text/vnd.graphviz"
       end
     end
   end
--- a/app/controllers/smart_answers_controller.rb
+++ b/app/controllers/smart_answers_controller.rb
@@ -11,6 +11,12 @@ class SmartAnswersController < ApplicationController
   rescue_from SmartAnswer::FlowRegistry::NotFound, with: :error_404
   rescue_from SmartAnswer::InvalidNode, with: :error_404
 
+  content_security_policy only: :visualise do |p|
+    # The script used to render the visualise tool requires eval execution
+    # unfortunately
+    p.script_src(*p.script_src, :unsafe_eval)
+  end
+
   def index
     @flows = flow_registry.flows.sort_by(&:name)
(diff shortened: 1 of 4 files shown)


app/controllers/smart_answers_controller.rb
```diff
@@ -11,6 +11,12 @@ class SmartAnswersController < ApplicationController
   rescue_from SmartAnswer::FlowRegistry::NotFound, with: :error_404
   rescue_from SmartAnswer::InvalidNode, with: :error_404
 
+  content_security_policy only: :visualise do |p|
+    # The script used to render the visualise tool requires eval execution
+    # unfortunately
+    p.script_src(*p.script_src, :unsafe_eval)
```
benthorner, line 17: It's good this is specific to the controller action. Can we make it specific to the file executed, as well? (It looks like `*p.script_src` would apply to any file.)

kevindew: It doesn't actually work like that, but I can see how you came to that impression.

It's somewhat unclear from the Rails DSL but what this does is append an additional rule to those already set for script. We need an additional rule added for script which is unsafe_eval and that isn't specified to a particular host or script just a general rule.

The eventual output would be like: `script-src 'self' *.publishing.service.gov.uk *.www.gov.uk www.gov.uk *.dev.gov.uk www.google-analytics.com ssl.google-analytics.com stats.g.doubleclick.net www.signin.service.gov.uk *.ytimg.com www.youtube.com www.youtube-nocookie.com hmrc-uk.digital.nuance.com 'unsafe-inline' 'unsafe-eval'`

benthorner: Ah, I see. Thanks for explaining. It looks like one [can have more than one CSP header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy#Multiple_content_security_policies), to accommodate multiple `script-src` tags, but only to make the policy stricter, which isn't compatible with the 'exception' scenario we have here. And [Rails doesn't support it](https://github.com/rails/rails/blob/master/actionpack/lib/action_dispatch/http/content_security_policy.rb#L166) anyway.

kevindew: Oh cool I didn't know you could do the multiple header approach.

PR #4481, opened 2020-04-22
Always show VAT deferral scheme

theseanything:
All users will see the VAT deferral scheme in the results page for the business coronavirus support finder.

### DIFF
--- a/lib/smart_answer/calculators/business_coronavirus_support_finder_calculator.rb
+++ b/lib/smart_answer/calculators/business_coronavirus_support_finder_calculator.rb
@@ -14,9 +14,6 @@ module SmartAnswer::Calculators
       job_retention_scheme: ->(calculator) {
         calculator.paye_scheme == "yes"
       },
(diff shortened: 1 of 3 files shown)


lib/smart_answer/calculators/business_coronavirus_support_finder_calculator.rb
```diff
@@ -14,8 +14,8 @@ class BusinessCoronavirusSupportFinderCalculator
       job_retention_scheme: ->(calculator) {
         calculator.paye_scheme == "yes"
       },
-      vat_scheme: ->(calculator) {
-        calculator.annual_turnover != "under_85k"
+      vat_scheme: ->(_calculator) {
```
benthorner, line 17: We could just remove this entirely: there's a precedent for that [here](https://github.com/alphagov/smart-answers/blob/master/lib/smart_answer_flows/business-coronavirus-support-finder/outcomes/results.govspeak.erb#L180).

PR #4499, opened 2020-05-01
Adding coronavirus information to parental pay calculators

OrcaTom:
Adding hints about using normal levels of pay rather than any furloughed pay (due to COVID-19) to calculate entitlements.


OrcaTom:
Review and fact check complete

PR #4502, opened 2020-05-04
Support HTML content within templates

theseanything:
Here we aim to add support for more content types within named `render_content_for` blocks. Previously it wasn't possible to specify the content type within the template. e.g content_for :body is always rendered as Govspeak.

This PR adds the format option to render_content_for method within templates.
```erb
<% render_content_for :body, format: html %>
...content...
<% end %>
```
This new option supports `:govspeak`, `:text` and `:html`. Named blocks (e.g. `:title`, `:body` etc) have individual defaults, which are outlined in the docs. Otherwise all blocks default to `:govspeak`. 

To enable this the following changes have been made:
- Removed the ability to show unprocessed govspeak markdown by appending `.txt` on URLs. This was a debugging feature, however cannot be supported as content is pre-processed into the required format before being added to the main view. It also seemed to add unnecessary complexity when the source code is available to read.
- Mandatory `error_` prefix for error message names. This is to allow a default format of :text to apply to that error message content.


test/unit/format_capture_helper_test.rb
```diff
@@ -0,0 +1,87 @@
...
+    block.call
+  end
+
+  def render(*_args, &block)
+    "<g>" + block.call + "</g>"
+  end
+end
+
+class MockClass
+  include MockCaptureHelper
```
kevindew, line 18: Not quite sure why these methods are imported in since they don't seem to be used outside this, I think you'd have same effect defining these as methods of the class.

test/unit/format_capture_helper_test.rb
```diff
@@ -0,0 +1,87 @@
...
+    context "class is extended by helper" do
+      setup do
+        @test_obj = MockClass.new
+        @test_obj.extend(SmartAnswer::Helpers::FormatCaptureHelper)
+      end
+
+      should "call super with same arguments when no block given" do
+        MockClass.any_instance.expects(:content_for).with(:name, "content", { option: :option })
+
+        @test_obj.content_for(:name, "content", { option: :option }, &nil)
```
kevindew, line 41: I'm not sure this test is doing what it says it is. Since @test_obj is an instance of MockClass (and the only instance) it looks like this is calling a method on @test_obj and then asserting that exact same method was called.

test/unit/format_capture_helper_test.rb
```diff
@@ -0,0 +1,87 @@
...
+      should "render govspeak by default" do
+        assert_captured_content("content", nil, "<g><p>content</p></g>")
+      end
+
+      should "render plain text when format set" do
+        assert_captured_content("content", :text, "content")
+      end
+
+      should "does not modify content when unknown format set" do
+        assert_captured_content("  content  ", :unknown, "  content  ")
```
kevindew, line 63: Should it raise an error in this scenario? feels like a bug

test/unit/format_capture_helper_test.rb
benthorner, line 18: Can we inline this module into the class definition?

test/unit/format_capture_helper_test.rb
```diff
@@ -0,0 +1,87 @@
...
+
+      MockClass.any_instance.expects(:content_for).with(*args)
+
+      @test_obj.content_for(:name, options) { content }
+    end
+
+    context "class is extended by helper" do
+      setup do
+        @test_obj = MockClass.new
+        @test_obj.extend(SmartAnswer::Helpers::FormatCaptureHelper)
```
benthorner, line 35: Can we do this in the class definition?

test/unit/format_capture_helper_test.rb
```diff
@@ -0,0 +1,87 @@
...
+  def content_for(_name, content, _options, &_block)
+    content
+  end
+
+  def capture(&block)
+    block.call
+  end
+
+  def render(*_args, &block)
+    "<g>" + block.call + "</g>"
```
kevindew, line 13: It has taken me a bit of a while to work out how this works - I wonder if it can be made clearer. As I understand we can assume that this is govspeak because that is the only time we call render?

It might be a bit easier to follow if it was along the lines of:

```
  def render(component, &block)
    if component == "govuk_publishing_components/components/govspeak"
      "<govspeak>#{block.call}</govspeak>"
    else
       block.call
    end
  end
```

lib/smart_answer/helpers/format_capture_helper.rb
```diff
@@ -0,0 +1,64 @@
+module SmartAnswer
+  module Helpers::FormatCaptureHelper
+    DEFAULT_FORMATS = {
+      govspeak: [/^body$/, /^post_body$/, /^next_steps$/],
+      text: [/^title$/, /^meta_description$/, /^hint$/, /^label$/, /^suffix_label$/, /^error_*./],
+    }.freeze
+
+    def content_for(name, content = nil, options = {}, &block)
+      if block_given?
```
benthorner, line 9: This confused me until I found [the interface we're trying to conform to](https://github.com/rails/rails/blob/157920aead96865e3135f496c09ace607d5620dc/actionview/lib/action_view/helpers/capture_helper.rb#L155). We should comment about that.

I think we should also match the normal behaviour of the interface, so this works with passed content; although it's not required, it confused me when trying to understand how this method is used.

test/unit/format_capture_helper_test.rb
```diff
@@ -0,0 +1,87 @@
...
+
+      should "call super with contents as options when no content given" do
+        MockClass.any_instance.expects(:content_for).with(:name, "", { option: :option })
+
+        @test_obj.content_for(:name, { option: :option }) { "" }
+      end
+
+      should "render govspeak by default" do
+        assert_captured_content("content", nil, "<g><p>content</p></g>")
+      end
```
kevindew, line 52: Might want to also make clear that it's wrapped in govspeak component.

lib/smart_answer/helpers/format_capture_helper.rb
```diff
@@ -0,0 +1,64 @@
...
+        options = content if content
+        content = capture(&block)
+
+        format = options.fetch(:format, default_format(name))
```
benthorner, line 13: Minor: we ought to delete this key from the options, before we pass to the parent.

app/views/smart_answers/question.html.erb
```diff
@@ -19,19 +19,13 @@
 
           <% if question.body.present? && show_body %>
             <article role="article">
-              <%= render "govuk_publishing_components/components/govspeak", {
-              } do %>
-                <%= question.body %>
-              <% end %>
```
kevindew, line 25: Yay this looks great 🥇 Wonder why they all have an empty object spanning two lines 🤔

test/unit/format_capture_helper_test.rb
```diff
@@ -82,6 +86,10 @@ def assert_captured_content(content, format, expected)
       should "not wrap empty content with govspeak html tags" do
         assert_captured_content("", :govspeak, "")
       end
+
+      should "remove extra trailing newlines from html" do
+        assert_captured_content("<div>content</div>\n\n\n\n", :html, "<div>content</div>")
```
kevindew, line 91: Where do these come from and how come we are concerned? can they be sorted at the source at all?

lib/smart_answer/erb_renderer.rb
```diff
@@ -14,22 +14,22 @@ def initialize(template_directory:, template_name:, locals: {}, helpers: [])
...
-      content = rendered_view.content_for(key) || ""
-      content = strip_leading_spaces(content.to_str)
-      normalize_blank_lines(content).chomp.html_safe
+      rendered_view.content_for(key) || ""
     end
 
     def content_for(key)
-      content = rendered_view.content_for(key) || ""
-      content = strip_leading_spaces(content.to_str)
-      GovspeakPresenter.new(content).html
```
benthorner: We should be able to delete this class now.

doc/smart-answers/erb-templates/question-templates.md
```diff
@@ -93,5 +98,3 @@ end
   The values represent % by weight
 <% end %>
 ```
```
kevindew, line 100: ⭐ for updating the docs

lib/smart_answer/helpers/format_capture_helper.rb
```diff
@@ -0,0 +1,64 @@
...
+        render_text(content)
+      else
+        content
+      end
+    end
+
+    def render_govspeak(content)
+      content = strip_leading_spaces(content)
+      content = Govspeak::Document.new(content, sanitize: false).to_html
+      content = content.chomp.html_safe
```
benthorner, line 48: Can you help me understand why we need to `chomp` here? I think it's new.

lib/smart_answer/erb_renderer.rb
```diff
@@ -14,22 +14,22 @@ def initialize(template_directory:, template_name:, locals: {}, helpers: [])
       @template_directory = template_directory
       @template_name = template_name
       @locals = locals
-      lookup_context = ActionView::LookupContext.new([@template_directory, FlowRegistry.instance.load_path])
+      default_view_paths = ActionController::Base.view_paths.paths.map(&:to_s)
```
benthorner, line 9: Can you help me understand this line?

app/presenters/api_presenter.rb
```diff
@@ -22,7 +22,7 @@ def payload
       if current_node.title.present?
         outcome_text = current_node.title
       else
-        strip_tags(current_node.body(html: true)).strip.lines.first
+        strip_tags(current_node.body).strip.lines.first
       end
 
       {
```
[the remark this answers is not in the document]
theseanything: Ah good catch on the comment! - I'll take a look and maybe remove in another PR.

lib/smart_answer/helpers/format_capture_helper.rb
```diff
@@ -0,0 +1,64 @@
+module SmartAnswer
+  module Helpers::FormatCaptureHelper
```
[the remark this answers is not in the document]
theseanything: Ah yeah good point! I'll try and re-organise this better - I think I was just trying to the code in to another file, because the erb_renderer.rb was getting super long.

lib/smart_answer/helpers/format_capture_helper.rb
```diff
@@ -0,0 +1,64 @@
...
+    DEFAULT_FORMATS = {
+      govspeak: [/^body$/, /^post_body$/, /^next_steps$/],
+      text: [/^title$/, /^meta_description$/, /^hint$/, /^label$/, /^suffix_label$/, /^error_*./],
+    }.freeze
```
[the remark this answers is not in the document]
theseanything: Do you mean explicitly stating the format on `content_for :body, format: :html` calls in the template?

lib/smart_answer/helpers/format_capture_helper.rb
```diff
@@ -0,0 +1,64 @@
...
+
+    def content_for(name, content = nil, options = {}, &block)
```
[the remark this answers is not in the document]
theseanything: Yup agree, I was trying limit breaking changes how existing templates are currently written, but probably safer to give it a new name and not try to deal with Rails internals.

lib/smart_answer/helpers/format_capture_helper.rb
theseanything: I'm going to try and rename this, rather than override `content_for` - should also help simplify the code as we don't have worry about the different ways `content_for` gets used.

lib/smart_answer/helpers/format_capture_helper.rb
theseanything: Ah I think the chomp was to remove newline between the content and govspeak tags from the component (line 50) i.e.
```html
<div class="govspeak-tag">
<p>content</p>
(chomp)
</div>
```

lib/smart_answer/helpers/format_capture_helper.rb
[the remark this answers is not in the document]
kevindew: Yup - though I'm a little biased because I quite liked the approach I used in https://github.com/alphagov/smart-answers/commit/92a14e31ea0ae5b297a3ffd96ee9a91448ca2315#diff-2079a30c43bb35352ff86cd242687980R25-R39 but that's probably just me being self satisfied with my own ideas.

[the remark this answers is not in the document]
theseanything: What sort of cryptic side effects could you see happening?

Overall, I was trying limit breaking changes to how templates are written. I'm keen to not add the burden of having to know what format a particular key should be when someone writes a template. Especially when certain keys must have a specific format e.g. :title is always :text. I was thinking because this behaviour is documented it should be clear what's going on. 

Also if you don't think block's this PR, we could create an issue for and revisit later?

test/unit/format_capture_helper_test.rb
theseanything: Yup - added this in.

test/unit/format_capture_helper_test.rb
theseanything: Had a look, turns out extend in the class definition will add the methods as Class methods than object methods. So will need call it on the instance object.

lib/smart_answer/helpers/format_capture_helper.rb
theseanything: So I've refactored it into a new method, not overriding content_for so hopefully that clarifies things a bit and removes the dependency to match the behaviour.

lib/smart_answer/helpers/format_capture_helper.rb
theseanything: Add this and a test

lib/smart_answer/erb_renderer.rb
theseanything: Done!

lib/smart_answer/helpers/format_capture_helper.rb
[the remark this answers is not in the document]
theseanything: I've moved them under the ErbRenderer namespace.

lib/smart_answer/helpers/format_capture_helper.rb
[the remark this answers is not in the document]
kevindew: Cryptic for me would be that code that appears quite similar has different effects. For example:

```
<% content_for :body do %>
  **Bold text**
<% end %>
```

produces a different outcome to:

```
<% content_for :label do %>
  **Bold text**
<% end %>
```
---

Your point makes sense about not wanting to change too much and it's not something I think is blocking. Though I do think you still get advantages when for a field like title that can only be text because the code is also telling you it can only be text. If you take an example of say an error message that you want to add a link or emphasis too you then need to know where to look up the output since the code doesn't tell you what it's going to do.

lib/smart_answer/erb_renderer.rb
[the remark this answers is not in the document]
theseanything: Ahah yup, definitely pinched that. Had a look and can't see an obvious nicer solution. Essentially what we are trying to do is get the govuk_publishing_components views in the view path (which is system dependent). So could do something like:
```ruby
lookup_context = ActionView::LookupContext.new([
    @template_directory,
    FlowRegistry.instance.load_path,
    File.expand_path("app/views", GovukPublishingComponents::Config.gem_directory),
])
```

But felt like this is making assumptions about where the views live inside `GovukPublishingComponents`

theseanything: Yup - my understanding when rails initialises it creates a list of view paths (places to search for templates  e.g. "app/views") when calling methods like `render`. GovukPublishingComponents gem adds its own views paths to this list. 

So here we are grabbing the view paths ("app/view", "./path/to/gem/views") that would normally be available to ActionController (i.e. normal Rails controller context). ErbRenderer is making its own ActionView::Base object (the thing responsible for rendering templates etc.), separate to the one that ActionController uses. 

We need to give ActionView::Base a list of view paths (within LookupContext) when we create a new instance. We want to copy the view paths from what we normally have in the controller to our new ActionView::Base object. Ultimately we are trying get the GovukPublishingComponents gem's views paths for our new ActionView::Base object. Admittedly we are assuming that the gems paths have been added and blindly copying everything.

Context for that line specifically:

View paths are logically just paths (strings)  - but here we are just unwrapping them from Rails objects:
```ruby
path_set = ActionController::Base.view_paths # Gets a ActionView::PathSet object (the one to be used by the ActionView::Base associated with ActionController::Base i.e. normal Rails context)
paths = path_set.paths # Get an Array of ActionView::OptimizedFileSystemResolver objects
path_strings = paths.map(&:to_s) # Calling to_string method on each OptimizedFileSystemResolver objects (return the @path attribute)
```

lib/smart_answer/helpers/format_capture_helper.rb
```diff
@@ -0,0 +1,64 @@
...
+      content = strip_leading_spaces(content)
+      content = Govspeak::Document.new(content, sanitize: false).to_html
+      content = content.chomp.html_safe
+
+      content.present? ? render("govuk_publishing_components/components/govspeak") { content } : ""
+    end
+
+    def render_text(content)
+      content = strip_leading_spaces(content)
+      normalize_blank_lines(content).strip.html_safe
```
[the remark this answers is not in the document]
theseanything: Yup - removed!

test/unit/format_capture_helper_test.rb
theseanything: Hopefully the larger refactor of the tests makes this clear now?

theseanything:
@kevindew @benthorner Thanks for the review and all the comments! Appreciate this becoming quite large and there's a lot going on - if its helpful, I can try split off smaller PRs? (The commits may be easier to follow)

> * We've got a whole ton of files with govspeak.erb extensions that can now be mixed media, shall we change all of those to .erb

Yup - happy for me to change that in a separate PR?

> * There is this somewhat quirky feature in Smart Answers where it takes the text of answers and puts them into the content item as hidden search terms (For example, the details->hidden_search_terms array in:  https://www.gov.uk/api/content/business-coronavirus-support-finder) I'm not sure how that works and/or if it uses the text handling

We should be good! The hidden_search_terms array is generated [here](https://github.com/alphagov/smart-answers/blob/57b41bea27ed793e51b101467bd4616a4265e324/app/presenters/flow_registration_presenter.rb#L66) - looks like it strips out HTML and is getting strings from the Outcome and Question Presenters, which were always rendered Govspeak anyway.

> * Another quirky thing we've got is the MarriageAbroadOutcomeFlattener, which I believe Platform Health may use, I don't know if that'll be negatively impacted at all? It's super weird so would be great to remove it

Should be good too! - it just generates partials from a yaml file, but the views that use them e.g [here](https://github.com/alphagov/smart-answers/blob/57b41bea27ed793e51b101467bd4616a4265e324/lib/smart_answer_flows/marriage-abroad/outcomes/outcome_marriage_abroad_in_country.govspeak.erb) them has been updated. The templates from which the partials are generated in the MarriageAbroadOutcomeFlattener aren't effected.

test/unit/format_capture_helper_test.rb
benthorner: Aha, yes, I see. We should be able to `include` it in the class definition, though. That would be more consistent with how we use helper modules outside of tests, too.

lib/smart_answer/helpers/format_capture_helper.rb
benthorner: Yes, that's much easier to understand. Do we still need the `if` check? I think we still have a semantic problem that this is only one half of the normal `content_for` method, which stores and retrieves content to render.

The new name is (therefore) confusing for me, since we're not actually `render`ing anything where the method is called; instead, the block is captured and a separate call must be made later to actually do the rendering onto the page.

The only name I could think of that was less ambiguous was `format_content_for`, since it can only apply to content going "in", and doesn't make sense for rendering "out" e.g. `<%= format_content_for :body %>`.

I know it sounds like a minor issue. Since this method is used all over the app, I think it makes sense to have a little more discussion on what it's called, to see if we can avoid confusing people in the future, who won't know about the helper initially.

lib/smart_answer/erb_renderer.rb
benthorner: Ah, I see. Thanks for the explanation of the mechanism. I was just missing the part at the beginning: we want to support rendering GOV.UK Publishing Components. Can add this to [the commit message](https://github.com/alphagov/smart-answers/pull/4502/commits/0983a2353b4d5456fda2c38e6f478bb664624c7a)? It's somewhat tangential to enabling raw HTML.

lib/smart_answer/helpers/format_capture_helper.rb
benthorner: Regarding [your main comment](https://github.com/alphagov/smart-answers/pull/4502#issuecomment-624840109), I'd be content for this to go into another PR.

lib/smart_answer/erb_renderer.rb
theseanything: Yup - added to the[ commit message.](https://github.com/alphagov/smart-answers/pull/4502/commits/96bdfb4dd63cd1e6e813d48292b727fdc33f0725)

lib/smart_answer/helpers/format_capture_helper.rb
theseanything: Yup - removed the `if block_given?` line.

So there might be a mis-understand, the `render_content_for` method does "render" (convert govspeak into HTML and wrap in the tags, and call any other erb render functions etc...) the content before storing to be accessed later by `content_for` - thats done on line 12 of the method.

app/presenters/api_presenter.rb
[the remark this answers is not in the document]
theseanything: #4509

test/unit/format_capture_helper_test.rb
theseanything: So for this case its consistent with how we are using the Helper modules in ErbRenderer. We only want to add these methods to a particular instance of ActiveView::Base used by ErbRenderer, don't want to include them for all instances.

test/unit/format_capture_helper_test.rb
theseanything: Agreed - the templates themselves should just be formatted properly. Not an existing problem as HTML format is new, so I've removed this behaviour.

lib/smart_answer/helpers/format_capture_helper.rb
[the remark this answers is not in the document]
theseanything: Captured in #4510

test/unit/format_capture_helper_test.rb
```diff
@@ -0,0 +1,87 @@
...
+  end
+
+  def render(*_args, &block)
+    "<g>" + block.call + "</g>"
+  end
+end
+
+class MockClass
+  include MockCaptureHelper
+end
```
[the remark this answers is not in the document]
theseanything: Class definition now a anonymous class within the setup function of the tests.

lib/smart_answer/helpers/format_capture_helper.rb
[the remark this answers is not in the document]
theseanything: The chomp was to remove extra space between the end of the user inputted govspeak and the wrapping tags from govuk_publishing_components:
```
<govuk_publishing_components>
<p>blah</p>
...(chomp)...
</govuk_publishing_components>
```

Again probably could be fixed at source.

test/unit/format_capture_helper_test.rb
theseanything: Now directly written the class definition.

test/unit/format_capture_helper_test.rb
theseanything: Yup this was some spooky behaviour - its setting up an expectation for the super method (i.e. MockClass#content_for), but here we are calling the (FormatCaptureHelper#content_for method). But with renaming `content_for` to `render_content_for` this should be clearer.

lib/smart_answer/helpers/format_capture_helper.rb
benthorner: > Yup - removed the if block_given? line.

👍 

> So there might be a mis-understand

I think I didn't express myself well. The name of the method implies it returns something, but that's not actually true. It's really a "setter". On balance, the original `content_for` is sufficiently unintuitive that there's not a lot we can do in this space.

Even `format_content_for` is still vague. My concern is that it could be easily confused with the Rails `content_for` method. Something completely different, like `smartanswer_block` would avoid that, since it's clearly specific to the app.

However, none of these suggestions accurately convey what the method does (that's also different to the Rails one), so I'll just resolve this suggestion.

test/unit/format_capture_helper_test.rb
benthorner: OK. I see where you're coming from 👍.

test/unit/format_capture_helper_test.rb
theseanything: Yup - now raises an `InvalidFormatType` exception.

lib/smart_answer/erb_renderer/format_capture_helper.rb
```diff
@@ -0,0 +1,67 @@
+module SmartAnswer
+  module ErbRenderer::FormatCaptureHelper
+    class InvalidFormatType < StandardError; end
```
kevindew, line 3: Minor, you probably want this to inherit from RuntimeError

lib/smart_answer/erb_renderer/format_capture_helper.rb
```diff
@@ -0,0 +1,67 @@
...
+      end
+    end
+
+    def render_govspeak(content)
+      content = strip_leading_spaces(content)
+      content = Govspeak::Document.new(content, sanitize: false).to_html
+      content = content.chomp.html_safe
+
+      content.present? ? render("govuk_publishing_components/components/govspeak") { content } : ""
+    end
```
kevindew, line 48: I wonder if we'll want to have a render_govspeak helper too so that we can drop chunks of govspeak into HTML sections.

lib/smart_answer/erb_renderer/question_options_helper.rb
```diff
@@ -0,0 +1,11 @@
+module SmartAnswer
+  module ErbRenderer::QuestionOptionsHelper
+    def options(options = nil)
+      if options
+        @options = options
+      else
+        @options || {}
+      end
+    end
```
kevindew, line 9: I know this is something you didn't introduce, but this makes makes my eyes pop a bit for the ambiguity of this method.

theseanything:
Thanks @kevindew!

> Wowsers - this is a huge PR (> 900 files 😵). I've only really given a lot of them a scroll through.

Yeah sorry, most of its the renaming from `content_for`.

> I do find myself a bit baffled by the render_content_for method, the naming seems a bit confusing outside of govspeak. Though I understand this naming has been what has dragged on this PR and I don't want to block your team over it. What's next for the naming discussion once this is merged?

Yup, so I've opened issue #4510 to continue discussion. I was initial hesitant to merge this, but think it's fine given the previous behaviour had the same issues (i.e. `content_for` implicitly rendered Govspeak for some sections and plaintext for others.) But there is definitely room for improvement on the syntax for these templates.

lib/smart_answer/erb_renderer/format_capture_helper.rb
theseanything: Yup - tho, will do in another PR

lib/smart_answer/erb_renderer/question_options_helper.rb
theseanything: Yup - can look at fixing in another PR

kevindew:
Great stuff - looks like this is good to merge.

PR #4512, opened 2020-05-11
Add tracking to change your answer links

theseanything:
This adds the necessary data attributes for the trackClick modules from Static to send events to GA. This will allows us to see which answers users are changing.

### DIFF
--- a/app/views/smart_answers/_previous_answer.html.erb
+++ b/app/views/smart_answers/_previous_answer.html.erb
@@ -3,16 +3,23 @@
   <td class="govuk-table__cell">
     <% if question.multiple_responses? %>
       <ul class="govuk-list">
-        <% question.response_labels(@presenter.accepted_responses[number_questions_so_far]).each do |label| %>
+        <% question.response_labels(accepted_response).each do |label| %>
           <li><%= label %></li>
         <% end %>
       </ul>
     <% else %>
-      <%= question.response_label(@presenter.accepted_responses[number_questions_so_far]) %>
+      <%= question.response_label(accepted_response) %>
     <% end %>
   </td>
   <td class="govuk-table__cell">
-    <%= link_to @presenter.change_collapsed_question_link(number_questions_so_far + 1), class: "govuk-link" do %>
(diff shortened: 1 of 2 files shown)


app/views/smart_answers/_previous_answer.html.erb
```diff
@@ -12,7 +12,14 @@
     <% end %>
   </td>
   <td class="govuk-table__cell">
-    <%= link_to @presenter.change_collapsed_question_link(number_questions_so_far + 1), class: "govuk-link" do %>
+    <%= link_to @presenter.change_collapsed_question_link(number_questions_so_far + 1),
+      class: "govuk-link",
+      data: {
+        "module": "track-click",
+        "track-category": "Smart Answer Change Link",
+        "track-action":  question.title,
```
benthorner, line 20: This is a bit different to [the other PR](https://github.com/alphagov/smart-answers/pull/4513/files#diff-85c63786058d6c48626a6c1cc46ba973R29), where the `action` really is an "action". Is this just something we have to do to get the data out? I'm wondering if we could do a combined string in the label, as a way to keep it consistent.

theseanything: Yup - make sense! I've made it a single string in the event label.

app/views/smart_answers/_previous_answers.html.erb
```diff
@@ -13,7 +13,13 @@
     </thead>
     <tbody class="govuk-table__body">
       <% @presenter.collapsed_questions.each_with_index do |question, number_questions_so_far| %>
-        <%= render partial: 'previous_answer', locals: {question: question, number_questions_so_far: number_questions_so_far} %>
```
benthorner, line 16: Just to confirm, as it's not mentioned in the commit: we're intentionally losing the `number_questions_so_far` local? It does appear to be unused in the template.

theseanything: Yup! To make it consistent we are passing in all the required attributes from `@presenter`.

PR #4516, opened 2020-05-13
Fix tracking events on coronavirus business support results

theseanything:
This removes tracking attribute from external links, as they are already tracked via the external-link module from Static. We also make the tracking events for internal links consistent with that modules where the URL is the action and the label includes the link text. Following the pattern used [here](https://github.com/alphagov/static/blob/master/doc/analytics.md#external-link-tracking-external-link-trackerjs).


theseanything:
> Code looks good.
> 
> Is this approach of category of "Internal Link Clicked" and such consistent with some sort of standard we have? if so is there a resource we can link to that explains it or shows precedence? That would probably be useful to have in the commit message.

Yup - making it consistent with the External Link Clicked pattern outlined in [Static](https://github.com/alphagov/static/blob/master/doc/analytics.md#external-link-tracking-external-link-trackerjs). Add that to the commit message.

PR #4543, opened 2020-05-26
Add email link to the business coronavirus support results

theseanything:
This adds a link for users to sign up to email alerts for all coronavirus business funding and support. This re-uses styling from the transition page and add custom CSS for the email-link app component.

### DIFF
--- a/app/assets/stylesheets/smart_answers.scss
+++ b/app/assets/stylesheets/smart_answers.scss
@@ -80,3 +80,21 @@ $is-ie: false !default;
   @include govuk-font($size: 16);
   color: $govuk-secondary-text-colour;
 }
+
+.app-c-email-link {
+  display: block;
+  @include govuk-responsive-margin(8, "top");
+  padding: govuk-spacing(6);
+  background-color: govuk-colour("light-grey", $legacy: "grey-4");
+}
+
+.app-c-email-link__icon {
+  position: relative;
+  float: left;
+  margin-right: .5em;
+  top: 2px;
+}
+
+.app-c-email-link__title {
+  margin-bottom: govuk-spacing(2);
+}
--- a/lib/smart_answer_flows/business-coronavirus-support-finder/outcomes/results.govspeak.erb
+++ b/lib/smart_answer_flows/business-coronavirus-support-finder/outcomes/results.govspeak.erb
@@ -326,3 +326,18 @@
      data-track-label="coronavirus business support page"
      href="/coronavirus/business-support">coronavirus business support page</a>.
 <% end %>
+
+<% render_content_for :body, format: :html do %>
+  <div class="app-c-email-link">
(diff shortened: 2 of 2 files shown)


theseanything:
Thanks! 💯 agree - definitely should be a component.

 >Is the two body blocks intended to be a long term thing? I can see why we've done that for this one now though I feel the append nature might be a bit confusing.

Yeah - this follows the documented behaviour of multiple "content_for" blocks, so was hoping it might not be too confusing.

It's tricky because the email-link box shouldn't be wrapped by the govspeak div. Condensing everything into single `render_content_for` block would need the govspeak content wrapped in by `render "govuk_publishing_components/govspeak"` block - which I think is somewhat confusing as we also provide the `render_content_for` block to render govspeak in the first place.

kevindew:
> Thanks! 💯 agree - definitely should be a component.
> 
> > Is the two body blocks intended to be a long term thing? I can see why we've done that for this one now though I feel the append nature might be a bit confusing.
> 
> Yeah - this follows the documented behaviour of multiple "content_for" blocks, so was hoping it might not be too confusing.
> 
> It's tricky because the email-link box shouldn't be wrapped by the govspeak div. Condensing everything into single `render_content_for` block would need the govspeak content wrapped in by `render "govuk_publishing_components/govspeak"` block - which I think is somewhat confusing as we also provide the `render_content_for` block to render govspeak in the first place.

Yeah it seems a tricky situation if we have over 90% of content in one format and a bit in another. 

I had hoped we'd approach the mixing with something like:

```
<% html_for :body do %>

  <%= render_govspeak do %>
    small bit of govspeak
  <% end %>

  <html-bits>

  <%= render_govspeak do %>
    another small bit of govspeak
  <% end %>
<% end %>
```

which seems common to approaching views (as per [content publisher](https://github.com/alphagov/content-publisher/blob/51417c5f1effe4ef5b4bc70dc8df708fa69801b3/app/views/publisher_information/publisher_updates.html.erb) but I'd agree it'd probably feel quite weird for this where we have a huge pile of govspeak in the block. 

Cleanest thing might be a partial e.g:

```
<% html_for :body do %>
  <%= render_govspeak(render("schemes", calculator: calculator)) %>
  <html-for-component>
<% end %>
```

However, don't let this discussion block you merging. Mostly musing about how this ties into new patterns.

alex-ju:
Adding a visual capture, for posterity.

Screenshot 2020-05-27 at 11 13 26

Agree with Kevin, this should be a component on its own (especially since it's being used across frontend applications) but it's hard to justify how is it different from [subscription-links](https://components.publishing.service.gov.uk/component-guide/subscription-links) and when you should use one or the other. Maybe a designer can help us with this answer.

andysellick:
@alex-ju for reference this is being built as a component in `collections` at the moment... https://github.com/alphagov/collections/pull/1757

We're also having a wider conversation (hopefully) about similar components and whether they should be merged. Maybe something worth talking about at the next catchup?

PR #4551, opened 2020-05-28
Fix issue of ErbRenderer#content_for returning nil

theseanything:
ErbRenderer#content_for was changed in commit 87ba54b97d30d2578cd0a8d1c540c48b67450249 to be delegate to the ActiveView::Base#content_for method. However this introduced a behaviour change whereby calls for named content that were not present in rendered_view returned nil instead of "". This introduced a bug where the ContentItemSyncer sent an invalid schema to the Publishing API preventing updates to start pages.

### DIFF
--- a/lib/smart_answer/erb_renderer.rb
+++ b/lib/smart_answer/erb_renderer.rb
@@ -1,7 +1,5 @@
 module SmartAnswer
   class ErbRenderer
-    delegate :content_for, to: :rendered_view
-
     def initialize(template_directory:, template_name:, locals: {}, helpers: [])
       @template_directory = template_directory
       @template_name = template_name
@@ -27,6 +25,10 @@ module SmartAnswer
 
     def relative_erb_template_path
       erb_template_path.relative_path_from(Rails.root).to_s
+    end
+
+    def content_for(name)
+      rendered_view.content_for(name) || ""
     end
 
   private
--- a/test/unit/erb_renderer_test.rb
+++ b/test/unit/erb_renderer_test.rb
@@ -113,6 +113,16 @@ Hello world
       end
     end
 
+    test "#content_for returns an empty string when content is missing" do
+      erb_template = ""
+
+      with_erb_template_file("template-name", erb_template) do |erb_template_directory|
+        renderer = ErbRenderer.new(template_directory: erb_template_directory, template_name: "template-name")
+
+        assert_equal renderer.content_for(:key), ""
+      end
+    end
+
     test "#option_text returns option text for specified key" do
       erb_template = "<% options(option_one: 'option-one-text', option_two: 'option-two-text') %>"
 


kevindew:
Do we know where this problem occurs in the ContentItemSyncer? I wonder if in the places where we require strings and not nils we run `.to_s` to show that we need a string at that point. 

It feels a little bit fishy to restore this part of the content_for method as it means this class is deviating from the subclass and the commonly understood behaviour of `content_for`

theseanything:
Good point - yeah I wasn't super happy un-delegating `content_for`.

This specific issue involves the `StartPageContentItem` and possibly `FlowContentItem` who's responsibility it is to build the payload to send to the Publishing API. The payload is built using methods on presenters [e.g.](https://github.com/alphagov/smart-answers/blob/99118ce776392cb832f10452c698bb54d0ca6624/app/presenters/start_node_presenter.rb#L10) that just call `content_for`. Didn't want to add defaults for every method:
```ruby
class StartNodePresenter
  def title
    @renderer.content_for(:title) || ""
  end

  def ab_title
    @renderer.content_for(:ab_title) || ""
  end

  def meta_description
    @renderer.content_for(:meta_description) || ""
  end

  def body
    @renderer.content_for(:body) || ""
  end

etc...
```

Adding `to_s` becomes very repetitive and tricky to keep consistent:
```ruby
def payload
    {
      base_path: base_path,
      title: flow_presenter.title.to_s, # <- shouldn't and probably is never nil
      description: flow_presenter.description.to_s,  # <-
      update_type: "minor",
      details: {
        external_related_links: external_related_links,
        introductory_paragraph: [
          {
            content: flow_presenter.start_page_body.to_s,  # <-
            content_type: "text/govspeak",
          },
        ],
        more_information: [
          content: flow_presenter.start_page_post_body.to_s,  # <- the only likely place for nil
          content_type: "text/govspeak",
        ],
        transaction_start_link: base_path + "/y",
        start_button_text: flow_presenter.start_page_button_text.to_s,  # <- weird because this is never nil anyway
        hidden_search_terms: flow_presenter.flows_content.to_s,  # <-
      },
etc...
```

ErbRenderer isn't a subclassed and we aren't overriding any behaviour, so wasn't super worried about the behaviour difference in `content_for` from ActiveView::Base. I think there is still definitely room to improve and its unfortunate to regress. We've currently got broken functionality - i.e. updating start pages that I'd like to fix before launching into a potentially larger refactor (but maybe there's an obvious easy fix?). Reverting back to existing behaviour felt safer for now.

kevindew:
Sure, can understand you not wanting to add repetitive things around. I figure you could add a method in each presenter of like `def fetch_content(name); @renderer.content_for(name).to_s; end` but you'd still need to do that in each presenter which would be a bit repetitive.

Sorry about the subclass point, I got muddled between ErbRenderer::FormatCaptureHelper being part of a view class hierarchy and this sitting alongside it 😕

theseanything:
Thanks @kevindew! Ah good idea - a method like that would also let us remove the a lot the other methods in the presenter.

PR #4552, opened 2020-05-28
Switch asset path and host

kevindew:
Trello: https://trello.com/c/oNEtjIVp/99-serve-static-assets-from-www-hostname

This configures this app to serve assets from the current host by
default rather than the GOVUK_ASSET_ROOT environment variable. This is
to support the switch of assets from assets.publishing.service.gov.uk to
www.gov.uk.

The host for assets can be specified with an ASSET_HOST env var which is
intended to be used in development environments where router isn't set
up for proxying. Unlike GOVUK_ASSET_ROOT this is expected to be app
specific hence why it doesn't have a GOVUK prefix.

### DIFF
--- a/config/initializers/slimmer.rb
+++ b/config/initializers/slimmer.rb
@@ -1,7 +1,3 @@
 SmartAnswers::Application.configure do
   config.slimmer.logger = Rails.logger
-
-  if Rails.env.development?
-    config.slimmer.asset_host = ENV["STATIC_DEV"] || Plek.new.find("static")
-  end
 end
--- a/config/application.rb
+++ b/config/application.rb
@@ -44,7 +44,11 @@ module SmartAnswers
     config.filter_parameters += [:password]
 
     # Path within public/ where assets are compiled to
-    config.assets.prefix = "/smartanswers"
+    config.assets.prefix = "/assets/smartanswers"
+
+    # allow overriding the asset host with an enironment variable, useful for
+    # when router is proxying to this app but asset proxying isn't set up.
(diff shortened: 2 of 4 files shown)


config/initializers/slimmer.rb
```diff
@@ -1,7 +1,3 @@
 SmartAnswers::Application.configure do
   config.slimmer.logger = Rails.logger
-
-  if Rails.env.development?
-    config.slimmer.asset_host = ENV["STATIC_DEV"] || Plek.new.find("static")
-  end
```
theseanything, line 6: Would this impact the use of `./startup.sh --live`? Which sets:
```
PLEK_SERVICE_STATIC_URI=${PLEK_SERVICE_STATIC_URI-assets.publishing.service.gov.uk}
```
or do we just need to change that to something like?
```
 PLEK_SERVICE_STATIC_URI=gov.uk \
```

kevindew: Thanks for taking a look 👍 

No, that doesn't need to be changed. A confusing quirk of the assets domain is it hosts 3 things:

- Rails static assets for each frontend app
- asset-manager app
- static app

This work only removes the first case of these and the weirdness that static (the app) is hosted on that domain still remains - [some more info here](https://github.com/alphagov/govuk-puppet/pull/10360). Really those files should be served off a static.publishing.service.gov.uk host to clean up some confusion, but hopefully it won't be too long until we kill static.

theseanything: Oh right - make sense! Thanks for the explanation!

PR #4561, opened 2020-06-04
Ensure link is correctly shown

beccapearce:
The link for checking the C02 emissions was not being shown correctly
on the is vehicle green page in the simplified expenses checker.
found here:
/simplified-expenses-checker/y/car/using_home_for_business/new/30000.0
 it was being shown as html instead of being changed into a hyperlink.

This is because hints are rendered as text and not html. Here the
content has been changed from hint to body so it can be processed as
html and shown correctly.

before: 
image (1)

after:
image (2)


thomasleese:
Nice catch!

PR #4564, opened 2020-06-08
Add new Business Coronavirus Risk Assessment flow

theseanything:
This adds a new flow to help users carry out a risk assessment around coronavirus for their business.

### DIFF
--- a/lib/smart_answer_flows/business-coronavirus-risk-assessment/outcomes/results.govspeak.erb
+++ b/lib/smart_answer_flows/business-coronavirus-risk-assessment/outcomes/results.govspeak.erb
@@ -1,0 +1,85 @@
+<% html_for :body do %>
+  <div class="govspeak-wrapper">
+<% end %>
+
+<% govspeak_for :body do %>
+  This guidance can help you carry out your risk assessment to make sure you keep employees and other people on site safe when opening during coronavirus (COVID-19).
+
+  <% if calculator.show?(:risk_assessment) %>
+    ## Record your risk assessment
+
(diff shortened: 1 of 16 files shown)


lib/smart_answer_flows/business-coronavirus-risk-assessment/outcomes/results.govspeak.erb
```diff
@@ -0,0 +1,85 @@
+<% html_for :body do %>
+  <div class="govspeak-wrapper">
+<% end %>
```
theseanything, line 3: We have an issue - where we need to override the default govspeak css, however we don't want to effect other smart answers. (`govspeak-wrapper` is a pre-existing css class that has the styles we want). The style changes we need are increasing the heading sizes and styling the `<hr>` more closely to the design system with larger margins.


It feels somewhat hacky and potentially confusing to have this wrapper div and also have it stated in a separate `html_for` block. It does work having the div inside the `govspeak_for` block, as it prevents the inner markdown from being rendered. This is also a lot of content to have to re-write in HTML, which also impacts how easy it is for content designers to update it.

Also thought about making the `render_govspeak` method public that excepts a block and using that to wrap the govspeak content. 

```erb
<%= render_govspeak do %>
  This guidance can help you carry out your risk assessment to make sure you keep employees and other people on site safe when opening during coronavirus (COVID-19).
<% end %>
```

But that could be possible confusing too, and raises question as to why don't we just use govuk_publishing_component: `render("govuk_publishing_components/components/govspeak") `

theseanything:
Closing in favour of #4573

PR #4579, opened 2020-06-15
Add internal link tracking helper

theseanything:
This adds a JS function that sets up GA tracking for internal links i.e. links within the same domain. It add an event listener to each link with a relevant href and sends a GA event if clicked. This is called on results pages to track user clicks and is needed by certain smart answer flows e.g.
business-coronavirus-support-finder.

This helper function makes it easier to enable internal link tracker, instead of having to manually add the correct data attribute to links to enable tracking. It will also help prevent future links from being missing in tracking, if the data attribute are forgotten to be added.

### DIFF
--- a/app/assets/javascripts/helpers.js
+++ b/app/assets/javascripts/helpers.js
@@ -19,3 +19,34 @@ function linkToTemplatesOnGithub() {
     element.removeAttr('data-debug-template-path');
   });
 };
+
+function trackInternalLinks (rootSelector) {
+  rootSelector = rootSelector || 'body';
+  var currentHost = document.location.protocol + '//' + document.location.hostname;
+  var internalLinkSelector = 'a[href^="' + currentHost + '"], a[href^="/"]';
+
+  $(rootSelector).find(internalLinkSelector).on('click', trackClickEvent);
+
+  function trackClickEvent (evt) {
+    var $link = getLinkFromEvent(evt);
+    var options = {transport: 'beacon'};
+    var href = $link.attr('href');
+    var linkText = $.trim($link.text());
+
+    if (linkText) {
+      options.label = linkText;
+    }
+
+    GOVUK.analytics.trackEvent('Internal Link Clicked', href, options);
+  };
+
+  function getLinkFromEvent (evt) {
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/helpers.js
```diff
@@ -19,3 +19,34 @@ function linkToTemplatesOnGithub() {
     element.removeAttr('data-debug-template-path');
   });
 };
+
+function trackInternalLinks (rootSelector) {
+  rootSelector = rootSelector || 'body';
+  var currentHost = document.location.protocol + '//' + document.location.hostname;
+  var internalLinkSelector = 'a[href^="' + currentHost + '"], a[href^="/"]';
+
+  $(rootSelector).on('click', internalLinkSelector, trackClickEvent);
```
DilwoarH, line 28: This adds a listener to the root - I think what will be better is if you did:
```suggestion
  $(rootSelector).find(internalLinkSelector).on('click', trackClickEvent);
```

app/views/smart_answers/result.html.erb
```diff
@@ -11,7 +11,7 @@
 
     <% outcome = @presenter.current_node %>
 
-    <div class="govuk-!-margin-bottom-6" data-debug-template-path="<%= outcome.relative_erb_template_path %>">
+    <div id="result-body" class="govuk-!-margin-bottom-6" data-debug-template-path="<%= outcome.relative_erb_template_path %>">
```
DilwoarH, line 14: I would avoid using IDs and using classes
```suggestion
    <div class="govuk-!-margin-bottom-6 result-body" data-debug-template-path="<%= outcome.relative_erb_template_path %>">
```

app/views/smart_answers/result.html.erb
```diff
@@ -43,6 +43,7 @@
           page: page
         };
         GOVUK.analytics.trackEvent('Smart Answer', 'Completed', options);
+        trackInternalLinks('div#result-body');
```
DilwoarH, line 46:
```suggestion
        trackInternalLinks('.result-body');
```

PR #4589, opened 2020-06-18
Fix wrapping of COVID-19 to not break on wrapping

theseanything:
This replaces the hyphen with a unicode non-breaking hyphen to prevent wrapping in the middle of the word COVID-19.


andysellick:
Don't know if it would work in this instance, but `&#8209;` is the character code for a non-breaking hyphen.

theseanything:
> Don't know if it would work in this instance, but `&#8209;` is the character code for a non-breaking hyphen.

Ah so do I need to have encoded version instead of the literal character? The literal character appears to work. It's a bit tricky because the templates are not HTML, they are just strings or govspeak. So if I add `&#8209;` it renders as "&#8209;" instead of the character.

PR #4590, opened 2020-06-18
Remove last updated date on pages

theseanything:
This removes "Last updated" date on the bottom of smart answer pages as it is inaccurate and potentially confusing for user.

This feature was implemented to show the last updated date to be the date when the application was last deployed. This means the last_updated date reflected any change to any smart answer and may have not even been a user facing change. This may give the user a false sense of confidence that the content in the smart answer is up to date.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,10 +1,4 @@
 module ApplicationHelper
-  def last_updated_date
-    File.mtime(Rails.root.join("REVISION")).to_date
-  rescue StandardError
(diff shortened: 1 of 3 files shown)


DilwoarH:
@theseanything - do we have anything to replace this timestamp with?

theseanything:
> @theseanything - do we have anything to replace this timestamp with?

@DilwoarH  So no, unfortunately just removing it entirely because theres no easy way at the moment to find out with a specific smart answer flow has been updated.

PR #4594, opened 2020-06-19
Update workplace exception flows

danacotoran:
## What
Update coronavirus employee risk assessment flow to reflect the updates to the guidance

## Why
Because of retail opening, there have been updates to the guidance. So we need to update the list in "Where do you work" and then we need to update the logic for exception flows. 

https://trello.com/c/dz47JOPD

https://trello.com/c/K8cKLCLa


lib/smart_answer_flows/coronavirus-employee-risk-assessment/questions/is_your_workplace_an_exception.govspeak.erb
```diff
@@ -1,22 +1,47 @@
+<%
+  render_terse_options =  false
+  workplace =  calculator.where_do_you_work
+  if workplace == "museums_or_galleries" || workplace == "cinema" || workplace == "indoor_attraction"
+    render_terse_options = true
```
DilwoarH, line 5: it might be better to just list the options here

```
case workplace
when "museums_or_galleries", "cinema", "indoor_attraction"
  options = {
    "yes": "Yes",
    "no": "No",
  }
else
  options = {
    "yes": "Yes, I work at one of these exceptions",
    "no": "No, I do not work at one of these exceptions",
  }
end
```

lib/smart_answer_flows/coronavirus-employee-risk-assessment/questions/is_your_workplace_an_exception.govspeak.erb
```diff
@@ -1,22 +1,47 @@
...
-  Is your workplace one of these exceptions?
+  <% if workplace == "museums_or_galleries" %>
+    Is your place of work a retail gallery?
+  <% elsif workplace == "cinema" %>
+    Is your place of work a drive-in cinema?
+  <% elsif workplace == "indoor_attraction" %>
+    Is your workplace an outdoor area of a zoo, safari park, wildlife centre or other animal attraction?
+  <% else %>
+    Is your workplace one of these exceptions?
+  <% end %>
```
DilwoarH, line 17: similarly here we could probably change all of this to:

```
<% case workplace %>
<% when "museums_or_galleries" %>
    Is your place of work a retail gallery?
  <% when "cinema" %>
    Is your place of work a drive-in cinema?
  <% when "indoor_attraction" %>
    Is your workplace an outdoor area of a zoo, safari park, wildlife centre or other animal attraction?
  <% else %>
    Is your workplace one of these exceptions?
  <% end %>
```

danacotoran: Thanks for the tip @DilwoarH, your suggestion simplifies the code 👍

PR #4606, opened 2020-06-24
Add GitHub PR template

theseanything:
This adds a PR template to notify users that merged changes will be automatically deployed to production.

e.g.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,3 @@
+:warning: Only merge changes if you are happy for them to go live. :warning: 
+
+This application is now [continuously deployed](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request). Merged changes are automatically deployed to staging and production.


.github/pull_request_template.md
```diff
@@ -0,0 +1,5 @@
+# What
+
+# Why
+
+:warning: This application is now continuously deployed. Merged changes are automatically deployed to staging and production. Only merge changes if you are happy for them to go live. :warning:
```
benthorner, line 5: We should refer to the docs here, to provide more context and pointers.

.github/pull_request_template.md
```diff
@@ -0,0 +1,5 @@
+# What
```
benthorner, line 1: This seems a bit opinionated. At least, this isn't how I prefer to write PRs. Could we make this change separately?

.github/pull_request_template.md
theseanything: Sure can we just add this in for now and update later?

benthorner: Hmm...I don't think this is that urgent. I was just asking to add a link to here: https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request

.github/pull_request_template.md
```diff
@@ -0,0 +1,3 @@
+:warning: Only merge changes if you are happy for them to go live. :warning:
```
DilwoarH, line 1:
```suggestion
## What
<!-- Description of the change being made -->

## Why
<!-- What are the reasons behind this change being made? -->

## Visual Changes
<!-- If the change results in visual changes, show a before and after -->


:warning: Only merge changes if you are happy for them to go live. :warning: 
```

theseanything: So I had originally added something like this, but Ben pointed out that it [maybe out of scope for this PR](https://github.com/alphagov/smart-answers/pull/4606#discussion_r444783642).

Seems like people may have differing views about PR templatse and I think it needs a bit more discussion, so I'm going to leave it for a future PR. For now it'd be good to have at least the warning about continuous delivery.

PR #4608, opened 2020-06-24
Improve print styles

theseanything:
## What
This PR improves the page styles for printing by:
- hiding the "change your answer" column in the previous answers table
- add consistent padding to cells of the previous answers table
- spans body text of result to the full width of the page

## Visual changes
image

image

### DIFF
--- a/app/assets/stylesheets/smart_answers.scss
+++ b/app/assets/stylesheets/smart_answers.scss
@@ -4,6 +4,15 @@ $is-ie: false !default;
 // The more generic SmartAnswer CSS is in Static _multi_step.scss
 
 .smart_answer {
+  @include govuk-media-query($media-type: print) {
+    .previous-answers-table {
+      th:last-child,
+      td:last-child {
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/smart_answers.scss
```diff
@@ -4,6 +4,28 @@ $is-ie: false !default;
 // The more generic SmartAnswer CSS is in Static _multi_step.scss
 
 .smart_answer {
+  @include govuk-media-query($media-type: print) {
```
DilwoarH, line 7: We should create a print CSS for the future - it will help keep things separated. 

```
<link rel="stylesheet" media="print" href="print.css">
```

theseanything: Good idea! - turns out there is already a print.scss, so I've move the styles into that file.

DilwoarH: nice!

PR #4609, opened 2020-06-24
Update retail flow

danacotoran:
Update retail exceptions question

Once users select “Retail” as their option in “Where do you work” question, they should get an “exception” question:

Your workplace can be open unless it is inside another building that is closed (for example, a gift shop inside a museum that is closed and customers cannot access it from outside).

Is your workplace inside another closed building?

- Yes (if yes, send them to - you should go to work)
- No  (If no, send them to - your workplace should be closed)

https://trello.com/c/SUfu8KbA

### DIFF
--- a/lib/smart_answer_flows/coronavirus-employee-risk-assessment.rb
+++ b/lib/smart_answer_flows/coronavirus-employee-risk-assessment.rb
@@ -84,8 +84,13 @@ module SmartAnswer
         option :no
 
         next_node do |response|
-          if response == "yes"
+          work_in_retail = calculator.where_do_you_work == "retail"
+          if response == "yes" && work_in_retail
+            outcome :workplace_should_be_closed
+          elsif response == "yes"
             question :are_you_shielding?
+          elsif work_in_retail
+            outcome :go_back_to_work
           else
             outcome :workplace_should_be_closed
           end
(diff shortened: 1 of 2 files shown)


lib/smart_answer_flows/coronavirus-employee-risk-assessment.rb
```diff
@@ -85,7 +85,13 @@ def define
 
         next_node do |response|
           if response == "yes"
-            question :are_you_shielding?
+            if calculator.where_do_you_work == "retail"
+              outcome :workplace_should_be_closed
+            else
+              question :are_you_shielding?
+            end
+          elsif calculator.where_do_you_work == "retail"
```
DilwoarH, line 93: This might be better
```ruby
          work_in_retail = calculator.where_do_you_work == "retail"

          if response == "yes" && work_in_retail
             outcome :workplace_should_be_closed
          elsif response == "yes"
             question :are_you_shielding?
          elsif work_in_retail
             outcome :go_back_to_work
          else
            outcome :workplace_should_be_closed
          end
```

danacotoran: Updated 👌🏼

PR #4653, opened 2020-07-13
Move arrested abroad flow logic to calculator class

theseanything:
This removes deprecated calculate, pre-calculate and save_input_as methods from the flow definition.

### DIFF
--- a/lib/smart_answer/calculators/arrested_abroad.rb
+++ b/lib/smart_answer/calculators/arrested_abroad.rb
@@ -3,12 +3,10 @@ module SmartAnswer::Calculators
     # created for the help-if-you-are-arrested-abroad calculator
     attr_reader :data
 
-    def initialize
-      @data = self.class.prisoner_packs
-    end
+    PRISONER_PACKS = YAML.load_file(Rails.root.join("config/smart_answers/prisoner_packs.yml")).freeze
 
     def generate_url_for_download(country, field, text)
-      country_data = @data.find { |c| c["slug"] == country }
+      country_data = PRISONER_PACKS.find { |c| c["slug"] == country }
       return "" unless country_data
 
       url = country_data[field]
@@ -26,16 +24,12 @@ module SmartAnswer::Calculators
       end
     end
 
-    def self.prisoner_packs
-      @prisoner_packs ||= YAML.load_file(Rails.root.join("config/smart_answers/prisoner_packs.yml"))
-    end
-
     def countries_with_regions
       %w[cyprus]
     end
 
     def get_country_regions(slug)
-      @data.find { |c| c["slug"] == slug }["regions"]
+      PRISONER_PACKS.find { |c| c["slug"] == slug }["regions"]
     end
   end
 end
--- a/lib/smart_answer/calculators/arrested_abroad.rb
+++ b/lib/smart_answer/calculators/arrested_abroad.rb
@@ -81,9 +81,9 @@ module SmartAnswer::Calculators
     end
 
     def has_extra_downloads
-      [police, judicial, consul, prison, lawyer, benefits, doc, pdf].count { |x|
-        x != ""
-      }.positive? || countries_with_regions.include?(country)
+      extra_downloads = [police, judicial, consul, prison, lawyer, benefits, doc, pdf]
+
+      extra_downloads.any?(&:present?) || countries_with_regions.include?(country)
     end
 
     def region_downloads
--- a/lib/smart_answer/calculators/arrested_abroad.rb
+++ b/lib/smart_answer/calculators/arrested_abroad.rb
@@ -85,13 +85,12 @@ module SmartAnswer::Calculators
     end
 
     def region_downloads
-      links = []
-      if COUNTRIES_WITH_REGIONS.include?(country)
-        regions = get_country_regions
-        regions.each_value do |val|
-          links << "- [#{val['url_text']}](#{val['link']})"
-        end
+      return "" unless COUNTRIES_WITH_REGIONS.include?(country)
+
+      links = get_country_regions.values.map do |region|
+        "- [#{region['url_text']}](#{region['link']})"
       end
+
       links.join("\n")
     end
 
--- a/lib/smart_answer/calculators/arrested_abroad.rb
+++ b/lib/smart_answer/calculators/arrested_abroad.rb
@@ -1,35 +1,104 @@
 module SmartAnswer::Calculators
   class ArrestedAbroad
     # created for the help-if-you-are-arrested-abroad calculator
-    attr_reader :data
+    attr_reader :country
 
     PRISONER_PACKS = YAML.load_file(Rails.root.join("config/smart_answers/prisoner_packs.yml")).freeze
 
-    def generate_url_for_download(country, field, text)
-      country_data = PRISONER_PACKS.find { |c| c["slug"] == country }
-      return "" unless country_data
+    def initialize(country)
+      @country = country
+    end
 
-      url = country_data[field]
-      output = []
-      if url
-        urls = url.split(" ")
-        urls.each do |u|
-          new_link = "- [#{text}](#{u})"
-          new_link += '{:rel="external"}' if u.include? "http"
-          output.push(new_link)
-        end
-        output.join("\n")
-      else
-        ""
+    def country_data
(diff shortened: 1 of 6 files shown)


lib/smart_answer/calculators/arrested_abroad.rb
```diff
@@ -1,14 +1,12 @@
...
+    attr_accessor :country
 
-    def initialize
-      @data = self.class.prisoner_packs
-    end
+    PRISONER_PACKS = YAML.load_file(Rails.root.join("config/smart_answers/prisoner_packs.yml")).freeze
 
     def generate_url_for_download(country, field, text)
-      country_data = @data.find { |c| c["slug"] == country }
+      country_data = PRISONER_PACKS.find { |c| c["slug"] == country }
```
reggieb, line 9: I think `country_data` should be a separate method. That way it can be reused (for example in `get_country_regions`), and would also ensure a consistent way of identifying the current country data from the country slug.

lib/smart_answer/calculators/arrested_abroad.rb
```diff
@@ -26,16 +24,80 @@ def generate_url_for_download(country, field, text)
...
-      @prisoner_packs ||= YAML.load_file(Rails.root.join("config/smart_answers/prisoner_packs.yml"))
-    end
-
     def countries_with_regions
       %w[cyprus]
     end
 
     def get_country_regions(slug)
-      @data.find { |c| c["slug"] == slug }["regions"]
+      PRISONER_PACKS.find { |c| c["slug"] == slug }["regions"]
```
reggieb, line 32: If `country_data` is defined as a separate method this simplified to:
```ruby
def country_regions
  country_data["regions"]
end
```

lib/smart_answer/calculators/arrested_abroad.rb
```diff
@@ -1,14 +1,12 @@
...
     # created for the help-if-you-are-arrested-abroad calculator
-    attr_reader :data
+    attr_accessor :country
 
-    def initialize
-      @data = self.class.prisoner_packs
-    end
+    PRISONER_PACKS = YAML.load_file(Rails.root.join("config/smart_answers/prisoner_packs.yml")).freeze
 
     def generate_url_for_download(country, field, text)
```
reggieb, line 8: Why is country being passed in when it's an attribute of the instance?

This class could be simplified if it was written as an object and not a collection of functional snippets.

lib/smart_answer/calculators/arrested_abroad.rb
```diff
@@ -1,14 +1,12 @@
 module SmartAnswer::Calculators
   class ArrestedAbroad
     # created for the help-if-you-are-arrested-abroad calculator
-    attr_reader :data
+    attr_accessor :country
```
reggieb, line 4: `country` should be set of initialization. It is fundamental to the way the class works. So:
```ruby
attr_reader :country

def initialize(country)
  @country = country
end
```

lib/smart_answer/calculators/arrested_abroad.rb
reggieb: This method should be rewritten to remove the use of single letter variables on multi-line code, and simply the code:
```ruby
def country_data
  @country_data ||= PRISONER_PACKS.find { |c| c["slug"] == country }
end

def generate_url_for_download(field, text)
  return "" unless country_data && country_data[field]

  urls = country_data[field].split(" ")
  output = urls.map do |url|
    new_link = "- [#{text}](#{url})"
    new_link += '{:rel="external"}' if url.include? "http"
  end
  output.join("\n")
end
```

lib/smart_answer/calculators/arrested_abroad.rb
```diff
@@ -26,16 +24,80 @@ def generate_url_for_download(country, field, text)
...
+    def organisation
+      location.fco_organisation
+    end
+
+    def country_name
+      location.name
+    end
+
+    def pdf
+      generate_url_for_download(country, "pdf", "Prisoner pack for #{country_name}")
```
reggieb, line 51: If `generate_url_for_download` is rewritten as I've suggested, `country` no longer needs to be passed into the method.

lib/smart_answer/calculators/arrested_abroad.rb
```diff
@@ -26,16 +24,80 @@ def generate_url_for_download(country, field, text)
...
+    end
+
+    def lawyer
+      generate_url_for_download(country, "lawyer", "English speaking lawyers and translators/interpreters in #{country_name}")
+    end
+
+    def has_extra_downloads
+      [police, judicial, consul, prison, lawyer, benefits, doc, pdf].count { |x|
+        x != ""
+      }.positive? || countries_with_regions.include?(country)
```
reggieb, line 85: I'd suggest:
```ruby
has_extra_downloads
  return true if [police, judicial, consul, prison, lawyer, benefits, doc, pdf].any?(&:present?)

  countries_with_regions.include?(country)
end
```

lib/smart_answer/calculators/arrested_abroad.rb
```diff
@@ -26,16 +24,80 @@ def generate_url_for_download(country, field, text)
...
+    def region_downloads
+      links = []
+      if countries_with_regions.include?(country)
+        regions = get_country_regions(country)
+        regions.each_value do |val|
+          links << "- [#{val['url_text']}](#{val['link']})"
+        end
+      end
+      links.join("\n")
+    end
```
reggieb, line 97:
```suggestion
def region_downloads
  return "" unless countries_with_regions.include?(country)
  
  links = country_regions.values.map do |region|
    "- [#{region['url_text']}](#{region['link']})"
  end
  
  links.join("\n")
end
```
    
```

File: lib/smart_answer/calculators/arrested_abroad.rb [hunk 8]
```diff
@@ -26,16 +24,80 @@ def generate_url_for_download(country, field, text)
...
+
+    def transfer_back
+      %w[austria belgium croatia denmark finland hungary italy latvia luxembourg malta netherlands slovakia].exclude?(country)
```
reggieb, line 100: This list of countries should probably be defined in a constant at the top of the class, as it's a configuration option and likely to change.

lib/smart_answer_flows/help-if-you-are-arrested-abroad.rb
```diff
@@ -7,76 +7,14 @@ def define
...
-          if arrested_calc.countries_with_regions.include?(country)
-            regions = arrested_calc.get_country_regions(country)
-            regions.each_value do |val|
-              links << "- [#{val['url_text']}](#{val['link']})"
-            end
-          end
-          links
+        on_response do |response|
+          self.calculator = Calculators::ArrestedAbroad.new
+          calculator.country = response
```
reggieb, line 17: This would be better:
```ruby
self.calculator = Calculators::ArrestedAbroad.new(response)
```

lib/smart_answer/calculators/arrested_abroad.rb
```diff
@@ -1,14 +1,12 @@
 module SmartAnswer::Calculators
   class ArrestedAbroad
     # created for the help-if-you-are-arrested-abroad calculator
```
reggieb, line 3: Is this comment still relevant?

reggieb (approved):
Looking good but I think more work is needed on the calculator.

SORRY - I didn't mean to submit this as an approval - but rather just as a collection of comments!

lib/smart_answer/calculators/arrested_abroad.rb
theseanything: Maybe not - but removing it didn't seem in scope for this PR.

lib/smart_answer/calculators/arrested_abroad.rb
theseanything: Done!

lib/smart_answer/calculators/arrested_abroad.rb
theseanything: yup - done the suggested change!

theseanything:
Thanks for the review @reggieb! I've updated the PR with your suggested changes - good catch, originally I wasn't really concerned with the logic - just a was copying and pasting from the flow definition. But it was in pretty bad shape, so thought it was worth the effort here!

reggieb (approved):
Looks good. I'm glad my comments were helpful.

PR #4654, opened 2020-07-13
Remove deprecated methods from am I getting the minimum wage flow

theseanything:
Calculate and pre-calculate methods have been deprecated. Here we replace them with logic within the on_response method block.

### DIFF
--- a/lib/smart_answer_flows/am-i-getting-minimum-wage.rb
+++ b/lib/smart_answer_flows/am-i-getting-minimum-wage.rb
@@ -14,15 +14,11 @@ module SmartAnswer
         option "current_payment"
         option "past_payment"
 
-        calculate :calculator do |response|
-          if response == "past_payment"
-            Calculators::MinimumWageCalculator.new(date: Date.parse("2019-04-01"))
-          else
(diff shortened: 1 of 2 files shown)


lib/smart_answer_flows/minimum-wage-calculator-employers.rb
```diff
@@ -36,10 +36,6 @@ def define
 
       # Q3
       value_question :how_old_are_you?, parse: Integer do
-        precalculate :age_title do
-          "How old are you?"
```
benthorner, line 40: Should there be a corresponding inline-into-view change for this?

theseanything: Oddly not - [the view doesn't use this value at all](https://github.com/alphagov/smart-answers/blob/bc167d2e94674549b93b31e8881b9c0a9bd7892a/lib/smart_answer_flows/minimum-wage-calculator-employers/questions/how_old_are_you.erb#L2).

But this change snuck in here - I'm going to remove it and re-raise apart of a PR to refactor the minimum wage calculator for employers flow

PR #4686, opened 2020-07-20
Add GitHub Actions workflow for CI

theseanything:
This PR adds a GitHub Action workflow to carry out CI tasks. This is apart of the migration from Jenkins to GitHub Actions set out by RFC [#123](https://github.com/alphagov/govuk-rfcs/blob/master/rfc-123-github-actions-ci.md). Future work outside this PR will be done to add functionality to tag releases and trigger deployments.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,51 @@
+name: CI
+on: [push, pull_request]
+jobs:
+  tests:
+    name: Test
+    runs-on: ubuntu-latest
+    env:
+      RAILS_ENV: test
+      GOVUK_CONTENT_SCHEMAS_PATH: vendor/govuk-content-schemas
+    steps:
+      - name: Checkout repository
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,51 @@
+name: CI
+on: [push, pull_request]
+jobs:
+  tests:
+    name: Test
```
thomasleese, line 5: I think this has to be named `test` otherwise we end up with a branch protection rule that requires `test` to pass, whereas actually it's `Test` that ends up passing.

https://github.com/alphagov/govuk-saas-config/blob/aebb18d9cd32664c94d2bb63d832804ffec222b8/github/lib/configure_repo.rb#L104

thomasleese: https://github.com/alphagov/smart-answers/pull/4692 should fix this.

PR #4691, opened 2020-07-22
Update state pension age question and statement for some users

edwardkerry:
State pension age for people born on or after 6th December 1953 is no longer based on being male or female, so this smart-answer no longer asks in this case and instead directs straight to the outcome.

This PR also removes some outcome text regarding the state pension age being under review for dates of birth that are outside the review.

[Trello](https://trello.com/c/ykYW1UoP/2052-3-remove-question-and-statement-for-some-users-state-pension-age-calculator)


theseanything:
Still under content review.

PR #4709, opened 2020-07-28
Allow 'Next step' button text to be changed

gclssvglx:
Allow the text on the `Next step` button to be changed via configuration in the smart answers Flow class.  

Defaults to `Next step` unless overridden.

For example...

```
module SmartAnswer
  class MySmartAnswerFlow < Flow
    def define
      # ...
      next_button_text Continue
      # ...
    end
  end
end
```

[Trello](https://trello.com/c/WFH0TtG2/372-change-next-step-buttons-to-continue)

### DIFF
--- a/app/views/smart_answers/question.html.erb
+++ b/app/views/smart_answers/question.html.erb
@@ -36,7 +36,7 @@
 
         <input type="hidden" name="next" value="1" />
         <%= render "govuk_publishing_components/components/button", {
-          text: "Next step",
+          text: @presenter.flow.next_button_text,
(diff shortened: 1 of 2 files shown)


theseanything:
@gclssvglx thanks for the PR! So it'd be preferable if the button label was specified in the question `erb` template e.g.
```erb
<%= text_for :next_button %>
Continue
<% end %>
```

This is because Smart Answer tries to keep separation of concerns, where view logic is kept out of the flow definition.

theseanything:
Replaced by #4719

PR #4710, opened 2020-07-28
Allows the previous answers section to be hidden

gclssvglx:
Allow the previous answers display to be hidden via configuration in the smart answers Flow class.  

Defaults to `true` unless overridden.

For example...

```
module SmartAnswer
  class MySmartAnswerFlow < Flow
    def define
      # ...
      show_previous_answers false
      # ...
    end
  end
end
```

[Trello](https://trello.com/c/KLGRL90U/386-allow-previous-answers-section-to-be-shown-hidden-as-needed)

### DIFF
--- a/app/views/smart_answers/question.html.erb
+++ b/app/views/smart_answers/question.html.erb
@@ -42,6 +42,8 @@
       </div>
     <% end %>
 
-    <%= render 'previous_answers' %>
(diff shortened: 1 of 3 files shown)


theseanything:
@gclssvglx not sure we need this PR anymore? Shall we close it?

PR #4714, opened 2020-07-28
Add escape button

gclssvglx:
Add escape button for vulnerable users.  When clicked, the user is taken to the BBC News website and the session data is removed.

Page caching is disabled so that any attempt to return to the smart answer returns no stored or cached data.

The button is only enabled if the flow is session-based and the `use_escape_button` flow variable is set to true.

- Add route and tests to remove session data
- Add escape button to flow pages
- Add configuration option to show button if flow is session-based

[Trello](https://trello.com/c/9cxZlrUr/367-add-leave-this-site-button-to-triage-tool-smart-answer-prototype)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -36,6 +36,7 @@ end
 
 group :development, :test do
   gem "byebug"
+  gem "jasmine", "~> 3.6.0"
   gem "pry"
   gem "rubocop-govuk"
 end
@@ -43,6 +44,7 @@ end
 group :test do
   gem "ci_reporter"
   gem "govuk_test"
+  gem "jasmine_selenium_runner", "~> 3.0.0", require: false
   gem "minitest"
   gem "minitest-focus"
   gem "mocha", require: false
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -7,7 +7,9 @@
 //= require govuk_publishing_components/components/radio
 //= require govuk_publishing_components/components/step-by-step-nav
 //= require helpers
+//= require components/escape-link
 
-$(document).ready(function() {
-  $('#current-error').focus();
-});
+window.addEventListener('DOMContentLoaded', function () {
+  var error = document.getElementById('current-error');
+  if (error) { error.focus() };
+})
--- a/app/controllers/session_answers_controller.rb
+++ b/app/controllers/session_answers_controller.rb
@@ -1,4 +1,6 @@
 class SessionAnswersController < ApplicationController
+  before_action :set_cache_headers
+
   def show
     @title = presenter.title
     @content_item = ContentItemRetriever.fetch(name) if presenter.finished?
@@ -10,7 +12,23 @@ class SessionAnswersController < ApplicationController
     redirect_to session_flow_path(id: params[:id], node_name: next_node_name)
   end
 
+  def destroy
+    session.delete(:responses)
+
+    if params[:ext_r] == "true"
+      redirect_to "https://bbc.co.uk/news"
+    else
+      redirect_to "/#{params[:id]}"
+    end
+  end
+
 private
+
+  def set_cache_headers
+    response.headers["Cache-Control"] = "no-cache, no-store"
+    response.headers["Pragma"] = "no-cache"
+    response.headers["Expires"] = "Mon, 01 Jan 1990 00:00:00 GMT"
+  end
 
   def presenter
     @presenter ||= begin
--- a/app/controllers/session_answers_controller.rb
+++ b/app/controllers/session_answers_controller.rb
@@ -67,6 +67,7 @@ private
 
     :question
   end
+  helper_method :page_type
 
   def next_node_name
     presenter.current_state.current_node.to_s
--- a/app/views/smart_answers/_escape-link.html.erb
+++ b/app/views/smart_answers/_escape-link.html.erb
@@ -1,0 +1,5 @@
+<%
+  data_attributes ||= {}
+  data_attributes[:module] = 'app-escape-link'
+%>
+<%= link_to "Leave this site", destroy_session_flow_path(ext_r: "true"), class: "app-c-escape-link govuk-link", rel: "nofollow noreferrer noopener", target: "_blank", data: data_attributes %>
--- a/app/views/smart_answers/question.html.erb
+++ b/app/views/smart_answers/question.html.erb
@@ -4,6 +4,12 @@
   <meta name="robots" content="noindex">
 <% end %>
 
+<% content_for :escape_link do %>
+  <% if @presenter.show_escape_link? %>
+    <%= render "smart_answers/escape-link" %>
+  <% end %>
+<% end %>
+
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
     <%= render 'smart_answers/debug' %>
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -15,6 +15,7 @@ Rails.application.routes.draw do
         format: false
   end
 
+  get "/:id/destroy_session", to: "session_answers#destroy", as: :destroy_session_flow
   get "/:id/:node_name", to: "session_answers#show", as: :session_flow
   get "/:id/:node_name/next", to: "session_answers#update", as: :update_session_flow
 end
--- a/lib/smart_answer/flow.rb
+++ b/lib/smart_answer/flow.rb
@@ -2,6 +2,8 @@ require "ostruct"
 
 module SmartAnswer
   class Flow
+    class NonSessionBasedFlow < StandardError; end
+
     attr_reader :nodes
     attr_accessor :need_content_id
     attr_writer :status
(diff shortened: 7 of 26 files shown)


app/views/smart_answers/landing.html.erb
```diff
@@ -38,4 +38,10 @@
       <% end %>
     </article>
   </div>
+
+  <div class="govuk-grid-column-one-third govuk-!-margin-top-9">
+    <% if @presenter.show_escape_link? %>
+      <%= render "smart_answers/escape-link" %>
+    <% end %>
+  </div>
```
theseanything, line 46: I don't think we need this - it won't be rendered on the start page. Also in production the start page is rendered by [frontend](https://github.com/alphagov/frontend), so this won't actually change how the start page looks live.

app/views/smart_answers/question.html.erb
```diff
@@ -39,4 +39,10 @@
 
     <%= render 'smart_answers/previous_answers' %>
   </div>
+
+  <div class="govuk-grid-column-one-third govuk-!-margin-top-9">
+    <% if @presenter.show_escape_link? %>
+      <%= render "smart_answers/escape-link" %>
+    <% end %>
+  </div>
```
theseanything, line 47: This doesn't quite look right on Safari - the link appears at the top of the page. Maybe putting it the layout, nested under https://github.com/alphagov/smart-answers/blob/a6d91ff8868a690d852e118cf32c415823595eb3/app/views/layouts/application.html.erb#L31

config/routes.rb
```diff
@@ -17,4 +17,5 @@
 
   get "/:id/:node_name", to: "session_answers#show", as: :session_flow
   get "/:id/:node_name/next", to: "session_answers#update", as: :update_session_flow
+  delete "/:id", to: "session_answers#destroy", as: :delete_session_flow
```
theseanything, line 20: This doesn't quite make sense as a URL. Sending a delete action to /<flow_name>, you'd expect it to delete the flow ahah. If we are changing to a link, I think also it needs to be a GET request. So maybe something like: 
```suggestion
  get "/:id/clear-session", to: "session_answers#clear_session", as: :clear_session
```

theseanything (changes requested):

theseanything:
Could we also break this work down into a few commits? (Add clear session route, add escape link component, add configuration option)

app/views/smart_answers/_escape-link.html.erb
```diff
@@ -0,0 +1 @@
+<%= button_to "Leave this site", { controller: :session_answers, action: :destroy, escape: true }, method: :delete, class: "app-c-escape-button", rel: "nofollow noreferrer noopener" %>
```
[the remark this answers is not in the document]
gclssvglx: Hi @theseanything - resolved - thanks 👍

app/assets/stylesheets/_escape-link.scss
```diff
@@ -0,0 +1,23 @@
...
+    background-color: govuk-colour('black');
+    color: govuk-colour('white');
+  }
+
+  &:hover,
+  &:focus {
+    background-color: $govuk-focus-colour;
+    color: $govuk-focus-text-colour;
+  }
+}
```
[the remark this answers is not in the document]
gclssvglx: Hi @theseanything - resolved - thanks 👍

app/views/smart_answers/landing.html.erb
gclssvglx: Hi @theseanything - resolved - thanks 👍

app/views/smart_answers/question.html.erb
gclssvglx: Hi @theseanything - resolved - thanks 👍

config/routes.rb
gclssvglx: Hi @theseanything - resolved - thanks 👍

config/routes.rb
```diff
@@ -15,6 +15,7 @@
         format: false
   end
 
+  get "/:id/clear_session", to: "session_answers#destroy", as: :delete_session_flow
```
theseanything, line 18: Nit: Think it'd be clearer to use a single consistent verb here, instead of "clear",  "destroy" and "delete"

test/controllers/session_answers_controller_test.rb
```diff
@@ -51,4 +51,35 @@ def params
       assert_match(/govuk-error-message/, response.body)
     end
   end
+
+  context "GET /:id/clear_session" do
+    setup do
+      get update_session_flow_path(id: flow_name, node_name: nodes[0]), params: params
+    end
+
+    should "be successful when the ext_r option is present and true" do
```
theseanything, line 60: Nit: we could be more specific about what successful is here?
```suggestion
    should "redirect to external site when the ext_r option is present and true" do
```

theseanything: And also for the other usages of "successful" on the test descriptions below

lib/smart_answer/flow.rb
```diff
@@ -48,6 +48,20 @@ def use_session?
...
+    end
+
+    def use_escape_button?
+      return false unless use_session?
+
+      ActiveModel::Type::Boolean.new.cast(@use_escape_button)
+    end
+
+    def show_escape_link?
+      use_session? && use_escape_button?
```
theseanything, line 62: Nit: we could raise an exception when escape button is enable, but the flow doesn't use sessions. Might be less confusing for future developers - as currently fail silently and they might not understand why.

app/controllers/session_answers_controller.rb
```diff
@@ -10,8 +12,24 @@ def update
...
+    end
+  end
+
 private
 
+  def set_cache_headers
+    response.headers["Cache-Control"] = "no-cache, no-store"
+    response.headers["Pragma"] = "no-cache"
+    response.headers["Expires"] = "Mon, 01 Jan 1990 00:00:00 GMT"
+  end
```
theseanything, line 31: Can we have tests that the cache headers are being set?

app/assets/javascripts/application.js
```diff
@@ -1,3 +1,4 @@
+//= require jquery
```
theseanything, line 1: This is a big new requirement - do we absolutely need it/ what are we using it for?

Gemfile
```diff
@@ -36,6 +36,8 @@ end
 
 group :development, :test do
   gem "byebug"
+  gem "jasmine", "~> 3.6.0"
+  gem "jasmine_selenium_runner", "~> 3.0.0", require: false
```
theseanything, line 40: Do we need this for development? or just testing?

spec/javascripts/components/escape-link.spec.js
```diff
@@ -0,0 +1,36 @@
+/* eslint-env jasmine, jquery */
+/* global EscapeLink */
+
+describe('Escape link component', function () {
```
theseanything, line 4: Are the jasmine tests being run when we call `bundle exec rake`? (thats what our CI calls to do testing)

theseanything (changes requested):

app/assets/javascripts/application.js
gclssvglx: Hi @theseanything - I totally understand.  The issue is the Jasmine tests - from the original find support app - require it!  And as we're simply copying over...

Gemfile
gclssvglx: Hi @theseanything - If you need/want to run the jasmine tests by hand (`bundle exec rake jasmine:ci` or via the browser - `bundle exec rake jasmine`) which can be very useful on occasion, then the `jasmine` gem needs to be in both `development` and `test`.  The `jasmine_selenium_runner` gem can be moved into `test`.

theseanything: Hmmm not sure I understand, because its a test dependency no matter how I run the test. For example "mintiest" is in the :test group, but I can still run tests manually.

app/assets/javascripts/application.js
theseanything: So my worry is that this means jquery is now loaded by all smart answers, regardless of whether they use them. Can we tweak the code to not use jquery?

app/controllers/session_answers_controller.rb
gclssvglx: Hi @theseanything - thanks.  Tests added.

lib/smart_answer/flow.rb
gclssvglx: Hi @theseanything - thanks.  Exception added and called. Test updated.

config/routes.rb
gclssvglx: Hi @theseanything - thanks.  Route renamed, tests updated.  Link updated - in later commit.

test/controllers/session_answers_controller_test.rb
gclssvglx: Hi @theseanything - thanks.  Wording updated.

app/assets/javascripts/application.js
gclssvglx: I don't believe the escape_link code or tests use JQuery at all.  So I'm a tad confused too.  It seems to be a Jasmine requirement as without it you get `Uncaught ReferenceError: $ is not defined`!

Gemfile
gclssvglx: Hi @theseanything - thanks.  I've moved these into the `test` only section.  This means that it is no longer possible to run `bundle exec rake jasmine:ci` directly.  However.... (see https://github.com/alphagov/smart-answers/pull/4714#discussion_r486211581)

spec/javascripts/components/escape-link.spec.js
gclssvglx: Hi @theseanything - thanks.  I've added a rake task to run the Javascript tests and updated the `Rakefile` to call that automatically.  So, it's now possible to just run `bundle exec rake` and the JS tests will run as part of that.

app/assets/javascripts/application.js
theseanything: Removed the jQuery code and removed the dependency.

theseanything (approved):

PR #4715, opened 2020-07-28
Add find support form

gclssvglx:
Convert the coronavirus find support service (otherwise known as the triage tool) from a standalone service to a smart answer service.

[Trello](https://trello.com/c/vlwPnMuk/394-convert-existing-triage-tool-to-smart-answers-format)

### DIFF
--- a/lib/smart_answer/calculators/coronavirus_find_support_calculator.rb
+++ b/lib/smart_answer/calculators/coronavirus_find_support_calculator.rb
@@ -1,0 +1,96 @@
+module SmartAnswer::Calculators
+  class CoronavirusFindSupportCalculator
+    attr_accessor :nation,
+                  :need_help_with,
+                  :feel_safe,
+                  :afford_rent_mortgage_bills,
+                  :afford_food,
+                  :get_food,
+                  :able_to_go_out,
+                  :self_employed,
+                  :worried_about_work,
+                  :have_somewhere_to_live,
+                  :have_you_been_evicted,
+                  :are_you_off_work_ill,
+                  :have_you_been_made_unemployed,
+                  :mental_health_worries
+
+    def has_results?
+      need_help_with.present?
+    end
+
+    def needs_help_with?(given_help_item)
+      return false unless has_results?
+
+      need_help_with.split(",").include? given_help_item
+    end
+
+    def needs_help_in?(given_nation)
+      return false if nation.blank?
+
+      nation.split(",").include? given_nation
+    end
+
+    def user_feels_unsafe?
+      needs_help_with?("feeling_unsafe") && feel_safe != "yes"
+    end
+
+    def user_cannot_pay_their_bills?
+      needs_help_with?("paying_bills") && afford_rent_mortgage_bills != "no"
+    end
+
+    def user_cannot_get_food?
+      needs_help_with?("getting_food") && (afford_food != "no" && get_food != "yes")
+    end
+
+    def user_is_worried_about_going_to_work?
+      needs_help_with?("going_to_work") && worried_about_work != "no"
+    end
+
+    def user_is_unemployed?
+      needs_help_with?("being_unemployed") && (
+        (self_employed != "yes" && have_you_been_made_unemployed != "no") || are_you_off_work_ill == "yes"
+      )
+    end
+
+    def user_needs_somewhere_to_live?
+      needs_help_with?("somewhere_to_live") && (
+        have_somewhere_to_live != "yes" || have_you_been_evicted != "no"
+      )
+    end
+
+    def user_has_mental_health_worries?
+      needs_help_with?("mental_health") && mental_health_worries != "no"
+    end
+
+    def next_question(current_node)
+      nodes = %i[
+        need_help_with?
+        feel_safe?
+        afford_rent_mortgage_bills?
+        get_food?
+        are_you_off_work_ill?
+        worried_about_work?
+        have_you_been_evicted?
+      ]
+
+      if nodes.slice(0..0).include?(current_node) && needs_help_with?("feeling_unsafe")
+        :feel_safe?
+      elsif nodes.slice(0..1).include?(current_node) && needs_help_with?("paying_bills")
+        :afford_rent_mortgage_bills?
+      elsif nodes.slice(0..2).include?(current_node) && needs_help_with?("getting_food")
+        :afford_food?
+      elsif nodes.slice(0..3).include?(current_node) && needs_help_with?("being_unemployed")
+        :self_employed?
+      elsif nodes.slice(0..4).include?(current_node) && needs_help_with?("going_to_work")
+        :worried_about_work?
+      elsif nodes.slice(0..5).include?(current_node) && needs_help_with?("somewhere_to_live")
+        :have_somewhere_to_live?
+      elsif nodes.slice(0..6).include?(current_node) && needs_help_with?("mental_health")
+        :mental_health_worries?
+      else
+        :nation?
+      end
+    end
+  end
+end
--- a/lib/smart_answer_flows/coronavirus-find-support.rb
+++ b/lib/smart_answer_flows/coronavirus-find-support.rb
@@ -1,0 +1,277 @@
+module SmartAnswer
+  class CoronavirusFindSupportFlow < Flow
+    def define
+      name "coronavirus-find-support"
+      start_page_content_id "6a6ab3c9-4612-4764-8f2f-2c534c3c6b19"
+      flow_content_id "31d81949-aa15-48cf-a7f1-f0d0a670e8db"
+      status :draft
+
+      # ======================================================================
+      # What do you need help with because of coronavirus?
+      # ======================================================================
+      checkbox_question :need_help_with? do
+        option :feeling_unsafe
+        option :paying_bills
+        option :getting_food
+        option :being_unemployed
+        option :going_to_work
+        option :somewhere_to_live
+        option :mental_health
+        option :not_sure
+
+        on_response do |response|
+          self.calculator = Calculators::CoronavirusFindSupportCalculator.new
+          calculator.need_help_with = response
+        end
+
+        next_node do
+          question calculator.next_question(:need_help_with?)
+        end
+      end
+
+      # ======================================================================
+      # Do you feel safe where you live?
+      # ======================================================================
+      multiple_choice :feel_safe? do
+        option :yes
+        option :yes_but_i_am_concerned_about_others
+        option :no
+        option :not_sure
+
+        on_response do |response|
+          calculator.feel_safe = response
+        end
+
+        next_node do
+          question calculator.next_question(:feel_safe?)
+        end
+      end
+
+      # ======================================================================
+      # Are you finding it hard to afford rent, your mortgage or bills?
+      # ======================================================================
+      multiple_choice :afford_rent_mortgage_bills? do
+        option :yes
+        option :no
+        option :not_sure
+
+        on_response do |response|
+          calculator.afford_rent_mortgage_bills = response
+        end
+
+        next_node do
+          question calculator.next_question(:afford_rent_mortgage_bills?)
+        end
+      end
+
+      # ======================================================================
+      # Are you finding it hard to afford food?
+      # ======================================================================
+      multiple_choice :afford_food? do
+        option :yes
+        option :no
+        option :not_sure
+
+        on_response do |response|
+          calculator.afford_food = response
+        end
+
+        next_node do
+          question :get_food?
+        end
+      end
+
+      # ======================================================================
+      # Are your able to get food?
+      # ======================================================================
+      multiple_choice :get_food? do
+        option :yes
+        option :no
+        option :not_sure
+
+        on_response do |response|
+          calculator.get_food = response
+        end
+
+        next_node do
+          if calculator.get_food == "yes"
+            question calculator.next_question(:get_food?)
+          else
+            question :able_to_go_out?
+          end
+        end
+      end
+
+      # ======================================================================
+      # Are you able to leave your home for food, medicine, or health reasons?
+      # ======================================================================
+      multiple_choice :able_to_go_out? do
+        option :yes
+        option :no_i_have_coronavirus
+        option :no_i_have_a_medical_condition
+        option :no_i_have_a_disability
+        option :no_for_another_reason
+
+        on_response do |response|
+          calculator.able_to_go_out = response
+        end
+
+        next_node do
+          question :self_employed?
+        end
+      end
+
+      # ======================================================================
+      # Are you self-employed or a sole trader?
+      # ======================================================================
+      multiple_choice :self_employed? do
+        option :yes
+        option :no
+        option :not_sure
+
+        on_response do |response|
+          calculator.self_employed = response
+        end
+
+        next_node do
+          if calculator.self_employed == "yes"
+            question :worried_about_work?
+          else
+            question :have_you_been_made_unemployed?
+          end
+        end
+      end
+
+      # ======================================================================
+      # Have you been told to stop working?
+      # ======================================================================
+      multiple_choice :have_you_been_made_unemployed? do
+        option :yes_i_have_been_made_unemployed
+        option :yes_i_have_been_put_on_furlough
+        option :no
+        option :not_sure
+
+        on_response do |response|
+          calculator.have_you_been_made_unemployed = response
+        end
+
+        next_node do
+          if %w[yes_i_have_been_made_unemployed yes_i_have_been_put_on_furlough].include? calculator.have_you_been_made_unemployed
+            question :worried_about_work?
+          else
+            question :are_you_off_work_ill?
+          end
+        end
+      end
+
+      # ======================================================================
+      # Are you off work because you're ill or self-isolating?
+      # ======================================================================
+      multiple_choice :are_you_off_work_ill? do
+        option :yes
+        option :no
+
+        on_response do |response|
+          calculator.are_you_off_work_ill = response
+        end
+
+        next_node do
+          question calculator.next_question(:are_you_off_work_ill?)
+        end
+      end
+
+      # ======================================================================
+      # Are you worried about going in to work?
+      # ======================================================================
+      multiple_choice :worried_about_work? do
+        option :yes
+        option :no
+        option :not_sure
+
+        on_response do |response|
+          calculator.worried_about_work = response
+        end
+
+        next_node do
+          question calculator.next_question(:worried_about_work?)
+        end
+      end
+
+      # ======================================================================
+      # Do you have somewhere to live?
+      # ======================================================================
+      multiple_choice :have_somewhere_to_live? do
+        option :yes
+        option :yes_but_i_might_lose_it
+        option :no
+        option :not_sure
+
+        on_response do |response|
+          calculator.have_somewhere_to_live = response
+        end
+
+        next_node do
+          question :have_you_been_evicted?
+        end
+      end
+
+      # ======================================================================
+      # Have you been evicted?
+      # ======================================================================
+      multiple_choice :have_you_been_evicted? do
+        option :yes
+        option :yes_i_might_be_soon
+        option :no
+        option :not_sure
+
+        on_response do |response|
+          calculator.have_you_been_evicted = response
+        end
+
+        next_node do
+          question calculator.next_question(:have_you_been_evicted?)
+        end
+      end
+
+      # ======================================================================
+      # Are you worries about your mental health or someone else's mental health?
+      # ======================================================================
+      multiple_choice :mental_health_worries? do
+        option :yes
+        option :no
+        option :not_sure
+
+        on_response do |response|
+          calculator.mental_health_worries = response
+        end
+
+        next_node do
+          question :nation?
+        end
+      end
+
+      # ======================================================================
+      # Where do you live?
+      # ======================================================================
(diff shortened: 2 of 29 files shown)


test/integration/smart_answer_flows/coronavirus_find_support_test.rb
```diff
@@ -0,0 +1,106 @@
...
+      assert_current_node :have_somewhere_to_live?
+      add_response "no"
+      assert_current_node :have_you_been_evicted?
+      add_response "yes"
+      assert_current_node :mental_health_worries?
+      add_response "yes"
+      assert_current_node :nation?
+      add_response "england,scotland,wales,northern_ireland"
+      assert_current_node :results
+    end
```
reggieb, line 104: I think there are three two-way branches in the logic, so I think that means six possible happy pathways through the flow. As there are so few specs, there should at least be one for each happy pathway.

lib/smart_answer_flows/coronavirus-find-support/outcomes/_feeling_unsafe.erb
```diff
@@ -0,0 +1,61 @@
+## Feeling unsafe
+
+### If you do not feel safe where you live or you’re worried about someone else
+
+If you’re in immediate danger call 999 and ask for the Police.
+
+If you’re in danger and unable to talk on the phone, call 999, then press 55 when prompted.
+
+Call the National Domestic Abuse Helpline on <a href="tel:0808 2000 247" class="govuk-link">0808 2000 247</a>, you can get advice in different languages. You can also <a href="https://www.nationaldahelpline.org.uk/Chat-to-us-online" class="govuk-link">get help online</a>.
```
reggieb, line 9: These links need a class defined. So need to be HTML. Hmm.

lib/smart_answer_flows/coronavirus-find-support.rb
```diff
@@ -0,0 +1,277 @@
+module SmartAnswer
+  class CoronavirusFindSupportFlow < Flow
+    def define
+      name "coronavirus-find-support"
```
theseanything, line 4: So this becomes the base path - we need to check with the team if we are happy with this. Ideally, if we had something different it'd prevent collisions with the existing path and make things a bit easier to transition.

gclssvglx: Ah!  Thanks for pointing that out @theseanything - I wondered what purpose that served.  👍

test/integration/smart_answer_flows/coronavirus_find_support_test.rb
gclssvglx: Thanks @reggieb 👍  - I was thinking that I'd missed a number of paths through the app.  I'll add more tests and appropriate logic.

lib/smart_answer_flows/coronavirus-find-support/outcomes/_being_unemployed.erb
```diff
@@ -0,0 +1,128 @@
+## Being unemployed
+
+### If you’ve been made redundant or unemployed, or put on temporary leave (on furlough)
+
+<a href="https://www.gov.uk/guidance/coronavirus-covid-19-what-to-do-if-you-were-employed-and-have-lost-your-job">What to do if you’ve lost your job</a>
```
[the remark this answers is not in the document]
gclssvglx: Thanks @reggieb - yes, my bad - will fix!

test/unit/calculators/coronavirus_find_support_calculator_test.rb
```diff
@@ -0,0 +1,534 @@
+require_relative "../../test_helper"
+
+module SmartAnswer::Calculators
+  class CoronavirusFindSupportCalculatorTest < ActiveSupport::TestCase
+    setup do
+      @calculator = CoronavirusFindSupportCalculator.new
+    end
+
+    context "#needs_help_with?" do
```
reggieb, line 9: Pedantic perhaps, but the content of this block described the behaviour of the `needs_help_with?` instance method of the class. So I think these blocks should be `describe` rather than `context`. So:
```ruby
describe "#needs_help_with?" do
```
The same for the other instance method blocks.

lib/smart_answer/calculators/coronavirus_find_support_calculator.rb
```diff
@@ -0,0 +1,66 @@
...
+    end
+
+    def needs_help_in?(given_nation)
+      return false if nation.blank?
+
+      nation.split(",").include? given_nation
+    end
+
+    def user_feels_unsafe?
+      needs_help_with?("feeling_unsafe") and feel_safe != "yes"
```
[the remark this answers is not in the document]
gclssvglx: Thanks @reggieb - will fix 👍

lib/smart_answer/calculators/coronavirus_find_support_calculator.rb
```diff
@@ -0,0 +1,66 @@
...
+      needs_help_with?("getting_food") and (afford_food != "no" and get_food != "yes")
+    end
+
+    def user_is_worried_about_going_to_work?
+      needs_help_with?("going_to_work") and worried_about_work != "no"
+    end
+
+    def user_is_unemployed?
+      needs_help_with?("being_unemployed") and (
+        (self_employed != "yes" and have_you_been_made_unemployed != "no") or are_you_off_work_ill == "yes"
```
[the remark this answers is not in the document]
gclssvglx: Thanks @reggieb - will fix 👍

test/unit/calculators/coronavirus_find_support_calculator_test.rb
```diff
@@ -0,0 +1,534 @@
...
+      end
+
+      should "return false if nation is nil" do
+        @calculator.nation = nil
+        assert_equal @calculator.needs_help_in?("one"), false
+      end
+    end
+
+    context "#user_feels_unsafe?" do
+      context "user has selected feeling_unsafe" do
```
reggieb, line 71: I think this should be:
```ruby
    describe "#user_feels_unsafe?" do
      context "user has selected feeling_unsafe" do
```
The outer block is describing the behaviour of the method, and the inner block wraps a section where that behaviour is altered by an external (to the method) change to the context within which the method is being called.

reggieb:
Please ignore my earlier comment  

> I also think there should be a spec for the new calculator.

The Github view had hidden the calculator spec because the file is so large, and I failed to spot that. I've now gone through the calculator specs and have added a couple of very minor comments.

lib/smart_answer_flows/coronavirus-find-support.rb
```diff
@@ -0,0 +1,277 @@
...
+
+        next_node do
+          question :nation?
+        end
+      end
+
+      # ======================================================================
+      # Where do you live?
+      # ======================================================================
+      checkbox_question :nation? do
```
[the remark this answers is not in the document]
gclssvglx: Thanks @theseanything - no, it has to handle folks on borders, those who cross borders, and multiple dwellings in multiple locations.

lib/smart_answer_flows/coronavirus-find-support.rb
```diff
@@ -0,0 +1,277 @@
...
+        option :not_sure
+
+        on_response do |response|
+          self.calculator = Calculators::CoronavirusFindSupportCalculator.new
+          calculator.need_help_with = response
+        end
+
+        next_node do
+          question :feel_safe?
+        end
```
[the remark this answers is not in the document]
gclssvglx: Thanks @theseanything - yes, something I missed (DOH!) - am in the process of fixing.

theseanything:
@gclssvglx so sorry accidentally closed the PR!

test/unit/calculators/coronavirus_find_support_calculator_test.rb
gclssvglx: Thanks @reggieb - I agree, it should be `describe`.  However, `describe` throws an error!

test/unit/calculators/coronavirus_find_support_calculator_test.rb
gclssvglx: Thanks @reggieb - I agree, it should be `describe`.  However, `describe` throws an error!

lib/smart_answer_flows/coronavirus-find-support.rb
```diff
@@ -0,0 +1,277 @@
...
+      # ======================================================================
+      multiple_choice :able_to_go_out? do
+        option :yes
+        option :no_1
+        option :no_2
+        option :no_3
+        option :no_4
+
+        on_response do |response|
+          calculator.able_to_go_out = response
```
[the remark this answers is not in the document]
gclssvglx: Thanks @theseanything - my bad - I forgot to add a whole bunch of paths through the flow.  Will fix.

lib/smart_answer_flows/coronavirus-find-support.rb
```diff
@@ -0,0 +1,277 @@
...
+          question :have_somewhere_to_live?
+        end
+      end
+
+      # ======================================================================
+      # Do you have somewhere to live?
+      # ======================================================================
+      multiple_choice :have_somewhere_to_live? do
+        option :yes
+        option :yes_but
```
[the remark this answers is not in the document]
gclssvglx: Thanks @theseanything - yes, they could be better - will fix. 👍

lib/smart_answer_flows/coronavirus-find-support.rb
```diff
@@ -0,0 +1,277 @@
...
+          end
+        end
+      end
+
+      # ======================================================================
+      # Have you been told to stop working?
+      # ======================================================================
+      multiple_choice :have_you_been_made_unemployed? do
+        option :yes_1
+        option :yes_2
```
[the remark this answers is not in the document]
gclssvglx: Thanks @theseanything - yes, they could be better - will fix. 👍

lib/smart_answer_flows/coronavirus-find-support.rb
```diff
@@ -0,0 +1,277 @@
...
+          question :feel_safe?
+        end
+      end
+
+      # ======================================================================
+      # Do you feel safe where you live?
+      # ======================================================================
+      multiple_choice :feel_safe? do
+        option :yes
+        option :yes_but
```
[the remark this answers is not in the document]
gclssvglx: Thanks @theseanything - yes, they could be better - will fix. 👍

test/unit/calculators/coronavirus_find_support_calculator_test.rb
reggieb: That's weird. The minitest documentation suggests it should work. https://github.com/seattlerb/minitest#specs-

reggieb: What happens if you add `require 'minitest/spec'` to the start of the file?

lib/smart_answer_flows/coronavirus-find-support/coronavirus_find_support.erb
```diff
@@ -0,0 +1,23 @@
...
+  * paying bills, rent, or your mortgage
+  * getting food
+  * being made redundant or unemployed, or not having any work
+  * what to do if you’re worried about going in to work
+  * having somewhere to live
+  * mental health and wellbeing, including information for children
+
+  <div role="note" aria-label="Information" class="application-notice info-notice">
+    <p>If you need medical help, <a rel="external" href="https://www.nhs.uk/conditions/coronavirus-covid-19/check-if-you-have-coronavirus-symptoms/">go to NHS 111 online</a>.</p>
+  </div>
```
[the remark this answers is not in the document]
gclssvglx: Thanks @reggieb - will turn into Markdown.

lib/smart_answer_flows/coronavirus-find-support/outcomes/_being_unemployed.erb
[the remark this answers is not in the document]
gclssvglx: Thanks @reggieb - will turn into Markdown.

lib/smart_answer_flows/coronavirus-find-support/outcomes/_feeling_unsafe.erb
gclssvglx: Thanks @reggieb - will turn into Markdown.

lib/smart_answer_flows/coronavirus-find-support/coronavirus_find_support.erb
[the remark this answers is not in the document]
gclssvglx: Thanks @theseanything  - will turn into Markdown.

test/unit/calculators/coronavirus_find_support_calculator_test.rb
gclssvglx: Either I get `undefined method 'assert_current_node'` or `undefined method 'need_help_with='` (depending on the spec file).  Looks like the issue is a bunch of `include`s have been added to `ActiveSupport::TestCase`.  I don't believe it's a valuable use of time to fix this, given it's just a semantic difference between `context` and `describe`.

reggieb: I agree

test/integration/smart_answer_flows/coronavirus_find_support_test.rb
```diff
@@ -0,0 +1,2540 @@
...
+require "smart_answer_flows/coronavirus-find-support.rb"
+
+class CoronavirusFindSupportFlowTest < ActiveSupport::TestCase
+  include FlowTestHelper
+
+  setup do
+    setup_for_testing_flow SmartAnswer::CoronavirusFindSupportFlow
+  end
+
+  context "all combinations of all options" do
```
reggieb, line 13: I don't think all the options need to be tested here. Following a discussion with Graham, I think there are 7 key questions that lead to changes in the user journey through the flow. Each of these 7 is independent of the others (if user selects X they are taken through subsection x no matter how they answered questions Y and/or Z). I would suggest therefore that the following flows are tested:

- One each for the 7 options feeling_unsafe paying_bills getting_food being_unemployed going_to_work, somewhere_to_live and mental_health, where only that option is selected and therefore only a single subsection of the flow is inserted between the start and the end of the flow.
- One for the journey where none of the options are selected.
- One for the journey where all the options are selected (to check they combine together corrently)

gclssvglx: Thanks @reggieb 👍  

Generally, I feel if you can test all possible options, then you should.  But I totally understand that isn't always a pragmatic approach.  

I'm not 100% happy to remove all but the 9 tests you indicate, as I feel we are no longer testing all paths and transitions which could expose us to 'incorrect' passing tests.  I don't want to create the potential for subtle and difficult to fix and find bugs!

Having said that, I do see the issue with the volume of tests and the maintenance of them, and I share your concerns on that.

Can we compromise?  A single test - that covers all 128 paths and transitions, it will also handle all changes we might make to the flow, the `next_node` blocks and the questions.  Of course, this approach requires a loop that goes around each possible option, transition and test them in-turn - this is how I generated the 128 tests in the first instance - you didn't think I typed them all in right ;-)  I know this type of test is more difficult to maintain and debug, and requires sufficient output to identify a failing test when one is generated.  I'm also aware that some may find this approach undesirable - it's probably not the "GDS way"!  However, in this instance it may be the best - most pragmatic approach.

reggieb: OK - really controversial. How about you wrapping up the script you used to generate this spec, in a rake task - so that it can be easily replicated and updated. Would that be an acceptable compromise? That way you get the coverage of all happy journey plus a way to maintain it.

gclssvglx: More than happy with that approach - good shout!  Will add it 👍

gclssvglx: Hi @reggieb I've added the test generation script.  

Initially, I attempted to create a generic script to create tests for all combinations of any and all smart answers.  While I feel this is possible and I got a fair way to doing it, I don't feel it should be done as part of this story and will take more time than I feel we have right now.

lib/smart_answer_flows/coronavirus-find-support.rb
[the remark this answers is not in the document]
theseanything: Would there be a way to create a function to calculate the next node from `current question` and `needs_help_with` values? Might help remove the repetition in the `next_node` blocks - which I could see being a pain to refactor if the sections changed. Also would allow us to write unit tests - rather than relying on integration tests to check every path.

lib/tasks/maintenance.rake
```diff
@@ -0,0 +1,88 @@
+namespace :maintenance do
+  desc "Create full combination flow tests for the Coronavirus Find Support smart answer."
```
reggieb, line 2: The description is fine, but I think some comments would be good - even just an example of usage.

lib/tasks/maintenance.rake
```diff
@@ -0,0 +1,88 @@
...
+      paying_bills
+      getting_food
+      being_unemployed
+      going_to_work
+      somewhere_to_live
+      mental_health
+    ]
+
+    test_counter = 0
+    file = File.open(ENV["output_file"], "w")
```
reggieb, line 18: I think it is better to use a block with `File.open` rather than depending on `file.close`. I believe the block version is a little better at dealing with an unexpected halt (as closing occurs no matter how the block is terminated), and it is also less susceptible to the `file.close` being accidentally removed by a later edit. Having said that the garbage collector should sort this out at the end of the task call anyway - so not essential.

So:

```ruby
File.open(ENV["output_file"], "w") do |file|
```
Then on line 85, replace `file.close` with `end`

lib/smart_answer_flows/coronavirus-find-support.rb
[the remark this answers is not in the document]
gclssvglx: Umm!  Thanks @theseanything should be possible - I'll take a look.  Thanks for the suggestion 👍

test/integration/smart_answer_flows/coronavirus_find_support_test.rb
theseanything: Sorry I'm late to this thread - I few a concerns about having integration tests for every path.
- Integration tests are expensive - how much extra time does it take to run these test?
- Is the added complexity of rake task to generate the test file worth the benefit we have from these extra tests?
- How do future developer know to run the rake task, rather than update the test file directly?
- Does this scale? What happens if we were to add another more questions or options? I can see us easily ending up with 1000+ scenarios if we were to add a more questions or options.
- We don't test every possible case in other smart answer flows, adding inconsistency in how we write tests, is this going to be confusing for future developers?
- What are we testing here exactly? Is it the next_node routing logic? If we made a help function (suggest in a comment above) - we could write unit tests to cover that more comprehensively.

lib/smart_answer_flows/coronavirus-find-support/outcomes/_paying_bills.erb
```diff
@@ -0,0 +1,45 @@
...
+  [What to do if you cannot pay your bills or mortgage, including if you cannot go out to pay them (Citizens Advice)](https://www.citizensadvice.org.uk/debt-and-money/if-you-cant-pay-your-bills-because-of-coronavirus/)
+<% end %>
+
+<% if calculator.needs_help_in?("wales") %>
+  [What to do if you cannot pay your rent or mortgage (Shelter Cymru)](https://sheltercymru.org.uk/get-advice/coronavirus/)
+
+  [What to do if you cannot pay your bills or mortgage, including if you cannot go out to pay them (Citizens Advice)](https://www.citizensadvice.org.uk/wales/debt-and-money/if-you-cant-pay-your-bills-because-of-coronavirus/)
+<% end %>
+
+#### Support and advice
```
theseanything, line 33: Might not be an issue - but h3 and h4 are styled the same for govspeak.

lib/smart_answer_flows/coronavirus-find-support.rb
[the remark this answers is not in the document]
gclssvglx: Hi @theseanything - thanks for the suggestion, I prefer the new approach 👍   

However, there is one slight issue - because (like `benefit-cap-calculator`) the smart answer is using a method to obtain the next question in the `next_node` block, I have had to add it to the `GraphPresenter` `EXEMPTIONS_LIST`.  The flow does still actually render (again like the `benefit-cap-calculator`) albeit it not quite correctly.  I feel this is a bug with the `Flow` not being able to run or follow the path correctly.  Something that probably needs addressing, but not as part of this PR.

lib/tasks/maintenance.rake
gclssvglx: This file is no longer required.

lib/tasks/maintenance.rake
gclssvglx: This file is no longer required.

lib/smart_answer_flows/coronavirus-find-support/outcomes/_paying_bills.erb
gclssvglx: I don't understand why would we do that? It doesn't make sense.  I guess if it is a problem we can make the H4 an H5 instead, assuming H5's aren't styled the same as H3 too!

PR #4722, opened 2020-07-30
Make app consistent in upgrade

cdccollins:

This makes Smart Answers more consistent with the other apps that we've been upgrading to Rails 6. 

One thing that I began to look at but seemed incredibly complicated was the autoloading. It isn't great to autoload everything in the lib file, as it then autoloads rake tasks into production. Best practice is to move files to app/lib and require files where necessary (see https://github.com/rails/rails/issues/37835). However, this app is quite complicated and I'm not sure what the best way to go about reworking this is.


EGiataganas:
Hey @cdccollins this looks great, I've compared the PR changes with [railsdiff][1] and there are some changes that didn't happen with `bundle exec rails app:update` that could also be done:

1. Remove `bin/bundle`, `bin/update`
2. Update [system! "bin/rails db:setup"][2] with `system! "bin/rails db:prepare"`
3. In the `config/environments/test.rb` the `config.cache_classes = true` has changed to `false` and there is `config.action_view.cache_template_loading = true`
4. Remove `config/initializers/new_framework_defaults.rb` as per the reason in alphagov/travel-advice-publisher@8b0745e
5. In the `config/puma.rb` Rails 6 introduce `has max_threads_count` and `min_threads_count`. It will be good to update it
6. In the `config/spring.rb` the `%w[]` needs to be updated to `Spring.watch()`
```ruby
Spring.watch(
  ".ruby-version",
  ".rbenv-vars",
  "tmp/restart.txt",
  "tmp/caching-dev.txt"
)
```
7. Update `app/views/layouts/application.html.erb`to include `csp_meta_tag` after `csrf_meta_tags`. Maybe it will be good to have a `csrf_meta_tags`, `csp_meta_tag`, `stylesheet_link_tag` and then `javascript_include_tag`

[1]: http://railsdiff.org/5.2.4.3/6.0.3.2
[2]: https://github.com/alphagov/smart-answers/blob/master/bin/setup#L24

EGiataganas (approved):

PR #4738, opened 2020-08-04
Remove deprecated methods from uk benefits abroad

reggieb:

Updates uk benefits abroad flow to remove methods `calculate` and `save_input_as`.

Processing has been moved to SmartAnswer::Calculators::UkBenefitsAbroadCalculator.

Also:

- Adds ListValidator.call which carries out required formatting steps
  within class rather than by object passing data in.

[Trello ticket](https://trello.com/c/YsfICJPz/415-update-deprecated-syntax-for-existing-smart-answer-flow)

### DIFF
--- a/lib/list_validator.rb
+++ b/lib/list_validator.rb
@@ -1,11 +1,18 @@
 class ListValidator
+  def self.call(constraint: {}, test: [])
+    test.map!(&:to_sym)
+    new(constraint.keys).all_valid?(test)
+  end
+
+  attr_reader :list
+
   def initialize(list)
     @list = list
   end
 
   def all_valid?(elements)
-    elements.present? &&
-      elements.is_a?(Array) &&
-      elements.all? { |element| @list.include?(element) }
+    return false unless elements.present? && elements.is_a?(Array)
+
+    (elements - list).empty?
   end
 end
--- a/lib/smart_answer_flows/uk-benefits-abroad.rb
+++ b/lib/smart_answer_flows/uk-benefits-abroad.rb
@@ -21,9 +21,6 @@ module SmartAnswer
         on_response do |response|
           self.calculator = uk_benefits_abroad_calculator
           calculator.going_abroad = (response == "going_abroad")
-
-          self.going_abroad = calculator.going_abroad
-          self.already_abroad = calculator.already_abroad
         end
 
         next_node do
@@ -47,8 +44,7 @@ module SmartAnswer
         option :income_support
 
         on_response do |response|
-          self.benefit = response
-          calculator.benefit = benefit
+          calculator.benefit = response
         end
 
         next_node do |response|
@@ -62,7 +58,7 @@ module SmartAnswer
             question :db_how_long_abroad?
           elsif response == "tax_credits"
             question :eligible_for_tax_credits?
-          elsif going_abroad
+          elsif calculator.going_abroad
             if response == "jsa"
               question :jsa_how_long_abroad? # Q3 going_abroad
             elsif response == "pension"
@@ -70,7 +66,7 @@ module SmartAnswer
             elsif response == "income_support"
               question :is_how_long_abroad? # Q32 going_abroad
             end
-          elsif already_abroad
+          elsif calculator.already_abroad
             if response == "jsa"
               question :which_country?
             elsif response == "pension"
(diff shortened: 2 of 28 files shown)


lib/smart_answer_flows/uk-benefits-abroad.rb
```diff
@@ -17,38 +17,13 @@ def define
       multiple_choice :going_or_already_abroad? do
         option :going_abroad
         option :already_abroad
-        save_input_as :going_or_already_abroad
 
-        on_response do
+        on_response do |response|
           self.calculator = uk_benefits_abroad_calculator
```
theseanything, line 22:
```suggestion
          self.calculator = Calculators::UkBenefitsAbroadCalculator.new
```
For consistency I think we want instantiate the calculator here. This also every request receives a new calculator - and prevent issues of state persisting between different requests.

lib/smart_answer_flows/uk-benefits-abroad.rb
```diff
@@ -17,38 +17,13 @@ def define
...
-
-        calculate :already_abroad do
-          going_or_already_abroad == "already_abroad"
-        end
+          calculator.going_abroad = (response == "going_abroad")
 
-        calculate :already_abroad_text_two do
-          " or permanently" if already_abroad
+          self.going_abroad = calculator.going_abroad
+          self.already_abroad = calculator.already_abroad
```
theseanything, line 26:
```suggestion
```
If we are saving state to the calculator - I don't think we need this. We can just reference the calculator for these attributes.

lib/smart_answer_flows/uk-benefits-abroad.rb
```diff
@@ -71,16 +46,9 @@ def define
...
-          if benefit == "disability_benefits"
-            "How long will you be abroad for?"
-          elsif going_or_already_abroad == "going_abroad"
-            "How long are you going abroad for?"
-          else
-            "How long will you be living abroad for?"
-          end
+        on_response do |response|
+          self.benefit = response
+          calculator.benefit = benefit
```
theseanything, line 51:
```suggestion
          calculator.benefit = response
```

Again for consistency with other smart answers - I think we can just save the state to the calculator.

lib/smart_answer_flows/uk-benefits-abroad.rb
```diff
@@ -118,10 +86,7 @@ def define
       country_select :which_country?, additional_countries: additional_countries, exclude_countries: exclude_countries do
         on_response do |response|
           calculator.country = response
-        end
-
-        calculate :country_name do
-          (WorldLocation.all + additional_countries).find { |c| c.slug == calculator.country }.name
+          self.country_name = (WorldLocation.all + additional_countries).find { |c| c.slug == calculator.country }.name
```
theseanything, line 89: Could this be method on the calculator?

lib/smart_answer_flows/uk-benefits-abroad.rb
```diff
@@ -562,7 +527,9 @@ def define
         option :less_than_a_year_other
         option :more_than_a_year
 
-        save_input_as :how_long_abroad_jsa
+        on_response do |response|
+          self.how_long_abroad_jsa = response
```
theseanything, line 531: for consistency - save state to a calculator.

reggieb: I can't find where this parameter is used so I'll remove it.

lib/smart_answer_flows/uk-benefits-abroad.rb
reggieb: `uk_benefits_abroad_calculator` is used 8 times within the flow. It used in blocks like this (lines 288-290):
```ruby
uk_benefits_abroad_calculator.state_benefits.each_key do |benefit|
  option benefit
end
```

lib/smart_answer_flows/uk-benefits-abroad.rb
reggieb: Thank you - I've moved them to the calculator (they are used a lot within the flow - hence me not moving them to calculator initially)

lib/smart_answer_flows/uk-benefits-abroad.rb
reggieb: Thank you - I've moved it to the calculator

lib/smart_answer_flows/uk-benefits-abroad.rb
reggieb: Thank you - I've moved them to the calculator

lib/list_validator.rb
```diff
@@ -1,11 +1,18 @@
 class ListValidator
+  def self.call(constraint: {}, test: [])
+    test.map!(&:to_sym)
+    new(constraint.keys).all_valid?(test)
+  end
```
theseanything, line 5: Do we have any tests for this behaviour?

theseanything: (Nit: would it possible to condense this class into a single function?)

lib/smart_answer_flows/uk-benefits-abroad.rb
theseanything: In these cases could we refer to the class constant rather than a instance method (there's also no reason for this to be an instance method) 
```ruby
Calculators::UkBenefitsAbroadCalculator::STATE_BENEFITS.each_key do |benefit|
  option benefit
end
```

lib/list_validator.rb
reggieb: As it's a refactor I believe the current tests should still be valid.

I did wonder if it would be easier to just add a method to the calculator as it's only used in one, but thought that pehaps that was a change too far.

reggieb: I've added unit test for the `.call` method

PR #4763, opened 2020-08-12
Add getting_food group of nodes

gclssvglx:
- Add form class
- Refactor existing form class instances
- Add getting_food question group

Decided against creating a generic `options` method and passing key and value labels as this requires knowledge that checkbox and radio's have different keys.  Instead have created 2 methods - `checkbox_options` and `radio_options`.  This allows the knowledge of checkbox or radio to remain in the template.
 
[Trello](https://trello.com/c/c4lZK28m/431-session-controller-creating-gettingfood-group-of-nodes-in-session-flow)

### DIFF
--- a/app/forms/form.rb
+++ b/app/forms/form.rb
@@ -1,0 +1,49 @@
+class Form
+  include ActiveModel::Model
+  include ActiveModel::Validations
+
+  NotImplementedError = Class.new(StandardError)
+
+  attr_accessor :session
+  attr_reader :params, :question_name, :flow_name
+
+  class << self
+    attr_reader :flow_name, :node_name
+
+    def answer_flow(flow_name)
+      @flow_name = flow_name
+    end
+
+    def answer_node(node_name)
+      @node_name = node_name
+    end
+  end
+
+  delegate :flow_name, :node_name, to: :class
+
+  def initialize(params, session)
+    @params = params
+    @session = session
+  end
+
+  def checkbox_options
+    options.each_with_object([]) do |(key, value), array|
+      array << { label: value, value: key.to_s }
+    end
+  end
+
+  def radio_options
+    options.each_with_object([]) do |(key, value), array|
+      array << { text: value, value: key.to_s }
+    end
+  end
+
+  def options
+    raise NotImplementedError, "The options method has not been defined"
+  end
+
+  def save
+    session[:flow_name] ||= {}
+    session[:flow_name][:node_name] = params[:node_name]
+  end
+end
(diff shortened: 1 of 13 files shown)


app/forms/form.rb
```diff
@@ -0,0 +1,37 @@
+class Form
+  include ActiveModel::Model
+  include ActiveModel::Validations
+
+  NotImplementedError = Class.new(StandardError)
+
+  attr_accessor :session
+  attr_reader :params, :question_name, :flow_name
+
+  def initialize(params, session, flow_name, question_name)
```
reggieb, line 10: I think the way `initialize` is used in the subclasses (reliance on `super`) could be improved. If the following is added to this class:
```ruby
    class << self
      attr_reader :flow_name, :node_name

      def part_of_flow(flow_name)
        @flow_name = flow_name
      end

      def form_for_node(node_name)
        @node_name = node_name
      end
    end

    delegate :flow_name, :node_name, to: :class
```

The flow and node could be set in each subclass like this and `initialize` would not have to be overwritten:
```ruby
class FooForm < Form
  part_of_flow :foo 
  form_for_node :bar 
 
   .....
```
With that in place you get:
```
  FooForm.flow_name => :foo 
  FooForm.node_name => :bar 
  foo = FooForm.new({}, {})
  foo.flow_name => :foo 
  foo.node_name => :bar 
```

app/views/layouts/session_answers.html.erb
```diff
@@ -10,4 +10,6 @@
   </div>
 </main>
 
+<%= @form.session[:session_answers].inspect %>
```
reggieb, line 13: That's debugging code. I don't think it should be committed.

app/forms/form.rb
reggieb: And to clarify, `initialize` within `Flow` would then be:
```ruby
  attr_accessor :session
  attr_reader :params

  def initialize(params, session)
    @params = params
    @session = session
  end
```

app/views/layouts/session_answers.html.erb
gclssvglx: Fixed - thanks 👍

app/forms/form.rb
gclssvglx: Thanks @reggieb - Have updated as per your suggestion, but I didn't like your attr names (sorry!) - as this is all about `answers`.  I've also read [Yehua Katz on `<< self`](https://yehudakatz.com/2009/11/15/metaprogramming-in-ruby-its-all-about-the-self/) and think I need to investigate further!

PR #4764, opened 2020-08-12
Make page titles unique

maxgds:
https://trello.com/c/SPXc1Bak/267-update-smart-answers-to-have-unique-page-titles

This updates question pages to have the question in the title and
the outcome page to specify the results in the title. Previously
all titles were that of the smart answer meaning multiple pages
shared a title which is a WCAG fail.

The pattern for this matches changes being made in Simple Smart Answers here: https://github.com/alphagov/frontend/pull/2438

Tests have been updated to check the title, they are checking for substrings because the whitespace on the full title is a little erratic, although it renders out fine in the browser title.

Note that this refers to the HTML title tag, not the page headings. This can be seen on the browser tab, or in the source as indicated in these screenshots:

### Results page:

Screenshot 2020-08-13 at 09 25 32
Path: /additional-commodity-code/y/0/0/0/60

### Question page:

Screenshot 2020-08-13 at 09 22 22
Path: /additional-commodity-code/y/0/0/0

### Error page:

Screenshot 2020-08-13 at 09 20 09
Path: /additional-commodity-code/y?next=1

=====

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -1,7 +1,19 @@
+<%
+    answer_title = yield :title
+    question_title = yield :question_title
+    outcome_title = yield :outcome_title
+    if question_title.present?
+      title = "#{question_title} - #{answer_title}"
+    elsif outcome_title.present?
+      title = "#{outcome_title} - #{answer_title}"
+    else
+      title = answer_title
+    end
+%>
 <!DOCTYPE html>
 <html>
   <head>
-    <title><%= yield :title %> - GOV.UK</title>
+    <title><%= title %> - GOV.UK</title>
     <%= stylesheet_link_tag "application", :media => "all" %>
     <%= stylesheet_link_tag "print.css", media: "print" %>
     <%= javascript_include_tag "test-dependencies" if Rails.env.test? %>
--- a/app/views/smart_answers/question.html.erb
+++ b/app/views/smart_answers/question.html.erb
@@ -1,13 +1,12 @@
+<% question = @presenter.current_node %>
+<% content_for :question_title do %><% if question.error.present? %>Error - <% end %><%= question.title %><% end %>
 <% content_for :head do %>
   <meta name="robots" content="noindex">
 <% end %>
 
-<% question = @presenter.current_node %>
-
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
     <%= render 'debug' %>
-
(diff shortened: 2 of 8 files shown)


danacotoran:
@maxgds Just wanted to ask if we are deliberately formatting page title differently for smart answers vs simple smart answers. 

For simple smart answers (as seen [here](https://github.com/alphagov/frontend/pull/2438)) it looks like we're going with a  `Question - Simple smart answer name - GOV.UK` format. 
Are we avoiding doing that with smart answers because the title would be ridiculously long in some instances (I might be answering my own question here...😆)?

maxgds:
> @maxgds Just wanted to ask if we are deliberately formatting page title differently for smart answers vs simple smart answers.
> 
> For simple smart answers (as seen [here](https://github.com/alphagov/frontend/pull/2438)) it looks like we're going with a `Question - Simple smart answer name - GOV.UK` format.
> Are we avoiding doing that with smart answers because the title would be ridiculously long in some instances (I might be answering my own question here...😆)?

The length shouldn't matter, and we should stay in line, I had gone by what the card I was working on said, but it's now been updated. I've also added the Error scenario and extended some error tests to check for the title too.

danacotoran:
@maxgds, looks good for the error and question scenarios. 👍  However, I've noticed that on the outcome page there is still a discrepancy with how we're building `title` for simple smart answer.

Let's take [this outcome page](https://smart-answer-fix-titles-hfbcsf.herokuapp.com/coronavirus-employee-risk-assessment/y/places_of_worship/yes/maybe) as an example: we can see the title is currently `Check if you should go back into work - GOV.UK`

Should it be `Get help to work from home - Check if you should go back into work - GOV.UK` for consistency with how it's been formatted for simple smart answers?

maxgds:
That's odd, it's working for some but not others. (although the working ones may want re-ordering). I'll investigate.

maxgds:
> Let's take [this outcome page](https://smart-answer-fix-titles-hfbcsf.herokuapp.com/coronavirus-employee-risk-assessment/y/places_of_worship/yes/maybe) as an example: we can see the title is currently `Check if you should go back into work - GOV.UK`

It turns out that sometimes a title hasn't been set for the outcome and what appears to be the title in the linked outcome page is actually part of the body. Since this can't be determined reliably as the body could contain anythings, I'm falling back to a title of "Outcome" where none is supplied. I've added a fixture and test to check for this.

PR #4792, opened 2020-08-18
Fix escaped text being sent to Publishing API

theseanything:
This marks content within a text_for block in a ERB smart answer template as HTML safe (e.g. title, meta description). This prevents characters being escaped, which can be sent to Publishing API and cause the escaped characters to be displayed. For example the use of apostrophes in a title on a start page. The content item contained a title with the escape sequence `&#39;` instead of a `'` and was shown to the users.

### DIFF
--- a/lib/smart_answer/erb_renderer/format_capture_helper.rb
+++ b/lib/smart_answer/erb_renderer/format_capture_helper.rb
@@ -13,7 +13,7 @@ module SmartAnswer
       content = capture_content(&block)
       content = strip_leading_spaces(content)
       content = normalize_blank_lines(content)
-      content_for(name, content.strip)
+      content_for(name, content.strip.html_safe)
     end
 
     def govspeak_for(name, &block)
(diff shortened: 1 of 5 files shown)


lib/smart_answer/erb_renderer/format_capture_helper.rb
```diff
@@ -13,7 +13,7 @@ def text_for(name, &block)
       content = capture_content(&block)
       content = strip_leading_spaces(content)
       content = normalize_blank_lines(content)
-      content_for(name, content.strip)
+      content_for(name, content.strip.html_safe)
```
reggieb, line 16: Using `html_safe` makes me a little nervous, but I can't see any place in the code where user text input would passed through this. It would be nice to use [`sanatize`](https://api.rubyonrails.org/classes/ActionView/Helpers/SanitizeHelper.html#method-i-sanitize) to do a more targeted "fix", but there doesn't seem to be a straight forward way to do that when it's character replacement rather than a tag or attribute being sanitized. So this does look like the best option.

reggieb (approved):
LGTM

PR #4793, opened 2020-08-18
Update flow to handle branching

reggieb:
- Use hash to define flows
- Add flow unit test
- Pass session to flow so that branching can be based on session data

[Trello ticket](https://trello.com/c/XTaHue3a/443-session-controller-update-flow-to-handle-branching)

### DIFF
--- a/app/session_flows/coronavirus_find_support_flow.rb
+++ b/app/session_flows/coronavirus_find_support_flow.rb
@@ -20,39 +20,33 @@ class CoronavirusFindSupportFlow
   def initialize(node_name, session = {})
     @node_name = node_name.to_sym
     @session = session || {}
-    state_machine.restore!(node_name)
   end
 
   def next_node
-    state_machine.next_node
-    state_machine.state
+    nodes[node_name]
   end
 
-  def state_machine
-    @state_machine ||= FiniteMachine.new(self, alias_target: :flow) do
-      initial :need_help_with
-
-      event :next_node, {
-        need_help_with: flow.next_group_start,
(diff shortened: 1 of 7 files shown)


reggieb:
Currently needs rebasing before it can be merged. That will need quite a lot of work - so as development is being frozen on the *session_answers* this branch may need to be left unmerged.

app/session_flows/coronavirus_find_support_flow.rb
```diff
@@ -1,32 +1,85 @@
...
+      self_employed: node_after_self_employed,
+      have_you_been_made_unemployed: node_after_have_you_been_made_unemployed,
+      are_you_off_work_ill: next_group_start,
+      worried_about_work: next_group_start,
+      have_somewhere_to_live: :have_you_been_evicted,
+      have_you_been_evicted: next_group_start,
+      mental_health_worries: :nation,
+      nation: :results,
+    }
   end
```
gclssvglx, line 46: Hi @reggieb Nice!  I like this approach, much cleaner.  And all the flow logic is in one place.

test/unit/session_flows/coronavirus_find_support_flow_test.rb
```diff
@@ -0,0 +1,175 @@
+require "test_helper"
+
+class CoronavirusFindSupportFlowTest < ActiveSupport::TestCase
```
gclssvglx, line 3: Very nice set of tests.  Thanks.

gclssvglx (approved):
@reggieb Nice one!  More than happy to freeze the dev at this point.

PR #4806, opened 2020-08-24
Add captions to the find support flow

gclssvglx:
Add question group names where appropriate - as `caption` blocks - to the find support smart answer flow.

[Trello](https://trello.com/c/LBnGF4hq/453-add-sections-to-question-caption-for-coronavirus-find-support)

### DIFF
--- a/lib/smart_answer_flows/coronavirus-find-support/questions/afford_food.erb
+++ b/lib/smart_answer_flows/coronavirus-find-support/questions/afford_food.erb
@@ -1,4 +1,4 @@
-<% text_for :hint do %>
+<% text_for :caption do %>
   Getting food
 <% end %>
 
--- a/lib/smart_answer_flows/coronavirus-find-support/questions/afford_rent_mortgage_bills.erb
+++ b/lib/smart_answer_flows/coronavirus-find-support/questions/afford_rent_mortgage_bills.erb
@@ -1,4 +1,4 @@
-<% text_for :hint do %>
+<% text_for :caption do %>
   Paying bills
 <% end %>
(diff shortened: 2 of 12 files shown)


lib/smart_answer_flows/coronavirus-find-support/questions/nation.erb
```diff
@@ -1,3 +1,7 @@
+<% text_for :caption do %>
+  &nbsp;
+<% end %>
```
gclssvglx, line 3: Hi @theseanything - for this page and `need_help_with` there is no group, other text we can use and the flow name is not shown.  Do you know a better way that `&nbsp;`?  Again, it works - but feels hacky.

reggieb: Wouldn't it be better to add a default behaviour for `caption` so that when not given in the erb, it inserts a non-breaking space.

reggieb: Looking again. This is the method that needs updates:

_app/presenters/question_presenter.rb_
```ruby
def caption
  return @renderer.content_for(:caption) if @renderer.content_for(:caption).present?
 
  @flow_presenter.title
end
```
I think by passing in the non-breaking space, it forcing the rendering of the content_for(:caption) which then renders that space.

gclssvglx: Have decided to go with the default flow title for now.

theseanything (approved):

PR #4814, opened 2020-08-26
Add option to hide captions on questions

theseanything:
This adds the following option that can be specified in question templates to hide the caption text above a question.

```erb
<% hide_caption true %>
```

The default behaviour is the caption is always shown, which is the existing behaviour.

To achieve this, I refactored the caption HTML fragment into a separate partial to make it easier to modify.

### DIFF
--- a/app/presenters/question_presenter.rb
+++ b/app/presenters/question_presenter.rb
@@ -34,6 +34,7 @@ class QuestionPresenter < NodePresenter
   end
 
   def caption
+    return nil if @renderer.hide_caption
     return @renderer.content_for(:caption) if @renderer.content_for(:caption).present?
 
     @flow_presenter.title
--- a/app/views/smart_answers/inputs/_country_select_question.html.erb
+++ b/app/views/smart_answers/inputs/_country_select_question.html.erb
@@ -1,4 +1,4 @@
-<span class="govuk-caption-l"><%= question.caption %></span>
+<%= render "smart_answers/inputs/caption", { question: question } %>
 <%= render "govuk_publishing_components/components/select", {
   id: "response",
   label: question.title,
--- a/app/views/smart_answers/inputs/_country_select_question.html.erb
+++ b/app/views/smart_answers/inputs/_country_select_question.html.erb
@@ -1,4 +1,4 @@
-<%= render "smart_answers/inputs/caption", { question: question } %>
+<%= render "smart_answers/inputs/caption", question: question %>
 <%= render "govuk_publishing_components/components/select", {
   id: "response",
   label: question.title,
--- a/lib/smart_answer/erb_renderer/question_options_helper.rb
+++ b/lib/smart_answer/erb_renderer/question_options_helper.rb
@@ -3,5 +3,9 @@ module SmartAnswer
     def options(new_options = nil)
       @options = new_options || @options || {}
     end
+
+    def hide_caption(hide_caption = false)
+      @hide_caption = hide_caption || @hide_caption
+    end
   end
 end
--- a/test/unit/checkbox_question_presenter_test.rb
+++ b/test/unit/checkbox_question_presenter_test.rb
@@ -33,6 +33,7 @@ module SmartAnswer
     end
 
     test "#caption returns the given caption when a caption is given" do
+      @renderer.stubs(:hide_caption).returns(false)
       @renderer.stubs(:content_for).with(:caption).returns("caption-text")
 
       assert_equal "caption-text", @presenter.caption
--- a/test/unit/erb_renderer_test.rb
+++ b/test/unit/erb_renderer_test.rb
@@ -71,6 +71,36 @@ module SmartAnswer
       end
     end
 
+    test "#hide_caption returns true if set as true in the view" do
+      erb_template = "<% hide_caption true %>"
+
+      with_erb_template_file("template-name", erb_template) do |erb_template_directory|
+        renderer = ErbRenderer.new(template_directory: erb_template_directory, template_name: "template-name")
+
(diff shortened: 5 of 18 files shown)


lib/smart_answer/erb_renderer/question_options_helper.rb
```diff
@@ -3,5 +3,14 @@ module ErbRenderer::QuestionOptionsHelper
     def options(new_options = nil)
       @options = new_options || @options || {}
     end
+
+    def show_caption?
+      @show_caption = true if @show_caption.nil?
+      @show_caption
+    end
```
reggieb, line 10: Ignore this if you like, but I'd prefer:
```ruby
def show_caption?
  return true if @show_caption.nil?

  @show_caption
end
```

test/unit/checkbox_question_presenter_test.rb
```diff
@@ -13,6 +13,7 @@ class CheckboxQuestionPresenterTest < ActiveSupport::TestCase
       @renderer.stubs(:option).with(:option1).returns("Option 1")
       @renderer.stubs(:option).with(:option2).returns({ label: "Option 2", hint_text: "Hint 2" })
       @renderer.stubs(:option).with(:option3).returns({ label: "Option 3" })
+      @renderer.stubs(:show_caption?).returns(true)
```
reggieb, line 16: Is this needed if the default behaviour is to return `true`?

test/unit/erb_renderer_test.rb
```diff
@@ -71,6 +71,26 @@ class ErbRendererTest < ActiveSupport::TestCase
...
+
+    test "#show_caption? returns true if not set in the view" do
+      erb_template = ""
+
+      with_erb_template_file("template-name", erb_template) do |erb_template_directory|
+        renderer = ErbRenderer.new(template_directory: erb_template_directory, template_name: "template-name")
+
+        assert renderer.show_caption?
+      end
+    end
```
reggieb, line 92: Should there be another test for:
```ruby
erb_template = "<% show_caption true %>"
```

reggieb (commented):
Looks fine, but I wonder if everything wouldn't be simpler if the flag was `hide_caption`. That way the default behaviour (`@hide_caption == nil`) would produce the desired behaviour without having to distinguish between `@hide_caption == nil` and `@hide_caption == false`

test/unit/checkbox_question_presenter_test.rb
theseanything: So unfortunately, you need to stub the method because "@renderer" in the context is just a stub object, not actually a renderer. We need to stub the hide_caption method because it's called when the we test `presenter.caption`. But I'll move the stub closer to where its needed (L39)

test/unit/erb_renderer_test.rb
theseanything: Added!

lib/smart_answer/erb_renderer/question_options_helper.rb
theseanything: So I've refactored this as hide_caption to simplify things as you suggested!

theseanything:
> Looks fine, but I wonder if everything wouldn't be simpler if the flag was `hide_caption`. That way the default behaviour (`@hide_caption == nil`) would produce the desired behaviour without having to distinguish between `@hide_caption == nil` and `@hide_caption == false`

Yup agree - I've refactored the PR to be hide_caption instead.

app/presenters/question_presenter.rb
```diff
@@ -34,6 +34,7 @@ def hint
   end
 
   def caption
+    return nil if @renderer.hide_caption
```
reggieb, line 37: You don't need to specify the `nil`. That's the default behaviour or return:
```ruby
return if @renderer.hide_caption
```

app/views/smart_answers/inputs/_country_select_question.html.erb
```diff
@@ -1,4 +1,4 @@
-<span class="govuk-caption-l"><%= question.caption %></span>
+<%= render "smart_answers/inputs/caption", { question: question } %>
```
reggieb, line 1: I don't think you need the curly brackets in these render calls. This should work:
```ruby
<%= render "smart_answers/inputs/caption", question: question %>
```

lib/smart_answer/erb_renderer/question_options_helper.rb
```diff
@@ -3,5 +3,9 @@ module ErbRenderer::QuestionOptionsHelper
     def options(new_options = nil)
       @options = new_options || @options || {}
     end
+
+    def hide_caption(hide_caption = false)
+      @hide_caption = hide_caption || @hide_caption
+    end
```
reggieb, line 9: I prefer the setter (`hide_caption`) and getter (`hide_caption?`) pairing you had previously

reggieb (approved):
Looks good. A couple of very minor suggests which you can ignore if you prefer.

app/presenters/question_presenter.rb
theseanything: I think I'm going to leave the explicit nil there for readability.

app/views/smart_answers/inputs/_country_select_question.html.erb
theseanything: Thanks! Removed the extra brackets around these.

lib/smart_answer/erb_renderer/question_options_helper.rb
theseanything: I do to - but I think I'm going to keep as it consistent with pattern of options above.

PR #4830, opened 2020-09-08
Update session store so that it works if page is returned to

reggieb:
This iteration of the session store system works for all happy paths that I can find. It will both complete a journey, and allow a user to return to an earlier part of the flow and alter an input.

The main outstanding issue is that error handling is not working. In that, if you submit a form without entering anything the page is shown again, but without an error message.

- Adds handling of data in the form of a hash (params) rather than an array, within the `Flow`.
- Adds testing of the hash handling of `Flow`
- Uses a `SessionStore` to manage data storage with session. The initial benefit is that it simplifies namespacing the session data (so that data entry from one flow won't overwrite that from another)
- Adds `SessionStore` tests
- Updates landing page and return to flow links to work with session store.

[Trello ticket](https://trello.com/c/NUE5PJkG/469-update-session-store-so-that-it-works-if-page-is-returned-to)

### DIFF
--- a/app/presenters/flow_presenter.rb
+++ b/app/presenters/flow_presenter.rb
@@ -5,7 +5,7 @@ class FlowPresenter
 
   attr_reader :params, :flow
 
-  delegate :need_it, :button_text, :use_session?, to: :flow
+  delegate :need_it, :button_text, :use_session?, :questions, to: :flow
   delegate :title, to: :start_node
 
   def initialize(params, flow)
@@ -23,7 +23,12 @@ class FlowPresenter
   end
 
   def current_state
-    @current_state ||= @flow.process(all_responses)
+    @current_state ||= if use_session?
+                         requested_node = params[:node_name] unless params[:next]
+                         @flow.resolve_state(params[:responses], requested_node)
+                       else
+                         @flow.process(all_responses)
+                       end
   end
 
   def name
@@ -75,13 +80,17 @@ class FlowPresenter
     @start_node ||= StartNodePresenter.new(node)
   end
 
-  def change_collapsed_question_link(question_number)
-    smart_answer_path(
-      id: @params[:id],
-      started: "y",
-      responses: accepted_responses[0...question_number - 1],
-      previous_response: accepted_responses[question_number - 1],
-    )
+  def change_collapsed_question_link(question_number, question)
+    if use_session?
+      session_flow_path(params[:id], node_name: question.node_name)
+    else
+      smart_answer_path(
+        id: @params[:id],
+        started: "y",
+        responses: accepted_responses[0...question_number - 1],
+        previous_response: accepted_responses[question_number - 1],
+      )
+    end
   end
 
   def normalize_responses_param
@@ -90,6 +99,8 @@ class FlowPresenter
       []
     when Array
       params[:responses]
+    when ActionController::Parameters
+      current_state.responses
     else
       params[:responses].to_s.split("/")
     end
--- a/app/views/smart_answers/landing.html.erb
+++ b/app/views/smart_answers/landing.html.erb
@@ -18,7 +18,7 @@
 
         <%= render "govuk_publishing_components/components/button", {
           text: start_node.start_button_text,
-          href: smart_answer_path(@name.to_s, started: "y"),
+          href: start_of_flow_path(@presenter),
           rel: "nofollow",
           start: true,
           margin_bottom: true,
--- a/lib/smart_answer/flow.rb
+++ b/lib/smart_answer/flow.rb
@@ -131,19 +131,34 @@ module SmartAnswer
       responses.inject(start_state) do |state, response|
         return state if state.error
 
-        begin
-          state = node(state.current_node).transition(state, response)
-          node(state.current_node).evaluate_precalculations(state)
-        rescue BaseStateTransitionError => e
-          if e.is_a?(LoggedError)
-            GovukError.notify e
-          end
+        transistion_state(state, response)
+      end
+    end
 
-          state.dup.tap do |new_state|
-            new_state.error = e.message
-            new_state.freeze
-          end
-        end
+    def resolve_state(responses, requested_node)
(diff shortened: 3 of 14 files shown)


lib/smart_answer/flow.rb
```diff
@@ -147,6 +147,30 @@ def process(responses)
       end
     end
 
+    def resolve_state(responses, requested_node)
+      state = start_state
+
+      until state.nil?
+        return state if state.error
+        return state if state.current_node.to_s == requested_node
```
theseanything, line 143:
```suggestion
        return state if state.current_node.to_s == requested_node && !responses.key?(state.current_node.to_s)
```
Do we need to check if there's been attempt to respond previously?

reggieb: That looks to work!

app/presenters/flow_presenter.rb
```diff
@@ -23,7 +23,12 @@ def finished?
   end
 
   def current_state
-    @current_state ||= @flow.process(all_responses)
+    @current_state ||= if use_session?
+                         requested_node = params[:node_name] unless params[:next]
+                         @flow.resolve_state(params[:responses], requested_node)
+                       else
+                         @flow.process(all_responses)
+                       end
```
gclssvglx, line 31: Instead of an `if...else` here (and again in `change_collapsed_question_link`) would sub-classing the `FlowPresenter` and overriding these two methods be worthwhile?

app/views/smart_answers/landing.html.erb
```diff
@@ -1,4 +1,11 @@
 <% start_node = @presenter.start_node %>
+<%
+  start_href = if @presenter.use_session?
+                 session_flow_path(@presenter.name, node_name: @presenter.flow.questions.first.name)
+               else
+                 smart_answer_path(@presenter.name, started: "y")
+               end
```
gclssvglx, line 7: If you go with the suggestion of a `SessionFlowPresenter` above - could this be a method on that class?  If not, maybe a helper method?

test/controllers/session_answers_controller_test.rb
```diff
@@ -37,17 +37,17 @@ def params
...
+  #    end
+  #
+  #    should "render error successfully" do
+  #      assert_response :success
+  #    end
+  #
+  #    should "display error" do
+  #      assert_match(/govuk-error-message/, response.body)
+  #    end
+  #  end
```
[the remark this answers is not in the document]
reggieb: These tests have been reinstated as the error behaviour is now working as expected.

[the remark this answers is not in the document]
gclssvglx: 👍

app/presenters/flow_presenter.rb
reggieb: That's an interesting idea. However, I think that should be done as a separate PR where it's impact can be considered in isolation.

app/views/smart_answers/landing.html.erb
theseanything: Or at the least could we move this to a method in the presenter? Just trying to keep logic out of views if possible.

theseanything (changes requested):
Thanks Rob! Could we tidy up commits? 
- Squash "Fix misspelling in landing page" into the commit that introduced the code?
- Separate out the work that fixes the start page start link into a separate commit (with tests)
- Separate out the work that fixes the "Previous responses" section into a separate commit (with tests)
- Squash "Add tests of responses hash handling by flow" tweaks with other commits
- Squash "Add clause to flow#resolve_state so empty input errors correctly" into the initial commit implementing `resolve_state` method

app/views/smart_answers/landing.html.erb
reggieb: I agree - I left the code there as that seems to be a pattern used a lot across GDS apps.

I'll move it to the `CurrentQuestionHelper` as there are already a lot of methods defining paths by presenter, there, and nothing like it in `FlowPresenter`.

lib/smart_answer/flow.rb
```diff
@@ -131,19 +131,34 @@ def process(responses)
...
+      state = node(state.current_node).transition(state, response)
+      node(state.current_node).evaluate_precalculations(state)
+    rescue BaseStateTransitionError => e
+      if e.is_a?(LoggedError)
+        GovukError.notify e
+      end
+
+      state.dup.tap do |new_state|
+        new_state.error = e.message
+        return new_state.freeze
```
theseanything, line 161: Do we need the return here?

app/presenters/flow_presenter.rb
```diff
@@ -75,13 +80,17 @@ def start_node
...
+    if use_session?
+      session_flow_path(params[:id], node_name: question.node_name)
+    else
+      smart_answer_path(
+        id: @params[:id],
+        started: "y",
+        responses: accepted_responses[0...question_number - 1],
+        previous_response: accepted_responses[question_number - 1],
+      )
+    end
```
theseanything, line 93: Are there tests for this?

app/presenters/flow_presenter.rb
```diff
@@ -90,6 +99,8 @@ def normalize_responses_param
       []
     when Array
       params[:responses]
+    when ActionController::Parameters
+      current_state.responses
```
theseanything, line 103: Are there tests for this?

theseanything (changes requested):
Looks good - just a few missing tests I think

lib/smart_answer/flow.rb
reggieb: I've removed it

app/presenters/flow_presenter.rb
reggieb: Tests added

app/presenters/flow_presenter.rb
reggieb: tests added

theseanything (approved):

PR #4852, opened 2020-09-15
Add extra event tracking to smart answers (inc find Coronavirus Support)

DilwoarH:
## What

Add extra event tracking to smart answers before Find Coronavirus Support is migrated so we can track how users interact with the tool.

We will need to add:

- Tracking to the question/answer combinations

To add

Pages
eventCategory : question_answer
eventAction: [selection 
eventLabel: [question]

Link to doc here with more details https://docs.google.com/spreadsheets/d/1EhSbmUgrZfkyGLdwsGnRZ_HedxwvuCdWrHAuGPdKjH4/edit?usp=sharing

Some smart answers collect information that could be considered personal. for example: https://www.gov.uk/state-pension-age/y/age which is stripped out in the URL before it reaches Google Analytics. This will also need to apply to the event data. 

## Why

So we know how users are interacting with the tool so we can identify points of failure and report on what users select

https://trello.com/c/QR4lSsFX/477-add-extra-event-tracking-to-smart-answers-inc-find-coronavirus-support

### DIFF
--- a/app/assets/javascripts/modules/track-responses.js
+++ b/app/assets/javascripts/modules/track-responses.js
@@ -1,0 +1,93 @@
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
+(function (global, GOVUK) {
+  'use strict'
+
+  GOVUK.Modules.TrackResponses = function () {
+    this.start = function (element) {
+      track(element[0])
+    }
+
+    function getQuestionKey(submittedForm) {
+      return submittedForm.getAttribute('data-question-key');
+    }
+
+    function getResponseLabelsForMultipleChoice(submittedForm) {
+      var labels = []
+      var checkedOptions = submittedForm.querySelectorAll('input:checked')
+
+      if (checkedOptions.length) {
+        checkedOptions.forEach(function (checkedOption) {
+          var checkedOptionId = checkedOption.getAttribute('id')
+          var checkedOptionLabel = submittedForm.querySelectorAll('label[for="' + checkedOptionId + '"]')
+          
+          var eventLabel = checkedOptionLabel.length
+            ? checkedOptionLabel[0].innerText.trim()
+            : checkedOption.value
+
+          labels.push(eventLabel)
+        })
+      } else {
+        labels.push("no response")
+      }
+
+      return labels
+    }
+
+    function getResponseLabelsForSelectOption(submittedForm) {
+      var labels = []
+      var selectInputs = submittedForm.querySelectorAll('select')
+
+      if (selectInputs.length) {
+        selectInputs.forEach(function (select) {
+          var value = select.value
+
+          if (value) {
+            var label = select.options[select.selectedIndex].innerHTML
+            var eventLabel = label.length ? label : value
+            labels.push(eventLabel)
+          } else {
+            labels.push("no response")
+          }
+        })
+      }
+
+      return labels
+    }
+
+    function getResponseLabels(submittedForm) {
+      var responseLabels = []
+      var questionType = submittedForm.getAttribute('data-type')
+      
+      switch (questionType) {
+        case 'checkbox_question':
+        case 'multiple_choice_question':
+          responseLabels = getResponseLabelsForMultipleChoice(submittedForm)
+          break
+        
+        case 'country_select_question':
+          responseLabels = getResponseLabelsForSelectOption(submittedForm)
+          break
+        
+        default:
+          break
+      }
+
+      return responseLabels
+    }
+
+    function track (element) {
+      element.addEventListener('submit', function (event) {
+        var submittedForm = event.target
+        var questionKey = getQuestionKey(submittedForm)
+        var responseLabels = getResponseLabels(submittedForm)
+
+        responseLabels.forEach(function(label){
+          var options = { transport: 'beacon', label: label }
+          GOVUK.analytics.trackEvent('question_answer', questionKey, options)
+        })
+      })
+    }
+  }
+})(window, window.GOVUK)
--- a/app/views/smart_answers/_previous_answers.html.erb
+++ b/app/views/smart_answers/_previous_answers.html.erb
@@ -2,7 +2,17 @@
   <table class="govuk-table previous-answers-table">
     <caption class="govuk-table__caption">
       <h2 class="govuk-heading-m">Previous answers</h2>
-      <p class="govuk-body"><%= link_to "Start again", restart_flow_path(@presenter), :class => "govuk-link" %></p>
+      <p class="govuk-body">
+      <%= link_to "Start again", 
+        restart_flow_path(@presenter), 
+        :class => "govuk-link",
+        :data => {
+          module: "track-click",
+          "track-action": "Start again",
+          "track-category": "StartAgain",
+          "track-label": @presenter.current_node.title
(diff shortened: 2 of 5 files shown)


app/assets/javascripts/modules/track-qa-choices.js
```diff
@@ -0,0 +1,83 @@
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
+(function (global, GOVUK) {
+  'use strict'
+
+  GOVUK.Modules.TrackQaChoices = function () {
```
theseanything, line 7: Could we rename this to something a bit more descriptive? I feel future devs might not understand what "qa" is
```suggestion
  GOVUK.Modules.TrackResponses = function () {
```

app/assets/javascripts/modules/track-qa-choices.js
```diff
@@ -0,0 +1,83 @@
...
+          case 'multiple_choice_question':
+              triggerEventForMultipleChoice(submittedForm);
+            break;
+          
+          case 'country_select_question':
+            triggerEventForSelectOption(submittedForm);
+            break;
+          
+          default:
+            break;
```
theseanything, line 78: How come we aren't supporting other question types?

app/assets/javascripts/modules/track-qa-choices.js
```diff
@@ -0,0 +1,83 @@
...
+            ? checkedOptionLabel
+            : checkedOption.value
+
+          var options = { transport: 'beacon', label: eventLabel }
+
+          GOVUK.analytics.trackEvent('question_answer', questionKey, options)
+        })
+      } else {
+        // Skipped questions
+        var options = { transport: 'beacon', label: 'no choice' }
```
theseanything, line 34: Nit:
```suggestion
        var options = { transport: 'beacon', label: 'no response' }
```
Just to keep terminology consistent

app/assets/javascripts/modules/track-qa-choices.js
```diff
@@ -0,0 +1,83 @@
...
+          var label = select.options[select.selectedIndex].innerHTML
+          var eventLabel = label.length ? label : value
+
+          var options = { transport: 'beacon', label: eventLabel }
+
+          GOVUK.analytics.trackEvent('question_answer', questionKey, options)
+        })
+      } else {
+        // Skipped questions
+        var options = { transport: 'beacon', label: 'no choice' }
```
theseanything, line 56: Nit:
```suggestion
        var options = { transport: 'beacon', label: 'no response' }
```

app/views/smart_answers/_previous_answers.html.erb
```diff
@@ -2,7 +2,12 @@
   <table class="govuk-table previous-answers-table">
     <caption class="govuk-table__caption">
       <h2 class="govuk-heading-m">Previous answers</h2>
-      <p class="govuk-body"><%= link_to "Start again", restart_flow_path(@presenter), :class => "govuk-link" %></p>
+      <p class="govuk-body"><%= link_to "Start again", restart_flow_path(@presenter), :class => "govuk-link", :data => {
+        module: "track-click",
+        "track-action": "Start again",
+        "track-category": "StartAgain",
```
theseanything, line 8: Do we need to check if a PA if these labels are suitable?

app/assets/javascripts/modules/track-qa-choices.js
```diff
@@ -0,0 +1,83 @@
...
+            triggerEventForSelectOption(submittedForm);
+            break;
+          
+          default:
+            break;
+        }
+      })
+    }
+  }
+})(window, window.GOVUK)
```
theseanything, line 83: Is it possible to add tests for this?

theseanything (changes requested):

app/views/smart_answers/_previous_answers.html.erb
DilwoarH: This is what it said on the spreadsheet (in the trello)

app/assets/javascripts/modules/track-qa-choices.js
DilwoarH: We are only tracking pre-defined responses at the moment as we have no way of stripping out PII

app/assets/javascripts/modules/track-responses.js
```diff
@@ -0,0 +1,93 @@
...
+          responseLabels = getResponseLabelsForMultipleChoice(submittedForm)
+          break;
+        
+        case 'country_select_question':
+          responseLabels = getResponseLabelsForSelectOption(submittedForm)
+          break;
+        
+        default:
+          break;
+      }
```
reggieb, line 75: Suggestion that can be ignored - it might be a little clear if rather than setting `var responseLabels = [];` it is just declared on line 60 (`var responseLabels`). Then on set ` responseLabels = []` on line 74 so that `[]` is obviously the default behaviour:
```javascript
default: responseLabels = []
break
```

app/assets/javascripts/modules/track-responses.js
```diff
@@ -0,0 +1,93 @@
...
+
+      return responseLabels;
+    }
```
reggieb, line 78: Generally `;` aren't being used to terminate lines in this code - which is fine. However, they are present on line 60 at the end of the `var` statement, and after the `break` statements on lines 67, 71, and 74. Are they necessary, and if not I think they should be removed for consistency.

app/views/smart_answers/_previous_answers.html.erb
```diff
@@ -2,7 +2,12 @@
...
+        "track-label": @presenter.current_node.title
+      } %></p>
```
reggieb, line 10: Very minor, but I think statements should either be multi-line (with each clause on a new line) or single line and not a combination of the two as here. I'd suggest this would be a little clearer:
```
<p class="govuk-body">
  <%= link_to "Start again", 
    restart_flow_path(@presenter), 
    class: "govuk-link", 
    data: {
      module: "track-click",
      "track-action": "Start again",
      "track-category": "StartAgain",
      "track-label": @presenter.current_node.title
    } 
  %>
</p>

File: app/views/smart_answers/question.html.erb [hunk 10]
```diff
@@ -13,7 +13,11 @@
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
     <%= render 'smart_answers/debug' %>
-    <%= form_tag current_question_path(@presenter), :method => :get do %>
+    <%= form_tag current_question_path(@presenter), :method => :get, :data => {
+      module: "track-responses",
+      type: question.partial_template_name,
+      "question-key": question.title
+    } do %>
```
reggieb, line 20: Again - I think this mix of single and multi-line presentation makes the statement harder to read as on first glance it looks like `module`, `type` and `question-key` are parameters of the parent `form_tag` statement rather than belonging to the inner `data` object. If the statement is totally multi-line the hierarchy of parameters becomes clearer:
```
<%= form_tag(
      current_question_path(@presenter), 
      method: :get, 
      data: {
        module: "track-responses",
        type: question.partial_template_name,
        "question-key": question.title
      } 
    ) do 
%>
```

reggieb (approved):
Looks good to me - Some pedantic comments that can be ignored.

theseanything (approved):

PR #4857, opened 2020-09-17
Fix destroy session

reggieb:
- Add method to `SessionStore` to remove data for flow from session
- Use new method in session controller destroy action
- Use mock expectations of session store to test session storage in
  controller test
- Test details of session store deletion in session store unit test

[Trello card](https://trello.com/c/VTW7MQrX/491-fix-destroy-session-route-to-clear-the-session-data)

### DIFF
--- a/test/unit/services/session_store_test.rb
+++ b/test/unit/services/session_store_test.rb
@@ -43,4 +43,19 @@ class SessionStoreTest < ActiveSupport::TestCase
       assert_equal [:a, node_name, :c], session[flow_name].keys
     end
   end
+
+  context "#clear" do
+    should "remove entries from session" do
+      session_store.add_response(response)
+      session_store.clear
+      assert_equal({}, session)
+    end
+
+    should "not change other data in session" do
+      other_data = { foo: :bar }
+      session.merge! other_data
+      session_store.clear
+      assert_equal other_data, session
+    end
+  end
 end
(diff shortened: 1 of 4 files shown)


test/unit/services/session_store_test.rb
```diff
@@ -43,4 +43,19 @@ def session_store
       assert_equal [:a, node_name, :c], session[flow_name].keys
     end
   end
+
+  context "#clear" do
+    should "remove entries from session" do
+      assert_not_equal({}, session)
```
theseanything, line 49: Why are we making an assertion before we've done the action?

theseanything (approved):

test/unit/services/session_store_test.rb
reggieb: The hash being empty is the default behaviour. So checking that the hash is empty after it is cleared, could mean it has been cleared, but it could also mean that it was empty to start with and nothing changed. So checking the hash is not empty beforehand makes sure the clearing is actually doing something. I was hoping to use `assert_changes` but that didn't seem to work.

I'd add, the test this code replaces is an example of the problem of using just a test of emptiness. The test passed because the session was empty at the end of the test. The problem was that session was always empty, so in fact the test wasn't testing the intended behaviour.

theseanything: Why can we not set the session not be empty, rather asserting it? So we no for certain the we are calling "#clear" on sessions that has data?

reggieb: I've changed the assert to a specific call that ensures the session isn't empty

theseanything (approved):

PR #4861, opened 2020-09-18
Fix node names in url string

reggieb:
What
Node names in the URL currently render with question marks and underscores instead of dashes. The question marks get HTML escaped which add unreadable text the URL. For GOV.UK we prefer to use dashes vs underscores.

e.g.

/flow-name/s/node_name%3F should be /flow-name/s/node-name

How

- Removed `?` from node names in flow definitions
- Use a dasherized version of node name in urls. With `node_path` used to identify the version of node name used in url paths.

[trello ticket](https://trello.com/c/3A0bLXHz/496-fix-node-names-in-url-string)

### DIFF
--- a/app/controllers/session_answers_controller.rb
+++ b/app/controllers/session_answers_controller.rb
@@ -2,7 +2,7 @@ class SessionAnswersController < ApplicationController
   before_action :set_cache_headers
 
   def start
-    redirect_to session_flow_path(id: params[:id], node_name: next_node_name)
+    redirect_to session_flow_path(id: params[:id], node_slug: next_node_slug)
   end
 
   def show
@@ -13,7 +13,7 @@ class SessionAnswersController < ApplicationController
 
   def update
     add_new_response_to_session
-    redirect_to session_flow_path(id: params[:id], node_name: next_node_name)
+    redirect_to session_flow_path(id: params[:id], node_slug: next_node_slug)
   end
 
   def destroy
@@ -52,9 +52,13 @@ private
   def session_store
     @session_store ||= SessionStore.new(
       flow_name: name,
-      current_node: params[:node_name],
+      current_node: node_name,
       session: session,
     )
+  end
+
+  def node_name
+    @node_name ||= params[:node_slug].underscore if params[:node_slug].present?
   end
 
   def flow_registry
@@ -66,15 +70,15 @@ private
   end
 
   def page_type
-    return :landing if params[:node_name].blank?
+    return :landing if node_name.blank?
     return :result if presenter.finished?
 
     :question
   end
   helper_method :page_type
 
-  def next_node_name
-    presenter.current_state.current_node.to_s
+  def next_node_slug
+    presenter.current_state.current_node.to_s.dasherize
   end
 
   def debug?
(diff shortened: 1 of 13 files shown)


app/presenters/node_presenter.rb
```diff
@@ -10,4 +10,8 @@ def initialize(node, flow_presenter, state = nil, _options = {}, _params = {})
   def node_name
     @node.name
   end
+
+  def node_path
+    node_name.to_s.dasherize
+  end
```
theseanything, line 16: Do we have unit tests for this?

app/controllers/session_answers_controller.rb
```diff
@@ -2,7 +2,7 @@ class SessionAnswersController < ApplicationController
   before_action :set_cache_headers
 
   def start
-    redirect_to session_flow_path(id: params[:id], node_name: next_node_name)
+    redirect_to session_flow_path(id: params[:id], node_path: next_node_path)
```
theseanything, line 5: Not sure if node_path describes this variable well. Were there any other name you were considering?

theseanything (changes requested):

app/presenters/node_presenter.rb
reggieb: I'm not too happy with the name, but couldn't think of a better alternative. `node_path_segment` perhaps. Happy to consider something else.

app/controllers/session_answers_controller.rb
reggieb: I'm not too happy with the name, but couldn't think of a better alternative. node_path_segment perhaps. Happy to consider something else.

app/presenters/node_presenter.rb
reggieb: The method is not unit tested directly, but it is tested indirectly in many of the tests where the node in the path has changed from underscore to dashed form. So yes it is tested.

reggieb: I've added a test for the node presenter which also tests node name

app/controllers/session_answers_controller.rb
reggieb: I've replaced `node_path` with `node_slug`

theseanything (dismissed):

node_modules/.yarn-integrity
```diff
@@ -0,0 +1,12 @@
...
+  "modulesFolders": [
+    "node_modules"
+  ],
+  "flags": [],
+  "linkedModules": [],
+  "topLevelPatterns": [],
+  "lockfileEntries": {},
+  "files": [],
+  "artifacts": {}
+}
```
theseanything, line 12: Can we remove this file? Not sure why this is include as part of this PR?

yarn.lock
```diff
@@ -0,0 +1,4 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
+# yarn lockfile v1
+
+
```
theseanything, line 4: Why is this file being included as part of this PR?

theseanything (changes requested):
Few extra files snuck into this PR. Whilst they maybe worth checking to the repo - adding them in these commits and this PR is out of scope.

theseanything (approved):

PR #4867, opened 2020-09-22
Fix no-results text & sub-section display issues on the results page of the Coronavirus Find Support flow

gclssvglx:
## What & Why

This change fixes the following two issues that are affecting users of the Coronavirus Find Support flow result page.

1. It is possible to reach the results page without the no-results text being shown. An example of this behaviour is:

- "What do you need help with because of coronavirus?" : "Feeling unsafe where you live, or what to do if you’re worried about the safety of another adult or child"
- "Do you feel safe where you live?" : "Yes"
- "Where do you want to find information about?" : "England"

The expected response is that the user is shown the no-results text. However, a blank page is shown.

2. Sub-sections are not being correctly shown/hidden as appropriate and based on the users given responses. An example of this behaviour is:

- "What do you need help with because of coronavirus?" : "Getting food"
- "Are you finding it hard to afford food?" : "Yes"
- "Are you able to get food?" : "Yes"
- "Where do you want to find information about?" : "England"

The expected response is that the user is shown the "Getting food" group and the "If you’re finding it hard to afford food" sub-section, but not the "If you’re unable to get food" sub-section.  However, both sub-sections are displayed.

Integration tests have been added to expose any change in the correct behaviour of both of these issues.

## Trello cards

* [Fix no results page for find support](https://trello.com/c/eOrbGfyA/501-fix-no-results-page-for-find-support)
* [Fix showing/hiding sub sections on the find support results page](https://trello.com/c/6n82URX5/502-fix-showing-hiding-sub-sections-on-the-find-support-results-page)

### DIFF
--- a/lib/smart_answer/calculators/coronavirus_find_support_calculator.rb
+++ b/lib/smart_answer/calculators/coronavirus_find_support_calculator.rb
@@ -14,52 +14,72 @@ module SmartAnswer::Calculators
                   :have_you_been_made_unemployed,
                   :mental_health_worries
 
-    def has_results?
-      return false if need_help_with.blank?
+    GROUPS = {
+      feeling_unsafe: [:feel_safe],
+      paying_bills: [:afford_rent_mortgage_bills],
+      getting_food: %i[afford_food get_food],
+      being_unemployed: %i[have_you_been_made_unemployed self_employed],
+      going_in_to_work: %i[worried_about_work are_you_off_work_ill],
+      somewhere_to_live: %i[have_somewhere_to_live have_you_been_evicted],
+      mental_health: [:mental_health_worries],
+    }.freeze
 
-      need_help_with != "none"
+    SECTIONS = {
+      feel_safe: lambda { |calculator|
+        calculator.needs_help_with?("feeling_unsafe") && calculator.feel_safe != "yes"
+      },
+      afford_rent_mortgage_bills: lambda { |calculator|
+        calculator.needs_help_with?("paying_bills") && calculator.afford_rent_mortgage_bills != "no"
+      },
+      afford_food: lambda { |calculator|
+        calculator.needs_help_with?("getting_food") && calculator.afford_food != "no"
+      },
+      get_food: lambda { |calculator|
+        calculator.needs_help_with?("getting_food") && calculator.get_food != "yes"
+      },
+      have_you_been_made_unemployed: lambda { |calculator|
+        calculator.needs_help_with?("being_unemployed") && calculator.have_you_been_made_unemployed != "no"
+      },
+      self_employed: lambda { |calculator|
+        calculator.needs_help_with?("being_unemployed") && calculator.self_employed != "yes"
+      },
+      worried_about_work: lambda { |calculator|
+        calculator.needs_help_with?("going_in_to_work") && calculator.worried_about_work != "no"
+      },
+      are_you_off_work_ill: lambda { |calculator|
+        calculator.needs_help_with?("going_in_to_work") && calculator.are_you_off_work_ill == "yes"
+      },
+      have_somewhere_to_live: lambda { |calculator|
+        calculator.needs_help_with?("somewhere_to_live") && calculator.have_somewhere_to_live != "yes"
+      },
+      have_you_been_evicted: lambda { |calculator|
+        calculator.needs_help_with?("somewhere_to_live") && calculator.have_you_been_evicted != "no"
+      },
+      mental_health_worries: lambda { |calculator|
+        calculator.needs_help_with?("mental_health") && calculator.mental_health_worries != "no"
+      },
+    }.freeze
+
+    def show_group?(name)
+      GROUPS[name].map { |section| show_section?(section) }.any?
+    end
+
+    def show_section?(name)
+      SECTIONS[name].call(self)
+    end
+
+    def has_results?
+      GROUPS.keys.map { |group| show_group?(group) }.any?
     end
 
     def needs_help_with?(given_help_item)
-      return false unless has_results?
+      return false if need_help_with.blank?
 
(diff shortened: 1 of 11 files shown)


lib/smart_answer/calculators/coronavirus_find_support_calculator.rb
```diff
@@ -14,14 +14,45 @@ class CoronavirusFindSupportCalculator
...
+      },
+      somewhere_to_live: lambda { |calculator|
+        calculator.needs_help_with?("somewhere_to_live") && (
+          calculator.have_somewhere_to_live != "yes" || calculator.have_you_been_evicted != "no"
+        )
+      },
+      mental_health: lambda { |calculator|
+        calculator.needs_help_with?("mental_health") && calculator.mental_health_worries != "no"
+      },
+    }.freeze
```
reggieb, line 44: This feels a little over complicated. I think the same result could be achieved if `RULES` was simply a list of the existing methods:
```ruby
RULES = %i[
  user_feels_unsafe?
  user_cannot_pay_their_bills?
  user_cannot_get_food?
  user_is_worried_about_going_to_work?
  user_is_unemployed?
  user_needs_somewhere_to_live?
  user_has_mental_health_worries?
]
```
And then `has_results?` would simplify to:
```ruby
def has_results?
  RULES.any? { |rule| __send__(rule) }
end
```
Also each rule method could then be memoized so that they only need to be run once per instance:
```ruby
def user_feels_unsafe?	
  @user_feels_unsafe? ||= needs_help_with?("feeling_unsafe") && feel_safe != "yes"	
end
```
As at the moment each rule is always run twice (to determine `has_results?`) and when called individually.

lib/smart_answer/calculators/coronavirus_find_support_calculator.rb
```diff
@@ -14,14 +14,45 @@ class CoronavirusFindSupportCalculator
...
-    def has_results?
-      return false if need_help_with.blank?
+    RULES = {
+      feeling_unsafe: lambda { |calculator|
+        calculator.needs_help_with?("feeling_unsafe") && calculator.feel_safe != "yes"
+      },
+      paying_bills: lambda { |calculator|
+        calculator.needs_help_with?("paying_bills") && calculator.afford_rent_mortgage_bills != "no"
+      },
+      getting_food: lambda { |calculator|
```
theseanything, line 24: So I think these rules need to be for sub sections - for example "Getting food" section is made up of the following subsections:

- If you’re finding it hard to afford food
- If you’re unable to get food

The both need rules to be shown/hidden based on their respective questions.

theseanything (changes requested):
Could we also tidy up the commit message with a more descriptive subject line and body?

lib/smart_answer/calculators/coronavirus_find_support_calculator.rb
gclssvglx: Thanks @theseanything 👍  - please see updated changes.

lib/smart_answer/calculators/coronavirus_find_support_calculator.rb
```diff
@@ -14,52 +14,72 @@ class CoronavirusFindSupportCalculator
...
+      getting_food: %i[afford_food get_food],
+      being_unemployed: %i[have_you_been_made_unemployed self_employed],
+      going_in_to_work: %i[worried_about_work are_you_off_work_ill],
+      somewhere_to_live: %i[have_somewhere_to_live have_you_been_evicted],
+      mental_health: [:mental_health_worries],
+    }.freeze
 
-    def needs_help_with?(given_help_item)
-      return false unless has_results?
+    SECTIONS = {
```
reggieb, line 27: I think `SECTIONS` is just adding complication to this class.

Defining all the calculator actions in one hash makes it more difficult to use and test each in isolation. It also prevents memory optimisation of each action. `SECTIONS` effectively replaces a standard ruby instance object with a complicated hash.

lib/smart_answer/calculators/coronavirus_find_support_calculator.rb
```diff
@@ -14,52 +14,72 @@ class CoronavirusFindSupportCalculator
...
 
-      need_help_with.split(",").include? given_help_item
+    def show_group?(name)
+      GROUPS[name].map { |section| show_section?(section) }.any?
     end
 
-    def needs_help_in?(given_nation)
-      nation == given_nation
+    def show_section?(name)
+      SECTIONS[name].call(self)
```
reggieb, line 68: If `SECTION` is removed and each action is defined as a method (as currently) this can be simplified to:
```ruby
def show_section?(name)
  __send__(name)
end
```

lib/smart_answer/calculators/coronavirus_find_support_calculator.rb
```diff
@@ -14,52 +14,72 @@ class CoronavirusFindSupportCalculator
...
-    end
+    GROUPS = {
+      feeling_unsafe: [:feel_safe],
+      paying_bills: [:afford_rent_mortgage_bills],
+      getting_food: %i[afford_food get_food],
+      being_unemployed: %i[have_you_been_made_unemployed self_employed],
+      going_in_to_work: %i[worried_about_work are_you_off_work_ill],
+      somewhere_to_live: %i[have_somewhere_to_live have_you_been_evicted],
+      mental_health: [:mental_health_worries],
+    }.freeze
```
reggieb, line 25: This constant is good - it clearly defines the relationship between calculation actions and how they are grouped.

reggieb (commented):
I like the way `GROUPS` has been defined. It is a good use of a constant and has a clear simple job of defining how calculator actions are grouped.

However, I think `SECTIONS` makes this class harder to use, optimise, modify, and test.

theseanything (approved):

PR #4868, opened 2020-09-22
Fix inconsistencies with content in coronavirus find support flow

reggieb:
Changes have been made to the currently live *govuk-coronavirus-find-support* app which this flow is replacing. These changes
need to be made in the coronavirus find support flow to maintain consistency while the app is still live.

- Added "Feeling safe" and "If you do not feel safe where you live or
  you’re worried about someone else" headers for section results page
- Change results section header "Paying bills" to "Paying your rent,
  mortgage, or bills"
- Change results section header "Going in to work" to "Being worried
  about working, or off work because you’re self-isolating"
- Remove "Support and advice" header from the "If you do not have
  somewhere to live or might become homeless" section on results page

[Trello ticket](https://trello.com/c/AEEPgyCy/503-content-fixes-to-align-with-current-tool)

### DIFF
--- a/lib/smart_answer_flows/coronavirus-find-support/outcomes/_feeling_unsafe.erb
+++ b/lib/smart_answer_flows/coronavirus-find-support/outcomes/_feeling_unsafe.erb
@@ -1,3 +1,7 @@
+## Feeling unsafe
+
+### If you do not feel safe where you live or you’re worried about someone else
+
 If you’re in immediate danger call 999 and ask for the Police.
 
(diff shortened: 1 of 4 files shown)


gclssvglx (approved):
LGTM 👍

theseanything (changes requested):
Could we update the commit message and PR description to say which flow this is effecting?  Not sure "vs the current live tool" will make sense to others. There is also a conflict that needs resolving.

theseanything (approved):

PR #4869, opened 2020-09-23
Fix change your answer links in session flows

reggieb:
Replacing the param :node_name with :node_slug meant node name was not
being passed to flow presenter. This fix passes the node name into params
before they are passed to the flow presenter.

`params[:node_name]` is passed from `FlowPresenter` to `Flow` where it is
used in the method `resolve_state` as `requested_node`. It needs to be 
present for the correct current state to be identified

An integration test is also added that ensures that the "Change" behaviour
works as expected.

[trello card](https://trello.com/c/MkLq2h6A/506-fix-change-your-answer-links)

### DIFF
--- a/test/integration/session_answers_test.rb
+++ b/test/integration/session_answers_test.rb
@@ -1,0 +1,32 @@
+require "test_helper"
+GovukTest.configure
+
+class SessionAnswersTest < ActionDispatch::SystemTestCase
+  setup do
+    Capybara.current_driver = :headless_chrome
+  end
+
+  teardown do
+    Capybara.use_default_driver
+  end
+
+  test "Change link returns previously visited page" do
+    visit "coronavirus-find-support/s"
+    within "legend" do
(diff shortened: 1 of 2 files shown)


test/integration/session_answers_test.rb
```diff
@@ -0,0 +1,32 @@
+require "test_helper"
+GovukTest.configure
+
+class SessionAnswersTest < ActionDispatch::SystemTestCase
+  setup do
+    Capybara.current_driver = :headless_chrome
```
reggieb, line 6: `:headless_chrome` is configured by `GovukTest` and is the [recommended driver documented for that gem](https://github.com/alphagov/govuk_test#usage). Other tests in this app are using `:rack_test` which doesn't seem to work headlessly in a default Rails integration test (`ActionDispatch::SystemTestCase`). So using headless chrome here and then reverting to the app default (`:rack_test`) afterwards.

theseanything (approved):

PR #4873, opened 2020-09-23
Fix the "Not sure" user response to the "need_help_with" question flow

gclssvglx:
## What

When the user selects the `Not sure` option on the `What do you need help with?` question, the user should be routed through all question groups. However, it skips all the questions and goes straight to the `Nations` question.

## Why

To make the behaviour consistent with the current tool.

[Trello](https://trello.com/c/mEkHxd3H/513-fix-the-not-sure-flow-in-needs-help-with-question)

### DIFF
--- a/test/unit/calculators/coronavirus_find_support_calculator_test.rb
+++ b/test/unit/calculators/coronavirus_find_support_calculator_test.rb
@@ -354,6 +354,11 @@ module SmartAnswer::Calculators
         @calculator.need_help_with = nil
         assert_equal @calculator.needs_help_with?("one"), false
       end
+
+      should "return true if the none item has been chosen" do
+        @calculator.need_help_with = "none"
+        assert_equal @calculator.needs_help_with?("one"), true
+      end
     end
 
     context "#needs_help_in?" do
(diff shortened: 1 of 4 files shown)


test/unit/calculators/coronavirus_find_support_calculator_test.rb
```diff
@@ -380,6 +385,11 @@ class CoronavirusFindSupportCalculatorTest < ActiveSupport::TestCase
 
     context "#next_question" do
       context "user is on the need_help_with? node" do
+        should "return feel_safe? when 'none' has been chosen" do
```
reggieb, line 388: Pedantic but `should "return feel_safe when ....` no question mark now on the question

reggieb (approved):
LGTM. A very minor comment that can be ignored.

test/unit/calculators/coronavirus_find_support_calculator_test.rb
gclssvglx: Thanks @reggieb Umm - there are quite a few of these.  I'll create a new PR for their removal rather than do that here.

PR #4876, opened 2020-09-25
Update start page content id to the current one

reggieb:
We need to update the find support flow's start page content id to be the
same as the current start page at
www.gov.uk/api/content/find-coronavirus-support.

[Trello ticket](https://trello.com/c/EULvB9l1/515-update-start-page-content-id-to-the-current-one)


gclssvglx (approved):
LGTM 👍

theseanything (approved):

PR #4885, opened 2020-09-25
Fix self-employment logic in find coronavirus support flow calculator

reggieb:
Update logic so that self-employed section is displayed by user who has selected
self-employed or is not sure about being self-employed.

Also fix bug where merge after flow name renamed has not updates an integration test

[Trello card](https://trello.com/c/Xulfz8pX/521-fix-incorrect-logic-for-when-sections-are-being-shown)

### DIFF
--- a/lib/smart_answer/calculators/find_coronavirus_support_calculator.rb
+++ b/lib/smart_answer/calculators/find_coronavirus_support_calculator.rb
@@ -38,6 +38,8 @@ module SmartAnswer::Calculators
         calculator.needs_help_with?("getting_food") && calculator.get_food != "yes"
       },
       have_you_been_made_unemployed: lambda { |calculator|
+        return false if calculator.have_you_been_made_unemployed.blank?
+
         calculator.needs_help_with?("being_unemployed") && calculator.have_you_been_made_unemployed != "no"
       },
       self_employed: lambda { |calculator|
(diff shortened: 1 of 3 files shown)


lib/smart_answer/calculators/find_coronavirus_support_calculator.rb
```diff
@@ -38,10 +38,10 @@ class FindCoronavirusSupportCalculator
         calculator.needs_help_with?("getting_food") && calculator.get_food != "yes"
       },
       have_you_been_made_unemployed: lambda { |calculator|
-        calculator.needs_help_with?("being_unemployed") && calculator.have_you_been_made_unemployed != "no"
+        calculator.needs_help_with?("being_unemployed") && calculator.have_you_been_made_unemployed?
```
theseanything, line 41: Nit: how come the extra method?

```suggestion
        return false if have_you_been_made_unemployed.blank?
        
        calculator.needs_help_with?("being_unemployed") && calculator.have_you_been_made_unemployed != "no"
```

theseanything (approved):

lib/smart_answer/calculators/find_coronavirus_support_calculator.rb
reggieb: I think `SECTIONS` is over-complicated and was trying to avoid further complicating it.

PR #4892, opened 2020-09-29
Remove previous answers from the results page

gclssvglx:
What

We should remove the previous answers section on the results page, but keep the start again link. This makes the behaviour of navigation consistent with the current tool.

Why

There is a concern about the behaviour of the previous answers section on the results page of the find coronavirus support flow. This flow uses sessions, which means answers to all questions are remembered and the user does not have to re-answer subsequent questions if going back in the flow to change an answer. This could be potentially confusing if the user expects to see related questions that are dependent on an answer to an earlier question, i.e. need help with -> getting food, expects to see questions about getting food.

[Trello](https://trello.com/c/MJWNmKBy/525-remove-previous-answers-from-the-results-page)

### DIFF
--- a/app/presenters/flow_presenter.rb
+++ b/app/presenters/flow_presenter.rb
@@ -16,6 +16,7 @@ class FlowPresenter
            :questions,
            :use_escape_button?,
            :show_escape_link?,
+           :hide_previous_answers_on_results_page?,
            to: :flow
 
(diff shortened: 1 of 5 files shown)


reggieb (approved):
The parameter name is a little long, but clear. LGTM

theseanything (approved):

PR #4902, opened 2020-10-01
Add broken links checking in smart answers

reggieb:
Add `LinkChecker` class
- Takes arrays of texts with urls and creates a report on them
- The report is generated by link-checker-api
  https://github.com/alphagov/link-checker-api

Add `BrokenLinkReport`
- Use BrokenLinkReport.call to generate a summary report on all the flows in lib/smart_answer_flows
- Use an instance of the class to run reports on individual flows or test folders, by passing into `new` the path of the folder containing the flow erb files
- Summary report provides summary, errors, suggested fix for each link, in erb file text, that is broken or may be broken
- Output is grouped by child folder name - which will match flow name if run at lib/smart_answer_flows

Example output (current report for find-coronavirus-support section of report)
```
Flow: find-coronavirus-support
====================
https://www.readyscotland.org/coronavirus/where-to-find-additional-support/
Bad Redirect

https://southallblacksisters.org.uk
Slow page
Contact the site administrator to see if they have an issue that can be fixed.

https://phw.nhs.wales/topics/latest-information-on-novel-coronavirus-covid-19/staying-well-at-home/how-are-you-feeling/
Security problem

https://www.nationaldahelpline.org.uk/Chat-to-us-online
Security problem
Contact the site administrator to see if they have an issue that can be fixed.

https://rapecrisis.org.uk/get-help
Bad Redirect
Link directly to: https://rapecrisis.org.uk/get-help/
```
The report can be generated for all the flows like this:
```ruby
BrokenLinkReport.for_erb_files_at Rails.root.join("lib/smart_answer_flows")
```

[Trello ticket](https://trello.com/c/36bzKnNO/400-broken-links-checking-in-smart-answers)

### DIFF
--- a/lib/broken_link_report.rb
+++ b/lib/broken_link_report.rb
@@ -1,0 +1,67 @@
+class BrokenLinkReport
+  def self.for_erb_files_at(flows_root_path)
+    new(flows_root_path).summary_report
+  end
+
+  attr_reader :flows_root_path
+  def initialize(flows_root_path)
+    @flows_root_path = flows_root_path
+  end
+
+  def folder_paths
+    path_to_items_within_root_folder = File.join(flows_root_path, "*")
+    Dir.glob(path_to_items_within_root_folder).select { |file| File.directory?(file) }
+  end
+
+  def folders
+    @folders ||= folder_paths.map { |folder_path| Folder.new(folder_path) }
+  end
+
+  def summary_report
+    @summary_report ||= begin
+      report = folders.each_with_object([]) do |folder, content|
+        next unless folder.report
+        next if folder.totals.links == folder.totals.ok
+
+        content << "Flow: #{folder.name}"
+        content << "===================="
+        folder.links.each do |link|
+          next unless link.status == :broken || link.status == :caution
+
+          content << link.uri
+          content << link.problem_summary
+          content << link.errors.join("\n") if link.errors.present?
+          content << link.suggested_fix if link.suggested_fix.present?
+          content << ""
+        end
+        content << ""
+      end
+      report.join("\n")
+    end
+  end
+
+  class Folder
+    attr_reader :path
+    def initialize(path)
+      @path = path
+    end
+
+    def name
+      @name ||= File.basename(path)
+    end
+
+    def erb_files
+      Dir.glob(File.join(path, "**/*.erb"))
+    end
+
+    def texts
+      erb_files.map { |file_path| File.read(file_path) }
+    end
+
+    def link_checker
+      @link_checker ||= LinkChecker.new(texts)
+    end
+    delegate :report, to: :link_checker
+    delegate :totals, :links, to: :report, allow_nil: true
+  end
+end
--- a/lib/link_checker.rb
+++ b/lib/link_checker.rb
@@ -1,0 +1,56 @@
+class LinkChecker
+  attr_reader :texts
+
+  def initialize(texts)
+    @texts = texts
+  end
+
+  def urls
+    @urls ||= extract_urls
+  end
+
+  def report
+    return if urls.blank?
+
+    @report ||= generate_report
+  end
+
+private
+
+  def extract_urls
+    return Set[] if texts.blank?
+
+    urls = texts.map { |text| URI.extract(text, %w[http https]) }
+    urls.flatten!
+    urls.map! { |uri| remove_unmatched_closing_bracket_from_end(uri) }
+    urls.to_set
+  end
+
+  def remove_unmatched_closing_bracket_from_end(text)
+    text.gsub!(/\)[^\)\w]?$/, "") if unmatched_closing_bracket?(text)
+    text
+  end
+
+  def unmatched_closing_bracket?(text)
+    text.count("(") < text.count(")")
+  end
+
+  def generate_report
+    link_report = link_checker_api.create_batch(urls, checked_within: 5)
+    wait_time = 0.5
+
+    while link_report.status == :in_progress
+      sleep(wait_time)
+      link_report = link_checker_api.get_batch(link_report.id)
+      wait_time *= 2
+    end
+    link_report
+  end
+
+  def link_checker_api
+    @link_checker_api ||= GdsApi::LinkCheckerApi.new(
+      Plek.current.find("link-checker-api"),
+      bearer_token: ENV["LINK_CHECKER_API_BEARER_TOKEN"],
+    )
+  end
+end
--- a/test/unit/link_checker_test.rb
+++ b/test/unit/link_checker_test.rb
@@ -1,0 +1,62 @@
+require "test_helper"
+require "gds_api/test_helpers/link_checker_api"
+
+class LinkCheckerTest < ActiveSupport::TestCase
+  include GdsApi::TestHelpers::LinkCheckerApi
+
+  def url
+    @url ||= "http://example.com"
+  end
+
+  def text
+    @text ||= %(There is a link <a href="#{url}">here</a>)
+  end
+
+  def link_checker
+    @link_checker ||= LinkChecker.new([text])
+  end
+
+  context "#urls" do
+    should "return set of urls in text" do
+      assert_equal Set[url], link_checker.urls
+    end
+
+    context "when text contains no links" do
+      should "return an empty set" do
+        @text = "No links"
+        assert_equal Set[], link_checker.urls
+      end
+    end
+
+    context "when text contains link markup" do
+      should "return set of urls in text" do
+        @text = "There is a link [here](#{url})"
+        assert_equal Set[url], link_checker.urls
+      end
+    end
+
+    context "when text contains link markup and url end with a bracket" do
+      should "return set of urls in text" do
+        @url = "http://example.com/this(thing)"
+        @text = "There is a link [here](#{url})"
+        assert_equal Set[url], link_checker.urls
+      end
+    end
+
+    context "when stings missing" do
+      should "return an empty set" do
+        @link_checker = LinkChecker.new(nil)
+        assert_equal Set[], link_checker.urls
+      end
+    end
+  end
+
+  context "#report" do
+    should "generate a report" do
+      stub_link_checker_api_create_batch(
+        uris: [url], checked_within: 5, status: :completed,
+      )
+      assert_equal url, link_checker.report.links.first.uri
+    end
+  end
+end
--- a/test/fixtures/smart_answer_flows/flow-with-links/outcomes/results.erb
+++ b/test/fixtures/smart_answer_flows/flow-with-links/outcomes/results.erb
@@ -1,0 +1,14 @@
+<% text_for :title do %>
+  Here be links
+  <%= link_to "This link", "http://example.com/one" %>
+<% end %>
+
+<% govspeak_for :body do %>
+  This is some text.
+  [This is a link](https://example.com/two)
+  Some more text
+<% end %>
+
+<% html_for :body do %>
+  There is a link <a href="http://example.com/three">here</a>
+<% end %>
--- a/test/unit/broken_link_report_test.rb
+++ b/test/unit/broken_link_report_test.rb
@@ -1,0 +1,150 @@
+require "test_helper"
+require "minitest/autorun"
+
+class BrokenLinkReportTest < ActiveSupport::TestCase
+  def path
+    # Defaulting path to only flow-with-links flow to limit scope of most tests (simpler and faster)
+    @path ||= Rails.root.join "test/fixtures/smart_answer_flows/flow-with-links"
+  end
+
+  def broken_link_report
+    @broken_link_report ||= BrokenLinkReport.new(path)
+  end
+
+  def folder
+    @folder ||= BrokenLinkReport::Folder.new(path)
+  end
+
+  def erb_file_path
+    File.expand_path("outcomes/results.erb", path)
+  end
+
+  def erb_file_content
+    File.read erb_file_path
+  end
+
+  def totals
+    @totals ||= {
+      links: 1,
+      ok: 0,
+    }
+  end
+
+  def status
+    @status ||= :broken
+  end
+
+  def link
+    @link ||= OpenStruct.new(
+      status: status,
+      uri: "http://example.com/foo/bar",
+      problem_summary: "Whoops",
+      errors: ["It broke"],
+      suggested_fix: "Contact webmaster",
+    )
+  end
+
+  def links
+    @links ||= [link]
+  end
+
+  def report
+    @report ||= OpenStruct.new(
+      totals: OpenStruct.new(totals),
+      links: links,
+    )
+  end
+
+  def mock_link_checker
+    OpenStruct.new(report: report)
+  end
+
+  context ".for_erb_files_at" do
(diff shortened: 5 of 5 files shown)


lib/link_checker.rb
```diff
@@ -0,0 +1,73 @@
...
+  def extract_urls
+    case data
+    when String
+      uri = URI.extract(data, %w[http https])
+
+      # markdown urls trailing bracket included in uri so removing here
+      uri.map { |u| unmatched_closing_bracket?(u) ? u.gsub(/\)\W?$/, "") : u }
+    when Array
+      urls = data.map do |datum|
+        self.class.extract_urls(datum)
```
theseanything, line 38: Slight confused as to why we need to create multiple new LinkChecker objects to rather than a single object to handle all the links?

lib/link_checker.rb
```diff
@@ -0,0 +1,73 @@
...
+
+    while link_report.status == :in_progress
+      sleep(wait_time)
+      link_report = link_checker.get_batch(link_report.id)
+      wait_time *= 2
+    end
+    link_report
+  end
+
+  def link_checker
```
theseanything, line 80: Nit: 
Found this confusing as to whether it was returning a LinkChecker object or client for the Link Checker API
```suggestion
  def link_checker_api
```

lib/link_checker.rb
theseanything: Do we use the `self.extract_urls` publicly? Not sure we need that class method - we could simplify the interface to this class to just `for_files_at(path)` and only deal with "data" as an array.

lib/link_checker.rb
```diff
@@ -0,0 +1,86 @@
...
+
+  attr_reader :data
+
+  def initialize(data)
+    @data = data
+  end
+
+  def urls
+    # Gather urls in a Set to allow simple combination and duplicate removal
+    @urls ||= extract_urls.to_set
```
theseanything, line 26: Could extract urls just return a set, instead of having a separate method?

lib/link_checker.rb
```diff
@@ -0,0 +1,86 @@
...
+  end
+
+  def extract_urls
+    case data
+    when String
+      uri = URI.extract(data, %w[http https])
```
theseanything, line 32: Does this pick only absolute paths? Do we want to check for relative paths too?

lib/link_checker.rb
```diff
@@ -0,0 +1,86 @@
...
+    link_report
+  end
+
+  def link_checker
+    @link_checker ||= GdsApi::LinkCheckerApi.new(
+      Plek.current.find("link-checker-api"),
+      bearer_token: self.class.link_checker_api_bearer_token,
+    )
+  end
+end
```
theseanything, line 86: Can we make methods private that aren't being used publicly?

test/unit/link_checker_test.rb
```diff
@@ -0,0 +1,94 @@
...
+    @data ||= text
+  end
+
+  def link_checker
+    @link_checker ||= LinkChecker.new(data)
+  end
+
+  def fixtures_path
+    Rails.root.join "test/fixtures/smart_answer_flows/flow-with-links"
+  end
```
theseanything, line 25: Can we put these in [test setup method](https://api.rubyonrails.org/classes/ActiveSupport/Testing/SetupAndTeardown.html)?

test/unit/link_checker_test.rb
```diff
@@ -0,0 +1,94 @@
...
+      end
+    end
+
+    context "when data missing" do
+      should "return an empty set" do
+        @link_checker = LinkChecker.new(nil)
+        assert_equal Set[], link_checker.urls
+      end
+    end
+  end
```
theseanything, line 52: Can we test different edges cases of text where the URL might be in markdown, HTML a tag, or include brackets in the URL?

lib/link_checker.rb
```diff
@@ -0,0 +1,86 @@
...
+    @urls ||= extract_urls.to_set
+  end
+
+  def extract_urls
+    case data
+    when String
+      uri = URI.extract(data, %w[http https])
+
+      # markdown urls trailing bracket included in uri so removing here
+      uri.map { |u| unmatched_closing_bracket?(u) ? u.gsub(/\)\W?$/, "") : u }
```
theseanything, line 35: Might be worth extracting this out into its own method, so it easy to unit test.

test/unit/link_checker_test.rb
```diff
@@ -0,0 +1,94 @@
...
+    should "extract string delimited urls" do
+      link_checker = LinkChecker.for_files_at(fixtures_path)
+      assert link_checker.urls.include?("http://example.com/one")
+    end
+
+    should "extract url from govspeak markdown" do
+      link_checker = LinkChecker.for_files_at(fixtures_path)
+      assert link_checker.urls.include?("https://example.com/two")
+    end
+  end
```
theseanything, line 89: I feel like here, we've got a weird hybrid test of integration vs unit test. These tests are covering a lot of functionality including the ability to read in from files. i.e. can we read in a file path, can we read multiple files etc... The testing of the actual text parsing for URLs seems quite cumbersome - and would be easier if there was a discrete method for extract URLs from a string.

theseanything (changes requested):

lib/link_checker.rb
```diff
@@ -0,0 +1,73 @@
...
+    def extract_urls(data)
+      new(data).extract_urls
+    end
+
+    def link_checker_api_bearer_token
+      ENV["LINK_CHECKER_API_BEARER_TOKEN"]
+    end
+
+    def for_files_at(path)
+      path = File.join(path, "**/*.erb") unless path.to_s.include?("*")
```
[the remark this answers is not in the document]
reggieb: It allows a path to be passed in with a wildcard. So `LinkChecker.for_file_at('some/path/**/*.html')` would work

lib/link_checker.rb
```diff
@@ -0,0 +1,73 @@
...
+      data = Dir.glob(path).map { |filename| File.read(filename) }
+      new(data)
+    end
+  end
+
+  attr_reader :data
+
+  def initialize(data)
+    @data = data
```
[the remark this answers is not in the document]
reggieb: It could be a string or an array of strings. It's the data being tested. It's difficult to think of a succinct name. It's the input data.

lib/link_checker.rb
reggieb: If you look at the way it was done in [govuk-coronavirus-find-support](https://github.com/alphagov/govuk-coronavirus-find-support/blob/master/lib/link_checker.rb#L29) you can see doing the `.to_set` outside that method simplifies the code.

lib/link_checker.rb
reggieb: [`URI.extract`](https://ruby-doc.org/stdlib-2.7.1/libdoc/uri/rdoc/URI/RFC2396_Parser.html#method-i-extract) retrieves full uri. This is testing the links out of the app. Relative paths can be easily tested within the app tests.

lib/link_checker.rb
reggieb: I believe it's better OOP practice to have one object instant per computation rather than to reuse an instance method on a single instance. It avoid mutability issues and is a better pattern.

lib/link_checker.rb
```diff
@@ -0,0 +1,73 @@
...
+      link_report = link_checker.get_batch(link_report.id)
+      wait_time *= 2
+    end
+    link_report
+  end
+
+  def link_checker
+    @link_checker ||= GdsApi::LinkCheckerApi.new(
+      Plek.current.find("link-checker-api"),
+      bearer_token: self.class.link_checker_api_bearer_token,
```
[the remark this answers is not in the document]
reggieb: I wrapped it in a class method to make it easier to stub in tests - but in fact the bearer token is not need in test so it can be removed.

test/unit/link_checker_test.rb
reggieb: We could - but it's a lot less flexible to do that.

test/unit/link_checker_test.rb
reggieb: There is a discrete method for extracting URLs from a string. Its [`URI.extract`](https://ruby-doc.org/stdlib-2.7.0/libdoc/uri/rdoc/URI/RFC2396_Parser.html#method-i-extract) which is part of the standard Ruby library.

lib/link_checker.rb
reggieb: Done

test/unit/link_checker_test.rb
reggieb: Done

reggieb:
@theseanything Thank you for the review. I've refactors and thanks to your remarks I think the code has simplified significantly. Could you take another look please.

lib/link_checker.rb
```diff
@@ -0,0 +1,75 @@
...
+  end
+
+  def remove_unmatched_closing_bracket_from_end(text)
+    text.gsub!(/\)\W?$/, "") if unmatched_closing_bracket?(text)
+    text
+  end
+
+  def unmatched_closing_bracket?(text)
+    text.count("(") < text.count(")")
+  end
```
theseanything, line 55: Not sure this needs to be separate method? I can't see this being used anywhere else expect once in the `remove_unmatched_closing_bracket_from_end` method

test/unit/link_checker_test.rb
theseanything: Yup - so there's a scoped functionality around being able to extract urls from text - i.e. making sure we are calling url.extract and doing bracket removal. But that functionality isn't part of the `for_files_at` method we are testing here. We should be asserting that we are getting a LinkChecker object with `texts` attribute set to the correct value. We are testing more than we need too which will make them fragile - e.g. if the `urls` breaks, then all these tests would fails, which isn't necessarily indicative of an issue with `for_fails_at` method

theseanything (changes requested):

theseanything:
@reggieb I think the commit message could be tidied up. Not sure the summary line is accurate of all the changes being made. I struggle to understand from reading why the changes are being made. Potentially might make sense to have multiple commits - one for the adding the LinkChecker class and then another one to add the rake task.

lib/link_checker.rb
reggieb: I think it makes the code easier to read.

lib/broken_link_report.rb
```diff
@@ -0,0 +1,63 @@
+class BrokenLinkReport
+  attr_reader :flows_root_path
+  def initialize(flows_root_path)
+    @flows_root_path = flows_root_path
+  end
+
+  def folder_paths
+    path_to_items_within_root_folder = File.join(flows_root_path, "*")
+    Dir.glob(path_to_items_within_root_folder).select { |file| File.directory?(file) }
```
theseanything, line 9: Does the flows_root_path ever change? Could we simplify this to -

```suggestion
  def folder_paths
    flows_path = Rails.root.join("lib/smart_answer_flows/*")
    Dir.glob(flows_path).select { |file| File.directory?(file) }
```

lib/broken_link_report.rb
```diff
@@ -0,0 +1,63 @@
...
+    def texts
+      erb_files.map { |file_path| File.read(file_path) }
+    end
+
+    def link_checker
+      @link_checker ||= LinkChecker.new(texts)
+    end
+    delegate :report, to: :link_checker
+    delegate :totals, :links, to: :report, allow_nil: true
+  end
```
theseanything, line 62: Can we make methods private?

lib/broken_link_report.rb
```diff
@@ -0,0 +1,63 @@
...
+
+  def folder_paths
+    path_to_items_within_root_folder = File.join(flows_root_path, "*")
+    Dir.glob(path_to_items_within_root_folder).select { |file| File.directory?(file) }
+  end
+
+  def folders
+    @folders ||= folder_paths.map { |folder_path| Folder.new(folder_path) }
+  end
```
theseanything, line 14: Think this a private method?

lib/broken_link_report.rb
```diff
@@ -0,0 +1,63 @@
+class BrokenLinkReport
```
theseanything, line 1: Maybe consider a module rather than a class? This logic appears quite functional as there's no need to save state to objects.

lib/broken_link_report.rb
reggieb: It makes it harder to test. In test the flow root is at "test/test/fixtures/smart_answer_flows" and "test/fixtures/smart_answer_flows/flow-with-links" thereby testing against folders with limited content and with a relatively fixed content.

lib/broken_link_report.rb
reggieb: I think it best practice in Ruby to do most of the work at the instance level. At the class/module level most operation are carried out against the class singleton - so there is potential for mutation and threading issues.

lib/broken_link_report.rb
reggieb: I don't think so because they are used by the parent class in building the report.

lib/broken_link_report.rb
reggieb: It also makes it easier to extend the behaviour (for example, if want a report of just one flow you can set the root as the flow folder).

lib/broken_link_report.rb
theseanything: I think `erb_files`is - I can't see it being used by anything else

lib/broken_link_report.rb
reggieb: I've create a class method `call` that I think is the best compromise. It provides a simple interface to provide the basic default report of all the current flows, whilst leaving the instance object available to run fine level reports (on individual flow for example) and testing.

theseanything (commented):
Are you adding the Rake task and the Zendesk ticket API call in this PR or a separate PR?

lib/broken_link_report.rb
reggieb: If you start writing your code at the module level you lose a lot of the elegance of Ruby. The code becomes simply a collection of functional snippets, and as such plays into Ruby's weaknesses. Elixir is a far better language to use if you want to go that way.

The problem with working at the module level is that all actions happen against the module's class instance singleton. That means you can't store state between methods without running the risk of another process using the same methods to mutate the state in a way you weren't expected. So the only alternative is to pass state into each method.

With a multi-step action, passing the state into each method becomes messy because you have to have an external entity to maintain the state between calls, and to split state data up so that each method only gets what it needs (or has to manage data it doesn't need).

What rapidly happens is that the action or method maintaining state become more and more complicated and difficult to test and debug.

With an instance, you are playing to Ruby's strengths. One method can call another without having to pass the state directly into the method because the instance will maintain state. So it makes it a lot easier to write simple more elegant code. It makes it easier to break processes down into smaller steps allowing the process to be interrogated to check functionality. It makes it easy to comply to good coding practices such as restricting each method to a single action.

reggieb: I've renamed `call` as `for_erb_files_at` and change the functionality so that the path is passed in. I think this addresses you're main concern that the class method lacked meaning.

theseanything (approved):

PR #4909, opened 2020-10-08
Fix issue where node name doesn't matching page

reggieb:
Add redirect in `session_answers#show action` if requested node not current
The redirect is to the node determined by the flow presenter from the session data.

[Trello ticket](https://trello.com/c/QpXnPT8z/546-fix-issue-where-node-name-doesnt-matching-page)


gclssvglx (approved):
LGTM 👍

PR #4915, opened 2020-10-13
Refactor InheritsSomeoneDiesWithoutWillFlow to remove deprecated methods

reggieb:
- Remove deprecated methods `calculate`, `save_input_as`, and `precalculate`
- Store responses in new calculator `InheritsSomeoneDiesWithoutWillCalculator`
- Determine boolean to response in calculator
- Where `next_step` is fixed for `outcome` - define it in _erb_ file rather than with `precalculate`

[Trello ticket](https://trello.com/c/kkRK9kOP/545-remove-deprecated-flow-methods-from-inherits-someone-dies-without-will)

### DIFF
--- a/lib/smart_answer_flows/inherits-someone-dies-without-will.rb
+++ b/lib/smart_answer_flows/inherits-someone-dies-without-will.rb
@@ -16,10 +16,10 @@ module SmartAnswer
         option :scotland
         option :"northern-ireland"
 
-        save_input_as :region
-
-        calculate :next_steps do
-          %i[wills_link inheritance_link]
+        on_response do |response|
+          self.calculator = Calculators::InheritsSomeoneDiesWithoutWillCalculator.new
+          calculator.region = response
+          calculator.next_steps = %i[wills_link inheritance_link]
         end
 
         next_node do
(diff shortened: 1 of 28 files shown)


lib/smart_answer_flows/inherits-someone-dies-without-will/outcomes/outcome_25.erb
```diff
@@ -3,5 +3,5 @@
 <% end %>
 
 <% govspeak_for :next_steps do %>
-  <%= render partial: 'next_step_links', locals: { next_steps: next_steps } %>
+  <%= render partial: 'next_step_links', locals: { next_steps: calculator.next_steps } %>
```
theseanything, line 6: As far as I can see `next_steps` is a constant for this flow?

```suggestion
  <%= render partial: 'next_step_links', locals: { next_steps: [:ownerless_link] } %>
```

lib/smart_answer_flows/inherits-someone-dies-without-will.rb
```diff
@@ -395,13 +357,15 @@ def define
...
 
-        next_node do |response|
-          case response
-          when "yes"
+        next_node do
+          if calculator.great_aunts_or_uncles?
             outcome :outcome_45
-          when "no"
+          else
+            calculator.next_steps = [:ownerless_link]
```
theseanything, line 368: I think this is a constant, tied specifically to that particular outcome page - we should probably set it there. Otherwise we need to make sure, every time we route to outcome_46 we need to set next_steps.

(Same for outcome_25 and outcome_67)

theseanything (changes requested):
Minor tweak - but looks good!

lib/smart_answer_flows/inherits-someone-dies-without-will/outcomes/outcome_25.erb
theseanything: and for outcome_46 and outcome_67

theseanything (approved):

PR #4937, opened 2020-10-22
Remove deprecated flow methods from maternity paternity calculator (part 1)

reggieb:
Refactors two of the four flows that comprise the maternity paternity calculator smart answer

Flows refactored:
`SmartAnswers::MaternityPaternityCalculatorFlow`
`SmartAnswer::MaternityPaternityCalculatorFlow::AdoptionCalculatorFlow`

- Use state to store attributes, and calculator to store logic
- Replace `save_as_input` with storing response on state object
- Replace nil assignment of attributes via calculator with direct assignment to state within `on_response`.
- Use validate blocks instead of raising `SmartAnswer::InvalidResponse`
- Ensure consistent storage of `monthly_pay_method` across flows
- Fix issue where stubbed calculator not behaving correctly
- Add :govuk_date_with_day time format, and use these named format in preference to individual `strftime` calls

[Trello ticket](https://trello.com/c/RpgFrtyo/549-remove-deprecated-flow-methods-from-maternity-paternity-calculator)

### DIFF
--- a/lib/smart_answer/calculators/maternity_pay_calculator.rb
+++ b/lib/smart_answer/calculators/maternity_pay_calculator.rb
@@ -105,11 +105,11 @@ module SmartAnswer::Calculators
     end
 
     def format_date(date)
-      date.strftime("%e %B %Y")
+      date.to_s(:govuk_date) if date
     end
 
     def format_date_day(date)
-      date.strftime("%A, %d %B %Y")
+      date.to_s(:govuk_date_with_day) if date
     end
 
     def payday_offset
--- a/lib/smart_answer_flows/maternity-paternity-calculator/adoption_calculator_flow.rb
+++ b/lib/smart_answer_flows/maternity-paternity-calculator/adoption_calculator_flow.rb
@@ -20,8 +20,8 @@ module SmartAnswer
           option :yes
           option :no
 
-          calculate :adoption_is_from_overseas do |response|
-            response == "yes"
+          on_response do |response|
+            self.adoption_is_from_overseas = (response == "yes")
           end
 
           next_node do
@@ -30,11 +30,9 @@ module SmartAnswer
         end
 
         date_question :date_of_adoption_match? do
-          calculate :match_date do |response|
-            response
-          end
-          calculate :calculator do
-            Calculators::AdoptionPayCalculator.new(match_date)
+          on_response do |response|
+            self.match_date = response
+            self.calculator = Calculators::AdoptionPayCalculator.new(match_date)
           end
 
           next_node do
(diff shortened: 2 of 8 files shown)


reggieb:
Please note that I have been unable to successfully refactor the `outcome` blocks in `SmartAnswer::MaternityPaternityCalculatorFlow::AdoptionCalculatorFlow` to remove `precalculate` calls.

In particular the `precalculate :pay_method` was problematic because `pay_method` is used separately in some parts of the flow and as an alias for `pay_pattern` in others. It also relies on methods being in the current context that are not simple to move into a calculator context without knock-on effects.

lib/smart_answer_flows/maternity-paternity-calculator/adoption_calculator_flow.rb
```diff
@@ -222,24 +178,14 @@ def define
           from { 2.years.ago(Time.zone.today) }
           to { 2.years.since(Time.zone.today) }
 
-          precalculate :payday_offset do
-            calculator.payday_offset
-          end
-
-          precalculate :payday_offset_formatted do
-            calculator.format_date_day payday_offset
-          end
```
theseanything, line 231: Are these being used by the question template? Can't see where these values are being set?

lib/smart_answer_flows/maternity-paternity-calculator/adoption_calculator_flow.rb
```diff
@@ -254,12 +200,8 @@ def define
           option :monthly
 
           on_response do |response|
-            calculator.pay_pattern = response
-          end
-
-          calculate :calculator do |response|
-            calculator.pay_method = response
-            calculator
```
theseanything, line 262: Woah this is odd.

lib/smart_answer/calculators/maternity_pay_calculator.rb
```diff
@@ -105,11 +105,11 @@ def valid_pay_pattern?
...
 
     def format_date(date)
-      date.strftime("%e %B %Y")
+      date.to_s(:govuk_date) if date
     end
 
     def format_date_day(date)
-      date.strftime("%A, %d %B %Y")
+      date.to_s(:govuk_date_with_day) if date
     end
```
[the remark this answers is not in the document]
theseanything: (Happy to leave it in for now)

theseanything (changes requested):
Yup - I think the direction this is going is good!

lib/smart_answer_flows/maternity-paternity-calculator/adoption_calculator_flow.rb
```diff
@@ -43,44 +41,22 @@ def define
...
-          calculate :employment_start do
-            calculator.a_employment_start
+            self.employment_start = calculator.a_employment_start
+            self.qualifying_week_start = calculator.adoption_qualifying_start
           end
 
-          calculate :qualifying_week_start do
-            calculator.adoption_qualifying_start
+          validate :error_message do |_response|
+            match_date < adoption_placement_date
```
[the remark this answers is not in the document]
reggieb: Yes I think I'm guilty of trying to be too clever there; just had a play and you are right.

```ruby
def process(n)
  p [
      n,
      !(n < 5),
      5 <= n,
      5 < n,
      n >= 5
    ]
end

(4..6).each {|n| process(n)}
```
Outputs:
```
[4, false, false, false, false]
[5, true, true, false, true]
[6, true, true, true, true]
```
I think of the two variants that work, `n >= 5` is closer to the original, so least likely to confuse, so I'll change it to: 
```ruby
adoption_placement_date >= match_date
```

lib/smart_answer_flows/maternity-paternity-calculator/adoption_calculator_flow.rb
reggieb: To be honest, I really struggled to find where some variables were used. A problem is that the system is so complicated that I had to fall back to relying on the tests, but the tests often test that a particular attribute is set as expected, without it being used. 

Due to the way flows behave, things are further complicated by the tests not testing an attribute in isolation. That is, to test C, the tests often test A and B first (particularly in "integration" tests), and often it is not clear whether A and/or B are prerequisites of C or are just set on the way. That can make it hard to cut an attribute test out.

As a result, I've tended to refactor to maintain the current behaviour and only "corrected" that behaviour if the change was simple and obvious.

lib/smart_answer_flows/maternity-paternity-calculator/adoption_calculator_flow.rb
reggieb: Absolutely. The relationship between `pay_method` and `pay_pattern` confused the hell out of me. Here they appear to be the same, but in other parts of the flow/logic they are most definitely not the same.

lib/smart_answer_flows/maternity-paternity-calculator/adoption_calculator_flow.rb
theseanything: Yup - that makes sense and is the safest!

theseanything (approved):

PR #4955, opened 2020-10-29
Remove Retail, Hospitality and Leisure Grant and Small Business Grant funds

pollygreen:
These schemes are no longer active. We need to remove them so the smart answer is accurate and users aren't confused.

### DIFF
--- a/lib/smart_answer_flows/business-coronavirus-support-finder/outcomes/results.erb
+++ b/lib/smart_answer_flows/business-coronavirus-support-finder/outcomes/results.erb
@@ -33,7 +33,7 @@
     $CTA
     ### Self-Employment Income Support Scheme
 
-    This scheme is now closed for the first grants. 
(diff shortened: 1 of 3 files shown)


lib/smart_answer_flows/business-coronavirus-support-finder/outcomes/results.erb
```diff
@@ -55,7 +55,8 @@
     $CTA
   <% end %>
 
-  <% if calculator.show?(:retail_hospitality_leisure_business_rates) %>
+  <% if calculator.show?(:
+  _hospitality_leisure_business_rates) %>
```
theseanything, line 59:
```suggestion
  <% if calculator.show?(:retail_hospitality_leisure_business_rates) %>
```

theseanything (approved):

PR #4980, opened 2020-11-11
Use whenever to schedule sending broken links report to Zendesk

reggieb:
Setup is triggered within Jenkinsfile

[whenever is a gem](https://github.com/javan/whenever) used to schedule cron tasks

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -25,6 +25,7 @@ gem "slimmer"
 gem "tilt"
 gem "uglifier"
 gem "uk_postcode"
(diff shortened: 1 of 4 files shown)


reggieb:
The main problem with this solution is that each server instance of the application would have it's own cron job, which would result in multiple reports being generated. So ... closing this.

reggieb:
PR is superseded by this: https://github.com/alphagov/govuk-puppet/pull/10837

PR #4981, opened 2020-11-12
Remove precalculate from shared maternity paternity flow

reggieb:
- Allows `PAYMENT_OPTIONS` hash to be called directly from class method to remove number of intermediate methods required.
- Adds `payment_options` instance method so that hash can be accessed easily via calculator instance in erb files.
- Fixes issue where Next button appears inline with `post_body` when present

[Trello ticket](https://trello.com/c/RpgFrtyo/549-remove-deprecated-flow-methods-from-maternity-paternity-calculator)

### DIFF
--- a/app/views/smart_answers/question.html.erb
+++ b/app/views/smart_answers/question.html.erb
@@ -37,7 +37,7 @@
 
         <%= render partial: "smart_answers/inputs/#{question.partial_template_name}", locals: { question: question } %>
 
-        <%= question.post_body %>
+        <%= content_tag(:div, question.post_body, class: "govuk-!-margin-bottom-4") if question.post_body.present? %>
 
         <input type="hidden" name="next" value="1" />
         <%= render "govuk_publishing_components/components/button", {
--- a/lib/smart_answer/calculators/maternity_pay_calculator.rb
+++ b/lib/smart_answer/calculators/maternity_pay_calculator.rb
@@ -73,8 +73,12 @@ module SmartAnswer::Calculators
       @ssp_stop = @expected_week.weeks_after(-4).begins_on
     end
 
-    def self.payment_options(period)
-      PAYMENT_OPTIONS.fetch(period, {})
+    def self.payment_options
+      PAYMENT_OPTIONS
+    end
+
+    def payment_options
+      self.class.payment_options
     end
 
     def number_of_payments
--- a/lib/smart_answer_flows/maternity-paternity-calculator/shared_adoption_maternity_paternity_flow.rb
+++ b/lib/smart_answer_flows/maternity-paternity-calculator/shared_adoption_maternity_paternity_flow.rb
@@ -1,22 +1,15 @@
 module SmartAnswer
   class SharedAdoptionMaternityPaternityFlow < Flow
     def define
-      payment_options_weekly = Calculators::MaternityPayCalculator.payment_options("weekly")
-      payment_options_every_2_weeks = Calculators::MaternityPayCalculator.payment_options("every_2_weeks")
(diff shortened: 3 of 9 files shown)


gclssvglx (approved):
LGTM 👍

reggieb:
Sorry - just given it a final test and this needs more work.

app/views/smart_answers/question.html.erb
```diff
@@ -37,7 +37,7 @@
 
         <%= render partial: "smart_answers/inputs/#{question.partial_template_name}", locals: { question: question } %>
 
-        <%= question.post_body %>
+        <%= content_tag(:div, question.post_body, class: "govuk-!-margin-bottom-4") if question.post_body.present? %>
```
reggieb, line 40: Without this change:

next_step_without_div

After change

next_step_with_div

lib/smart_answer/calculators/maternity_pay_calculator.rb
```diff
@@ -73,8 +73,12 @@ def initialize(due_date, leave_type = "maternity")
...
 
-    def self.payment_options(period)
-      PAYMENT_OPTIONS.fetch(period, {})
+    def self.payment_options
+      PAYMENT_OPTIONS
+    end
+
+    def payment_options
+      self.class.payment_options
     end
```
gclssvglx, line 82: Do we really need both of these methods?  Can it be done with just one?

gclssvglx (commented):
Looking good, just a quick question...

app/views/smart_answers/question.html.erb
theseanything: Is this a bug introduced by the previous commit? or this a pre-existing issue?

lib/smart_answer_flows/maternity-paternity-calculator/shared_adoption_maternity_paternity_flow.rb
```diff
@@ -1,24 +1,17 @@
 module SmartAnswer
   class SharedAdoptionMaternityPaternityFlow < Flow
     def define
-      payment_options_weekly = Calculators::MaternityPayCalculator.payment_options("weekly")
-      payment_options_every_2_weeks = Calculators::MaternityPayCalculator.payment_options("every_2_weeks")
-      payment_options_every_4_weeks = Calculators::MaternityPayCalculator.payment_options("every_4_weeks")
-      payment_options_monthly = Calculators::MaternityPayCalculator.payment_options("monthly")
+      payment_options = Calculators::MaternityPayCalculator.payment_options
```
theseanything, line 4: Ah this is a nice tidy up!

lib/smart_answer/calculators/maternity_pay_calculator.rb
reggieb: Within the flow definitionthe method is called directly within the `define` and `radio` calls where the a calculator instance doesn't exist. So it's easier to use a class method there.

Within the erb files, an instance of the calculator is easily accessible, so it's easier to use that there; that is to use the instance method.

app/views/smart_answers/question.html.erb
reggieb: It's a pre-existing issue I believe. I spotted it when testing this code (it happens on the page immediately prior to the one this refactored code affects).

It needs fixing, but is so minor it seems silly to create a separate PR for it.

lib/smart_answer/calculators/maternity_pay_calculator.rb
gclssvglx: Going to resolve this following a conversation on the options here.

gclssvglx (approved):
Nice work @reggieb 👍

PR #5011, opened 2020-11-26
Remove deprecated methods from the calculate your holiday entitlement flow

gclssvglx:
## What

We should remove support for the `calculate`, `precalculate` and `save_input_as` methods in flow definition for calculate-your-holiday-entitlement flow. 

## Why

These methods have been deprecated and their removal will reduce the complexity of the codebase. 

[Trello](https://trello.com/c/rBAEdQ2Q/587-remove-deprecated-flow-methods-from-calculate-your-holiday-entitlement)

- Changes to start pages are **not** continuously deployed, and are [deployed using a different process](https://github.com/alphagov/smart-answers/blob/master/doc/smart-answer-flow-development/publishing.md).

### DIFF
--- a/lib/smart_answer/calculators/holiday_entitlement.rb
+++ b/lib/smart_answer/calculators/holiday_entitlement.rb
@@ -13,28 +13,62 @@ module SmartAnswer::Calculators
     DAYS_PER_LEAP_YEAR = 366.to_d
     STANDARD_WORKING_DAYS_PER_WEEK = 5.to_d
 
-    attr_reader :working_days_per_week,
-                :hours_per_week,
+    attr_accessor :calculation_basis,
+                  :holiday_period,
+                  :leave_year_start_date,
+                  :hours_per_shift
+
+    attr_reader :hours_per_week,
                 :start_date,
                 :leaving_date,
-                :leave_year_start_date,
+                :working_days_per_week,
                 :shifts_per_shift_pattern,
                 :days_per_shift_pattern
 
-    def initialize(working_days_per_week: 0,
-                   hours_per_week: 0,
-                   start_date: nil,
-                   leaving_date: nil,
-                   leave_year_start_date: nil,
-                   shifts_per_shift_pattern: 0,
-                   days_per_shift_pattern: 0)
-      @working_days_per_week = BigDecimal(working_days_per_week, 10)
-      @hours_per_week = BigDecimal(hours_per_week, 10)
-      @start_date = start_date
-      @leaving_date = leaving_date
-      @leave_year_start_date = leave_year_start_date || calculate_leave_year_start_date
-      @shifts_per_shift_pattern = BigDecimal(shifts_per_shift_pattern, 10)
-      @days_per_shift_pattern = BigDecimal(days_per_shift_pattern, 10)
+    def initialize
+      @leave_year_start_date = calculate_leave_year_start_date
+    end
+
+    def start_date=(date)
+      @start_date = date
+      @leave_year_start_date = calculate_leave_year_start_date
+    end
+
+    def leaving_date=(date)
+      @leaving_date = date
+      @leave_year_start_date = calculate_leave_year_start_date
+    end
+
+    def hours_per_week=(hours)
+      @hours_per_week = BigDecimal(hours, 10)
+    end
+
+    def working_days_per_week=(days)
+      @working_days_per_week = BigDecimal(days, 10)
+    end
+
+    def shifts_per_shift_pattern=(shifts)
+      @shifts_per_shift_pattern = BigDecimal(shifts, 10)
+    end
+
+    def days_per_shift_pattern=(days)
+      @days_per_shift_pattern = BigDecimal(days, 10)
+    end
+
+    def holiday_entitlement_hours
+      full_time_part_time_hours_and_minutes.first
+    end
+
+    def holiday_entitlement_minutes
+      full_time_part_time_hours_and_minutes.last
+    end
+
+    def hours_daily
+      compressed_hours_daily_average.first
+    end
+
+    def minutes_daily
+      compressed_hours_daily_average.last
     end
 
     def compressed_hours_daily_average
--- a/lib/smart_answer_flows/calculate-your-holiday-entitlement.rb
+++ b/lib/smart_answer_flows/calculate-your-holiday-entitlement.rb
@@ -15,14 +15,14 @@ module SmartAnswer
         option "annualised-hours"
         option "compressed-hours"
         option "shift-worker"
-        save_input_as :calculation_basis
 
-        calculate :leaving_date
-        calculate :leave_year_start_date
-        calculate :start_date
+        on_response do |response|
+          self.calculator = Calculators::HolidayEntitlement.new
+          calculator.calculation_basis = response
+        end
 
-        next_node do |response|
-          case response
+        next_node do
+          case calculator.calculation_basis
           when "days-worked-per-week", "hours-worked-per-week", "compressed-hours", "irregular-hours", "annualised-hours"
             question :calculation_period?
           when "shift-worker"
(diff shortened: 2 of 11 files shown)


lib/smart_answer_flows/calculate-your-holiday-entitlement.rb
```diff
@@ -210,128 +233,61 @@ def define
...
+          calculator.hours_per_shift = response
+        end
+
+        validate :error_no_hours_worked do
+          calculator.hours_per_shift.positive?
+        end
 
-          hours_per_shift
+        validate :error_over_24_hours_worked do
+          calculator.hours_per_shift.between?(1, 24)
```
reggieb, line 245: As the lower limit is already covered in the `:error_no_hours_worked` validate block, I think this should just cover the condition where `hours_per_shift` is greater that 24 as the title suggests. So
```ruby
calculator.hours_per_shift <= 24
```

lib/smart_answer/calculators/holiday_entitlement.rb
```diff
@@ -37,6 +41,48 @@ def initialize(working_days_per_week: 0,
       @days_per_shift_pattern = BigDecimal(days_per_shift_pattern, 10)
     end
```
reggieb, line 42: Looking at the flow where this calculator is used, passing attributes in to the calculator on initialization is only used in tests ..... and as you've removed most of those assignments from the test, I think the `initialize` method can be significantly simplified.

It seems a shame to do all that work in the tests to remove the assignment of attributes there, to then not reap the benefit by simplifying the code here.

lib/smart_answer_flows/calculate-your-holiday-entitlement.rb
```diff
@@ -124,22 +133,28 @@ def define
...
+        validate :error_end_date_outside_leave_year_range do
+          calculator.leaving_date.blank? || YearRange.new(begins_on: calculator.leave_year_start_date).include?(calculator.leaving_date)
+        end
+
+        validate :error_start_date_before_start_leave_year_date do
+          calculator.start_date.blank? || calculator.leave_year_start_date.before?(calculator.start_date)
+        end
+
+        validate :error_start_date_outside_leave_year_range do
+          calculator.start_date.blank? || YearRange.new(begins_on: calculator.leave_year_start_date).include?(calculator.start_date)
```
reggieb, line 153: Perhaps pedantic, but `if start_date` is different to `if start_date.present?`. As the first returns true if start_date is `" "` and the other false.

I think this should be:
```ruby
calculator.start_date.nil? || YearRange.new(begins_on: calculator.leave_year_start_date).include?(calculator.start_date)
```
Same for the `:error_start_date_before_start_leave_year_date` validate block.

The other validate blocks here are OK using `calculator.leaving_date.blank?` because the original was `if leaving_date.present?`

lib/smart_answer_flows/calculate-your-holiday-entitlement.rb
```diff
@@ -154,13 +169,16 @@ def define
...
+        end
 
-          raise InvalidResponse, :error_no_hours_worked if hours_per_week <= 0
+        validate :error_no_hours_worked do
+          calculator.hours_per_week.positive?
+        end
 
-          hours_per_week
+        validate :error_over_168_hours_worked do
+          calculator.hours_per_week.between?(1, 168)
```
reggieb, line 181: Again, `between?` is great where the range is being caught within one validation, but when it is split into two (lower and higher limits), the test for the higher limit should just test that, and not use `between?`.

reggieb (commented):
A lot of excellent work here. I've added some comments.

lib/smart_answer_flows/calculate-your-holiday-entitlement.rb
```diff
@@ -59,12 +62,14 @@ def define
...
-        calculate :working_days_per_week do |response|
-          working_days_per_week = response
-          raise InvalidResponse if working_days_per_week <= 0 || working_days_per_week > 7
+        on_response do |response|
+          calculator.working_days_per_week = response
+        end
 
-          working_days_per_week
+        validate :error_message do
+          calculator.working_days_per_week.between?(1, 7)
```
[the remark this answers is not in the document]
gclssvglx: Thanks 👍

lib/smart_answer/calculators/holiday_entitlement.rb
gclssvglx: Thanks @reggieb - this is much cleaner

lib/smart_answer/calculators/holiday_entitlement.rb
```diff
@@ -37,6 +41,48 @@ def initialize(working_days_per_week: 0,
...
 
+    def start_date=(date)
+      @start_date = date
+      @leave_year_start_date = calculate_leave_year_start_date
+    end
+
+    def leaving_date=(date)
+      @leaving_date = date
+      @leave_year_start_date = calculate_leave_year_start_date
```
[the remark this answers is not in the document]
gclssvglx: `calculate_leave_year_start_date` tests `worked_partial_year?` which tests the presence of both `start_date` and `leaving_date`.  Without running `calculate_leave_year_start_date` when either the `start_date` or `leaving_date` changes, we get a lot of failing tests.  When you dig into this it's because the `leave_year_start_date` is out.

lib/smart_answer/calculators/holiday_entitlement.rb
```diff
@@ -64,6 +110,8 @@ def formatted_full_time_part_time_days
       format_number(rounded_full_time_part_time_days)
     end
 
+    alias_method :holiday_entitlement_days, :formatted_full_time_part_time_days
```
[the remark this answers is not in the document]
gclssvglx: Thanks - I've updated the views and removed all `alias_method`'s

lib/smart_answer_flows/calculate-your-holiday-entitlement.rb
gclssvglx: Updated as per suggestion - thanks 👍

lib/smart_answer_flows/calculate-your-holiday-entitlement.rb
gclssvglx: Updated as per suggestion - thanks 👍

lib/smart_answer_flows/calculate-your-holiday-entitlement.rb
gclssvglx: Updated as per suggestion - thanks 👍

lib/smart_answer/calculators/holiday_entitlement.rb
```diff
@@ -37,6 +41,48 @@ def initialize(working_days_per_week: 0,
       @days_per_shift_pattern = BigDecimal(days_per_shift_pattern, 10)
     end
 
+    def start_date=(date)
+      @start_date = date
+      @leave_year_start_date = calculate_leave_year_start_date
```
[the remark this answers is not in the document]
gclssvglx: I agree!  However, we need to update the `leave_year_start_date` in `initialize` and when either the `start_date` or `leaving_date` is changed.  `initialize` is fine of course, and if this were an ActiveRecord model I'd simply check if these attributes are dirty and run a callback.  But, I don't believe there is a way to do that in a plain ol' ruby object.

lib/smart_answer/calculators/holiday_entitlement.rb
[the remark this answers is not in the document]
reggieb: OK - It will have to stay as it is then :shrug:

lib/smart_answer/calculators/holiday_entitlement.rb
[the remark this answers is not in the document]
reggieb: OK

reggieb (approved):
Looks good!

PR #5013, opened 2020-11-27
Add RSpec Feature test for coronavirus find support flow

reggieb:
Heading says it all really.

[Trello ticket](https://trello.com/c/KUryV0ib/635-add-rspec-integration-tests-for-find-support)

### DIFF
--- a/spec/features/smart_answer/find_coronavirus_support_flow_spec.rb
+++ b/spec/features/smart_answer/find_coronavirus_support_flow_spec.rb
@@ -1,0 +1,249 @@
+require "rails_helper"
+
+RSpec.feature "FindCoronavirusSupportFlow", type: :feature do
+  def start_flow_to_need_help_with
+    visit "/find-coronavirus-support"
+    expect(page).to have_selector("h1", text: headings[:flow_title])
+
+    click_govuk_start_button
+    expect(page).to have_selector("h1", text: headings[:need_help_with])
+  end
+
+  def continue_to_results
+    click_button "Continue"
+    expect(page).to have_selector("h1", text: headings[:nation])
+
+    choose "England"
+    click_button "Continue"
+    expect(page).to have_selector("h1", text: headings[:results])
+  end
+
+  let(:headings) do
+    {
+      flow_title: "Find out what support you can get if you’re affected by coronavirus",
+      need_help_with: "What do you need help with because of coronavirus?",
+      feel_unsafe: "Do you feel unsafe where you live?",
+      afford_rent_mortgage_bills: "Are you finding it hard to pay your rent, mortgage or bills?",
+      afford_food: "Are you finding it hard to afford food?",
+      get_food: "Are you able to get food?",
+      self_employed: "Are you self-employed, a freelancer, or a sole trader?",
+      have_you_been_made_unemployed: "Have you been made redundant or told to stop working?",
+      worried_about_work: "Are you worried about going in to work?",
+      are_you_off_work_ill: "Are you off work because you’re ill or self-isolating?",
+      have_somewhere_to_live: "Do you have somewhere to live?",
+      have_you_been_evicted: "Have you been evicted?",
+      mental_health_worries: "Are you worried about your mental health, or another adult or child’s mental health?",
+      nation: "Where do you want to find information about?",
+      results: "Information based on your answers",
+    }
+  end
+
+  let(:result_subheadings) do
+    {
+      feeling_unsafe: "Feeling unsafe",
+      paying_bills: "Paying your rent, mortgage, or bills",
+      getting_food: "Getting food",
+      being_unemployed: "Being made redundant or unemployed, or not having any work",
+      going_to_work: "Being worried about working, or off work because you’re self-isolating",
+      somewhere_to_live: "Having somewhere to live",
+      mental_health: "Mental health and wellbeing",
+    }
+  end
+
+  let(:need_help_with) do
+    {
+      feeling_unsafe: "Feeling unsafe where you live, or what to do if you’re worried about the safety of another adult or child",
+      paying_bills: "Paying your rent, mortgage, or bills",
+      getting_food: "Getting food",
+      being_unemployed: "Being made redundant or unemployed, or not having any work",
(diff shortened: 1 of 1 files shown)


spec/features/smart_answer/find_coronavirus_support_flow_spec.rb
```diff
@@ -0,0 +1,249 @@
...
+      feeling_unsafe: "Feeling unsafe where you live, or what to do if you’re worried about the safety of another adult or child",
+      paying_bills: "Paying your rent, mortgage, or bills",
+      getting_food: "Getting food",
+      being_unemployed: "Being made redundant or unemployed, or not having any work",
+      going_to_work: "Being worried about working, or off work because you’re self-isolating",
+      somewhere_to_live: "Having somewhere to live",
+      mental_health: "Mental health and wellbeing, including information for children",
+      none: "None of these",
+    }
+  end
```
gclssvglx, line 64: I understand why you've had to do this, but having to repeat these here feels wrong.  Should we not be making use of the `locales` files instead?

spec/features/smart_answer/find_coronavirus_support_flow_spec.rb
```diff
@@ -0,0 +1,249 @@
...
+
+    check need_help_with[:being_unemployed]
+    click_button "Continue"
+    expect(page).to have_selector("h1", text: headings[:self_employed])
+
+    choose "No"
+    click_button "Continue"
+    expect(page).to have_selector("h1", text: headings[:have_you_been_made_unemployed])
+
+    choose "Yes, I’ve been made redundant or I’m out of work, or might be soon"
```
gclssvglx, line 121: Same as above

spec/features/smart_answer/find_coronavirus_support_flow_spec.rb
```diff
@@ -0,0 +1,249 @@
...
+    choose "Yes, I am"
+    continue_to_results
+    expect(page).to have_selector("h2", text: result_subheadings[:mental_health])
+  end
+
+  scenario "None of these" do
+    start_flow_to_need_help_with
+    check need_help_with[:none]
+    continue_to_results
+    expect(page).to have_text("Based on your answers, there’s no specific information for you in this service at the moment")
```
gclssvglx, line 186: Same as above

spec/features/smart_answer/find_coronavirus_support_flow_spec.rb
```diff
@@ -0,0 +1,249 @@
...
+  def start_flow_to_need_help_with
+    visit "/find-coronavirus-support"
+    expect(page).to have_selector("h1", text: headings[:flow_title])
+
+    click_govuk_start_button
+    expect(page).to have_selector("h1", text: headings[:need_help_with])
+  end
+
+  def continue_to_results
+    click_button "Continue"
```
gclssvglx, line 13: This relies on the button being `Continue`.  However, there is the option to change the label on that button now.  Is there a way to mitigate having to update all occurrences of `Continue` in this file when someone changes the button text?  Could we use an ID or `locale` files?

gclssvglx (commented):
Hi @reggieb - looks great.  But I'm concerned we're not using the locale files.  Is there a reason why we can't do that?

reggieb:
Yes - I absolutely agree that using locale file would be the way to go. I had a [similar discussion with Sean in the previous RSpec PR](https://github.com/alphagov/smart-answers/pull/5000#discussion_r527702237).

The problem at the moment is that these texts are defined within the flow erb files and do not use locales, and I could not see a simple way to grab the text directly from the erb files.

Moving texts to translation locale files is a separate piece of work I believed.

gclssvglx:
> Yes - I absolutely agree that using locale file would be the way to go. I had a [similar discussion with Sean in the previous RSpec PR](https://github.com/alphagov/smart-answers/pull/5000#discussion_r527702237).
> 
> The problem at the moment is that these texts are defined within the flow erb files and do not use locales, and I could not see a simple way to grab the text directly from the erb files.
> 
> Moving texts to translation locale files is a separate piece of work I believed.

Yeah, I get that converting the flows to use `locale`s is a big chunk of work.  So long as it's on the radar...!

gclssvglx (approved):
@reggieb Nice tests - like it 👍

spec/features/smart_answer/find_coronavirus_support_flow_spec.rb
reggieb: "Continue" isn't the default button text. The default is "Next step".

OK - so say we use a translation file here. We have two next step labels, which need identifying with a key. I'd suggest `:continue` and `:next_step`. The code then becomes something like:

```ruby
click_button i18n.t('flow.next_button.continue')
```
If we change the button back to default we'd need to change each occurrence to:

```ruby
click_button i18n.t('flow.next_button.next_step')
```

We could wrap it all in a method. So we have something like:

```ruby
click_next_button(:continue)
```
But then we've obscured what is actually going on by one step - a step that I'd argue is unnecessary.

I am very conscious of the way RSpec feature tests have been implemented in other GDS projects; where everything is wrapped up in methods in what I call pseudo-cucumber. It makes it very hard to work out what the tests are actually doing. This is especially annoying when the standard RSpec and Capybara syntax is so descriptive without this unnecessary obscuration.

So I would argue that considering all the options, the clearest and simplest solution is just to use:
```ruby
click_button "Continue"
```
Everyone familiar with Capybara should understand what it's doing, and it is just as simple to change "Continue" to "Next Step" as it is `:continue` to `:next_step`.

Sure - where the text is long, use translation files. But for very basic actions like this, I'd argue that the simplest solution is the best.

gclssvglx: I was hoping we could do something like this...

```
# config/locales/en.yml
next_button: "Next step"
...
flows:
  find_support: 
    next_button: "Continue"
```

Then ...
```
# lib/smart_answer_flows/find-coronavirus-support.rb
module SmartAnswer
  class FindCoronavirusSupportFlow < Flow
    def define
      # ...
      button_text i18n.t('flows.find_support.next_button')
```

PR #5020, opened 2020-12-01
Ensure errors connecting to GdsApi are captured in Sentry

reggieb:
Update ContentItemRetriever to rescue any GdsApi exception. Use `GdsApi::BaseError` as [all `GdsApi` exceptions inherit from `BaseError`](https://github.com/alphagov/gds-api-adapters/blob/master/lib/gds_api/exceptions.rb)

Notify Sentry (via [GovukError](https://github.com/alphagov/govuk_app_config#manual-error-reporting)) that exception has occurred.

On 27th Nov we had an incident that cause the content store to return 5xx, which subsequently caused errors in Smart Answers. However, these errors did not appear to be captured by Sentry. 

[Trello ticket](https://trello.com/c/kendYDw0/637-understand-why-5xx-errors-from-the-content-store-arent-seen-in-sentry)

### DIFF
--- a/test/unit/services/content_item_retriever_test.rb
+++ b/test/unit/services/content_item_retriever_test.rb
@@ -43,29 +43,49 @@ class ContentItemRetrieverTest < ActiveSupport::TestCase
     end
 
     context "when content item can't be found" do
-      should "return empty content item hash" do
+      setup do
         response = { status: 404, body: {}.to_json }
         stub_request(:get, @request_url).to_return(response)
+      end
 
+      should "return empty content item hash" do
         assert_equal ContentItemRetriever.fetch(@slug), {}
+      end
+
+      should "notify Sentry" do
+        GovukError.expects(:notify)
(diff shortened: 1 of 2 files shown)


test/unit/services/content_item_retriever_test.rb
```diff
@@ -62,11 +62,15 @@ class ContentItemRetrieverTest < ActiveSupport::TestCase
...
-
+        stub_content_store_isnt_available
         assert_equal ContentItemRetriever.fetch(@slug), {}
       end
+
+      should "notifies Sentry" do
+        stub_content_store_isnt_available
+        GovukError.expects(:notify)
+        ContentItemRetriever.fetch(@slug)
+      end
```
theseanything, line 73: Should we be adding this expectation to the other tests too?

reggieb: Yes. That makes sense. I've updated the other specs. Thank you

theseanything (approved):

PR #5023, opened 2020-12-01
Use govuk design pattern for suffix labels on questions

reggieb:
Update money questions by moving suffix label to input component suffix rather than part of hint text

An example showing the effect of this change:

per_week_as_suffix

### DIFF
--- a/app/presenters/money_question_presenter.rb
+++ b/app/presenters/money_question_presenter.rb
@@ -6,7 +6,7 @@ class MoneyQuestionPresenter < QuestionPresenter
   end
 
   def hint_text
-    text = [body, hint, suffix_label].reject(&:blank?).compact.join(", ")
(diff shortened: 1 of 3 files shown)


app/views/smart_answers/inputs/_salary_question.html.erb
```diff
@@ -8,7 +8,7 @@
   id: "response_amount",
   heading_size: "l",
   width: 10,
-  hint: "in £",
+  suffix: "in £",
```
theseanything, line 11: I think maybe this should be using the prefix attribute instead: https://components.publishing.service.gov.uk/component-guide/input/with_prefix

but that's out of scope for this PR - so maybe just leave out the salary questions

reggieb: Thank you. I'll remove that.

theseanything (approved):

PR #5029, opened 2020-12-04
Fix an empty suffix showing in money questions if no suffix specified

reggieb:
A bug was introduced when money questions were set to use a suffix on the input component rather than adding the suffix text to the hint.

The bug is that an empty suffix box appears after the input.

empty_money_question_suffix

This update fixes that bug by only adding the suffix when present, and thus removes the empty box

[Trello](https://trello.com/c/70RZ8gLI/650-fix-bug-where-empty-money-input-suffix-box-appeared-if-no-suffix-was-specified)

### DIFF
--- a/app/presenters/question_presenter.rb
+++ b/app/presenters/question_presenter.rb
@@ -45,7 +45,8 @@ class QuestionPresenter < NodePresenter
   end
 
   def suffix_label
-    @renderer.content_for(:suffix_label)
+    content = @renderer.content_for(:suffix_label)
+    content.presence
   end
 
   def has_labels?
--- a/test/unit/question_presenter_test.rb
+++ b/test/unit/question_presenter_test.rb
@@ -48,6 +48,12 @@ module SmartAnswer
       assert_equal "suffix-label-text", @presenter.suffix_label
     end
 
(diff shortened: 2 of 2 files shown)


app/views/smart_answers/inputs/_money_question.html.erb
```diff
@@ -9,7 +9,7 @@
   heading_size: "l",
   width: 10,
   hint: question.hint_text,
-  suffix: question.suffix_label,
+  suffix: (question.suffix_label if question.suffix_label.present?),
```
theseanything, line 12: Good catch! Think if we put this logic into the question presenter https://github.com/alphagov/smart-answers/blob/68dd25abae9f324ca9627b5fa7be8b31d95e746b/app/presenters/question_presenter.rb#L48, we can test it too.

theseanything (changes requested):

app/views/smart_answers/inputs/_money_question.html.erb
reggieb: That makes sense - thanks. I've moved the modification to the presenter.

theseanything (approved):

PR #5055, opened 2020-12-21
Add self-isolation question to find coronavirus support

theseanything:
## What

Feedback showed that users were struggling with other things, such as access to food, when self-isolating. Are we covering these needs when users say they're self-isolating?

## Why

So that those that are self-isolating can get the help and support they need

[Trello](https://trello.com/c/A3P5vEXu/647-what-other-needs-do-users-have-when-they-self-isolate-are-we-meeting-them)


gclssvglx (approved):
LGTM 👍

PR #5066, opened 2020-12-24
End of transition Benefits abroad changes

deborahchua:
We have various changes to the logic, outcomes, and questions for the following benefits:

- Job Seekers Allowance (JSA)
- Winter Fuel Payment (WFP)
- Employment and Support Allowance (ESA)
- Industrial Injuries Disablement Benefit (IIDB)
- Benefits for carers and people with disabilities(DB)

We have also temporarily removed:
- Child Benefit
- Tax credits

See individual commit messaged for more information.

Trello card: https://trello.com/c/3LMuXVpZ/2255-5-benefits-abroad-smart-answer-end-of-transition-and-iidb-changes


lib/smart_answer/calculators/uk_benefits_abroad_calculator.rb
```diff
@@ -106,6 +106,10 @@ def employer_paying_ni_not_ssp_country_entitled?
       (COUNTRIES_OF_FORMER_YUGOSLAVIA + %w[barbados guernsey jersey israel turkey]).include?(country)
     end
 
+    def jersey_guernsey?
```
ChrisBAshton, line 114: Minor: This might be better renamed as `channel_islands?`? (Would need to rename the outcome too). It's better to give the name of the group of countries if possible, just like we do with EEA countries.

lib/smart_answer_flows/uk-benefits-abroad.rb
```diff
@@ -82,7 +82,7 @@ def define
           case calculator.benefit
           when "jsa"
             if calculator.already_abroad && calculator.eea_country?
-              outcome :jsa_eea_already_abroad_outcome # A3 already_abroad
+              outcome :jsa_eea_already_abroad_outcome
```
ChrisBAshton, line 85: Hmmm nothing has changed here except for removal of a comment, which suggests to me that the tests being removed/rewritten should be moved to a more relevant (previous) commit.

lib/smart_answer/calculators/uk_benefits_abroad_calculator.rb
```diff
@@ -106,6 +106,11 @@ def employer_paying_ni_not_ssp_country_entitled?
       (COUNTRIES_OF_FORMER_YUGOSLAVIA + %w[barbados guernsey jersey israel turkey]).include?(country)
     end
 
+    def social_security_countries?
```
ChrisBAshton, line 109: This seems to be a helper method in place of having to write `social_security_countries_jsa? && social_security_countries_iidb? && ...`.

Seems a good and useful idea, but rather than duplicate the country names I'd be tempted to call these submethods as above, so that there is less we have to keep in sync. E.g. if a country got added to `social_security_countries_jsa?` we'd also have to remember to update `social_security_countries?`, which feels like a potentially dangerous gap we could miss.

test/integration/smart_answer_flows/uk_benefits_abroad_test.rb
```diff
@@ -1275,6 +1275,14 @@ class UKBenefitsAbroadTest < ActiveSupport::TestCase
           assert_current_node :jsa_social_security_already_abroad_outcome
         end
       end
+      context "answer Albania" do # not EEA or SS country
```
ChrisBAshton, line 1824: Worth fixing this up to the previous commit IMO. Ideally we shouldn't commit new tests outside of the relevant app code change.

lib/smart_answer_flows/uk-benefits-abroad.rb
```diff
@@ -103,7 +103,8 @@ def define
           when "winter_fuel_payment"
             if calculator.country_eligible_for_winter_fuel_payment?
               if calculator.going_abroad
-                outcome :wfp_going_abroad_outcome # A9 going_abroad
+                question :worked_in_eea_or_switzerland?
+                # outcome :wfp_going_abroad_outcome # A9 going_abroad
```
ChrisBAshton: Comment has slipped in.

lib/smart_answer_flows/uk-benefits-abroad.rb
```diff
@@ -608,11 +609,19 @@ def define
         next_node do |response|
           case response
           when "before_jan_2021"
-            outcome :jsa_eea_going_abroad_maybe_outcome
+            if calculator.benefit == "jsa"
+              outcome :jsa_eea_going_abroad_maybe_outcome
+            elsif calculator.benefit == "winter_fuel_payment"
+              outcome :wfp_going_abroad_eea_maybe_outcome
+            end
           when "after_jan_2021"
```
ChrisBAshton: This can be shortened to `when "after_jan_2021", "no"`, to save you having to repeat the same outcome for each case.

test/integration/smart_answer_flows/uk_benefits_abroad_test.rb
```diff
@@ -303,18 +303,90 @@ class UKBenefitsAbroadTest < ActiveSupport::TestCase
...
         end
       end
+      # context "answer other country" do
+      #   setup do
+      #     add_response "albania"
+      #   end
+      #   should "take you to not eligible outcome" do
+      #     assert_current_node :wfp_not_eligible_outcome
+      #   end
+      # end
```
ChrisBAshton: Comment has slipped in.

test/integration/smart_answer_flows/uk_benefits_abroad_test.rb
```diff
@@ -379,6 +379,103 @@ class UKBenefitsAbroadTest < ActiveSupport::TestCase
...
+            should "ask has one of your parents ever lived in an EU country, Norway, Iceland, Liechtenstein or Switzerland?" do
+              assert_current_node :parents_lived_in_eea_or_switzerland?
+            end
+            context "answer Yes before January 2021" do # parents EU?
+              setup do
+                add_response :before_jan_2021
+              end
+              should "go to outcome WFP EEA maybe" do
+                assert_current_node :wfp_going_abroad_eea_maybe_outcome
+              end
```
ChrisBAshton, line 426: General point throughout this PR - you could make the code easier to scan by changing it to:

```ruby
should "go to outcome WFP EEA maybe" do
  add_response :before_jan_2021
  assert_current_node :wfp_going_abroad_eea_maybe_outcome
end
```

`setup` blocks are only really useful if you then have multiple `should` blocks. But if you only have one `should` block, you may as well put the code inline.

test/integration/smart_answer_flows/uk_benefits_abroad_test.rb
```diff
@@ -476,6 +476,15 @@ class UKBenefitsAbroadTest < ActiveSupport::TestCase
           end
         end
       end
+
+      context "answer Kosovo" do # SS country
```
ChrisBAshton, line 480: Similar point about fixing up to previous commit.

Also looks like it applies to the next four commits or so.

lib/smart_answer_flows/uk-benefits-abroad.rb
```diff
@@ -552,9 +552,9 @@ def define
 
         next_node do |response|
           if calculator.going_abroad && response == "esa_under_a_year_medical"
-            outcome :esa_going_abroad_under_a_year_medical_outcome # A27 going_abroad
+            outcome :esa_going_abroad_under_a_year_medical_outcome
           elsif calculator.going_abroad && response == "esa_under_a_year_other"
-            outcome :esa_going_abroad_under_a_year_other_outcome # A28 going_abroad
+            outcome :esa_going_abroad_under_a_year_other_outcome
```
ChrisBAshton, line 550: Nitpick: you've removed these comments, but not the one on line 559.
Happy to keep them or remove them either way, but we should probably be consistent throughout the file.

lib/smart_answer_flows/uk-benefits-abroad.rb
```diff
@@ -612,13 +612,15 @@ def define
               outcome :jsa_eea_going_abroad_maybe_outcome
             elsif calculator.benefit == "winter_fuel_payment"
               outcome :wfp_going_abroad_eea_maybe_outcome
+            elsif calculator.benefit == "esa"
+              outcome :esa_going_abroad_eea_outcome
             end
           when "after_jan_2021"
-            if calculator.benefit == "jsa" || calculator.benefit == "winter_fuel_payment"
+            if calculator.benefit == "jsa" || calculator.benefit == "winter_fuel_payment" || calculator.benefit == "esa"
```
ChrisBAshton: At this point it becomes easier to change to:

```suggestion
            if %w[esa jsa winter_fuel_payment].include?(calculator.benefit)
```

test/integration/smart_answer_flows/uk_benefits_abroad_test.rb
```diff
@@ -496,6 +496,137 @@ class UKBenefitsAbroadTest < ActiveSupport::TestCase
...
+        # end
+
+        # context "answer kosovo" do
+        #   setup do
+        #     add_response "kosovo"
+        #   end
+        #   should "take you to other outcome" do
+        #     assert_current_node :esa_going_abroad_eea_outcome
+        #   end
+        # end
```
ChrisBAshton: A couple more commented blocks have snuck in. Would be good to remove them and only add them when they're uncommented.

Or in this case, just don't comment them out at all - looks like it wasn't needed, if I've understood 21789ced3f55a85448e23e68fa28e677734e64ee.

lib/smart_answer_flows/uk-benefits-abroad.rb
```diff
@@ -102,10 +102,10 @@ def define
             end
           when "winter_fuel_payment"
             if calculator.country_eligible_for_winter_fuel_payment?
-              if calculator.going_abroad && calculator.country == "ireland"
+              if calculator.country == "ireland"
                 question :is_british_or_irish?
               else
-                outcome :worked_in_eea_or_switzerland? # A7 already_abroad
+                question :worked_in_eea_or_switzerland? # A7 already_abroad
```
ChrisBAshton, line 108: Minor: Ideally this would be fixed up to the commit where we originally added `outcome`, as it's always been a `question`.

ChrisBAshton (commented):
Well done for this substantial piece of work 👏  It's hard to coordinate so many different content changes!

The vast majority of my comments are minor/nitpicks, and I understand if on a PR this size it's too cumbersome to fix up commits etc, so it may be easier to push up a 'code review amends' commit, or just ignore some of the suggestions. I didn't see anything blocking merge though (with the possible exception of the new `social_security_countries?` method, which I worry might fall out of sync and cause bugs to slip through the cracks).

lib/smart_answer/calculators/uk_benefits_abroad_calculator.rb
beccapearce: Just looking into this. I agree this is a dangerous gap. I was looking at how to fix it. 

In the country question for those going abroad we use `social_security_countries_jsa?` to direct `COUNTRIES_OF_FORMER_YUGOSLAVIA guernsey jersey new-zealand` to ask if they have worked or lived int he EEA. 

but also in the country question we use `social_security_countries?` to direct `COUNTRIES_OF_FORMER_YUGOSLAVIA barbados bermuda canada guernsey jersey israel jamaica mauritius new-zealand philippines turkey usa` to `jsa_social_security_already_abroad_outcome`.

These countries are the same as used in the method `social_security_countries_bereavement_benefits?` however `social_security_countries?` is being used in the jsa journey so I don't really want to use the `social_security_countries_bereavement_benefits?` method as it may get confusing later if we just want to change the bereavement_benefits journey.

Not sure what would be the best way to distinguish the 2 methods from each other, would renaming the methods to `social_security_countries_jsa_going_abroad?` and `social_security_countries_jsa_already_abroad?` be an acceptable solution? Or is there a better direction to take this in?

ChrisBAshton: Ah, I understood `social_security_countries?` to mean "all countries which have some form of UK social security". In which case you'd just call all the relevant submethods and tack a `.uniq` on the end to filter out any duplicate country names. If it's not actually used like that, then I'd just remove it - better to refer explicitly to the relevant 'submethods' rather than rely on the wrong abstraction.

I've re-read your comment a few times and still don't really understand the nuances, so I suspect we're better off just removing the method if it isn't very helpful! But more generally... it sounds like perhaps you might want to consider a method with an argument:

```ruby
def social_security_countries_jsa(already_abroad: false)
  # return different lists depending on the `already_abroad` argument
end
```

beccapearce: Sorry I don't think I've explained it super well! But I really like using`def social_security_countries_jsa(already_abroad: false)` I think that makes it clearer! 

I'll make those changes now! Thank you! :)

lib/smart_answer_flows/uk-benefits-abroad.rb
```diff
@@ -601,14 +592,151 @@ def define
...
+            if calculator.benefit == "jsa"
+              outcome :jsa_eea_going_abroad_maybe_outcome
+            elsif calculator.benefit == "winter_fuel_payment"
+              outcome :wfp_going_abroad_eea_maybe_outcome
+            elsif calculator.benefit == "esa"
+              if calculator.going_abroad
+                outcome :esa_going_abroad_eea_outcome
+              else
+                outcome :esa_already_abroad_eea_outcome
+              end
```
ChrisBAshton, line 613: This (and several examples below) might benefit from being ternary statements:

```ruby
elsif calculator.benefit == "esa"
  outcome calculator.going_abroad ? : esa_going_abroad_eea_outcome : : esa_already_abroad_eea_outcome
```

Helps avoid the code from getting too 'nested' :)

lib/smart_answer_flows/uk-benefits-abroad.rb
```diff
@@ -601,14 +592,151 @@ def define
...
+            elsif calculator.benefit == "disability_benefits"
+              if calculator.going_abroad
+                outcome :db_going_abroad_eea_outcome
+              else
+                outcome :db_already_abroad_eea_outcome
+              end
+            end
+          when "after_jan_2021"
```
ChrisBAshton, line 652:
```suggestion
          when "after_jan_2021", "no"
```

Then remove lines 670-689 (unless I'm wrong about them being identical!).

lib/smart_answer_flows/uk-benefits-abroad.rb
```diff
@@ -601,14 +592,151 @@ def define
...
+          when "one_year_or_less"
+            if calculator.channel_islands?
+              outcome :jsa_channel_islands_outcome
+            else
+              outcome :jsa_social_security_going_abroad_outcome
+            end
+          when "more_than_one_year"
+            outcome :jsa_social_security_going_abroad_outcome
+          end
+        end
```
ChrisBAshton, line 733: With only two cases, and with a conditional that otherwise gives a default outcome, I think this could be simplified:

```ruby
next_node do |response|
  if response == "one_year_or_less" && calculator.channel_islands?
    outcome :jsa_channel_islands_outcome 
  else
    outcome :jsa_social_security_going_abroad_outcome
  end
end
```

test/integration/smart_answer_flows/uk_benefits_abroad_test.rb
ChrisBAshton: I think this point still stands - by far the biggest change in this PR is this test file, (2082 additions), which could probably be reduced to maybe 1000 additions by squishing some of the unnecessary `setup` blocks. The next person to edit this Smart Answer would thank you!

ChrisBAshton (approved):
Thanks Becca. I've had a look at the overall change, rather than commit by commit, and left a few comments below.

Aware that this is due soon and my comments aren't blocking, so happy to approve, though it would be great to slim down the uk_benefits_abroad_test.rb file in a new commit if you have time. (It's also low risk as it's not changing app behaviour).

Otherwise LGTM 👍

PR #5079, opened 2021-01-05
Update results.erb

DGarland5:
Added Tier 4 to Local Restrictions Support Grant (Closed) as the grant was extended to cover it.


hannahjump:
This isn't you, but do you think we could just say Tier 2, Tier 3, or Tier 4 rather than 'high' etc? I imagine that was added when we didn't know the style

hannahjump (approved):

PR #5093, opened 2021-01-11
Add integration test for the calculate-employee-redundancy-pay flow

gclssvglx:
A new spec for the calculate-employee-redundancy-pay flow

[Trello](https://trello.com/c/16rlsnnn/713-add-rspec-test-for-calculate-employee-redundancy-pay)


theseanything (changes requested):
Looks good! I think we should remove the all old mintiest integrations test with this PR, so we aren't running both

gclssvglx:
> Looks good! I think we should remove the all old mintiest integrations test with this PR, so we aren't running both

Yes, quite right - done.

PR #5098, opened 2021-01-14
Change default button text from `Next step` to `Continue`

theseanything:
This removes inconsistency between Smart Answer flows where the Find Coronavirus Support flow uses "Continue", where as all other flows use "Next step". Content designers have agreed that "Continue" is better wording for the button.


gclssvglx (approved):
LTGM - `Continue` always did make more sense than `Next step` to me.

PR #5120, opened 2021-01-21
Remove extra sector question in business support flow

theseanything:
This merges the "Is your a business a nightclub, dancehall or adult entertainment venue?" question as another option within the pre-existing sector question. The sector question has also been refactored to make it clear that users can select multiple options.


test/integration/smart_answer_flows/business_coronavirus_support_finder_test.rb
```diff
@@ -1,58 +0,0 @@
-require_relative "../../test_helper"
```
gclssvglx, line 1: I understand why this needs deleting - but it doesn't really relate to this PR.  Should it be a separate PR?

lib/smart_answer/calculators/business_coronavirus_support_finder_calculator.rb
```diff
@@ -65,11 +64,11 @@ def initialize
       },
       lrsg_open: lambda { |calculator|
         calculator.business_based == "england" &&
-          calculator.restricted_sector == "no"
+          !calculator.sectors.include?("nightclubs_or_adult_entertainment")
```
gclssvglx, line 67: Just a suggestion - would using `exclude?` be better here?  I know `exclude?` is defined in `ActiveSupport` - however I feel it reads better.

test/integration/smart_answer_flows/business_coronavirus_support_finder_test.rb
theseanything: Yup - so because I had changed the flow I'd need to update these tests to make them pass. Figured if they are going to be deleted anyways thats the easier change to make.

lib/smart_answer/calculators/business_coronavirus_support_finder_calculator.rb
theseanything: ahh nice! I'll update that

test/integration/smart_answer_flows/business_coronavirus_support_finder_test.rb
gclssvglx: No worries

lib/smart_answer/calculators/business_coronavirus_support_finder_calculator.rb
gclssvglx: Nice - reads much better 👍

PR #5124, opened 2021-01-25
Remove files related to marriage abroad flattening

thomasleese:
We don't need these anymore since we flattened all the country outcomes in https://github.com/alphagov/smart-answers/pull/3952. It's confusing to keep them around as it suggests that they're still being used when they're not.


theseanything:
Agree it's super confusing whether we need to use this or not - wondering whether we'd still need it if we changed content in one of the partials (e.g. `lib/smart_answer_flows/marriage-abroad/outcomes/_contact_nearest_embassy_of_country.erb`) - i.e. would the process without the flattener, be to change every country individually?

thomasleese:
> Agree it's super confusing whether we need to use this or not - wondering whether we'd still need it if we changed content in one of the partials (e.g. `lib/smart_answer_flows/marriage-abroad/outcomes/_contact_nearest_embassy_of_country.erb`) - i.e. would the process without the flattener, be to change every country individually?

As I understand it these partials aren't used at all anymore. The flattening process effectively expanded all the partials into the final outcomes, allowing content designers to make changes, but as you say it means if we want to make bulk changes we need to change each country individually.

I don't think we can use the flattening scripts anymore because the content designers might have made changes to the final outcome, so if a bulk change to one of the partials that were used in the past is required, we'll have to find a new way of making that change, rather than using the flattening scripts.

theseanything (approved):

PR #5133, opened 2021-01-29
Fix calculator

1pretz1:
Few changes to the self assessment calculator as requested for tax year 2019-20:

* Filing deadline pushed back until 28th February (late filing fee starting from 1st March)
* Late payment interest rates start on the 1st of Feb rather than 2nd
* Validation removed that ensures payment date is before filing date
* Content removed concerning outdated calculator (as it now should be up to date 😄) 

Trello:
https://trello.com/c/Uff9Ko9M/2339-fix-anomaly-in-estimate-your-penalty-for-late-self-assessment-tax-returns

### DIFF
--- a/lib/smart_answer/calculators/self_assessment_penalties.rb
+++ b/lib/smart_answer/calculators/self_assessment_penalties.rb
@@ -85,6 +85,8 @@ module SmartAnswer::Calculators
     end
 
     def valid_payment_date?
+      return true if tax_year == "2019-20"
+
       filing_date <= payment_date
     end
 
--- a/lib/smart_answer/calculators/self_assessment_penalties.rb
+++ b/lib/smart_answer/calculators/self_assessment_penalties.rb
@@ -139,12 +139,16 @@ module SmartAnswer::Calculators
     def interest
       return 0 if overdue_payment_days <= 0
 
-      days_with_penalty_interest = payment_deadline..(payment_date - 2)
+      days_with_penalty_interest = payment_deadline..interest_accrual_start_date_for_year
       interest_charges_per_day = days_with_penalty_interest.map do |date|
         calculate_interest_for_date(date)
       end
 
       SmartAnswer::Money.new(interest_charges_per_day.sum.round(2))
+    end
+
+    def interest_accrual_start_date_for_year
+      tax_year == "2019-20" ? payment_date - 1.day : payment_date - 2.days
     end
 
     def total_owed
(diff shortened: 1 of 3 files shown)


lib/smart_answer/calculators/self_assessment_penalties.rb
```diff
@@ -139,14 +139,18 @@ def late_filing_penalty
     def interest
       return 0 if overdue_payment_days <= 0
 
-      days_with_penalty_interest = payment_deadline..(payment_date - 2)
+      days_with_penalty_interest = payment_deadline..(payment_date_for_tax_year)
```
brucebolt, line 144: It took me quite a while to understand what was happening here with the -1 and -2.  Maybe we could define a new constant for `interest_accrual_start_date` (like we have `payment_deadline`) and use that instead of doing the -1 and -2?

brucebolt (commented):
Looks good.  I've added a comment about the interest accrual changes that might make this easier to work with when subsequent year changes are added,

lib/smart_answer/calculators/self_assessment_penalties.rb
brucebolt: Also, thought of a very slight improvement that could be made here: there's a `prev_day` method on `Date` which might be clearer than subtracting 1?

1pretz1: Yeah I didn't do a great job of making this code readable so thanks for highlighting it!

I can't use a constant as the variable we need (`payment_date`) is not available at the class level. I've restructured the method to be a bit more readable including your much better naming. I've gone with `day`/`days` terminology instead of `prev_day` due to the fact we need to state the day before yesterday too. Unless there is some magic `day_before_yesterday` method too.

Let me know if that's any clearer?

1pretz1:
Thanks for the review @brucebolt.

I've also bolted on another commit which should make a few tests a bit more readable.

lib/smart_answer/calculators/self_assessment_penalties.rb
```diff
@@ -139,14 +139,18 @@ def late_filing_penalty
...
+      days_with_penalty_interest = payment_deadline..interest_accrual_start_date_for_year
       interest_charges_per_day = days_with_penalty_interest.map do |date|
         calculate_interest_for_date(date)
       end
 
       SmartAnswer::Money.new(interest_charges_per_day.sum.round(2))
     end
 
+    def interest_accrual_start_date_for_year
+      tax_year == "2019-20" ? payment_date - 1.day : payment_date - 2.days
```
brucebolt, line 153: Thanks!  This is so much clearer.

brucebolt (approved):
Thanks for making the day subtraction clearer.

All looks good to me :+1:

PR #5151, opened 2021-02-04
Add RSpec feature test helpers

gclssvglx:
What

Add RSpec feature test helper methods

Why

We want to minimise the amount of boilerplate code in our feature tests

[Trello](https://trello.com/c/cTl73Hnf/763-add-rspec-helper-methods)

### DIFF
--- a/spec/features/smart_answer/am_i_getting_minimum_wage_spec.rb
+++ b/spec/features/smart_answer/am_i_getting_minimum_wage_spec.rb
@@ -43,7 +43,7 @@ RSpec.feature "SmartAnswer::AmIGettingMinimumWageFlow", type: :feature do
       visit "/am-i-getting-minimum-wage"
       expect(page).to have_selector("h1", text: headings.fetch(:flow_title))
 
-      click_govuk_start_button
+      click_link "Start now"
       expect(page).to have_selector("h1", text: headings.fetch(:what_would_you_like_to_check))
 
       choose "If you're getting the National Minimum Wage or the National Living Wage"
@@ -87,7 +87,7 @@ RSpec.feature "SmartAnswer::AmIGettingMinimumWageFlow", type: :feature do
       visit "/am-i-getting-minimum-wage"
       expect(page).to have_selector("h1", text: headings.fetch(:flow_title))
 
-      click_govuk_start_button
+      click_link "Start now"
       expect(page).to have_selector("h1", text: headings.fetch(:what_would_you_like_to_check))
 
       choose "If you're getting the National Minimum Wage or the National Living Wage"
@@ -139,7 +139,7 @@ RSpec.feature "SmartAnswer::AmIGettingMinimumWageFlow", type: :feature do
       visit "/am-i-getting-minimum-wage"
       expect(page).to have_selector("h1", text: headings.fetch(:flow_title))
 
-      click_govuk_start_button
+      click_link "Start now"
       expect(page).to have_selector("h1", text: headings.fetch(:what_would_you_like_to_check))
 
       choose "If you're getting the National Minimum Wage or the National Living Wage"
@@ -187,7 +187,7 @@ RSpec.feature "SmartAnswer::AmIGettingMinimumWageFlow", type: :feature do
       visit "/am-i-getting-minimum-wage"
       expect(page).to have_selector("h1", text: headings.fetch(:flow_title))
 
-      click_govuk_start_button
+      click_link "Start now"
       expect(page).to have_selector("h1", text: headings.fetch(:what_would_you_like_to_check))
 
       choose "If you're getting the National Minimum Wage or the National Living Wage"
@@ -224,7 +224,7 @@ RSpec.feature "SmartAnswer::AmIGettingMinimumWageFlow", type: :feature do
       visit "/am-i-getting-minimum-wage"
       expect(page).to have_selector("h1", text: headings.fetch(:flow_title))
 
-      click_govuk_start_button
+      click_link "Start now"
       expect(page).to have_selector("h1", text: headings.fetch(:what_would_you_like_to_check))
 
       choose "If an employer owes you payments from last year (April 2019 to March 2020)"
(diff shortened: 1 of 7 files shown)


spec/support/methods/feature_test_helper_methods.rb
```diff
@@ -0,0 +1,45 @@
...
+      raise "For the `date` type, `with` should be a hash" unless with.is_a?(Hash)
+
+      fill_in "response[day]", with: with[:day]
+      fill_in "response[month]", with: with[:month]
+      fill_in "response[year]", with: with[:year]
+    end
+    click_button "Continue"
+  end
+
+  def ensure_page_has(header: nil, subheader: nil, subheaders: [], text: nil)
```
theseanything, line 33: So I'm wondering if we need to the subheaders? Its only really the find-support flow that has them I think?

But if so you might be able to roll subheaders and subheaders in to a single argument - the expectation is somewhat the same.
```suggestion
  def ensure_page_has(header: nil, subheaders: [], text: nil)
```

spec/support/methods/feature_test_helper_methods.rb
```diff
@@ -0,0 +1,45 @@
...
+    when :check_boxes
+      raise "For the `check_boxes` type, `with` should be an array" unless with.is_a?(Array)
+
+      with.each do |answer|
+        check answer
+      end
+    when :input_text
+      fill_in "response", with: with
+    when :date
+      raise "For the `date` type, `with` should be a hash" unless with.is_a?(Hash)
```
theseanything, line 24: I wonder if being about to pass in a date string more convenient than a hash? Whilst a has is more explicit, feels like a lot more typing.

i.e. `{day: 31, month: 12, year: 2020}` => `"31-12-2020"`

etc... `date = Date.parse("31-12-2020"); date.month;`

spec/support/methods/feature_test_helper_methods.rb
```diff
@@ -0,0 +1,45 @@
...
+
+      fill_in "response[day]", with: with[:day]
+      fill_in "response[month]", with: with[:month]
+      fill_in "response[year]", with: with[:year]
+    end
```
theseanything, line 29: Might be helpful to have a catch all (`else`) clause on the switch that raises an error in case we input an invalid `type_of` by mistake. It was somewhat difficult to realise when I'd done this, because the errors in capybara are pretty confusing.

spec/support/methods/feature_test_helper_methods.rb
```diff
@@ -0,0 +1,45 @@
+module FeatureTestHelperMethods
+  def start(the_flow:, at:)
+    visit "/#{at}"
+    ensure_page_has(header: the_flow)
+    click_govuk_start_button
```
theseanything, line 5: I think we should get rid of the method `click_govuk_start_button`, because I don't think we will use it anywhere else and its a lot of overhead (separate helper file and module) just for this simple method
```suggestion
    click_link "Start now"
```

spec/support/methods/feature_test_helper_methods.rb
```diff
@@ -0,0 +1,45 @@
...
+      choose with
+    when :check_box
+      check with
+    when :check_boxes
+      raise "For the `check_boxes` type, `with` should be an array" unless with.is_a?(Array)
+
+      with.each do |answer|
+        check answer
+      end
+    when :input_text
```
theseanything, line 21: We should probably make these consistent with what they're called in smart answer so it's easier to remember what the question types are.
```suggestion
    when :value
```

spec/support/methods/feature_test_helper_methods.rb
```diff
@@ -0,0 +1,45 @@
...
+  def start(the_flow:, at:)
+    visit "/#{at}"
+    ensure_page_has(header: the_flow)
+    click_govuk_start_button
+  end
+
+  def answer(question:, of_type:, with:)
+    ensure_page_has(header: question)
+    case of_type
+    when :radio_button
```
theseanything, line 11:
```suggestion
    when :radio
```

spec/support/methods/feature_test_helper_methods.rb
```diff
@@ -0,0 +1,45 @@
...
+      choose with
+    when :check_box
+      check with
+    when :check_boxes
+      raise "For the `check_boxes` type, `with` should be an array" unless with.is_a?(Array)
+
+      with.each do |answer|
+        check answer
+      end
```
theseanything, line 20:
```suggestion
    when :checkbox
      raise "For the `checkbox` type, `with` should be an array" unless with.is_a?(Array)

      with.each do |answer|
        check answer
      end
```

spec/support/methods/feature_test_helper_methods.rb
gclssvglx: Yeah, I agree - will add this in.

spec/support/methods/feature_test_helper_methods.rb
gclssvglx: Yup, agree - will remove and update the `start` method.

spec/support/methods/feature_test_helper_methods.rb
gclssvglx: Ok - I thought you'd prefer them to be consistent with the flow.  I was thinking we should be consistent with the HTML naming convention.  I'm happy to stick with the flow naming.  Will update.

spec/support/methods/feature_test_helper_methods.rb
gclssvglx: As https://github.com/alphagov/smart-answers/pull/5151#discussion_r571971758

spec/support/methods/feature_test_helper_methods.rb
gclssvglx: As https://github.com/alphagov/smart-answers/pull/5151#discussion_r571971758

spec/support/methods/feature_test_helper_methods.rb
gclssvglx: I'd prefer to keep `subheaders` as they are used in the find-support flow and testing the `h1` there doesn't make sense or help much as it's always the same!  However, I've removed `subheader` and updated the code so that it will convert to Array - if a String is given.  This way at least there is only one optional param for `subheaders` now.

spec/support/methods/feature_test_helper_methods.rb
gclssvglx: Yup, makes sense and different date formats and validation better.  Good spot - thanks.

theseanything (approved):

PR #5166, opened 2021-02-10
Change language around CNI for Mexico

richardskipper:
https://trello.com/c/JRffxT41/3308-getting-married-abroad-mexico-amend-the-language-for-cni

Change to wording around CNIs, to make it clear that not all local register offices will require them before you get married.


callumknights (approved):

brucebolt (commented):
Should the apostrophes be typographic style (i.e. `’` instead of `'`)?  I'm not sure if it matters for this app or not?

callumknights:
> Should the apostrophes be typographic style (i.e. `’` instead of `'`)? I'm not sure if it matters for this app or not?

Having just spoke to Richard and double checked myself the deployment doesn't show any difference compared to what is currently live, so I would assume it doesn't matter for this app.

brucebolt:
> > Should the apostrophes be typographic style (i.e. `’` instead of `'`)? I'm not sure if it matters for this app or not?
> 
> Having just spoke to Richard and double checked myself the deployment doesn't show any difference compared to what is currently live, so I would assume it doesn't matter for this app.

Brilliant, that means the Govspeak parsing must be doing something here.

I'll merge this now.

PR #5173, opened 2021-02-11
Reset the GitHub Actions cache for bundler

theseanything:
We seem to have got some invalid state in our cache, and there is no way
to manually clear the cache.  We can only wait for it to expire (takes 7
days), or change the key.

Trying to build this commit with the current cache gives:

    LoadError: libffi.so.6: cannot open shared object file: No such file
    or directory -
    /home/runner/work/govuk-account-manager-prototype/govuk-account-manager-prototype/tmp/bundle/ruby/2.7.0/gems/ffi-1.14.2/lib/ffi_c.so

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,7 +24,7 @@ jobs:
         uses: actions/cache@v1
         with:
           path: vendor/bundle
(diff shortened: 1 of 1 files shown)


gclssvglx (approved):
LGTM

PR #5178, opened 2021-02-12
Reset the GitHub Actions cache for bundler

theseanything:
Previously tried to reset the cache for GitHub Actions. However, missed the restore key #5173

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,11 +21,11 @@ jobs:
         uses: ruby/setup-ruby@v1
 
       - name: Retrieve cached Gems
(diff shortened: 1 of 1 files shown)


gclssvglx (approved):
LGTM

kevindew:
@theseanything I'm not sure if it would have helped avoid your cache problem but it's probably better to switch from using "actions/cache@v2" to instead passing `with: { bunder_cache: true }` to "ruby/setup-ruby". It has it's own bundler cache that is preferred to be used instead of "actions/cache": https://github.com/ruby/setup-ruby#caching-bundle-install-automatically

PR #5198, opened 2021-02-22
Extend HMRC C-19 late payment penalties to 1st Apr

gclssvglx:
## What

Currently, the trigger date for self assessment late payment penalties
for 2019-20 is 2 March. Because of coronavirus, HMRC have decided to
extend it to 1 April. This means we need to change the calculation
formula, so that users inputting a date between 2 March and 1 April
don't have a late payment penalty listed on their outcome.

## Why

If we don't do this there's a risk that some users will think they need
to pay a late penalty payment when they don't. The late payment penalties
are quite a lot of money, so this could cause unnecessary stress for
users at a time when many may already be having financial problems. There's
also a reputational risk to HMRC and GOV.UK, and a risk to our relationship
with HMRC.

[Trello](https://trello.com/c/qcuggKKf/2072-content-change-in-self-assessment-penalty-calculator)

### DIFF
--- a/lib/smart_answer/calculators/self_assessment_penalties.rb
+++ b/lib/smart_answer/calculators/self_assessment_penalties.rb
@@ -153,6 +153,10 @@ module SmartAnswer::Calculators
       tax_year == "2019-20" ? payment_date - 1.day : payment_date - 2.days
     end
 
+    def first_late_payment_days
+      tax_year == "2019-20" ? 60 : 30
+    end
+
     def total_owed
       SmartAnswer::Money.new((estimated_bill.value + interest.to_f + late_payment_penalty.to_f).floor)
     end
@@ -162,7 +166,7 @@ module SmartAnswer::Calculators
     end
 
     def late_payment_penalty
-      if overdue_payment_days <= 30
+      if overdue_payment_days <= first_late_payment_days
         0
       elsif overdue_payment_days <= 181
         SmartAnswer::Money.new(late_payment_penalty_part.round(2))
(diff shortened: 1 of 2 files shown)


lib/smart_answer/calculators/self_assessment_penalties.rb
```diff
@@ -4,6 +4,7 @@ module SmartAnswer::Calculators
   class SelfAssessmentPenalties < OpenStruct
     ONLINE_FILING_DEADLINE_YEAR = SmartAnswer::YearRange.resetting_on("31 January").freeze
     ONLINE_FILING_DEADLINE_YEAR_FEB = SmartAnswer::YearRange.resetting_on("28 February").freeze
+    HMRC_C19_EXTENSION_PAYMENT_DEADLINE_YEAR = SmartAnswer::YearRange.resetting_on("1 March").freeze
```
1pretz1, line 7: I _think_ this should be the 31st of March, rather than the 1st, as interest payments now kick in on the 1st of April.

1pretz1: Bearing in mind that these dates denote the deadline, rather than the date the payment interest kicks in.

1pretz1:
Logic looks good @gclssvglx. Just a date change that is needed.

lib/smart_answer/calculators/self_assessment_penalties.rb
gclssvglx: 👍  Thanks @1pretz1 - good spot - fixed

1pretz1 (approved):

PR #5200, opened 2021-02-23
Simplify layout templates

theseanything:
This simplifies the view templates by removing an extra layout "smart_answer" layout which should be rendered by all pages.

This also moves partials into a "shared" folder to make it easier to see which templates are partials or full views. 


alex-ju (approved):
This looks like a good clean-up! As a sidenote, the escape-link used to be an actual component, with documentation and all, visible via the component guide, but that probably got lost while ported here – will raise a separate PR do extract it from here. The rest of them seem to be plain partials, great to see them grouped in a folder.

PR #5203, opened 2021-02-24
Simplify the rateable business property question

Rosa-Fox:
[Trello](https://trello.com/c/asQsJdKh/182-simplify-the-rateable-business-property-question)

- Simplifies the `What is the rateable value of your business' non-domestic property?` question
to `Does your business have any rateable non-domestic property?`.

- Updates the radio button options to be: "Yes", "No", "Not sure".

- The reason for this is that this question is used to display the `Business rates holiday for retail, hospitality and leisure` and the `Support for nursery businesses that pay business rates` schemes.

- These schemes don't require knowing specifically what the property value is. We previously had a scheme that did, but that has been removed.

- We also found that some users are not familiar with 'rateable non-domestic property value'. Therefore, adding a "not sure" option aims to help them progress and reduce drop off.

### Before

smart-answers-before

### After
smart-answers-after

### DIFF
--- a/lib/smart_answer/calculators/business_coronavirus_support_finder_calculator.rb
+++ b/lib/smart_answer/calculators/business_coronavirus_support_finder_calculator.rb
@@ -26,12 +26,12 @@ module SmartAnswer::Calculators
       },
       retail_hospitality_leisure_business_rates: lambda { |calculator|
         calculator.business_based == "england" &&
-          calculator.non_domestic_property != "none" &&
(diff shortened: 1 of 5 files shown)


kevindew (approved):
Code looks good to me 👍  - I imagine it's worth getting a product review to check it.

kevindew:
@DGarland5 has confirmed this looks good so I'll merge it.

PR #5205, opened 2021-02-25
Accessibility fix for hide this page

DilwoarH:
## What

The leave this site button does not properly explain what it does, so it could be more accessible for people using screen readers

We need to design the content for what this says, so that we can make sure people get the right idea of what is happening. 

See page 15 of the report - https://drive.google.com/drive/folders/13sklGDeiQv95CFXFglXZ_ka-NpOK5cGL

Also look at the discussion on the component in the design system

## Why

So people with accessibility needs can use the button 

## Standard used

https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA8.html

## Demo

https://smart-answer-accessibil-jeb3no.herokuapp.com/find-coronavirus-support/s/need-help-with

## Tested with Assistive Labs

Reads out:
```
Quick exit button, it opens BBC Weather in a new tab  link
```

## Dev work - Content change

Button content: 

Change from:
Leave this site

To:
Hide this page

Add hidden screen reader text:
Quick exit button, it opens BBC Weather in a new tab

Change link to: 
BBC Weather - https://www.bbc.co.uk/weather

 https://trello.com/c/FGFqviB4/180-accessibility-smart-answer-fix-the-leave-this-site-link-purpose

### DIFF
--- a/app/views/smart_answers/shared/_escape-link.html.erb
+++ b/app/views/smart_answers/shared/_escape-link.html.erb
@@ -2,4 +2,13 @@
   data_attributes ||= {}
   data_attributes[:module] = 'app-escape-link'
 %>
-<%= link_to "Leave this site", destroy_session_flow_path(ext_r: "true"), class: "app-c-escape-link govuk-link", rel: "nofollow noreferrer noopener", target: "_blank", data: data_attributes %>
+<%= link_to("Hide this page",
+  destroy_session_flow_path(ext_r: "true"),
+  class: "app-c-escape-link govuk-link",
+  rel: "nofollow noreferrer noopener",
+  target: "_blank",
+  data: data_attributes,
+  aria: {
+    label: "Hide this page. Quick exit button, it opens BBC Weather in a new tab"
+  }
+) %>
--- a/app/controllers/session_answers_controller.rb
+++ b/app/controllers/session_answers_controller.rb
@@ -26,7 +26,7 @@ class SessionAnswersController < ApplicationController
     session_store.clear
 
     if params[:ext_r] == "true"
-      redirect_to "https://bbc.co.uk/news"
+      redirect_to "https://www.bbc.co.uk/weather"
     else
       redirect_to "/#{params[:id]}"
     end
--- a/test/controllers/session_answers_controller_test.rb
+++ b/test/controllers/session_answers_controller_test.rb
@@ -97,7 +97,7 @@ class SessionAnswersControllerTest < ActionDispatch::IntegrationTest
 
     should "redirect to external sitewhen the ext_r option is present and true" do
       get destroy_session_flow_path(id: flow_name, ext_r: "true")
-      assert_redirected_to "https://bbc.co.uk/news"
+      assert_redirected_to "https://www.bbc.co.uk/weather"
     end
 
     should "remove the session data for the flow when escaping" do


DilwoarH:
Demo: https://smart-answer-accessibil-jeb3no.herokuapp.com/find-coronavirus-support/s/need-help-with

DilwoarH:
Hi @maxgds @injms @owenatgov,

I added you to this ticket as you all are probably best placed to look at my solution.

I used the following specification as a guideline on how to implement:
https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA8.html

This is a demo link:
https://smart-answer-accessibil-jeb3no.herokuapp.com/find-coronavirus-support/s/need-help-with

app/views/smart_answers/shared/_escape-link.html.erb
```diff
@@ -2,4 +2,13 @@
   data_attributes ||= {}
   data_attributes[:module] = 'app-escape-link'
 %>
-<%= link_to "Leave this site", destroy_session_flow_path(ext_r: "true"), class: "app-c-escape-link govuk-link", rel: "nofollow noreferrer noopener", target: "_blank", data: data_attributes %>
+<%= link_to("Hide this page",
+  destroy_session_flow_path(ext_r: "true"),
+  class: "app-c-escape-link govuk-link",
+  rel: "nofollow noreferrer noopener",
+  target: "_blank",
```
kevindew, line 9: Sorry I know you didn't introduce this. I'm wondering why this is set as `target="_blank"` as, for users without js, whereas I'd have thought we want to replace the current page.

app/views/smart_answers/shared/_escape-link.html.erb
```diff
@@ -2,4 +2,13 @@
...
+  data: data_attributes,
+  aria: {
+    label: "Quick exit button, it opens BBC Weather in a new tab"
+  }
```
kevindew, line 13: This looks much easier to read in a multiline approach 👍

kevindew (commented):
Code looks good to me, I've got a question about part of the original implementation which may be better fielded towards @gclssvglx

app/views/smart_answers/shared/_escape-link.html.erb
DilwoarH: I'm not sure what the reasoning behind it is but the behaviour is that it opens a new tab with the BBC weather and also replaces the current page with BBC weather.

One thing it doesn't do is rewrite the history, not sure if this was intentional.

kevindew: Sure, you've described there what the Javascript does. However this `target="_blank"` attribute doesn't seem to have a role in the JS version (since we [preventDefault](https://github.com/alphagov/smart-answers/blob/9e7d2081513f8e5f1d711b84f1d993e34e8380c4/app/assets/javascripts/components/escape-link.js#L15)).

Whereas without JS this tab is left open.

gclssvglx: I believe this is why `target="_blank"` was used...

> We have a 'leave this site' button on the triage tool currently. We should add this functionality, exactly how it works now, in the smart answer prototype.  [Trello](https://trello.com/c/9cxZlrUr/221-add-leave-this-site-button-to-triage-tool-smart-answer-prototype)

In light of the different structure between the original tool and smart answers - this approach and the decision to use `_blank` were queried at the time.

gclssvglx: Sorry, I know this doesn't help much!

hannahjump:
Thanks Dilwoar, this looks good from a content perspective. 
Just checking, will a screen reader read out the button text too, so `Hide this page, Quick exit button, it opens BBC Weather in a new tab`

app/views/smart_answers/shared/_escape-link.html.erb
kevindew: Thanks @gclssvglx, is the repo that this was copied from still around so we can check the commits here?

At the moment I'm assuming this "target=_blank" is probably a mistake as it'd still leave this on your screen for a desktop computer, however I'm not sure if someone made that as a conscious decision that it was better to do that than have a working back button.

DilwoarH:
> Thanks Dilwoar, this looks good from a content perspective.
> Just checking, will a screen reader read out the button text too, so `Hide this page, Quick exit button, it opens BBC Weather in a new tab`

No just `Quick exit button, it opens BBC Weather in a new tab link`

hannahjump:
> > Thanks Dilwoar, this looks good from a content perspective.
> > Just checking, will a screen reader read out the button text too, so `Hide this page, Quick exit button, it opens BBC Weather in a new tab`
> 
> No just `Quick exit button, it opens BBC Weather in a new tab link`
> 
> Did you want `Hide this page` too?

Yes please, can it say `Hide this page` too

app/views/smart_answers/shared/_escape-link.html.erb
gclssvglx: Hi @kevindew - my task on [this PR](https://github.com/alphagov/smart-answers/pull/4714) was to implement the same escape button functionality from [the find support service](https://github.com/alphagov/govuk-coronavirus-find-support) [as it stood at the time](https://github.com/alphagov/govuk-coronavirus-find-support/blob/c5d0260ea390511a17d3fe93c309e2e5a16438bb/app/views/components/_escape-link.html.erb#L5).  So, I do not believe it is my place to justify the original use of `target="_blank"`.

The original functionality was introduced [with this commit](https://github.com/alphagov/govuk-coronavirus-find-support/pull/258/commits/198f87ff1cbb07615e557e4e25c0f7620351510e) from [this PR](https://github.com/alphagov/govuk-coronavirus-find-support/pull/258).  This [response from the original author](https://github.com/alphagov/govuk-coronavirus-find-support/pull/258#issuecomment-637656553) may help to answer your concern/question.

To dispel your suggestion that I have implemented the [required functionality](https://trello.com/c/9cxZlrUr/367-add-leave-this-site-button-to-triage-tool-smart-answer-prototype) incorrectly, here is the [original code](https://github.com/alphagov/govuk-coronavirus-find-support/blob/c5d0260ea390511a17d3fe93c309e2e5a16438bb/app/views/components/_escape-link.html.erb#L5)...

```
<%= link_to text, href, class: "app-c-escape-link govuk-link", rel: "nofollow noreferrer noopener", target: "_blank", data: data_attributes %>
```

...and [my re-implementation.](https://github.com/alphagov/smart-answers/blob/627646c6346f864088ecf5e52e37da3cc0bf48e0/app/views/smart_answers/_escape-link.html.erb#L5).

```
<%= link_to "Leave this site", destroy_session_flow_path(ext_r: "true"), class: "app-c-escape-link govuk-link", rel: "nofollow noreferrer noopener", target: "_blank", data: data_attributes %>
```

kevindew: Thanks for those links @gclssvglx. Apologies, I wasn't suggesting you'd made the mistake in your copying more that a mistake may have occurred in the original implementation.

It would have been useful if your port commit messages had linked to where it was ported from, given the challenges we're having in establishing a "why".

gclssvglx: Yes, you're right @kevindew.  In hind-sight - It would have been useful to include links to original code.

kevindew: I've chatted to @alex-ju about it now. 

He indicated that the reason we have the target=_blank is to avoid any back history. The needs for users on mobile devices (where other tabs are not visible) was placed higher than the risks of someone on a machine where the tab would still be visible on screen. Also turns out it was documented here: https://github.com/alphagov/govuk-coronavirus-find-support/blob/c5d0260ea390511a17d3fe93c309e2e5a16438bb/app/views/components/docs/escape-link.yml#L4 (any reason we didn't port this as a component?)

@DilwoarH would you be able to update the commit that changes this to include the why's of this component so we have it documented in our commit history so we don't have any future confusion if we change this?

DilwoarH:
> > > Thanks Dilwoar, this looks good from a content perspective.
> > > Just checking, will a screen reader read out the button text too, so `Hide this page, Quick exit button, it opens BBC Weather in a new tab`
> > 
> > 
> > No just `Quick exit button, it opens BBC Weather in a new tab link`
> > Did you want `Hide this page` too?
> 
> Yes please, can it say `Hide this page` too

Hi @hannahjump  - it now says `Hide this page. Quick exit button, it opens BBC Weather in a new tab`

injms (approved):
This is a good improvement - I think it makes it clearer to screen reader users what will happen when they select the link.

I'd be interested in whether we're considering the feedback to implement this as a navigation region? This is placed just before the feedback component near the end of the document, so many assistive tech users might not ever be aware that it's there.

Regardless of the components place in the page, this is a definite improvement that should go live.

hannahjump:
> This is a good improvement - I think it makes it clearer to screen reader users what will happen when they select the link.
> 
> I'd be interested in whether we're considering the feedback to implement this as a navigation region? This is placed just before the feedback component near the end of the document, so many assistive tech users might not ever be aware that it's there.
> 
> Regardless of the components place in the page, this is a definite improvement that should go live.

@injms We have another card for navigation region :D ! agree about people being aware it's there.

PR #5211, opened 2021-03-01
Add merge rails_helper into spec_helper

theseanything:
This removes the individual `require rails_helper` for each RSpec test by merging `rails_helper.rb` into the `spec_helper.rb` that is automatically included.


kevindew:
Is there any chance we can get rid of the need for a rails helper and a spec helper? as per: https://docs.publishing.service.gov.uk/manual/conventions-for-rails-applications.html#testing-utilities ?

It always seems to end up being a pain having information in both when you need both nearly all the time.

theseanything:
@kevindew good point! Done and removed a lot of the generic comments, that we can find in RSpec documentation anyway.

spec/spec_helper.rb
```diff
@@ -1,94 +1,71 @@
...
-  #
-  #   # Seed global randomization in this process using the `--seed` CLI option.
-  #   # Setting this allows you to use `--seed` to deterministically reproduce
-  #   # test failures related to randomization by passing the same `--seed` value
-  #   # as the one that triggered the failure.
   Kernel.srand config.seed
 end
+
+# Without this GovukContentSchemaTestHelpers configuration, get `NoMethodError: undefined method `<<' for nil:NilClass`
+require "slimmer/test"
```
kevindew, line 65: Any chance we could move this line to above `RSpec.configure` it feels surprisingly to be after it.

spec/spec_helper.rb
```diff
@@ -1,94 +1,71 @@
...
+require "govuk-content-schema-test-helpers/test_unit"
+
+GovukContentSchemaTestHelpers.configure do |config|
+  config.schema_type = "publisher_v2"
+  config.project_root = Rails.root
+end
```
kevindew, line 71: I'm pretty sure this isn't actually used in rspec. I tried removing it locally and all the tests still passed.

kevindew (approved):
Thanks Sean, this is great. I've just put a couple of quick minor comments.

kevindew:
That was meant to be a comment rather than approve, but since they're minor changes it doesn't matter if you merge it without a re-review.

PR #5213, opened 2021-03-02
Rename SessionAnswersController to FlowController

theseanything:
The naming "SessionAnswers" is somewhat confusing as it's used uniquely here and isn't defined anywhere. It originally was named this to represent a SmartAnswer that used session for handling user responses. However this controller will in future not only support sessions, but also query string. This generalises the name and uses terminology we have defined.

[Trello card](https://trello.com/c/42XuYGir)


alex-ju (approved):
🎛️

PR #5221, opened 2021-03-03
Set the unpublish tasks to discard drafts

kevindew:
Trello: https://trello.com/c/kjtIlrE8/212-bs-smart-answer-set-up-a-prefix-redirect

By default the Publishing API will refuse to update the unpublishing of
content which has a draft. This can be problematic if we've temporarily
taken a Smart Answer offline and marked it as a draft.

This can be resolved by telling the Publishing API that we don't mind
them discarding the draft. This feature is normally in place so editors
draft work isn't lost, however since the draft is all part of this repo
we don't need to precious about it.

### DIFF
--- a/lib/tasks/publishing_api.rake
+++ b/lib/tasks/publishing_api.rake
@@ -33,6 +33,7 @@ namespace :publishing_api do
       args.content_id,
       type: "redirect",
       redirects: [redirect],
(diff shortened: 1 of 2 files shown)


leenagupte (approved):
This makes sense 👍

PR #5231, opened 2021-03-09
Add ability to specify custom view template

theseanything:
This allow the ability to specify a name of an alternative view template to render than the default for an outcome node. This allows the creation of custom results pages beyond the standard render govspeak body. This allow Smart Answer to be used for future pages such as Start a business, that have more dynamic and custom styled results pages.

This is an unnecessary methods as it is only called once within the controller itself. We can remove the logic for rendering the start page, as this is never executed, as start pages always rendered by SmartAnswerController.


alex-ju (approved):
Nice one!

PR #5233, opened 2021-03-09
Revert "D garland5 patch 4"

kevindew:
Reverts alphagov/smart-answers#5232

This isn't ready to go live yet and still needs a content 2i


leenagupte (approved):
The results page still seems to load so: 👍

PR #5240, opened 2021-03-18
Add ability to put Heroku preview app behind basic auth

thomasleese:
The username and password are set using environment variables which are configured through the Heroku UI.

This has been lifted from [Collections](https://github.com/alphagov/collections/blob/master/app/controllers/application_controller.rb#L11-L16) and the env variables have been added to smart-answers-preview.herokuapp.com already.

This is to prevent users from accidentally landing on a page in the review app and believing that it's GOV.UK.

[Trello Card](https://trello.com/c/65Fp8KmU/2425-3-set-username-password-for-heroku-previews-of-smart-answers)


brucebolt (approved):

alex-ju (approved):
Thanks for raising this, Tom! I added it recently to my to-do list. I think/hope this is the only frontend app that wasn't behind HTTP auth 🤞

PR #5243, opened 2021-03-19
Fix date range used to calculate parental pay leave

brucebolt:
The date range used by this calculator is incorrect.  We currently use tax years (6th April to 5th April), but should instead work from the first Sunday in April to the day before the Sunday in the next April.

This updates to use those dates and removes the hardcoded dates from the outcomes.

Trello card: https://trello.com/c/TdxjQXlz

### DIFF
--- a/lib/smart_answer/calculators/pay_leave_for_parents_calculator.rb
+++ b/lib/smart_answer/calculators/pay_leave_for_parents_calculator.rb
@@ -126,8 +126,8 @@ module SmartAnswer::Calculators
       earnings_employment == "yes" && work_employment == "yes"
     end
 
-    def paid_leave_is_in_tax_year?(year)
-      (paid_leave_period & SmartAnswer::YearRange.tax_year.starting_in(year)).number_of_days.positive?
+    def paid_leave_is_in_year?(year)
+      (paid_leave_period & SmartAnswer::YearRange.new(begins_on: first_day_in_year(year))).number_of_days.positive?
     end
 
     def paid_leave_period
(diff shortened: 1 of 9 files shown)


thomasleese (approved):
I do wonder if there's a way we can store all those pieces of text in a single table which gets updated and have something like:

```erb
<% (2013..2020).each do |year| %>
  <% if calculator.paid_leave_is_in_year?(year) %>
    Weekly rate (between <%= calculator.formatted_first_day_in_year(year) %> and <%= calculator.formatted_last_day_in_year(year) %>) | £1<%= calculator.formatted_lower_weekly_rate(year) %>, or 90% of their average weekly earnings (whichever is lower).
  <% end %>
<% end %>
```

But I think this is perhaps a change beyond this PR.

brucebolt:
> I do wonder if there's a way we can store all those pieces of text in a single table which gets updated and have something like:
> 
> ```
> <% (2013..2020).each do |year| %>
>   <% if calculator.paid_leave_is_in_year?(year) %>
>     Weekly rate (between <%= calculator.formatted_first_day_in_year(year) %> and <%= calculator.formatted_last_day_in_year(year) %>) | £1<%= calculator.formatted_lower_weekly_rate(year) %>, or 90% of their average weekly earnings (whichever is lower).
>   <% end %>
> <% end %>
> ```
> 
> But I think this is perhaps a change beyond this PR.

Yeah, I did start moving all this into YAML files, but it then felt like I'd be rewriting much of the calculator, so decided it would be out of scope for this work.

PR #5245, opened 2021-03-19
Add `/flow/` prefix routes for FlowController

theseanything:
This adds new routes using the `/flow/` segment prefix instead of `/s/` for the FlowController. The existing `/s/` prefix routes are still supported and continue to be the default for flow navigation.

It was agreed that "flow" is more descriptive than "s" as the segment name. "s" was originally thought to mean "session" when session based flows were introduced, however we'd like to generalise the FlowController to also support storing user responses in the query string.

This registers the new routes in the content item, so GOV.UK's router handles both sets of URLs whilst we transition to use `/flow/` prefix exclusively.

These changes are prerequisite work needed by the [agreed implementation plan](https://github.com/alphagov/smart-answers/blob/master/docs/arch/002-store-responses-in-query-string.md#suggested-steps-for-implementation) to migrate flows to use query parameters to store user responses. 

### DIFF
--- a/app/controllers/flow_controller.rb
+++ b/app/controllers/flow_controller.rb
@@ -1,5 +1,6 @@
 class FlowController < ApplicationController
   before_action :set_cache_headers
+  before_action :redirect_path_based_flows
 
   def start
     session_store.clear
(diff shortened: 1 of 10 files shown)


app/controllers/flow_controller.rb
```diff
@@ -34,6 +35,10 @@ def destroy
 
 private
 
+  def redirect_path_based_flows
+    redirect_to smart_answer_path(params[:id], started: "y") unless flow.use_session?
+  end
```
kevindew, line 40: 👍   Seems reasonable, the commit explanation was helpful to understand why.

kevindew (approved):
Looks good to me 👍

PR #5246, opened 2021-03-19
Refactor `use_session` flow option to `response_store`

theseanything:
This changes the boolean `use_session` parameter in flow definitions to `response_store`. The new parameter now accepts a symbol to indicate which response store to use. This will allow the flag to be used to specify more than one response storage type. 

The change for session based flows is the following:
`use_session true` => `responses_store :session`

This is prerequisite work for enable flows to store user responses in query parameters.

### DIFF
--- a/app/controllers/flow_controller.rb
+++ b/app/controllers/flow_controller.rb
@@ -36,7 +36,7 @@ class FlowController < ApplicationController
 private
 
   def redirect_path_based_flows
-    redirect_to smart_answer_path(params[:id], started: "y") unless flow.use_session?
+    redirect_to smart_answer_path(params[:id], started: "y") if flow.response_store.nil?
   end
 
   def set_cache_headers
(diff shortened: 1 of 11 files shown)


kevindew (approved):
This looks good too.

My only concern is there is quite a lot of `== :session` scattered around - but I couldn't think of an immediate nicer approach. I wonder about using [inquiry](https://apidock.com/rails/String/inquiry) so you could do response_store.session? but that'd conflict with the symbol based definition.

theseanything:
> My only concern is there is quite a lot of `== :session` scattered around - but I couldn't think of an immediate nicer approach. I wonder about using [inquiry](https://apidock.com/rails/String/inquiry) so you could do response_store.session? but that'd conflict with the symbol based definition.

Ah yes! I might not worry for now, as I think most of it will go when we transition to using a single controller!

PR #5254, opened 2021-03-23
Spike: code in terms of outcomes, rather than answers

ChrisBAshton:
Probably easiest to view the commit: a5ab26089b2dd77815d35b80c8960c86a08a1661

Following the logic of nested conditionals requires keeping track of lots of information. It can be very difficult to give an answer to the question "how do I reach this outcome?". It can also be difficult to deduce whether multiple conditions can lead to the same outcome, and what those conditions might be.

Instead of calculating in terms of the user's answers, it is much simpler to treat each outcome in turn, and calculate whether the user's answers should qualify for that outcome. Advantages:

- Flattens the logic structure, so that we have one long if-else block, rather than a heavily nested (complex) structure.
- The conditions for each outcome are defined in one place, rather than multiple paths of code being able to lead to the same outcome.
- It's now much easier to explain how each outcome is arrived at.

I considered taking this a step further and storing each outcome and its conditional logic in an array of all possible outcomes in a prioritised list. This would take us in more of a data structure direction, rather than arbitrary code, and we could pass the array to an interpreter which steps through each outcome and returns early when it's found the first outcome that qualifies. This approach seemed a bit too much of a step away from how things are done in Smart Answers, so one to consider for the future rather than implement right now.


lib/smart_answer_flows/maternity-paternity-pay-leave.rb
```diff
@@ -138,43 +138,57 @@ def define
...
+                ) || (
+                  calculator.employment_status_of_mother == "self-employed" ||
+                  calculator.employment_status_of_partner == "unemployed"
+                )
+              )
+            outcome :outcome_birth_nothing
+          elsif calculator.mother_worked_at_least_26_weeks == "no" && !calculator.two_carers?
             outcome :outcome_single_birth_nothing
-          else
-            outcome :mother_earned_at_least_390
```
ChrisBAshton, line 177: I think we probably ought to have a catch-all at the end:

```ruby
else
  raise "No outcome for the supplied answers"
end
```

ChrisBAshton:
Closing, as not intending to merge as-is. Superseded by #5347.

PR #5264, opened 2021-03-26
Update wording for register-a-death for commonwealth countries

benjamineskola:
This moves the `COMMONWEALTH_COUNTRIES` list from the marriage-abroad code to a place where it can be shared, and then uses it for the register-a-death calculation too.

The marriage-abroad list is actually not a complete list, so I've created a complete list then removed the ones to which marriage-abroad does not apply: Australia, Gambia, Kenya, Maldives, Mozambique, Rwanda, South Africa, and Tanzania. Feasibly there are reasons for all of these to be special cases, but it seems to make sense to have the commonwealth-countries list actually reflect the real membership.

https://trello.com/c/KxsNUhk7/2442-register-a-death-smart-answer-update-wording-for-commonwealth-countries


ChrisBAshton (approved):
LGTM - nice one 👏

benjamineskola:
A further oddity is that the marriage-abroad code doesn't actually seem to use that list at all — rather than checking membership of a list and then conditionally displaying text, it simply loads a different file based on the country name. So I'm intending to follow this with another PR to remove the redundant code — but it didn't seem to make sense to remove that list first and then add it back for register-a-death.

ChrisBAshton (approved):

PR #5267, opened 2021-03-26
Add custom result page for start a business

alex-ju:
### What
Build a custom template to render the results page for [next steps for your business flow](https://smart-answer-add-custom-jyc8pf.herokuapp.com/next-steps-for-your-business) with a default set of items.
This includes adding a [hero section](https://smart-answer-add-custom-jyc8pf.herokuapp.com/component-guide/hero), a [result item](https://smart-answer-add-custom-jyc8pf.herokuapp.com/component-guide/result-item) and a [result section](https://smart-answer-add-custom-jyc8pf.herokuapp.com/component-guide/result-section) component.

### Why
To be able to show the results in a similar way with the version tested with users.

### Visual changes
<table>
<tr><th>Before</th><th>After</th></tr>
<tr><td valign="top">

before

</td><td valign="top">

after

</td></tr>
</table>

[Trello card](https://trello.com/c/j1As1BQY)


app/views/layouts/application.html.erb
```diff
@@ -29,12 +29,15 @@
...
   </head>
   <body class="govuk-template__body">
-    <div class="govuk-width-container smart_answer" id="wrapper">
-      <main id="content" role="main">
-        <%= yield %>
-      </main>
-      <%= yield :escape_link %>
-      <%= render "govuk_publishing_components/components/feedback" %>
+    <div id="wrapper">
+      <%= yield :hero %>
```
theseanything, line 33: Does the hero need to be inside the main div? The [example of MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/main) includes the header

app/views/smart_answers/custom_result.html.erb
```diff
@@ -0,0 +1,25 @@
...
+  <% if outcome.title.present? %><%= outcome.title %><% else %>Outcome<% end %>
+<% end %>
+<% content_for :head do %>
+  <meta name="robots" content="noindex">
+<% end %>
+
+<% content_for :hero do %>
+  <%= render "components/hero", {
+    title: outcome.title,
+    description: outcome.description,
```
theseanything, line 13: I think ultimately we want to move away from ERB templates and use locale files.  (ERB templates create a lot of overhead and I think were originally introduced to support long form govspeak content). 

Instead of adding a new field for the ERB template - we could just use `t("next_steps_for_your_business.hero.description")` which should work out of the box. There would need to be longer term thinking about how we properly organise content - but this would be good enough for now without adding anymore additional code to ERB templates.

config/smart_answers/next_steps_for_your_business.yml
```diff
@@ -0,0 +1,31 @@
+sections:
+  - title: "Business tax"
+    results:
+      - title: "You need to register for corporation tax within 3 months of starting to do business, if you haven’t done so already."
+        url: "/corporation-tax"
+        description: "Register for Corporation Tax"
```
theseanything, line 6: Nit: Wondering whether we could flatten the data structure, and add "section_name" as an attribute on a result? This data organisation seem tightly coupled to the view and might make hard to make changes in future. (Not something blocking this PR tho)

app/views/smart_answers/custom_result.html.erb
```diff
@@ -20,6 +20,5 @@
   <div class="govuk-grid-column-full">
     <%= outcome.body %>
     <%= render 'smart_answers/shared/debug' %>
-    <%= render 'smart_answers/shared/previous_answers', hide_previous_answers: @presenter.hide_previous_answers_on_results_page? %>
```
theseanything: Can we remove this from the earlier commit to keep the history simple?

app/helpers/index_helper.rb
```diff
@@ -12,7 +12,8 @@ def live_link(flow_name, status)
   end
 
   def code_links(flow_name)
-    tag.p(link_to("Definition", "https://www.github.com/alphagov/smart-answers/blob/master/lib/smart_answer_flows/#{flow_name}.rb")) +
-      tag.p(link_to("Content files", "https://www.github.com/alphagov/smart-answers/blob/master/lib/smart_answer_flows/#{flow_name}"))
+    link_to("Definition", "https://www.github.com/alphagov/smart-answers/blob/master/lib/smart_answer_flows/#{flow_name}.rb") +
+      tag.br +
+      link_to("Content files", "https://www.github.com/alphagov/smart-answers/blob/master/lib/smart_answer_flows/#{flow_name}")
```
theseanything, line 17: Thanks! Sorry you had to see this horrendous attempt at frontend!

theseanything (changes requested):

app/views/smart_answers/custom_result.html.erb
```diff
@@ -0,0 +1,25 @@
+<% # this custom result template is currently tested on the next-steps-for-your-business flow %>
+<% outcome = @presenter.current_node %>
+<% content_for :outcome_title do %>
+  <% if outcome.title.present? %><%= outcome.title %><% else %>Outcome<% end %>
+<% end %>
+<% content_for :head do %>
+  <meta name="robots" content="noindex">
+<% end %>
```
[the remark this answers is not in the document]
alex-ju: We could move it to the main layout, but given we're now planning to render the index page with this app, it may not be the best idea.

app/views/components/docs/hero.yml
```diff
@@ -0,0 +1,13 @@
+name: Hero section
+description: The main, introductory section of a page
+body: |
+  Currently used the introduce the results, this element was borrowed from topic pages (such as [Education and childcare](https://www.gov.uk/coronavirus/education-and-childcare))
```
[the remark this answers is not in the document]
alex-ju: Yes, that would be the aim – I would wait to see how our final design would look like (we may decide against this blue header) before thinking of moving this upstream to the gem. The only other instance that has something similar is `collections`, but it's worth looking across all apps – there may be other areas as well.

app/views/smart_answers/custom_result.html.erb
alex-ju: I thought about that, but then how do we keep this `custom_results` template generic, so that other flows can use it in future? I'm uneasy about coupling this with a specific flow.

config/smart_answers/next_steps_for_your_business.yml
alex-ju: Wouldn't that mean we'll introduce some redundancy where multiple results are part of the same section? Then if we don't group them under a section how would we render multiple results under a single title?

app/assets/stylesheets/components/_hero.scss
```diff
@@ -0,0 +1,12 @@
...
+  @include govuk-font($size: 19);
+  @include govuk-responsive-margin(6, 'bottom');
+
+  background-color: govuk-colour('blue');
+  padding: govuk-spacing(9) 0 govuk-spacing(6);
+
+  *,
+  *:link {
+    color: govuk-colour('white');
+  }
```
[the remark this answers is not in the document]
alex-ju: Good spot!

app/views/layouts/application.html.erb
```diff
@@ -29,12 +29,15 @@
     <% end %>
   </head>
   <body class="govuk-template__body">
-    <div class="govuk-width-container smart_answer" id="wrapper">
-      <main id="content" role="main">
-        <%= yield %>
-      </main>
-      <%= yield :escape_link %>
-      <%= render "govuk_publishing_components/components/feedback" %>
+    <div id="wrapper">
```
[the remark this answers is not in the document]
alex-ju: Yes, that's the selector slimmer uses when stitching a page together

app/views/layouts/application.html.erb
alex-ju: The MDN example has the `<header>` outside the `<main>` element
```
<header>Gecko facts</header>

<main role="main">
  <p>Geckos are a group of usually small, usually nocturnal lizards. They are found on every continent except Australia.</p>
 
  <p>Many species of gecko have adhesive toe pads which enable them to climb walls and even windows.</p>
</main>
```

The challenge we have here with moving the hero element inside the `<main>` element is that we'll have to diverge from the Design System-based template, which constrains the main element to a container (see below) – and I would try to avoid this, especially with the [ongoing work to use this standard template across our frontend apps](https://github.com/alphagov/static/pull/2411).

```
<div class="govuk-width-container ">
  <main class="govuk-main-wrapper" id="main-content" role="main">
    <p>Geckos are a group of usually small, usually nocturnal lizards. They are found on every continent except Australia.</p>
 
    <p>Many species of gecko have adhesive toe pads which enable them to climb walls and even windows.</p>
  </main>
</div>
```

app/views/smart_answers/custom_result.html.erb
theseanything: Yup fair! I guess I was thinking that this would save us having to put in additional functionality into ERB rendering - however, agree we probably don't want to use locales until we have more concrete plan about how we will use them for all flows. Let's stick with your proposed changes for now.

config/smart_answers/next_steps_for_your_business.yml
theseanything: So yup, it'll be more verbose. I find that the nesting structure can make it difficult to read and can make it more complicated to make changes i.e. it's easier to change the "section_name" value rather than move the whole result. This might be further problematic if we need to add another layer of grouping e.g ""what you need to do soon", "you might be interested in" etc etc

(Also further down the line if this data is being published from another system, it's probably unlikely to be structured with the nesting.)

Logic for a view can handle grouping results into sections - 
```ruby
# Create a hash where the key is the section name and value is array of results
grouped_results = results.group_by { |result| results["section_name"] }

# Access them like so...
business_tax_results = grouped_results["Business tax"]

# or...
grouped_results.each do |section_name, results| .... 
```

theseanything: (Again not something blocking this PR, but we can think about in future)

theseanything (approved):

config/smart_answers/next_steps_for_your_business.yml
alex-ju: I see your point about this data being made available via a publishing system – I'll give it a go at flattening it and grouping based on a string key (for now – we may be able to record these attributes as entities in a distant future)

alex-ju: Have updated this as suggested – thanks for expanding on it!

PR #5271, opened 2021-03-31
Update results.erb

DGarland5:
Added Local Restrictions Support Grants (LRSG (Closed) Addendum)

### DIFF
--- a/lib/smart_answer/calculators/business_coronavirus_support_finder_calculator.rb
+++ b/lib/smart_answer/calculators/business_coronavirus_support_finder_calculator.rb
@@ -10,6 +10,7 @@ module SmartAnswer::Calculators
 
     def initialize
       @closed_by_restrictions = []
+      @sectors = []
     end
 
     RULES = {
@@ -19,10 +20,6 @@ module SmartAnswer::Calculators
       statutory_sick_rebate: lambda { |calculator|
         calculator.business_size == "0_to_249" &&
           calculator.paye_scheme == "yes"
-      },
-      christmas_pub_payment: lambda { |calculator|
-        calculator.business_based == "england" &&
-          calculator.sectors.include?("retail_hospitality_or_leisure")
       },
       retail_hospitality_leisure_business_rates: lambda { |calculator|
         calculator.business_based == "england" &&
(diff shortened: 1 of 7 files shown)


DGarland5 (commented):
Reviewed feedback and made updates

lib/smart_answer/calculators/business_coronavirus_support_finder_calculator.rb
```diff
@@ -65,6 +65,22 @@ def initialize
       additional_restrictions_grant: lambda { |calculator|
         calculator.business_based == "england"
       },
+      restart_grant: lambda { |calculator|
+        calculator.business_based == "england" &&
+          calculator.sectors.intersection(%w[retail_hospitality_or_leisure personal_care]).any?
+      },
+      council_grants: lambda { |calculator|
```
gclssvglx, line 72: Nice!  Simplifies things a lot 👍

lib/smart_answer_flows/business-coronavirus-support-finder/outcomes/results.erb
```diff
@@ -186,249 +186,247 @@
...
-      - of £51,000 or over, you may receive £6,000
+    - of £15,000 or under, you may receive £2,667
+    - between £15,000 and £51,000, you may receive £4,000
+    - of £51,000 or over, you may receive £6,000
 
-      [Find out more about the Restart Grant](https://www.gov.uk/guidance/check-if-youre-eligible-for-the-coronavirus-additional-restrictions-grant). You’ll need to apply through your local council.
-      $CTA
-    <% end %>
+    [Find out more about the Restart Grant](https://www.gov.uk/guidance/check-if-youre-eligible-for-the-coronavirus-additional-restrictions-grant). You’ll need to apply through your local council.
+    $CTA
```
gclssvglx, line 427: I wonder if all the CTA's should be in separate partials?  That would leave the `if` logic here and the content designers would no longer need to touch this page.  Should help avoid the need for large cleanups like this.

gclssvglx (approved):
Nice one!  No blockers, but I've made a suggestion that I feel would help in future.

lib/smart_answer_flows/business-coronavirus-support-finder/outcomes/results.erb
kevindew: Yup, that's a good suggestion - I think though it's somewhat delicate as to how much that would help or not.

It would definitely make this file easier to mentally manage but I could see it also make life harder for Dolly when she is adding schemes. For example a commit like https://github.com/alphagov/smart-answers/pull/5271/commits/16984043eb9f4e070f9f0623d34d1670cac745b2 would be a bit harder with extra files to manage.

PR #5262, opened 2021-03-26
Adds country notice to the business support smart answer

DilwoarH:
## WHAT
We get the blue banner up (and down) on the business support smart answer for two weeks for the purpose of the test.

## WHY
We are doing a before and after test and the banner is the key bit!

https://trello.com/c/Py8XjzRQ/275-implement-blue-banner-for-before-and-after-test

Screenshot 2021-03-26 at 01 12 15
Screenshot 2021-03-26 at 01 12 23

⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

PR #5277, opened 2021-04-06
Publish Smart Answers as a single content item

theseanything:
This PR refactors Smart Answer to now be represented by a single content item in the Publishing API.

Previously, Smart Answer was represented two separate content items; one the for start page and the one for the flow. This was because start pages were published by Smart Answers as "transaction" page and rendered by Frontend in production. This was to prevent divergence in code between start pages in Smart Answer and Frontend (although there has been little divergence over years).

However, having separate content items has meant we had to introduce an extra `/s/` or `/y/` slug into URLs to prevent the base_path's of the two content items conflicting. This made it difficult to refactor routing, which is needed to [migrate the existing flows to use query parameters to store user responses](https://github.com/alphagov/smart-answers/blob/master/docs/arch/002-store-responses-in-query-string.md#rename-s-prefix-to-flow).

This PR will bering the following benefits:
- Consolidating the content items, means the start page will now be rendered by Smart Answers instead of Frontend. This allows us remove the extra `/s/` or `/y/` slug in Smart Answer URLs and will make it easier to transition existing flows from path based to be query parameter based.
- This simplifies the content item presenters and refactors the publishing process to only publish a single content item.
- The ability to implement Smart Answer only features on start pages without effecting all transaction pages. For example [additional banners](https://github.com/alphagov/smart-answers/pull/5262) and the ability to pass query parameter onwards to the flow (functionality needed by the "Next steps for you business" flow).

Existing flow will continue to use the previous start page content id, as this will allow history to be maintained in the Publishing API. The new content item presenter makes use of the new `smart_answer` schema.

[Work is being done](https://github.com/alphagov/smart-answers/pull/5286) to ensure the start page view template in Smart Answers is feature complete with the transaction view template in Frontend. That will need to be merged before this PR. 

Following this PR, the existing flow content items will be manually unpublished with `vanish`.  


alex-ju (approved):
Looks good! I like the simplified version of `content_istem_syncer`

PR #5289, opened 2021-04-09
Remove external related links from Smart Answers

theseanything:
This removes the ability to set external related links for Smart Answers. There are no flows that currently use this functionality and should built into content tagger instead.

### DIFF
--- a/app/presenters/flow_content_item.rb
+++ b/app/presenters/flow_content_item.rb
@@ -10,10 +10,8 @@ class FlowContentItem
       base_path: flow_presenter.start_page_link,
       title: flow_presenter.title,
       update_type: "minor",
-      details: {
-        external_related_links: flow_presenter.external_related_links,
-      },
-      schema_name: "generic_with_external_related_links",
+      details: {},
+      schema_name: "generic",
       document_type: "smart_answer",
       publishing_app: "smartanswers",
       rendering_app: "smartanswers",
--- a/app/presenters/flow_presenter.rb
+++ b/app/presenters/flow_presenter.rb
@@ -128,10 +128,6 @@ class FlowPresenter
     end
   end
 
-  def external_related_links
-    @flow.external_related_links || []
-  end
-
   def flows_content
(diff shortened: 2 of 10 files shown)


barrucadu (approved):
Does the schema name need changing, or is `generic_with_external_related_links` still ok? https://github.com/alphagov/smart-answers/blob/remove-external-related-links/app/presenters/flow_content_item.rb#L14

huwd (approved):
This seems like a sensible tidy up if it's not being used. Good spot!
Approving, but would be good to get a sense check on the context?

How has this come up now? Is this a consequence of you moving `smart-answers` off onto it's own content Item? Something like:

> it previously shared a content item with something that used `external_related_link` but now it is its own thing we can tidy up something it doesn't use?

alex-ju (approved):

alex-ju:
needs a rebase though

theseanything:
> Does the schema name need changing, or is `generic_with_external_related_links` still ok? https://github.com/alphagov/smart-answers/blob/remove-external-related-links/app/presenters/flow_content_item.rb#L14

Good shout. Probably wouldn't have any effect, but have switched to `generic` schema so it's less confusing. Ultimately we are creating and [switching Smart Answers to use a new schema `smart_answer](https://github.com/alphagov/smart-answers/pull/5277/commits/658d58b7c9c1f0f28cdad975ab3163d0b32cb40e#diff-eeec9eb2f8facd60b34242016c9df141f9ccfb1b22fb51b2eefe1f489786a6edR18)`, which will be more appropriate longer term.

theseanything:
> This seems like a sensible tidy up if it's not being used. Good spot!
> Approving, but would be good to get a sense check on the context?
> 
> How has this come up now? Is this a consequence of you moving `smart-answers` off onto it's own content Item? 

@huwd - ah, so this was just something I'd noticed whilst doing this work. However, there's no dependencies on doing this - just a small tidy up, that would simplify the codebase and make it easier to maintain.

PR #5315, opened 2021-04-26
Get current response for question

leenagupte:

# What's changed?

Add a method to `FlowPresenter` to get the existing answer (if it exists) to the current question that the user is on on the flow.

The simplest questions have been updated to pre-populate their answers using this new method.

To make the changes easier to review, the other, more complex questions will be updated in separate PRs.

## Changes in this PR
- [x] When a user click back on a URL based Smart Answer then the previous questions show, with the previous answer pre-filled
- [x] When a user click back on a session based Smart Answer then the previous questions show, with the previous answer pre-filled
- [x] Country-select question
- [x] Money question
- [x] Postcode question
- [x] Value question

## Changes in other PRs
- [x] Checkbox question (#5316)
- [x] Radio button question (#5318)
- [x] Date question (#5319)
- [x] Salary question (#5320)

# Why?

There is currently a bug in session-based smart-answers, where answers to previous questions are not being populated even though the answers exist in the session.

Getting the responses in one place (`FlowPresenter`) means that in future PRs we can remove the `prefill_value_*` methods in the helper and instead have each `QuestionPresenter` get the value of the current question from `FlowPresenter` and have the presenter itself decide how it needs to work with the data in custom cases, e.g. checkboxes.

The benefit of this approach is that each presenter doesn't need to know what type of flow (url, session) the smart-answer is.

### DIFF
--- a/app/presenters/country_select_question_presenter.rb
+++ b/app/presenters/country_select_question_presenter.rb
@@ -1,12 +1,10 @@
 class CountrySelectQuestionPresenter < QuestionPresenter
-  include CurrentQuestionHelper
-
   def select_options
     @node.options.map do |option|
       {
         text: option.name,
         value: option.slug,
-        selected: option.slug == prefill_value_for(self),
+        selected: option.slug == response_for_current_question,
       }
     end
   end
--- a/app/presenters/flow_presenter.rb
+++ b/app/presenters/flow_presenter.rb
@@ -79,8 +79,15 @@ class FlowPresenter
   end
 
   def response_for_current_question
-    responses = params[:responses]
-    responses[current_state.current_node.to_s]
+    if response_store
+      responses = params[:responses]
+      responses[current_state.current_node.to_s]
+    elsif params[:previous_response].present?
+      params[:previous_response]
+    else
+      question_number = current_state.path.size
+      all_responses[question_number]
+    end
   end
 
   def current_question_number
--- a/app/presenters/flow_presenter.rb
+++ b/app/presenters/flow_presenter.rb
@@ -78,6 +78,11 @@ class FlowPresenter
     @node_presenters[node.name] ||= presenter_class.new(node, self, current_state, {}, params)
   end
 
+  def response_for_current_question
+    responses = params[:responses]
+    responses[current_state.current_node.to_s]
+  end
+
   def current_question_number
     current_state.path.size + 1
   end
--- a/test/unit/flow_presenter_test.rb
+++ b/test/unit/flow_presenter_test.rb
@@ -8,7 +8,12 @@ class FlowPresenterTest < ActiveSupport::TestCase
   setup do
     @flow = SmartAnswer::Flow.new do
       name "flow-name"
-      value_question :first_question_key
+      value_question :first_question_key do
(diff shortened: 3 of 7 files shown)


app/presenters/country_select_question_presenter.rb
```diff
@@ -1,12 +1,10 @@
 class CountrySelectQuestionPresenter < QuestionPresenter
-  include CurrentQuestionHelper
```
kevindew, line 2: Great to see this go 👍

test/unit/flow_presenter_test.rb
```diff
@@ -8,7 +8,15 @@ def flow_registry
...
-      value_question :first_question_key
+      value_question :first_question_key do
+        next_node { question :second_question_key }
+      end
+      value_question :second_question_key do
+        next_node { question :third_question_key }
+      end
+      value_question :third_question_key do
+        next_node { outcome :outcome_key }
+      end
```
kevindew, line 19: Does anything make use of this third question? From my quick glance it looked like it might be superfluous.

test/unit/flow_presenter_test.rb
```diff
@@ -100,32 +108,30 @@ def flow_registry
...
       assert_equal(
-        "/calculate-your-holiday-entitlement/y/days-worked-per-week?previous_response=starting",
+        "/#{@flow.name}/y/question-1-answer?previous_response=question-2-answer",
         flow_presenter.change_collapsed_question_link(2, questions.first),
       )
       assert_equal(
-        "/calculate-your-holiday-entitlement/y/days-worked-per-week?previous_response=starting",
+        "/#{@flow.name}/y/question-1-answer?previous_response=question-2-answer",
         flow_presenter.change_collapsed_question_link(2, questions.last),
       )
```
kevindew, line 125: This looks a nice improvement 👍

app/presenters/flow_presenter.rb
```diff
@@ -79,8 +79,15 @@ def presenter_for(node)
...
-    responses[current_state.current_node.to_s]
+    if response_store
+      responses = params[:responses]
+      responses[current_state.current_node.to_s]
+    elsif params[:previous_response].present?
+      params[:previous_response]
+    else
+      question_number = current_state.path.size
+      all_responses[question_number]
+    end
```
kevindew, line 90: I feel like object/responsiblity wise it'd make a bit more sense if the current response was part of the current_state of a smart answer, there are some things feel a bit clunky around this. As we already discussed this is something I'm going to try look into later on and isn't blocking this PR (it'd be intended to enhance it). 

I do like that this method encapsulates the clunkiness though and that all the views only need to reference response_for_current_question.

kevindew (commented):
Looks good to me. Nice commit messages too 👍  - I've just left a minor comment to address.

test/unit/flow_presenter_test.rb
leenagupte: Ah, yeah good spot! I'll remove and fix up.

leenagupte:
@kevindew Thanks for your review. I've amended the test setup and fixed up.

kevindew (approved):
Super 👍

PR #5316, opened 2021-04-26
Get response for checkboxes

leenagupte:
Follows on from: PR #5315 

# What's changed?

Add a method to `CheckboxQuestionPresenter` to get the current value of the checkbox.

## Changes in this PR
- [x] Checkbox question

## Changes in other PRs
- [x] When a user click back on a URL based Smart Answer then the previous questions show, with the previous answer pre-filled (#5315)
- [x] When a user click back on a session based Smart Answer then the previous questions show, with the previous answer pre-filled (#5315)
- [x] Country-select question (#5315)
- [x] Money question (#5315)
- [x] Postcode question (#5315)
- [x] Value question (#5315)
- [x] Radio button question (#5318)
- [x] Date question (#5319)
- [x] Salary question (#5320)

# Why?

There is currently a bug in session-based smart-answers, where answers to previous checkbox questions are not being populated even though the answers exist in the session. 

### DIFF
--- a/app/presenters/checkbox_question_presenter.rb
+++ b/app/presenters/checkbox_question_presenter.rb
@@ -34,4 +34,10 @@ class CheckboxQuestionPresenter < QuestionWithOptionsPresenter
       }
     end
   end
+
+  def checked?(value)
+    return false if response_for_current_question.blank?
+
+    response_for_current_question.include?(value)
+  end
 end
--- a/app/presenters/checkbox_question_presenter.rb
+++ b/app/presenters/checkbox_question_presenter.rb
@@ -1,6 +1,4 @@
 class CheckboxQuestionPresenter < QuestionWithOptionsPresenter
-  include CurrentQuestionHelper
-
   def response_labels(values)
     values.split(",").map do |value|
       if value == SmartAnswer::Question::Checkbox::NONE_OPTION
@@ -29,7 +27,7 @@ class CheckboxQuestionPresenter < QuestionWithOptionsPresenter
         label: option[:label],
         value: option[:value],
         hint: option[:hint_text],
-        checked: prefill_value_includes?(self, option[:value]),
+        checked: checked?(option[:value]),
         exclusive: option[:value] == "none" || nil,
       }
     end
--- a/test/unit/checkbox_question_presenter_test.rb
+++ b/test/unit/checkbox_question_presenter_test.rb
@@ -15,6 +15,7 @@ module SmartAnswer
       @renderer.stubs(:option).with(:option3).returns({ label: "Option 3" })
 
       @presenter = CheckboxQuestionPresenter.new(@question, nil, nil, renderer: @renderer)
+      @presenter.stubs(:response_for_current_question).returns(nil)
     end
 
     test "#response_labels returns option labels for responses" do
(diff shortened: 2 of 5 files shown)


app/presenters/checkbox_question_presenter.rb
```diff
@@ -34,4 +34,16 @@ def checkboxes
...
   end
+
+  def checked?(value)
+    return if response_for_current_question.blank?
+
+    response = response_for_current_question
+
+    if response_for_current_question.is_a?(String)
+      response = to_response(response_for_current_question)
+    end
```
kevindew, line 43: This string check feels pretty low level, do we know what circumstances it comes up as a concern? The past code looks like it is only needed on `params[:previous_response]` so I'm wondering if this can be dealt with within: https://github.com/alphagov/smart-answers/blob/d2f1da06b6b3c44ba97144baedec3db0757bd9de/app/presenters/flow_presenter.rb#L86 ?

app/presenters/checkbox_question_presenter.rb
```diff
@@ -34,4 +34,16 @@ def checkboxes
       }
     end
   end
+
+  def checked?(value)
+    return if response_for_current_question.blank?
```
kevindew, line 37: This should return false rather than nil since it is a question mark method.

test/unit/checkbox_question_presenter_test.rb
```diff
@@ -53,5 +53,17 @@ class CheckboxQuestionPresenterTest < ActiveSupport::TestCase
...
+      should "return true if value appears in string" do
+        @presenter.stubs(:response_for_current_question).returns("option1,option2")
+        assert @presenter.checked?("option2")
+      end
+
+      should "return true if value appears in array" do
+        @presenter.stubs(:response_for_current_question).returns(%w[option1 option2])
+        assert @presenter.checked?("option2")
+      end
+    end
```
kevindew, line 68: It looks like this test is missing the blank guard clause.

test/unit/checkbox_question_presenter_test.rb
```diff
@@ -15,6 +15,7 @@ class CheckboxQuestionPresenterTest < ActiveSupport::TestCase
       @renderer.stubs(:option).with(:option3).returns({ label: "Option 3" })
 
       @presenter = CheckboxQuestionPresenter.new(@question, nil, nil, renderer: @renderer)
+      @presenter.stubs(:response_for_current_question).returns(nil)
```
kevindew, line 18: It might be useful to explain why this was added in the commit message, I had to do a bit of digging to understand why this line was added.

kevindew (commented):
Great to see this moving forward, only significant comment is related to the string type checking.

app/presenters/checkbox_question_presenter.rb
leenagupte: Is the suggestion to call `to_response` on `params[:previous_response]` in the `FlowPresenter` regardless of the question type?

leenagupte: That could work, because for most cases it does nothing, and for the others, I'm having to call it anyway.

leenagupte: Although, `to_response` only exists in the QuestionPresenters, not in FlowPresenter, so I don't think I can easily use it.

leenagupte:
@kevindew Thanks for your review. I've addressed all of your comments and fixed up.

kevindew (approved):
Great 👍

PR #5319, opened 2021-04-26
Get response for date question

leenagupte:
Follows on from: PR #5315 

# What's changed?

Add a method to `DateQuestionPresenter` to get the current value of the date fields.

## Changes in this PR
- [x] Date question

## Changes in other PRs
- [x] When a user click back on a URL based Smart Answer then the previous questions show, with the previous answer pre-filled (#5315)
- [x] When a user click back on a session based Smart Answer then the previous questions show, with the previous answer pre-filled (#5315)
- [x] Country-select question (#5315)
- [x] Money question (#5315)
- [x] Postcode question (#5315)
- [x] Value question (#5315)
- [x] Checkbox question (#5316)
- [x] Radio button question
- [x] Salary question (#5320)

# Why?

There is currently a bug in session-based smart-answers, where answers to previous date questions are not being populated even though the answers exist in the session. 

### DIFF
--- a/app/presenters/date_question_presenter.rb
+++ b/app/presenters/date_question_presenter.rb
@@ -1,6 +1,4 @@
 class DateQuestionPresenter < QuestionPresenter
-  include CurrentQuestionHelper
-
   delegate :default_day,
            :default_month,
            :default_year,
@@ -15,15 +13,21 @@ class DateQuestionPresenter < QuestionPresenter
   end
 
   def selected_day
-    prefill_value_for(self, :day)
+    return if response_for_current_question.blank?
+
+    response_for_current_question[:day]
   end
 
   def selected_month
-    prefill_value_for(self, :month)
+    return if response_for_current_question.blank?
+
+    response_for_current_question[:month]
   end
 
   def selected_year
-    prefill_value_for(self, :year)
+    return if response_for_current_question.blank?
+
+    response_for_current_question[:year]
   end
 
 private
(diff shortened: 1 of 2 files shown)


app/presenters/date_question_presenter.rb
```diff
@@ -15,15 +13,24 @@ def response_label(value)
   end
 
   def selected_day
-    prefill_value_for(self, :day)
+    return if response_for_current_question.blank?
+    return response_for_current_question unless response_for_current_question.is_a? Hash
```
kevindew, line 17: Would we expect this to ever be valid and not be a hash? I wonder if this should just return nil if it's not a hash.

leenagupte:
@kevindew Thanks for the re-review, I've made the suggested change and fixed up.

app/presenters/date_question_presenter.rb
```diff
@@ -15,15 +13,24 @@ def response_label(value)
   end
 
   def selected_day
-    prefill_value_for(self, :day)
+    return if response_for_current_question.blank?
```
kevindew, line 16: Minor: you don't actually need these guard conditions now you've got the second one. Sorry I didn't twig on that when I first made the hash suggestion.

kevindew (approved):
Just a minor thing that I don't think will need another review.

app/presenters/date_question_presenter.rb
leenagupte: Thanks, I've fixed and fixed-up.

kevindew (approved):

PR #5329, opened 2021-04-27
Remove redundant or inappropriate tests

theseanything:
This PR removes the following tests:
- Duplicate unit tests for the part_year_profit_tax_credits calculator class. The functionality being tested is covered by the existing unit tests. These tests are unit tests placed in the integration folder.
- Removes checks for every flow to have a content id. This isn't testing code functionality, but rather data. These are of limited value as ultimately as non critical interactions with the Publishing API will fail. Ideally this should be check should be covered as functionality written into the `Flow` class, rather than as an integration "test".
- Removes check for making sure flows are visualisable. Ultimately if a flow isn't visualisable there is not much that can be done, and just requires extra work to add it to the EXEMPTIONS_LIST. 


kevindew (commented):
Nice one, these seem reasonable to remove - all seem rather curious. I am a little surprised that the content_id one seems to be the only thing that tries to check all flows are remotely valid, but that seems a whole other can of worms 🙈 

Is there anything that uses the EXEMPTIONS_LIST constant once this code is removed? and struck me as something only for this test.

theseanything:
> Is there anything that uses the EXEMPTIONS_LIST constant once this code is removed? and struck me as something only for this test.

I've added a commit to remove the EXEMPTION_LIST, which also fixes the whole point of the `visualisable?` method. 🤦🏻 The `visualisable?` was used to add a warning on the visualisation page to show a warning. However, that never happened because the "integration test" forced that to always be true. Now it'll show a warning for the exempt flows (benefits cap calc, find coronavirus support), which makes sense because the visualisations are broken.

kevindew:
> > Is there anything that uses the EXEMPTIONS_LIST constant once this code is removed? and struck me as something only for this test.
> 
> I've added a commit to remove the EXEMPTION_LIST, which also fixes the whole point of the `visualisable?` method. 🤦🏻 The `visualisable?` was used to add a warning on the visualisation page to show a warning. However, that never happened because the "integration test" forced that to always be true. Now it'll show a warning for the exempt flows (benefits cap calc, find coronavirus support), which makes sense because the visualisations are broken.

Yay - an improvement too 🎉  Thanks @theseanything

kevindew (approved):

PR #5331, opened 2021-04-28
Prefill "none" option in checkboxes

leenagupte:
Follows on from: #5316 

# What's changed?

Stop removing "none" options when results to checkbox questions are processed.

# Why?

In url-based flows, if a user selected "none" in a checkbox question, and then went back to change their answer, nothing would be populated, and it would appear as though the user had never answered at all.
This issue doesn't seem to affect session-based smart-answers.

### DIFF
--- a/lib/smart_answer/question/checkbox.rb
+++ b/lib/smart_answer/question/checkbox.rb
@@ -38,7 +38,7 @@ module SmartAnswer
       end
 
       def to_response(input)
-        input.split(",").reject { |v| v == NONE_OPTION }
+        input.split(",")
       end
 
       def none_option?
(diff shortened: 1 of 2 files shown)


lib/smart_answer/question/checkbox.rb
```diff
@@ -38,7 +38,9 @@ def parse_input(raw_input)
       end
 
       def to_response(input)
-        input.split(",").reject { |v| v == NONE_OPTION }
+        input.split(",")
+      rescue StandardError
+        nil
```
kevindew, line 43: Rather than use standard error, that might conceal other exceptions should we do `input.split(",") if input.respond_to?(:split)`

leenagupte: Thanks! I've fixed and fixed-up.

kevindew (approved):

Issue #5225, opened 2021-03-05
Improve organisation and naming of controllers and methods


theseanything:
As discussed in #5212, we'd like to improve the organisation and naming of the controllers and their methods. After that ADR has been implemented we will have the following controller structure:

SmartAnswerController
- index (Index page)
- show (Start page)
- visualise (Visualisation page)

FlowController
- start (Redirect to the first flow question)
- show (Question and outcome pages)
- update (Redirect to next question or outcome)
- destroy (Clear responses and redirect to start page)

@kevindew raised that it is confusing having "SmartAnswersController" and "FlowController", as it unclear what each is responsible for. This may be further exacerbated by the concepts "Smart Answer" and "Flow" not having been strictly defined or existing conflicting usage.

The other change proposed was moving the `visualise` method from SmartAnswerController to the FlowController, due to it being publicly available in production environment. Start page and index pages are not rendered or accessible from Smart Answers in production. However, it's not clear whether visualise is a feature we support for public use, outside its primary use as a tool to help flow development.

kevindew:
Thanks for opening this @theseanything 👍 

The area that I think is most surprising is the start actions in the FlowController. I find it rather surprising that this doesn't return a start page as that's such a closely related concept.

The other actions on the FlowController also seem to be a little surprising as they are named the same as Rails resources action names but don't perform in the same way. Convention would dictate that the they all have a resource id and used GET/PUT/DELETE methods respectively). 

I'd wonder if we can get the action naming closer to the domain of the flow? We could probably use `node` instead of `show` since it seems to always refer to a node, `answer` or `next` might cover update since it's always about providing an answer and the URL has /next in it. `destroy` could be `exit` or `leave` to reflect the leaving process. I imagine `index` could replace `show` since that seems conventionally Rails for the root of a resource.

I'd imagine since we do publicly offer `/visualise` and it's probably a can of worms to deal with "do we want to do it"? it's probably simplest to just move that to FlowController and defer the wider considerations.

I'd wonder with the remaining SmartAnswerController whether we'd want to separate that into a DevelopmentController for the index page, that is purely development? Then that'd leave just start page which could go in a StartPageController, I guess a single action of `index` would make sense?

theseanything:
> The area that I think is most surprising is the start actions in the FlowController. I find it rather surprising that this doesn't return a start page as that's such a closely related concept.

[...]

PR #5334, opened 2021-04-29
Remove `/s/` prefix for default routes to FlowController

theseanything:
This sets the defaults routes to the FlowController to not be prefixed with `/s/`. The prefix was introduced to deal with a constraint of publishing the start page and the flow as separate content items in the Publishing API. The two content items needed unique base paths. However, Smart Answers now render the start page and are represented by a single content item and this constraint is removed. Removing the prefix simplifies the url structure and removes confusion around the meaning of the prefix.

The old prefix routes will continue to be supported until existing traffic no longer uses them - then they can be removed entirely. This will automatically update the urls in the pages, so users who continue their user journey will begin to use the non-prefixed routes.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -16,13 +16,14 @@ Rails.application.routes.draw do
         format: false
   end
 
-  get "/:id/s/destroy_session", to: "flow#destroy", as: :destroy_flow
-  get "/:id/s", to: "flow#start", as: :start_flow
-  get "/:id/s/:node_slug", to: "flow#show", as: :flow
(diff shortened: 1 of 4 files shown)


kevindew (approved):
This looks good to me - and surprisingly simple. It'll be ace to lose those prefixes and this seems isolated enough that it'll hopefully not introduce unforeseen issues.

It'd be good to continue the [discussion on the path/method names](https://github.com/alphagov/smart-answers/issues/5225), I find the `destroy_session` one a bit jarring.

PR #5336, opened 2021-04-30
Use summary list for answers

alex-ju:
### What
 - Replace 'Your answers' table with summary list component
 - Replace 'Start again' link in table caption with a standard link

### Why

The current implementation fails WCAG 2.1 as follows:
 - 1.3.1 Info and Relationships (Level A) – [because of table markup and change links in table](https://trello.com/c/zLF7s8Tu)
 - 2.1.1 Keyboard (Level A) – [because of link in table caption](https://trello.com/c/ql82hggf)

### Visual changes

<table>
<tr><th>Before</th><th>After</th></tr>
<tr><td valign="top">

mobile-before

</td><td valign="top">

mobile-after

</td></tr>
<tr><td valign="top">

desktop-before

</td><td valign="top">

desktop-after

</td></tr>
</table>


app/views/smart_answers/shared/_previous_answers.html.erb
```diff
@@ -1,54 +1,62 @@
...
+      margin_bottom: 6
+    } %>
+  <% else %>
+    <%= render "govuk_publishing_components/components/heading", {
+      text: "Your answers",
+      heading_level: 2,
+      font_size: "m",
+      margin_bottom: 4
+    } %>
+  <% end %>
```
theseanything, line 16: I wonder if we could make these headings more consistent and simplify the code?

```ruby
<%= render "govuk_publishing_components/components/heading", {
  text: local_assigns[:hide_previous_answers] ? "Change your answers" : "Your answers",
  heading_level: 2,
  font_size: "m",
  margin_bottom: 4
} %>
```

Not sure the reasoning around the difference for margin and font - but from the git history, this was just lifted and shifted from another part of the page, where the larger font and margin may have made more sense. Don't think there is any reason not to make them consistent.

app/views/smart_answers/shared/_previous_answers.html.erb
```diff
@@ -1,54 +1,62 @@
...
-    <% end %>
-  </table>
+  <% unless local_assigns[:hide_previous_answers] %>
+    <% items = @presenter.collapsed_questions.each_with_index.map { |question, number_questions_so_far|
+      accepted_response = @presenter.accepted_responses[number_questions_so_far]
+
+      if question.multiple_responses?
+        value = render "govuk_publishing_components/components/list", {
+          items: question.response_labels(accepted_response)
+        }
```
theseanything, line 37: Is the intended output a `<ul>`? I don't think this works because the accepted responses is always a string, not an array of responses. For questions where `multiple_responses? == true`, the output is string with the responses joined with a `,` e.g. `"response1, response2, response3"` not `["response1", "response2", "response3"]`.

app/views/smart_answers/shared/_previous_answers.html.erb
```diff
@@ -1,54 +1,62 @@
...
+        edit: {
+          href: @presenter.change_collapsed_question_link(number_questions_so_far + 1, question),
+          data_attributes: {
+            "module": "gem-track-click",
+            "track-category": "Smart Answer Change Link",
+            "track-action":  "Link clicked",
+            "track-label": "#{question.title} / #{question.response_label(accepted_response)}"
+          }
+        }
+      }
```
theseanything, line 54: Nit: It'd be great to be able to push this logic into the presenter. If the question's knew their own responses, we wouldn't have to worry about matching up question presenters with their response, which would simplify things. 

But that's probably apart of a larger refactor.

theseanything: Actually, looks like this might be encompassed in https://github.com/alphagov/smart-answers/pull/5337

app/views/smart_answers/shared/_previous_answers.html.erb
alex-ju: Happy to do so! I assumed some design considerations from `find-coronavirus-support` and didn't want to mess with them.

<table>
<tr><th>Before</th><th>After</th></tr>
<tr><td valign="top">

before


</td><td valign="top">

after


</td></tr>
</table>

app/views/smart_answers/shared/_previous_answers.html.erb
alex-ju: Yes, the intention is to render a list of items when we deal with multiple items. `response_labels` seems to be [always returning an array as it splits by `,`](https://github.com/alphagov/smart-answers/blob/main/app/presenters/checkbox_question_presenter.rb#L2-L10).

app/views/smart_answers/shared/_previous_answers.html.erb
alex-ju: Agree will try and add it to the flow presenter then – I essential adapted it from the nested partial, clearly not the cleanest way.

alex-ju: ah, I just saw your second comment. Would be keen to refactor afterwards then.

app/views/smart_answers/shared/_previous_answers.html.erb
theseanything: Since the code was already split like this - happy for this to be something we can look at later.

app/views/smart_answers/shared/_previous_answers.html.erb
theseanything: 🤦🏻 Sorry you're right!! I was getting confused!

theseanything (approved):

PR #5342, opened 2021-05-04
Make question hints interface more consistent

theseanything:
This makes the interface and behaviour more consistent across different question types. Now all question access hint text via the `hint` attribute and pass it to relevant property on the component. This also unused fields `pretext` and `body` for some question types where they aren't used - in preference to use hint or post body text in future.

### DIFF
--- a/app/views/smart_answers/inputs/_salary_question.html.erb
+++ b/app/views/smart_answers/inputs/_salary_question.html.erb
@@ -8,7 +8,8 @@
   id: "response_amount",
   heading_size: "l",
   width: 10,
-  hint: "in £",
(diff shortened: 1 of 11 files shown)


app/views/smart_answers/inputs/_salary_question.html.erb
```diff
@@ -8,7 +8,8 @@
   id: "response_amount",
   heading_size: "l",
   width: 10,
-  hint: "in £",
+  hint: question.hint,
+  suffix: "£",
```
alex-ju, line 12: ❤️

alex-ju (approved):
🙌

PR #5347, opened 2021-05-06
Proposal: ADR-3 - "Outcome-centric Logic"

ChrisBAshton:
👉 [Rendered version](https://github.com/alphagov/smart-answers/blob/adr-3/docs/arch/adr-003-outcome-centric-logic.md) 👈 

### DIFF
--- a/docs/arch/adr-003-outcome-centric-logic.md
+++ b/docs/arch/adr-003-outcome-centric-logic.md
@@ -1,0 +1,112 @@
+# ADR 3: Outcome-centric Logic
+
+## Problem
+
+Some of our Smart Answer flows have become very difficult to read and understand.
+This is down to two factors: liberal use of nesting, and different branches of logic leading to the same outcomes.
+
+As a result, it can be very difficult to give an answer to the question "how do I reach outcome X?". It can also be difficult to deduce whether multiple conditions can lead to the same outcome.
+
+This makes the code harder to work with, slowing pace of delivery and increasing the likelihood of bugs.
+
+We've identified three Smart Answers which have these issues:
+
+- [pay-leave-for-parents](https://github.com/alphagov/smart-answers/blob/7cc0ca0c85fe9de04d859e705dbe9b9b5d402306/lib/smart_answer_flows/pay-leave-for-parents.rb#L251-L459)
+- [check-uk-visa](https://github.com/alphagov/smart-answers/blob/7cc0ca0c85fe9de04d859e705dbe9b9b5d402306/lib/smart_answer_flows/check-uk-visa.rb#L194-L221)
+- [uk-benefits-abroad](https://github.com/alphagov/smart-answers/blob/7cc0ca0c85fe9de04d859e705dbe9b9b5d402306/lib/smart_answer_flows/uk-benefits-abroad.rb#L163-L188)
+
+## Proposal
+
+Instead of calculating in terms of the user's answers, we should treat each outcome in turn, and calculate whether the user's answers should qualify for that outcome. Here is a [proof of concept](https://github.com/alphagov/smart-answers/commit/a5ab26089b2dd77815d35b80c8960c86a08a1661) for one portion of one of the offending Smart Answer flows.
+
+This approach:
+
+1. defines the required conditions on a per-outcome basis, so it's much easier to reason about how to get to a particular outcome
+1. flattens the heavily nested and complex structure, making it easier to make changes to the logic.
+
+If we look at the vast majority of Smart Answer flows which do _not_ suffer from this spaghetti code, such as the Register a Birth Smart Answer, the logic [is very easy to follow](https://github.com/alphagov/smart-answers/blob/7cc0ca0c85fe9de04d859e705dbe9b9b5d402306/lib/smart_answer_flows/register-a-birth.rb). Note that for each 'node', there is only one branch of logic to arrive at a particular outcome, and there is no nesting. In other words, **most Smart Answers already use an outcome-centric approach**.
+
+This proposal is to bring _all_ Smart Answers in line, by following these rules:
+
+```
+for every smart answer flow:
+  for every node in the flow:
+    - each outcome is only defined once
+    - there is only one level of indentation
+      (i.e. if/else and case statements are allowed, but no nesting)
+```
+
+This should bring all of our Smart Answers up to roughly the same readability standard, and we'll then have this ADR we can point our developers to, to prevent this from happening again.
+
+### Suggested steps for implementation
+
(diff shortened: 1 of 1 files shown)


kevindew:
Thanks for raising this Chris 🎉   it's always great to have fresh ideas on these things

Couple of quick thoughts from me: 

Firstly, I'm not sure this is really an architectural decision, it seems more localised to what are good code practices for dealing with logic. It'd seem quite reasonable to refactor the Smart Answers with the problem without needing an ADR, on just the basis that they are hard to understand.  The example given in the ADR of a [standard violating flow](https://github.com/alphagov/smart-answers/pull/5347/files#diff-8f37e9339d97c3d00ebf1922038eb645c4213f59633944ac81eeff0035a0a71fR54-R68) doesn't seem to be something that looks problematic to me (or seem to offer a particular readability advantage compared to [the alternative](https://github.com/alphagov/smart-answers/pull/5347/files#diff-8f37e9339d97c3d00ebf1922038eb645c4213f59633944ac81eeff0035a0a71fR72-R84)). If I had coded the first one I'd be quite surprised to be told about this ADR convention in a review, it'd feel a strangely dogmatic for something quite contextual.

Secondly, when I look at the proposed improvement from the spike: https://github.com/alphagov/smart-answers/commit/a5ab26089b2dd77815d35b80c8960c86a08a1661#diff-f08ad6e534d72243b8a2ee513a6ef4c2752f7152efc6130b1e3c2c4648e99f88R141-R191 I'm wondering if it is enough of an improvement to justify standardisation. I can see the change is noticeably better than what precedes it, but is still an extremely complex conditional and I would imagine the experience of unit testing it would be impractical, tedious, and miserable.  I'm wondering if there is actually a missing abstraction layer for these flows with challenging logic - perhaps at the calculator or an additional class (or multiple) - as a 50 line if/else block is always going to be a tough mental exercise to understand.

Modestly this could be along the lines of: 

```
next_node do
  if calculator.partner_started_working_before_continuity_start_date?
    question :partner_started_working_before_continuity_start_date
  elsif calculator.mat_leave_mat_pay?
    outcome :outcome_mat_leave_mat_pay
  elsif calculator.mat_allowance_14_weeks?
    outcome :outcome_mat_allowance_14_weeks
  # etc...
  end
```

Or going further:

```
next_node do
  resolver = MaternityParentalPayLeave::OutcomeResolver.new(calculator)
  if resolver.outcome?
    outcome resolver.outcome
  else
     question resolver.next_question
  end
end
```

docs/arch/adr-003-outcome-centric-logic.md
```diff
@@ -0,0 +1,112 @@
...
+If we look at the vast majority of Smart Answer flows which do _not_ suffer from this spaghetti code, such as the Register a Birth Smart Answer, the logic [is very easy to follow](https://github.com/alphagov/smart-answers/blob/7cc0ca0c85fe9de04d859e705dbe9b9b5d402306/lib/smart_answer_flows/register-a-birth.rb). Note that for each 'node', there is only one branch of logic to arrive at a particular outcome, and there is no nesting. In other words, **most Smart Answers already use an outcome-centric approach**.
+
+This proposal is to bring _all_ Smart Answers in line, by following these rules:
+
+```
+for every smart answer flow:
+  for every node in the flow:
+    - each outcome is only defined once
+    - there is only one level of indentation
+      (i.e. if/else and case statements are allowed, but no nesting)
```
ChrisBAshton, line 36:
```suggestion
      (i.e. if/else and case statements are allowed, but no nesting)
    - outcome logic is delegated to the calculator
```

docs/arch/adr-003-outcome-centric-logic.md
```diff
@@ -0,0 +1,112 @@
...
+```ruby
+next_node do |response|
+  if calculator.course_type == "uk-full-time" && response == "dental-medical-healthcare"
+    question :are_you_a_doctor_or_dentist?
+  elsif calculator.course_type == "uk-full-time"
+    outcome :outcome_uk_full_time_students
+  elsif calculator.course_type == "uk-part-time"
+    outcome :outcome_uk_all_students
+  else
+    outcome :outcome_eu_students
```
ChrisBAshton, line 81: With the new proposed rule, this code would look like:

```suggestion
  if calculator.should_ask_if_doctor_or_dentist?
    question :are_you_a_doctor_or_dentist?
  elsif calculator.uk_full_time_student?
    outcome :outcome_uk_full_time_students
  elsif calculator.uk_part_time_student?
    outcome :outcome_uk_all_students
  else
    outcome :outcome_eu_students
```

There is the question of what we should name these calculator methods - I note in your initial example you've [named the method the same as the outcome](https://github.com/alphagov/smart-answers/pull/5347#issuecomment-833439172), which read quite confusingly to me, though would probably help keep things consistent. I've proposed a more imperative naming style above, which feels a bit more natural and unit-testable. Would be interested in your thoughts.

ChrisBAshton (commented):
Thanks for the comments, @kevindew.

> The example given in the ADR of a [standard violating flow](https://github.com/alphagov/smart-answers/pull/5347/files#diff-8f37e9339d97c3d00ebf1922038eb645c4213f59633944ac81eeff0035a0a71fR54-R68) doesn't seem to be something that looks problematic to me (or seem to offer a particular readability advantage compared to [the alternative](https://github.com/alphagov/smart-answers/pull/5347/files#diff-8f37e9339d97c3d00ebf1922038eb645c4213f59633944ac81eeff0035a0a71fR72-R84)). If I had coded the first one I'd be quite surprised to be told about this ADR convention in a review, it'd feel a strangely dogmatic for something quite contextual.

That's fair. The example given was meant to concisely demonstrate what changes would happen if we applied the rules proposed in the ADR. I agree that the current code in that example isn't problematic, and thus we could use some discretion rather than apply the rules universally, though this restricts future possibilities such as ADR conformance code linting. That said, it would be good to get a feel for how effective this proposed approach is at keeping Smart Answers maintainable, before getting ahead of ourselves. So I'd be happy to add a note about using judgement rather than dogmatically applying the ADR - does that sound ok?

I like your suggestion of a named method tied to each unique outcome:

```ruby
if calculator.partner_started_working_before_continuity_start_date?
  question :partner_started_working_before_continuity_start_date
elsif calculator.mat_leave_mat_pay?
  outcome :outcome_mat_leave_mat_pay
# etc...
```

The above is actually in line with the ADR proposal, insofar as there is only one level of indentation, and each outcome appears only once per node. The difference is that the logic encapsulating each condition has been moved into the calculator, which seems a good call to make. I've always found Smart Answers to be quite a curious mix of delegating decisions to calculators and also making some decisions inside the flow class itself.

Would it be a reasonable tweak to the ADR to propose that there is a third rule - for logic to be defined inside the calculator - as in my comments below? As above, it would be a useful guide to follow, rather than a cast-iron rule every calculator has to follow.

PR #5348, opened 2021-05-06
Add GA event to track response submission using question keys

theseanything:
This PR adds an additional GA event that tracks user responses in reference to the node id of the question, instead of the question text. Tracking using the question text was fragile as minor changes to question text made it difficult to group similar events. Smart Answer's will continue to send the existing GA events for a user response in reference to the question text, to support any existing analytic analysis (e.g. find-coronavirus-support). Ideally in the longer term we will remove that old GA event.

This also fixes confusing code references to the `data-question-key` HTML attribute as the question key. However that was being set as the question text.

### DIFF
--- a/app/assets/javascripts/modules/track-responses.js
+++ b/app/assets/javascripts/modules/track-responses.js
@@ -9,8 +9,8 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
       track(element[0])
     }
 
-    function getQuestionKey (submittedForm) {
-      return submittedForm.getAttribute('data-question-key')
+    function getQuestionHeading (submittedForm) {
+      return submittedForm.querySelector('h1').innerText
     }
 
     function getResponseLabelsForRadio (submittedForm) {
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/modules/track-responses.js
```diff
@@ -9,8 +9,8 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
       track(element[0])
     }
 
-    function getQuestionKey (submittedForm) {
-      return submittedForm.getAttribute('data-question-key')
+    function getQuestionHeading (submittedForm) {
+      return submittedForm.querySelector('h1').innerText
```
alex-ju, line 13: Not sure I agree with the approach here. The advantage of using a data attribute is that it communicates the element is being used to pick data from, regardless of its semantics (now is a heading level 1, but tomorrow could be a heading level 2 – and the analytics script will fail silently).

Was expecting something like `querySelector('data-question-key').innerText`

alex-ju (approved):
One comment about querying a certain element type – otherwise looks good!
Is there a plan for follow-up work to converge the two events into one?

app/assets/javascripts/modules/track-responses.js
theseanything: Thats a good point. I'll add in another attribute that's data-question-text to support the existing events, without referencing the h1 e.g. `querySelector('data-question-text').innerText`

PR #5361, opened 2021-05-11
Open result links in new tab for "Next steps for your business"

theseanything:
We changed the behaviour so links on the results page open in a new tab or window. This is to help user return to the results page.


app/views/components/_result-item.html.erb
```diff
@@ -11,6 +11,8 @@
   <h4 class="govuk-heading-s">
     <%= link_to title, url,
     class: "govuk-link",
+    target: "_blank",
+    rel: "noopener noreferrer",
```
alex-ju, line 15: 👍

PR #5367, opened 2021-05-11
Updade custom results tracking action value

alex-ju:
We have 3 levels of information on the custom result page: group, topic, and result. We initially setup tracking values as `[topic_index].[result_index]`, but our PA finds it more useful to be `[group_index].[result_index]` so we're updating accordingly. To avoid confusion we explicitly name indexes at the component level and update the result-section component to render a set of sections to maintain component isolation.

[Trello card](https://trello.com/c/l0LQGvFP)


app/views/components/_result-sections.html.erb
```diff
@@ -0,0 +1,18 @@
...
+<% topics.each do |topic, results| %>
+  <section class="app-c-result-section">
+    <h3 class="govuk-heading-m"><%= topic %></h3>
+    <% results.each do |result, index| %>
+      <%= render "components/result-item", {
+        highlighted: highlighted,
+        group_index: group_index,
+        result_index: result_index
+      }.merge(result.symbolize_keys) %>
+      <% result_index += 1 %>
```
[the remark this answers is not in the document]
theseanything: Hmm, I can't think of a neater way to do that.

app/views/components/_result-sections.html.erb
```diff
@@ -0,0 +1,18 @@
+<%
+  highlighted ||= true
+  group_index ||= 1
+  result_index ||= 1
+%>
+<% topics.each do |topic, results| %>
+  <section class="app-c-result-section">
+    <h3 class="govuk-heading-m"><%= topic %></h3>
+    <% results.each do |result, index| %>
```
theseanything, line 9: Do we use the index here?
```suggestion
    <% results.each do |result| %>
```

theseanything (approved):

app/views/components/_result-sections.html.erb
alex-ju: Good spot – we certainly don't 😬

PR #5370, opened 2021-05-12
Fix VAT result logic for Next steps for your business

theseanything:
We have three results for VAT registration however we should only show one depending on which option the user selects for the annual turnover question.


alex-ju (approved):
Good spot!

PR #5375, opened 2021-05-14
SPIKE - Implement new testing strategy for smart answer

JonathanHallam:
The intention of this spike is to make sure that all logic is tested and to catch syntax errors on the start/question pages in the steps along the flow of a smart answer.

To do this we have created new integration tests using minitest in the following format;

```ruby
require "test_helper"

class MyFlowTest < ActionDispatch::IntegrationTest
  setup do
    stub_content_store_has_item("/my-flow")
  end

  context "test something" do
    should "render the results page" do
      get "/my-flow/y/all/params/to/results/page"
      assert_select "p", "some text that only exists on the results page"
    end
  end
end
```
This has several advantages over our current tests;
 * We are rendering the result page - without using Capybara.
 * We are concentrating on the flow outcome, not the flow path or route (although, by definition we are exercising the flow path). We're not explicitlly testing the routes, instead relying on the successful rendering of the results page to indicate that we've made it to the end of a route.  
 * We ensure that the tests will fail if there are any syntax errors (i.e. <% if true %>, but no <% end %>) on the start page, any of the interim question pages, or the result page.
 * We are only testing the happy path as this is all we need to test here having moved view logic into the calculator.
 * These tests execute significantly faster than RSpec tests that have similar coverage (see Test speed comparison). They are somewhat slower than existing Minitests but are testing content which was not done previously.

### How?

As a general principle, move as much of the content from the existing result page into a partial or partials.

Add public methods to the flow’s Calculator like so…

```ruby
def outcome_method
  content_array = Array.new
  content_array << "a_partial_name"
  content_array << "another_partial_name" if something == "something_else"
  content_array << "yet_another_partial_name"
end
```

On the result page…

```erb
<% govspeak_for :body do %>
  <%# some text that is always displayed specific to this outcome %>

  <% calculator.outcome_method.each do |part| %>
    <%= render partial: part %>
  <% end %>

  <%# some text that is always displayed specific to this outcome %>
<% end %>
```

### Why is this an improvement?

 * We have removed all of the logic from the result page and into the flow’s Calculator where it belongs.
     - This allows us to do individual unit tests for the methods we’ve created in the calculator. This improves test speed and clarity and also allows us to minimise tests on the frontend as we can be confident in our error handling and alerting before this time.
 * We have moved as much content as possible from the result page into a partial or partials.
     - This change means that we can more easily render the content. It also reduces the amount of logic we need to add to the calculator.
 * We are minimising the potential to break the result page by removing code logic statements.

### RSpec
```shell
Finished in 4 minutes 3.1 seconds (files took 5.56 seconds to load)
24 examples, 0 failures
```

### MiniTest - old version
```shell
Finished in 0.299327s, 434.3076 runs/s, 1530.0992 assertions/s.
130 runs, 458 assertions, 0 failures, 0 errors, 0 skips
```

### MiniTest - new version
```shell
Finished in 4.414292s, 1.1327 runs/s, 1.1327 assertions/s.
5 runs, 5 assertions, 0 failures, 0 errors, 0 skips
```

## Next steps

We’re aware that there’s a good degree of repetition in the outcomes for holiday entitlement smart answer. In the long/medium term this could be moved to a single outcome erb file. This would require the text at the beginning of each outcome to be moved into either its own partial or a new I18N file that could contain content for the whole flow.

### DIFF
--- a/lib/smart_answer/calculators/holiday_entitlement.rb
+++ b/lib/smart_answer/calculators/holiday_entitlement.rb
@@ -181,6 +181,40 @@ module SmartAnswer::Calculators
       (shifts_per_shift_pattern / days_per_shift_pattern * DAYS_PER_WEEK).round(10)
     end
 
+    def compressed
+      content_array = []
+      content_array << "your_employer_with_rounding"
+      content_array << "the_user_should_be_aware" if holiday_period == "starting"
+      content_array << "guidance_on_calculations"
+    end
+
+    def days_and_hours_per_week
+      content_array = []
+      content_array << "days_per_week_greater_than_five" if working_days_per_week > 5
+      content_array << "your_employer_with_rounding"
+      content_array << "the_user_should_be_aware" if holiday_period == "starting"
+      content_array << "guidance_on_calculations"
+    end
+
+    def irregular_and_annualised
+      content_array = []
+      content_array << "your_employer_with_rounding"
+      content_array << if holiday_period == "starting"
+                         "irregular_and_annualised_user_awareness"
+                       else
+                         "entitlement_restriction"
+                       end
+      content_array << "guidance_on_calculations"
+    end
+
+    def shift_worker
+      content_array = []
+      content_array << "shifts_per_week_greater_than_five" if shifts_per_week > 5
+      content_array << "shift_worker_your_employer_with_rounding"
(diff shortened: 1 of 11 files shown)


kevindew:
Superseded by https://github.com/alphagov/smart-answers/pull/5394

PR #5388, opened 2021-05-19
Remove with_cache_control_expiry test helper

leenagupte:
The `with_cache_control_expiry` method is only used in two tests, and is
doing the same job as a stub, so it's not really needed.

Also, it felt odd to be calling a `get` from within the setup, so that
has been moved to be called explicitly within in each test as well.


kevindew (approved):
Looks good, nice cleanup

Issue #5350, opened 2021-05-06
This application has both RSpec and Minitest testing frameworks, we only want one of them


kevindew:
Smart Answers is one of GOV.UK's oldest applications (born 2011), it pre-dates the GOV.UK convention of using RSpec for Ruby testing and instead uses Minitest for testing. It has a large test suite (approx 7,000 tests and 20,000 assertions) which can be quite difficult to work with.

One of the gaps in the Minitest suite is that the embedded Ruby template files that are part of a flow ([example](https://github.com/alphagov/smart-answers/blob/04602a364176b0523ae939b45ad573cd4be84204/lib/smart_answer_flows/am-i-getting-minimum-wage/outcomes/current_payment_above.erb)) are not rendered in the test suite. This leaves a coverage risk that any syntactic errors in these files are not caught by the test suite. In November 2020 [a new approach](https://github.com/alphagov/smart-answers/pull/5000) was taken that added RSpec to the test suite and integration tested the flows with Capybara. This new approach was precipitated on a migration of the test suite from RSpec to Minitest and not intended to be supplementary.

In March 2021 we (GOV.UK Covid Products and GOV.UK Start a Business) reconsidered the approach to migrate the whole test suite to RSpec. We concluded that we should not proceed with a migration to RSpec because we felt that the anticipated effort would outweigh the value generated, we also had concerns about the performance of a fully migrated RSpec test suite (the existing Minitest one benefits from [Rails parallel testing](https://edgeguides.rubyonrails.org/testing.html#parallel-testing)). Instead we would replace any RSpec tests with Minitest ones so that Smart Answers only uses the one testing framework.

The tasks we identified that needed to be done for this are:

- [x] Port over [the RSpec unit tests](https://github.com/alphagov/smart-answers/tree/04602a364176b0523ae939b45ad573cd4be84204/spec/lib) to Minitest
- [ ] Port over (or remove if already covered) [generic feature tests](https://github.com/alphagov/smart-answers/tree/04602a364176b0523ae939b45ad573cd4be84204/spec/features/smart_answers) to [engine integration tests](https://github.com/alphagov/smart-answers/tree/04602a364176b0523ae939b45ad573cd4be84204/test/integration/engine)
- [x] Decide on an approach to testing flows that involves rendering the view files - https://trello.com/c/n0yuPABN/803-placeholder-spike-into-how-smart-answer-flow-tests-can-test-erb-rendering
[...]

PR #5390, opened 2021-05-20
Refactor RSpec unit tests to Minitest

theseanything:
This is apart of migration of our existing RSpec testing to Minitest. We are consolidating our tests to a single testing framework as it was unclear which framework to write new tests in. These test we're written in RSpec in attempt to start a migration to RSpec, however now we've decided that the effort in migrating the existing MiniTest tests is not worth the benefits we gain from using RSpec. Particular concerned with the lack of parallel testing which would cause a significant slow down.

This partial address issue #5350, which discusses the wider migration of tests to MiniTest.


test/helpers/flow_helper_test.rb
```diff
@@ -0,0 +1,110 @@
+require "test_helper"
+
+class FlowHelperTest < ActionView::TestCase
```
kevindew, line 3: Should this be accompanied by a deletion of spec/helpers/flow_helper_spec.rb ?

kevindew (commented):
This looks great, thanks @theseanything

test/helpers/flow_helper_test.rb
theseanything: Good catch!! 🤦🏻

kevindew (approved):

PR #5391, opened 2021-05-21
Replace stubbing of WorldLocation `find` and `all`

leenagupte:
Trello: https://trello.com/c/xTCxOAD6

# What's changed?

Uses the helper methods in gds-api-adaptors to stub the WorldLocation data rather than stubbing the `find` and `all` method.

# Why?

WorldLocation is used to populate country lists in multiple smart-answers, I believe that stubbing the `all` method has led to some strange behaviour in the tests. 

The tests in `uk_benefits_abroad_test.rb` fail intermittently in CI [1].
I've tested this locally by adding a loop to the tests and running the "when going abroad" tests 50 times

e.g.
```
50.times do |i|
  context "when going abroad #{i}" do
```

I found that two of assertions in tests would always fail. They weren't the same test each time. The only thing the failures have in common is that the node would always drop back to the `:which_country?` node, implying that there was something wrong with the country data. Either a country hadn't been selected, or the more likely reason that the country selected in the test wasn't available in the list of countries returned by `WorldLocation.all`.

If a stub of the `all` method was indeed lingering about, then it would explain why the tests sometimes fail.

Removing the remaining stubs for the `find` and `all` method has made all of the tests in `uk_benefits_abroad_test` stable.

There are a few lingering concerns...

Firstly, why were the stubs leaking between tests?

Also, I'm not sure about the changes to the `MarriageAbroadCalculatorTest`. The stubbing feels a bit low-level. It feels like the test needs to know too much about what is returned from Worldwide API. Also, should the tests for `fco_organisation` even be in `MarriageAbroadCalculatorTest`? This method delegates to `WorldLocation` so it feels like the tests should be there.

[1]: https://ci.integration.publishing.service.gov.uk/job/smartanswers/job/main/5525/console

### DIFF
--- a/test/unit/calculators/marriage_abroad_calculator_test.rb
+++ b/test/unit/calculators/marriage_abroad_calculator_test.rb
@@ -254,15 +254,12 @@ module SmartAnswer
       end
 
       context "#world_location" do
-        setup do
-          @calculator = MarriageAbroadCalculator.new
-          @calculator.ceremony_country = "ceremony-country"
-        end
-
         should "return the world location for the given ceremony country" do
-          WorldLocation.stubs(:find).with("ceremony-country").returns("world-location")
+          stub_worldwide_api_has_location("world-location")
+          @calculator = MarriageAbroadCalculator.new
+          @calculator.ceremony_country = "world-location"
 
-          assert_equal "world-location", @calculator.world_location
+          assert_equal "world-location", @calculator.world_location.slug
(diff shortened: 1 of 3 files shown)


kevindew (approved):
This looks good to me 👍  nice code reduction.

Fingers crossed it gets those flaky tests 🤞 🤞  - thanks for taking on that battle ⭐

PR #5394, opened 2021-05-24
Smart Answer Flow integration test approach mini-RFC

kevindew:
As per: https://github.com/alphagov/smart-answers/issues/5350 we have started work to form a proposal on an integration testing approach for Smart Answer Flows that is able to exercise their ERB files. The purpose of this quick RFC is to get early feedback and input on the approach to help work towards a relatively quick consensus of how to move forward. Please provide feedback by May 28th.

Desirable properties of these tests are:

- Fast to run
- Assertions at view level to prove their execution
- Method to achieve full (or at least very high) coverage of flow files
- Easy to understand (well relatively, given how hard Smart Answers can be to understand)
- Relatively low on tedium to write and repetition to read (ideally length of test flow file is within 2x the flow file length)
- Be consistent between flows

An early thing we learnt is that testing the flows through controller execution (either via Rack, Chrome or Rails request testing) is an order of magnitude slower than exercising the flow directly. A basic comparison is 50 times slower (`Finished in 0.199304s, 55.1921 runs/s` Rails request tests vs `Finished in 9.626176s, 1.2466 runs/s` comparable flow specific integration tests).

The code associated with this PR offers example implementations of the testing approach applied to existing Smart Answers. - with some variance in syntax. We've taken the approach of writing tests without needing them, or their helper code, to actually execute. This is an experiment in writing the code we want to be reading. We want to then test this approach on a subsection of Smart Answers to see where the warts/annoyances lie - suggestions of quirky scenarios to consider are very welcome.

## How to feedback

This is a relatively early stage proposal that we're opening to feedback to help validate our initial direction. Please try keep to all comments actionable so we can avoid this piece of work stalling. 

Things we're wondering are:

1. Is this node based testing approach a goer? Please don't consider assertion syntax as part of your feelings towards this.
2. What parts of the syntax suggestions seem idiomatic and usable? What parts could be improved - please provide suggestions for how they can be improved.
3. Any suggestions to improve testing terminology in contexts / should? We've struggled a bit to find the right verb to describe the transition between nodes.
4. Any thoughts on whether we routinely check the copy from a view in basic scenarios? (I'm slowly veering towards just establishing that the copy isn't empty with ability to check text which would be used to exercise view logic as it'd be very [tedious to assert some](https://github.com/alphagov/smart-answers/blob/a1f1a6ef84077a8ba67b8af45c620d295eaa296d/lib/smart_answer_flows/uk-benefits-abroad.rb#L688-L760))
4. Are we missing any common patterns that should be tested?
5. Any particular flows you recommend to try this out with? (we've purposely not considered the stranger ones of marriage abroad and maternity-paternity-calculator yet to not skew this towards the unconventional)

## Feedback wanted: Node based approach

The approach we are experimenting with breaks up a test of a Smart Answer Flow into its constituent nodes i.e for every node there is a context block for it and this context block should exhaustively cover the possibilities of that node. These tests exercise the integration between a flow, calculator and ERB files. Lower level concerns (controllers, app views) should be exercised separately.

We think this is an improved way to test a flow because it breaks it down into manageable parts. The existing approach to testing a flow seems to be based on journeys through the entire flow - as there can be many permutations of a journey it's very hard to know if or where logic can be tested (this approach can lead to flows of [significant context depths](https://github.com/alphagov/smart-answers/blob/a1f1a6ef84077a8ba67b8af45c620d295eaa296d/test/integration/smart_answer_flows/uk_benefits_abroad_test.rb#L2394-L2457))

## Input wanted: Syntax for setup and assertions

Flow tests share a lot of commonality in what they assert and on, thus we've been experimenting with how the assertions we write can be both brief yet understandable, while also maintaining conventions of the [shoulda gem](https://github.com/thoughtbot/shoulda#overview). We've put together drafts of our assertion ideas below, all of which may come out differently once the realities of implementation are applied. For example, we've been experimented with adding our own Shoulda matchers (`should our_matcher(arg)`), without knowing the process of creating these but imagining that if they were hard to do we could instead have methods of `should_our_matcher(arg)`.

We have struggled on the right way to phrase terminology in these and would welcome suggestions.

### Testing a start page

```ruby
  context "start page" do
    should see_start_page.with_title("Calculate holiday entitlement")
  end
```

We're suggesting a custom matcher see_start_page - that asserts the title is what we expect. Some alternative approaches could be

more verbose alternative:

```ruby
should "have a start page" do
  assert_equal "Calculate holiday entitlement", start_page_title
end
```

or using render as a key term instead of see:
```ruby
should render_start_page.with(title: "Calculate holiday entitlement")
```

implicitly checks that start page has anything for a title not necessarily this Smart Answer
```ruby
should render_start_page
```

### Basic question

For a question with nothing quirky we probably only need to test that it renders ok and can go to other destinations:

```ruby
context "question :question_name?" do
  setup do
    # set-up pre-requisite data to view this node
    response = { earlier_question?: "value" }
  end
  
  should see_question(:question_name?).with_title("Question title")
  
  should accept_answer("answer 1").with_next_node(:next_question?)
  should accept_answer("answer 2").with_next_node(:outcome)
end
```

More verbose alternative is:

```ruby
context "question :question_name?" do
  setup do
    response = { earlier_question?: "value" }
  end
  
  should "render the question" do
    assert_equal :question_name?, node_name
    assert_equal "Question title", question_title
  end

  should "continue to next question when given answer 1" do
    add_response "answer 1"
    assert_equal :next_question?, node_name
  end
  
  should "continue to outcome when given answer 2" do
    add_response "answer 2"
    assert_equal :next_question?, node_name
  end
end
```

or a different syntactical approach:

```ruby
context "question :question_name?" do
  setup do
    set_responses { earlier_question?: "value" }
  end
  
  should render_question(:question_name?) # just implicitly check it has a title that isn't empty
  should have_next_node(:next_question?).on_response("answer 1")
  
  # other ideas
  should reach_question(:next_question?).on_response("answer 1")
  should reach_outcome(:my_outcome).on_response("answer 1")
end
```

### Complex next steps

There can, of course, be more complex scenarios for the next steps of a Smart Answer that can't be neatly summarised with a one line. These can be approached with regular context and should blocks. For example:

```ruby
context "question :calculation period?" do
  # ...
  context "when the answer is full-year" do
    setup do
      add_response "full-year"
    end

    should "take people with irregular hours to the :irregular_and_annualised_done outcome" do
      responses[:basis_of_calculation?] = "irregular-hours"
      assert_equal :irregular_and_annualised_done, next_node_name
    end

    # ...
   end
 end
```

Some alternative syntax options could be:

```ruby
# set current response with a << operator for consistency with other responses operations
responses << "full-year"

# more explanatory way to replace existing responses
replace_responses { basis_of_calculation?: "irregular-hours" }

# asserting more specifically on a node
assert_next_node :irregular_and_annualised_done

# asserting specifically that we've reached an outcome (which could implicitly or explicitly test rendering if we wished)
assert_reached_outcome :irregular_and_annualised_done
```

### Input Validation

There's various minor validations in Smart Answer flows that should be tested. The basic proposal is to do these mostly with context blocks and perhaps using a custom assertion for repetitive ones.

```ruby
context "when given an invalid day of the week" do
  should_not accept_answer(0)
  should_not accept_answer(8)
end

should_not_accept_a_response_before(responses[:what_is_your_starting_date?])
```

Some alternative ideas:

```ruby

should "validate input is a day of the week" do
  assert_valid_response 1
  assert_invalid_response 0
  assert_invalid_response 8
end

```

### Outcomes

Outcomes are a bit tricky to decide what to do testing wise as they have such a low footprint the flow file and often they can have logicless views. 

Our initial approach is to assert each of them rendered a particular part of the view:

```ruby
  context "outcome :irregular_and_annualised_done" do
    setup do
      responses = {
        basis_of_calculation?: "annualised-hours",
        calculation_period?: "full-year",
      }
    end

    should see_outcome(:irregular_and_annualised_done).with_text("The statutory holiday entitlement is 5.6 weeks holiday.")
  end
```

I'm pondering if this type of testing is only really appropriate for views with logic in them as their ability to render any outcome can be asserted at a question level by checking that a response leads to an outcome and that outcome has some text. 

Testing outcomes could then be reserved for the checking of particular view aspects. For example:

```ruby
  context "outcome :results" do
    setup do
      responses = { question_1?: "answer 1", question_2?: "answer 2" }
    end

    context "when question_1? was answered with 'blah'" do
       setup { replace_responses question_1?: "blah" }
       should "have blah section" do
          assert_matches /blah/, outcome_body
       end
    end
  end
```

## Next

Once we've collated feedback and completed discussions on this we'll consider how to prototype a Smart Answer fully with a feedback informed approach. Thanks in advance for any comments.


leenagupte (commented):
## Node based approach

I really like the idea of Node based tests as it allows you to see all of the responses that bring you to a particular point in the flow. I'd expect that adding all of the responses needed to each test would make the test file quite long, but as the tests as assertions won't be nested, I'm not concerned about that.

I am concerned that testing for specific copy in the tests will make it harder for content designers to make updates to content, as they'll run the risk of breaking tests and more developer intervention will be required. I like the idea of testing that a particular partial is called and that it contains some content. I think the only exception to this should be in the partials that contain conditional logic. For example, in the `uk_benefits_abroad` we probably want to include tests for "bermuda", "mauritius", "canada" and an other country [1].

As for which flows to test this out on, I think at one of each type? i.e. url (traditional), session and query-string. I would however add "benefit cap calculator" to the "gnarly" list as some of the questions are asked in a loop and could be tricky to test.

## Syntax
As I mentioned before, I have concerns about explicitly testing the content, so I think my preferred option for the start page is to just assert that it has been rendered, i.e. `should render_start_page`.

I think for the basic questions, the "verbose" alternative is the clearest. I think terminology like "accept_answer" makes it feel like it's testing validations, and I'm not sure if we should do that in these tests. Should they just test the happy path? 

For the complex next steps, is it possible to continue to use the `add_response` syntax rather than adding the responses to a list?
e.g.

```
should "take people with irregular hours to the :irregular_and_annualised_done outcome" do
  add_response "irregular-hours" 	
  assert_equal :irregular_and_annualised_done, next_node_name
end
```

I'm not sure if assigning responses to a key will work for traditional smart-answer and do wonder if a separate approach is needed for session and querystring smart-answers. I don't know if it's possible to change how the responses are stored in the tests without changing how they're stored in the smart-answers. Is there a story/PR to store the answers to traditional path-bases smartanswers in a Hash? I think until I understand what's possible, I can't give an opinion on this approach to testing questions.


[1]: https://github.com/alphagov/smart-answers/blob/a1f1a6ef84077a8ba67b8af45c620d295eaa296d/lib/smart_answer_flows/uk-benefits-abroad/outcomes/bb_going_abroad_ss_outcome.erb

kevindew:
One thing I've just found is that there are flow unit tests defined here: https://github.com/alphagov/smart-answers/tree/main/test/unit/smart_answer_flows they seem very incomplete compared to the integration tests so I'd suggest as part of this work we remove those to only have the integration tests (similar to how we would for controller tests)

theseanything:
Thanks everyone who worked on putting this together! Tried to organise my comments inline with the different sections you’ve asked for feedback on.  

## Node based approach

I think this is getting us closer to writing more comprehensive tests. What would be good is to define exactly what parts of the codebase we’d be testing? E.g. “next_node blocks”, “on_response_blocks”, “validation” blocks. That’d make easier to understand if we are missing coverage on any code defined as part of the flow.  

If we’ve got a clear idea of what parts of the codebase we are testing, I wonder if we could aim to have these closer to unit tests? (Because we want them to be cheap and comprehensive). At the moment, the examples feel like they’re in-between integration and unit tests in terms of what code they cover. Which makes it difficult to understand how you’d write them. How do we decide what assertions we should have per “context block”?

The other benefit about framing these as unit tests is it’d push us to have a well defined API for flow/node definitions.

For example:
If we say we a “question node” can have the following defined for it: 	- next_node
	- validate
	- on_response
	- title
	- body
	- etc etc

It easy to understand we just need to write unit tests for those blocks.

```ruby
context "question: nation" do
  context "when response is england" do
    setup do
      response = { nation: "england" }
    end
      
    context "#on_response" do
      it "should set the calculator attribute" do
        assert_equal state.calculator.nation, "england"
      end
    end

    context "#next_node" do
      it "should return next node as results" do
          assert_equal :results, next_node
      end

    context “#title" do
      it "should return the title" do
          assert_equal "Which nation do you live in?", node.title
      end
    end
  end
end
```

(This is just verbose testing syntax for the sake of explaining what I mean - but could make it more concise)

> We think this is an improved way to test a flow because it breaks it down into manageable parts. The existing approach to testing a flow seems to be based on journeys through the entire flow - as there can be many permutations of a journey it's very hard to know if or where logic can be tested (this approach can lead to flows of significant context depths)

Agree, we need a cheap way to test lots of cases. 

## Syntax

Not too much to say. I like the concise nature of the shoulda matchers. As I mentioned above I'm concerned about understanding how exactly to write tests for different nodes and what assertions I should be including.

## Testing copy
I’d recommend to keep the initial scope of this work manageable, would be to ignore checking exactly what the copy is. Maybe just the presence is enough for now and would trigger the erb file to render at least. 


Sorry if I’ve missed anything - thought I’d try and leave some feedback before Friday.

ChrisBAshton:
I really like the proposal - nice one for putting this together 👏 

I agree with the comments above that particular bits of copy should not be present in the test, as it makes it harder to make content changes and is also somewhat arbitrary (which bits of copy to use, and how much of it).

As for syntax, I'd lean towards the terse, given the sheer amount of logical paths to be tested (the test files in the PR are already dauntingly long). I quite liked your example snippet here (NB, I've removed the question title copy, as suggested above):

```ruby
context "question :question_name?" do
  setup do
    # set-up pre-requisite data to view this node
    response = { earlier_question?: "value" }
  end
  
  should see_question(:question_name?)
  
  should accept_answer("answer 1").with_next_node(:next_question?)
  should accept_answer("answer 2").with_next_node(:outcome)
end
```

I think we could make more use of custom DSL to remove some redundancy. For example, the `should see_question(:question_name?)` is already implied in the `context`, and is kind of confusing as a `should` call, as it seems to be asserting that our test setup works as expected, rather than testing the node itself. I think something like this could work better:

```ruby
node :question_name? do
  set_previous_responses {
    earlier_question?: "value"
  }

  has_answer("answer 1").leading_to_next_question(:next_question?)
  has_answer("answer 2").leading_to_outcome(:outcome)
end
```

You've identified a gap in our testing, which is that some outcomes have conditional logic that should be tested. In these cases, testing copy seems the way forward (as suggested by others above). With our custom DSL, that might look something like this:

```ruby
node :outcome do
  context "happy path" do
    set_previous_responses {
      question_1?: "answer 1",
      question_2?: "answer 2",
    }

    assert_no_matches /special case/, outcome_body
  end

  context "with a particular answer" do
    set_previous_responses {
      question_1?: "answer 1",
      question_2?: "something special",
    }

    assert_matches /special case/, outcome_body
  end
end
```

This could be DRY'd up with a `replace_responses question_2?: "something special"` as you've proposed, depending on how many responses are needed to render the outcome, and therefore how much duplication there would be. But if a particular answer has no bearing on the outcome copy (e.g. `question_1?` in this case), I'd be inclined to omit it, in which case there'd be no duplication.

In summary: ✅ for node-based testing, with a couple of suggestions for how we might want to write the tests.

Rosa-Fox:
The node based approach is good due to it not being too difficult to follow. It is also the only option that has met the performance criteria so far. 

I agree with not testing for specific text, unless the text is rendered conditionally. 

In terms of which flows to test out on, a mixture of complexity seems good. People that have worked more on this could choose more complex flows and others could choose more basic flows to see how easy the approach is to adopt. Then they could also try a more complex flow. 

I think this syntax is easier to read for testing the start page: 

`should render_start_page.with(title: "Calculate holiday entitlement")` 

For the basic question syntax I prefer the more verbose option, even though it will be more time consuming to write, because ultimately it will be less overwhelming to read. If it caused the tests to run slower, that could be a reason to use another option, in which case I prefer the first example `(should see_question(:question_name?).with_title("Question title")`. 

For the complex_next_steps I like this assertion: 
`assert_next_node :irregular_and_annualised_done` as opposed to `assert_equal :irregular_and_annualised_done, next_node_name` but think that is still fine.
 
I think `should_not accept_answer(0)` works ok, at first glance I was a bit confused as I thought the number related to a question number in the flow, but the context does explain it... so I think that was more a problem with me than it lol.

> “I'm pondering if this type of testing is only really appropriate for views with logic in them as their ability to render any outcome can be asserted at a question level by checking that a response leads to an outcome and that outcome has some text.”

Is it enough to just check a couple of journeys or do we need to have all possible journeys covered? I suppose just having an outcome test for views with logic could be a good compromise, though I imagine this is still a lot.

kevindew:
Thank you everyone for the feedback, lots of great points raised and super useful to get all the perspectives. 🎉 

The main consensus seems to be that the node based approach sounds good and that there isn't a driver to always be testing view copy (presence of some is enough) - indeed there are concerns about the overhead of work maintaining view test copy could be.

I'll try respond to specific areas of feedback, I'm very happy to chat verbally with anyone further too.

kevindew:
@leenagupte 

> I think terminology like "accept_answer" makes it feel like it's testing validations, and I'm not sure if we should do that in these tests. Should they just test the happy path?

They should test off the happy path, as we don't intend for there to be any other tests. Currently there is test/integration/smart_answer_flows and test/unit/smart_answer_flows where, in theory, integration could be entirely happy path and unit off. However in practice there's very low use of those unit tests, and it'd be a total pain to test it all twice.

> I'm not sure if assigning responses to a key will work for traditional smart-answer and do wonder if a separate approach is needed for session and querystring smart-answers. I don't know if it's possible to change how the responses are stored in the tests without changing how they're stored in the smart-answers. Is there a story/PR to store the answers to traditional path-bases smartanswers in a Hash? I think until I understand what's possible, I can't give an opinion on this approach to testing questions.

This should be possible when https://github.com/alphagov/smart-answers/pull/5337 is merged. That abstracts the state handling to be distinct from the flow presentation.

At a push I can't imagine it'd be too big a deal to have different handling for traditional and response store flows. I think the main casualty would be readability as you have to try grok which answer in a collection refers to which question.

> As for which flows to test this out on, I think at one of each type? i.e. url (traditional), session and query-string. I would however add "benefit cap calculator" to the "gnarly" list as some of the questions are asked in a loop and could be tricky to test.

Thanks - that sounds a good suggestion. Hopefully we can soon be at a point where we don't need to worry about their type for testing. I didn't know benefit cap calculator had a loop that sounds interesting 😬 .

kevindew:
@theseanything

>  I think this is getting us closer to writing more comprehensive tests. What would be good is to define exactly what parts of the codebase we’d be testing? E.g. “next_node blocks”, “on_response_blocks”, “validation” blocks. That’d make easier to understand if we are missing coverage on any code defined as part of the flow.

I like the idea of context blocks for `next_node` and `validation` blocks and have applied them (see https://github.com/alphagov/smart-answers/pull/5394/commits/0bd83f0f4dfa81f20862b54fc7272f5d92f2039d). I'm not convinced `on_response` is essential given the calculator assignments should be implicitly tested, so would see it as something optional for scenarios where logic is involved.

> At the moment, the examples feel like they’re in-between integration and unit tests in terms of what code they cover. Which makes it difficult to understand how you’d write them. How do we decide what assertions we should have per “context block”?

It's a good point that they are in-between unit and integration tests. I think they have to fulfil both roles really - in a very similar manner to how a Rails controller is tested -  as having two levels of tests would be repetitive and tedious, and hasn't worked for the existing thing.

I think your suggestion about blocks helps make it clearer about the sections to test. I'd propose `next_node` and `validations` are the essential aspects to test (given they're user visible) with others being circumstantial. So that keeps these with a slightly higher emphasis on integration test than unit.

One thing this does raise is where these tests should they live while fulfilling both functions. I'm considering that it'd be better to have the flow in a directory test/flows outside of the test/integration and test/unit - and with that move the flows to app/flows since they seem an odd thing for lib as they are. But I'll step away from that if it's a divisive path.

kevindew:
@ChrisBAshton 

> I think we could make more use of custom DSL to remove some redundancy.

Those are some nice ideas about a custom DSL, thanks for sharing them. I'm a bit concerned about how that would stretch the learning curve for these tests so I'm inclined to not venture beyond the shoulda matchers. The nice thing about staying relatively safe is that it's always easy to break out into basic testing for odder scenarios.

> For example, the should see_question(:question_name?) is already implied in the context

I think it reads better as `render_question(:question_name?)` to assert that given the responses you see a question page. However I do think you raise an interesting point about whether checking seeing/rendering is necessary or not. I've had a bit of a think and considered `assert_next_node` could do a lot of the lifting (and it's kind of necessary to avoid some very https://github.com/alphagov/smart-answers/blob/a1f1a6ef84077a8ba67b8af45c620d295eaa296d/lib/smart_answer_flows/uk-benefits-abroad.rb#L688-L760) but I think it's best to verify individual questions render still. Without that, everything becomes a bit confusing to understand as you need to consider questions from a different context block.

kevindew:
@Rosa-Fox 

> For the basic question syntax I prefer the more verbose option, even though it will be more time consuming to write, because ultimately it will be less overwhelming to read. If it caused the tests to run slower, that could be a reason to use another option, in which case I prefer the first example (should see_question(:question_name?).with_title("Question title").

I had a go at writing a demo one with verbose syntax and it was seemed rather tedious and heavy on reading repetition. I think the key thing is to get a syntax that reads like a text description. I found myself writing `should "have a next node of :question_name? for a response of 'answer'"` a lot so felt that `should have_next_node(:question_name?).for_response("answer")` was a good way to more succinctly convey the same meaning. Does that work for you?

> I think should_not accept_answer(0) works ok, at first glance I was a bit confused as I thought the number related to a question number in the flow, but the context does explain it... so I think that was more a problem with me than it lol.

I wasn't super happy with that looking again. and have tried the more traditional:

```
      should "be invalid for a day of the week above 7" do
        add_response "0"
        assert_current_node_is_error
      end
```

since we probably have to describe them anyway.

For functions handling the same validation I thought       `should_reject_a_response_*` was maybe a nicer way than `should_not_accept_answer`.

> Is it enough to just check a couple of journeys or do we need to have all possible journeys covered? I suppose just having an outcome test for views with logic could be a good compromise, though I imagine this is still a lot.

It's probably best to think of it in terms of code paths rather than journeys. There could be thousands of journeys through a flow, but every question maybe only has a handful of code paths. So I'd suggest the aim is to explore all the code paths - just so we've got the basic assurance that a minor syntax error won't take out a Smart Answer. I think we only need to explore them modestly though, given the low amounts of logic and the relatively high probability that they'll be tweaked by non-devs.

kevindew:
As an update I spent yesterday afternoon prototyping whether and how this would actually work; I put together two Smart Answers in the flow: https://github.com/alphagov/smart-answers/commit/87cb4eb74729138b4c308fec1f240e782650e4e0

Notes I recorded from this experiment were:

Some learnings from putting together this work:

> It's quite easy to accidentally get a question name wrong. I've found it was useful to have something that verifies a question is valid before setting them.
> 
> It's frustrating when you're trying to test something but the set-up questions fails (for example you've missed a required response). I've found it beneficial to have error checks prior to assertions to confirm the expected state - to support this we specify the node under test in the setup.
> 
> It can be quite powerful to have rich assertion functions that check state while doing the assertions. This seems to lead to relatively easy to debug code (because of very specific errors) but they could prove a little magical. I think they should be ok just for the sheer amount of repetition across Smart Answers.
> 
> The shoulda matchers were pretty easy to put together and nice to write, however they have a pain point that they're hard to run in isolation when they fail (as they don't have a proper line number) so I gradually shied away from them and will look at removing them in the next iteration.
> 
> It's confusing having concepts of current and next for a flow state as they only have one set of responses - what is current in one instance is next in another. This is an pre-existing challenge and I couldn't think of how to resolve it without massive duplications so it seems like something to live with.

I thought I'd share as I know Leena in particular was knowing how it'd work and it represents a few changes. I realise though that the WIP nature makes this proposal a little tricky to follow so I'll try wrap that aspect up soon and eventually create an ADR that's more explicit.

Any feedback on the implementation is welcome.

GDSNewt:
Really cool stuff Kevin!

Feedback regarding the node based approach has already been articulated well, broadly speaking i'm in agreement with Leena and Rosa. I'll also echo what they said with regards not testing for specific text in the tests when possible.

Syntax wise, I prefer the conciseness that the shoulda matchers offer. The verbose may be more informative, but I find them much less readable. I'd also echo Chris' point that a more verbose approach makes the code intimidatingly long at times.

kevindew:
@leenagupte @theseanything @ChrisBAshton @Rosa-Fox @GDSNewt

Thank you for your feedback ⭐ . I'm now closing this PR to supersede it with https://github.com/alphagov/smart-answers/pull/5440 (to try make this proposal not super confusing) and it'd be great for you to look over that and feedback as appropriate.

PR #5401, opened 2021-05-26
Rename references of multiple choice to radio

theseanything:
Previously the Question::MultipleChoice node class was renamed to Radio to better reflect the type of question and reduce confusion between checkbox type questions. However, the associated presenter and other reference to multiple choice questions weren't updated. This makes all reference to state "radio" to prevent confusion and make the codebase consistent.


kevindew (approved):
This looks good. I've always misinterpreted MultipleChoiceQuestion as accepting multiple choices from the user 👍

PR #5439, opened 2021-06-16
Parallelize tests based on number of processors

kevindew:
Rather than hard coding the number of workers that this should run on
this option uses the number of cores on the executing machine to decide
the distribution. At the time of me setting the original option I didn't
know that this option was available.

For most GOV.UK users this shouldn't make much difference as GDS issued
Macbook Pro's have 6 physical coress. But this should be more friendly to
any configurations that have a differing number of cores.


leenagupte (approved):
Looked up `:number_of_processors` [1]. That's a nifty option. LGTM 👍 

[1]: https://api.rubyonrails.org/v6.0.2.1/classes/ActiveSupport/TestCase.html

PR #5444, opened 2021-06-16
Change penalty dates for late Self Assessment payment

brucebolt:
The dates being used to calculate the penalty were two days early.  Therefore we need to add another 2 days to each payment penalty step start date.

Also added a new set of tests that cover all tax years and all penalty steps to ensure there are no glitches in how the calculator works (e.g. in leap years).  These are based on the table provided by HMRC in https://govuk.zendesk.com/agent/tickets/4571228.

Note that changing the payment date in the tests also changes the interest due, as the payment is being made 2 days later in both cases, therefore a number of tests have needed to be updated.

[Trello card](https://trello.com/c/91XYwlNb)

### DIFF
--- a/test/integration/smart_answer_flows/estimate_self_assessment_penalties_test.rb
+++ b/test/integration/smart_answer_flows/estimate_self_assessment_penalties_test.rb
@@ -185,16 +185,16 @@ class EstimateSelfAssessmentPenaltiesTest < ActiveSupport::TestCase
     # #5% fine + 1000pounds(previous fine)(band 3), taxdue > 6002pounds
     context "5% fine + 1000pounds(previous fine)(band 3), taxdue > 6002pounds" do
       setup do
-        add_response "2015-08-01"
-        add_response "2015-08-01"
+        add_response "2015-08-03"
+        add_response "2015-08-03"
         add_response "10000.00"
       end
       should "show results" do
         assert_equal 1500, current_state.calculator.late_filing_penalty
         assert_equal 10_000, current_state.calculator.estimated_bill
-        assert_equal 161.16, current_state.calculator.interest
+        assert_equal 162.95, current_state.calculator.interest
         assert_equal 1000, current_state.calculator.late_payment_penalty
-        assert_equal 12_661, current_state.calculator.total_owed_plus_filing_penalty
+        assert_equal 12_662, current_state.calculator.total_owed_plus_filing_penalty
       end
     end
     # #band 4 case 1
@@ -214,16 +214,16 @@ class EstimateSelfAssessmentPenaltiesTest < ActiveSupport::TestCase
     # #10% fine + 1000pounds(previous fine)(band 4), taxdue > 6002pounds
     context "10% fine + 1000pounds(previous fine)(band 4), taxdue > 6002pounds" do
       setup do
-        add_response "2016-02-01"
-        add_response "2016-02-01"
+        add_response "2016-02-03"
+        add_response "2016-02-03"
         add_response "10000.00"
       end
       should "show results" do
         assert_equal 2000, current_state.calculator.late_filing_penalty
         assert_equal 10_000, current_state.calculator.estimated_bill
-        assert_equal 325, current_state.calculator.interest
+        assert_equal 326.78, current_state.calculator.interest
         assert_equal 1500, current_state.calculator.late_payment_penalty
-        assert_equal 13_825, current_state.calculator.total_owed_plus_filing_penalty
+        assert_equal 13_826, current_state.calculator.total_owed_plus_filing_penalty
       end
     end
   end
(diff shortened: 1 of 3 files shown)


test/integration/smart_answer_flows/estimate_self_assessment_penalties_test.rb
```diff
@@ -250,4 +250,138 @@ class EstimateSelfAssessmentPenaltiesTest < ActiveSupport::TestCase
...
+        setup do
+          add_response tax_year
+          add_response "online"
+          add_response Date.parse(payment_dates[:step_1]).prev_day
+          add_response Date.parse(payment_dates[:step_1]).prev_day
+          add_response payment_amount
+        end
+
+        should "not apply the step 1 penalty" do
+          assert_not_equal expected_penalty[:step_1], current_state.calculator.late_payment_penalty
```
ChrisBAshton, line 326: Should this be more specific, e.g. 0? Or `n - expected_penalty[:step_1]`?

test/integration/smart_answer_flows/estimate_self_assessment_penalties_test.rb
```diff
@@ -250,4 +250,138 @@ class EstimateSelfAssessmentPenaltiesTest < ActiveSupport::TestCase
...
+
+    penalty_dates.each do |tax_year, payment_dates|
+      context "for a step 1 payment date in #{tax_year} (#{payment_dates[:step_1]})" do
+        setup do
+          add_response tax_year
+          add_response "online"
+          add_response payment_dates[:step_1]
+          add_response payment_dates[:step_1]
+          add_response payment_amount
+        end
```
ChrisBAshton, line 309: I'd be tempted to move this `setup` block outside of the `context` to make it clearer what changes between each unit test. I.e.

```ruby
setup do
  add_response tax_year
  add_response "online"
  add_response payment_date
  add_response payment_date
  add_response payment_amount
end

context "for a step 1 payment date in #{tax_year} (#{payment_dates[:step_1]})" do
  let(:payment_date) { payment_dates[:step_1] }
  # ...
end

context "for the day before a step 1 payment date in #{tax_year} (#{payment_dates[:step_1]})" do
  let(:payment_date) { Date.parse(payment_dates[:step_1]).prev_day }
  # ...
end
```

test/integration/smart_answer_flows/estimate_self_assessment_penalties_test.rb
```diff
@@ -250,4 +250,138 @@ class EstimateSelfAssessmentPenaltiesTest < ActiveSupport::TestCase
...
+        step_2: "2019-08-03",
+        step_3: "2020-02-03",
+      },
+      "2018-19": {
+        step_1: "2020-03-02",
+        step_2: "2020-08-02",
+        step_3: "2021-02-02",
+      },
+      "2019-20": {
+        step_1: "2021-04-02",
```
ChrisBAshton, line 295: Is this April rather than March because of this?

> Penalty 1 = 1st late payment penalty (normally payment 30 days late, but not for 2019/20 due to extension of payment date due to Covid)

If so, maybe an inline comment explaining the difference for this one? Also not 100% sure where you got the date, as it appears to be missing from the Trello card.

ChrisBAshton (commented):
LGTM - just wanted to double-check one of the dates, and also a suggestion for DRYing up the tests in a way that doesn't compromise the nice readability they have already.

test/integration/smart_answer_flows/estimate_self_assessment_penalties_test.rb
brucebolt: The dates were copied from the Zendesk ticket.  There's a table in the message dated 20 May 16:17.  I'll add a comment.

test/integration/smart_answer_flows/estimate_self_assessment_penalties_test.rb
brucebolt: Good idea.  I'll change the tests to `assert_equal` to the previous step's penalty (or zero, in the case of the first example).

test/integration/smart_answer_flows/estimate_self_assessment_penalties_test.rb
brucebolt: It would be great if I could do this, but I don't think minitest allows us to use `let` to pass variables into the `setup` in the same way we can with Rspec.  I've tried several different things, but it doesn't look like `let` is a feature.

ChrisBAshton: Ah, I'd not spotted that this is Minitest. Fair enough!

ChrisBAshton (approved):

PR #5460, opened 2021-06-29
Fix Child Benefit Tax Calculator repeated questions failing

kevindew:
Trello: https://trello.com/c/tiNo3QKW/2597-some-users-unable-to-submit-child-benefit-tax-calculator

This has been put together to address an issue discovered off the back of https://github.com/alphagov/smart-answers/pull/5337 which breaks the ChildBenefitTaxCalculator Flow. The ChildBenefitTaxCalculator has a quirky feature where a user can be asked the questions about child benefit start and end dates multiple times, depending on the number of children a user has. This feature relied on there not being a mapping between a question and an answer, as
the same question would have multiple different answers. This unfortunately meant that it wasn't compatible with the recently
introduced response store approach to [Smart Answers](https://github.com/alphagov/smart-answers/blob/59753722534a1f7d4d361bcf57bc1fa69f8a484f/docs/arch/002-store-responses-in-query-string.md).

The options to resolve this were rather limited, as there is no functionality to dynamically define questions for a Smart Answer (nor does this seem simple as the object lifecycle of a Flow would need to be
reconsidered). The approach I took to resolve this was to utilise the maximum number of children to define a node per child. This allowed there to no longer be dynamic nodes. This is perhaps a bit of a cop-out as it may be worth investigating whether dynamic nodes should be a feature, however as this is broken in production it was best to be pragmatic.

To implement this I've also added the ability to specify the template_name of a question and I've changed the way we store responses in tests so that the tests behave more like actual Smart Answers do in production.


test/unit/smart_answer_flows/child_benefit_tax_calculator_flow_test.rb
```diff
@@ -152,19 +152,37 @@ class ChildBenefitTaxCalculatorFlowTest < ActiveSupport::TestCase
...
+          should "go to the corresponding add_child_benefit_x_stop? question" do
+            assert_equal :add_child_benefit_2_stop?, @new_state.current_node_name
+            assert_node_exists :add_child_benefit_2_stop?
           end
         end
       end
 
       # Q3c
-      context "when answering how_many_children_part_year? question" do
+      context "when answering child_benefit_1_stop? question" do
```
1pretz1, line 181: Think this should be `"when answering add_child_benefit_1_stop? question"`

test/unit/smart_answer_flows/child_benefit_tax_calculator_flow_test.rb
```diff
@@ -200,19 +218,51 @@ class ChildBenefitTaxCalculatorFlowTest < ActiveSupport::TestCase
         end
       end
 
+      context "when answering a later child_benefit_x_stop? question" do
```
1pretz1, line 221: Same comment as above

1pretz1 (approved):
Changes look good to me, thanks for putting this together @kevindew ⭐ I've tested this with some real examples and it passed with flying colours 🌈 

Approving as all my comments are minor. My only other comment would be that it'd be nice for readability if the changing of `child_index` to `child_number` was in a separate commit as I think it muddies the main logic change.

lib/smart_answer/node.rb
```diff
@@ -9,13 +9,19 @@ def initialize(flow, name, &block)
...
+      @template_name = filesystem_friendly_name
       instance_eval(&block) if block_given?
     end
 
     delegate :to_sym, to: :name
 
     delegate :to_s, to: :name
 
+    def template_name(template_name = nil)
+      @template_name = template_name unless template_name.nil?
```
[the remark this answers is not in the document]
kevindew: Yeah - we use this method as an accessor as well here: https://github.com/alphagov/smart-answers/pull/5460/commits/9db29bc68b95afe0c3e8cba6cf7fb3ade9babec5#diff-766d8248f7ca1c2987930a41bb2aaf2bcca8a0689cbcfc2ed38dfa9c311bf59fR12

It'd be a bit easier to understand if it was actually an `attr_accessor` but there's already quite a high degree of precedent for this pattern: https://github.com/alphagov/smart-answers/blob/59753722534a1f7d4d361bcf57bc1fa69f8a484f/lib/smart_answer/flow.rb#L27-L40

test/unit/smart_answer_flows/child_benefit_tax_calculator_flow_test.rb
kevindew: Oops thanks for this and the one below.

kevindew:
Thanks for the review @1pretz1, you're right about the separate commits - I was being a bit lazy, I've unpicked them now.

lib/smart_answer/node.rb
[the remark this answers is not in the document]
1pretz1: Ah okay, I missed that bit. Thanks for explaining!

1pretz1 (approved):
LGTM! 🍏

PR #5472, opened 2021-07-05
Add StartNode class

theseanything:
This PR helps position the codebase to make start pages as first class nodes. Changes in this PR include:
- Adds a StartNode class for the StartNodePresenter class, so it's consistent with other node types.
- Refactors the node presenter initialisation code, to help ensure we maintain this 1 to 1 relationship between node and node presenter classes.

### DIFF
--- a/lib/smart_answer/node.rb
+++ b/lib/smart_answer/node.rb
@@ -24,7 +24,7 @@ module SmartAnswer
       @template_name
     end
 
-    def presenter(flow_presenter, state)
+    def presenter(flow_presenter: nil, state: nil)
       self.class::PRESENTER_CLASS.new(self, flow_presenter, state)
     end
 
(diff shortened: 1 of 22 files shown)


lib/smart_answer/node.rb
```diff
@@ -22,6 +24,10 @@ def template_name(template_name = nil)
       @template_name
     end
 
+    def presenter(flow_presenter, state)
```
alex-ju, line 27: Is it worth defining default parameters instead of passing empty arguments? (e.g. `presenter(nil, nil)` → `presenter()`)
```
    def presenter(flow_presenter ||= nil, state ||= nil)
```

alex-ju (approved):

lib/smart_answer/node.rb
theseanything: Good shout! I've added a commit to change the method to use kwargs and default values

PR #5487, opened 2021-07-12
Fix SharedAdoptionMaternityPaternityFlow namespacing

kevindew:

Unlike the other Flows defined in the maternity-paternity-calculator this class has not been defined within the MaternityPaternityCalculatorFlow namespace.

This inconsistency has caused problems on related work I'm doing to move the flows to test/flow and interface nicer with the Zeitwerk autoloader.

This is a much smaller change than this diff may suggest. [View without whitespace being included](https://github.com/alphagov/smart-answers/pull/5487/files?diff=split&w=1) for an easier review.


gclssvglx (approved):
Yeah, def prefer the `diff` view.  Looks good.

PR #5489, opened 2021-07-12
Remove SmartAnswer namespace from Flow classes

kevindew:

Note, this isn't as scary a diff as it may appear. The vast majority of the changes making up the diff are changes in tab indentation and can be hidden by the "hide whitespace changes" option. [👉  Direct link](https://github.com/alphagov/smart-answers/pull/5489/files?diff=split&w=1).

This takes the first step of the journey towards moving flows from `lib/smart_answer_flows` to `app/flows` by removing the unconventional usage of the `SmartAnswer` namespace for classes that aren't defined in a `lib/smart_answer` directory.

Doing this has allowed me to take a look at the autoloading and I've configured these classes to be autoloaded by Zeitwerk. This has allowed for the removal of some quirky reload practices and manual require statements. More information in individual commits.

### DIFF
--- a/lib/smart_answer/flow_registry.rb
+++ b/lib/smart_answer/flow_registry.rb
@@ -58,7 +58,7 @@ module SmartAnswer
       else
         require @load_path.join(name)
       end
-      namespaced_class = "SmartAnswer::#{class_prefix}Flow".constantize
+      namespaced_class = "#{class_prefix}Flow".constantize
       namespaced_class.build
     end
 
--- a/lib/smart_answer/flow_registry.rb
+++ b/lib/smart_answer/flow_registry.rb
@@ -53,11 +53,6 @@ module SmartAnswer
 
     def build_flow(name)
       class_prefix = name.tr("-", "_").camelize
-      if Rails.env.development?
-        load @load_path.join("#{name}.rb")
-      else
-        require @load_path.join(name)
-      end
       namespaced_class = "#{class_prefix}Flow".constantize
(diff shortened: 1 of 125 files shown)


lib/smart_answer/flow_registry.rb
```diff
@@ -53,12 +53,7 @@ def available?(name)
 
     def build_flow(name)
       class_prefix = name.tr("-", "_").camelize
-      if Rails.env.development?
-        load @load_path.join("#{name}.rb")
-      else
-        require @load_path.join(name)
-      end
```
theseanything, line 60: Will this still support automatic reloading of flows during development?

kevindew: Yup, they just don't need custom code for it as Zeitwerk takes care of it as part of Rails class reloading

It also fixes a problem that [any classes that use the `append` method](https://github.com/alphagov/smart-answers/blob/e36f68bedac1e35007f6e07dbbb7b2da95494be7/lib/smart_answer_flows/maternity-paternity-calculator.rb#L43) stop working when automatic reloading occurs and you get an error such as: `uninitialized constant SmartAnswer::MaternityPaternityCalculatorFlow::AdoptionCalculatorFlow`

gclssvglx (approved):
😱  125 changed files!!!  You're right once you view without whitespace changes - it's much better 👍   Actual changes all look great and make the code so much nicer.  An amazing bit of code cleanup 🧹  thanks for this.

PR #5506, opened 2021-07-20
Allow additional parameters to pass through flow

theseanything:
This adds the ability to pass permitted query parameters from the start page to rest of the flow nodes.

This functionality is needed by the "Next steps for you business" flow, which needs to use the "ct" and "crn" query parameters as filtering parameters on the results page.

To enable this functionality, this PR makes the following changes:
- Adds a new QueryParametersResponseStore subclass store encapsulate behaviour for query_parameter flows
- The QueryParametersResponseStore will store all parameters that are question node name or specified by additional parameters attribute in the flow definition
- Refactor forwarding parameters to be a method in the response store, rather than state.
- Changes the behaviour of the start page to only delete user submitted responses when starting a flow. So the additional parameters pass to the flow and not cleared.

### DIFF
--- a/app/presenters/flow_presenter.rb
+++ b/app/presenters/flow_presenter.rb
@@ -14,7 +14,7 @@ class FlowPresenter
 
   delegate :title, :meta_description, to: :start_node
   delegate :node_slug, to: :current_node
-  delegate :accepted_responses, :forwarding_responses, :current_response, to: :state
+  delegate :accepted_responses, :current_response, to: :state
 
   def initialize(flow, state)
     @flow = flow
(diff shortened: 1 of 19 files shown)


app/presenters/flow_presenter.rb
```diff
@@ -42,7 +42,7 @@ def start_node
     @start_node ||= @flow.start_node.presenter(flow_presenter: self)
   end
 
-  def change_answer_link(question)
+  def change_answer_link(question, forwarding_responses)
```
alex-ju, line 45:
```rb
def change_answer_link(question, forwarding_responses = {} )
```
non-blocker: I guess we could've made the second argument optional by setting an empty object as default.

alex-ju (approved):
Looks good! 👏

PR #5504, opened 2021-07-19
Rewrite MarriageAbroad flow test

kevindew:
Trello: https://trello.com/c/fU4RQYvF/494-foundational-work-to-implement-smart-answer-flow-test-adr

This ports the Marriage Abroad Flow tests to the style proposed in [1]
and was done as part of the initial experimentation work [2]. It then
deletes the previous integration and unit tests.

While creating this I discovered that, by the old tests not rendering
the views, that there was a missing stub for Sweden organisation which
I've added in.

I did do some optimisations to speed up these new tests. I found that by
stubbing only the worldwide API for the country under test, that there
was a significant increase in performance.

Before:

```
Finished in 210.360104s, 16.6001 runs/s, 49.8193 assertions/s.
3492 runs, 10480 assertions, 0 failures, 0 errors, 0 skips
```

After:

```
Finished in 39.723775s, 88.0077 runs/s, 176.4938 assertions/s.
3496 runs, 7011 assertions, 0 failures, 0 errors, 0 skips
```

There is still a bunch of low hanging fruit that could increase the
performance of these tests. There are (somewhat depressingly) 1,275 near
identical view files with the copy: 'It’s not possible to get legal
recognition for your same-sex relationship in $country_name', these
could be converted to a generic partial to save perhaps a third of the
tests ran here from needing to be run. I imagine there are various
optimisations available for countries with near identical copy.

[1]: https://github.com/alphagov/smart-answers/pull/5459
[2]: https://github.com/alphagov/smart-answers/pull/5440

⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

PR #5507, opened 2021-07-21
Add new question and outcome to marriage abroad flow

beccapearce:
Japan Vietnam and Panama offer consular opposite sex civil partnerships.

To provide this information in smart-answers this PR:
- Adds new question whether couples want to get married or register a
civil partnership
- Adds new outcomes for opposite sex civil partnerships for these countries

The `opposite_sex_civil_partnerships` outcomes for Panama and Vietnam (previously countries_with_18_outcomes) have been placed in the path `outcomes/panama/_opposite_sex_civil_partnership`. This differs from the typical pattern for outcomes in the marriage-abroad smart-answer. 

We deviated from the typical pattern since it would add around 9 outcomes for countries that previously has 18 outcomes. Since all of these outcomes would be the same it would create a large amount of duplication and make it more difficult to keep these files consistent with each other.

Trello card: 
https://trello.com/c/b6vvMECA/2600-2-add-new-question-and-outcome-to-japan-panama-and-vietnam-marriage-abroad-about-opposite-sex-civil-partnership

### DIFF
--- a/config/smart_answers/marriage_abroad_data.yml
+++ b/config/smart_answers/marriage_abroad_data.yml
@@ -130,7 +130,6 @@ countries_with_18_outcomes:
   - oman
   - pakistan
   - palau
-  - panama
   - papua-new-guinea
   - paraguay
   - peru
@@ -185,10 +184,12 @@ countries_with_18_outcomes:
   - uzbekistan
   - vanuatu
   - venezuela
-  - vietnam
   - western-sahara
   - yemen
   - zambia
+countries_with_19_outcomes:
+  - panama
+  - vietnam
 countries_with_2_outcomes:
   - australia
   - croatia
@@ -197,7 +198,6 @@ countries_with_2_outcomes:
   - france
   - india
   - ireland
-  - japan
   - luxembourg
   - philippines
   - south-africa
@@ -207,6 +207,8 @@ countries_with_2_outcomes:
   - turkey
   - usa
   - zimbabwe
+countries_with_3_outcomes:
+  - japan
 countries_with_2_outcomes_marriage_or_pacs:
   - monaco
   - new-caledonia
--- a/app/flows/marriage_abroad_flow.rb
+++ b/app/flows/marriage_abroad_flow.rb
@@ -97,10 +97,30 @@ class MarriageAbroadFlow < SmartAnswer::Flow
         calculator.sex_of_your_partner = response
       end
 
-      next_node do
-        if calculator.has_outcome_per_path?
+      next_node do |response|
+        if calculator.offers_consular_opposite_sex_civil_partnership? && response == "opposite_sex"
+          if calculator.ceremony_country == "vietnam" && calculator.partner_nationality == "partner_local"
+            outcome :outcome_marriage_abroad_in_country
+          else
+            question :marriage_or_civil_partnership?
+          end
+        elsif calculator.has_outcome_per_path?
           outcome :outcome_marriage_abroad_in_country
         end
+      end
+    end
+
+    # Q6
+    radio :marriage_or_civil_partnership? do
+      option :marriage
+      option :civil_partnership
+
+      on_response do |response|
+        calculator.type_of_ceremony = response
+      end
+
+      next_node do
+        outcome :outcome_marriage_abroad_in_country
       end
     end
 
--- a/app/flows/marriage_abroad_flow/outcomes/countries/japan/_opposite_sex_civil_partnership.erb
+++ b/app/flows/marriage_abroad_flow/outcomes/countries/japan/_opposite_sex_civil_partnership.erb
@@ -1,0 +1,64 @@
+You can register an opposite sex civil partnership at the British embassy in Tokyo under UK law (except for Scotland).
+
+You'll be asked to choose what part of the UK you want to register your civil partnership under. You can choose England, Wales or Northern Ireland. You cannot register it under Scotland.
+
+If you and your partner do not live in Japan, you need to have been in Japan for at least 21 days before you can register your civil partnership. You need to stay in Japan for the 21 days.
+
+^Check the [travel advice for Japan](/foreign-travel-advice/japan) before making any plans.
+
+##Prove you’re free to register a civil partnership
+
+You and your partner need to give notice and sign a declaration that you’re legally allowed to register a civil partnership.
+
+You need to have been in Japan for at least 7 full days before you can give notice. For example, if you arrived on Monday you cannot give notice until the following Tuesday.
+
+###Give notice of your civil partnership
+
+[Make an appointment at the embassy in Tokyo](https://www.book-consular-appointment.service.gov.uk/TimeSelection?location=152&service=6) to give notice of your civil partnership and set a date for your ceremony.
+
+It costs £50 to give notice. You can pay in the [local currency](/government/publications/japan-consular-fees) by card or in cash.
+
+You need to bring:
+
+- your and your partner’s passports
+- proof that you’ve been in Japan for at least 7 full days, for example a utility bill, bank statement, passport stamp or boarding pass
+
+If you or your partner have been married or in a civil partnership before, you’ll also need:
+
+- your [original decree absolute or final order](/copy-decree-absolute-final-order)  (this could be a printed PDF and covering email from the court) - if you’re divorced
+- your annulment certificate - if your marriage or civil partnership was annulled
+- your civil partnership dissolution - if your civil partnership was dissolved
+- your partner’s [death certificate](/order-copy-birth-death-marriage-certificate) and marriage certificate - if you’ve been widowed
+
+If your divorce, civil partnership dissolution or annulment took place outside the UK, you’ll need evidence that you or your ex partner lived in or were a national of that country at the time.
+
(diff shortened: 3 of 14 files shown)


kevindew:
Hey @beccapearce, just a heads up that this PR may get caught up in some churn and change in Smart Answers related to https://github.com/alphagov/smart-answers/blob/main/docs/arch/003-testing-flows.md - I've got a [PR up that rewrites Marriage Abroad tests](https://github.com/alphagov/smart-answers/pull/5504) at the moment.

Don't let this stop you merging if you work is ready before any of that stuff happens, but if you do end up caught with conflicts introduced by those changes do deal with feel free to reach out to me to help you get it sorted.

beccapearce:
Cool! Thanks for the heads up @kevindew !

benjamineskola:
There was some [feedback from Content](https://trello.com/c/b6vvMECA/2600-2-add-new-question-and-outcome-to-japan-panama-and-vietnam-marriage-abroad-about-opposite-sex-civil-partnership#comment-60f988edf865965f7f03d711) that most of the outcomes are located in subdirectories of [`lib/smart_answer_flows/marriage-abroad/outcomes/countries`](https://github.com/alphagov/smart-answers/tree/add-question-to-married-abroad/lib/smart_answer_flows/marriage-abroad/outcomes/countries/japan), whereas these new outcomes are separate which makes them harder to find and update.

I think [this](https://github.com/alphagov/smart-answers/blob/add-question-to-married-abroad/lib/marriage_abroad_outcome_flattener.rb#L64) is the method that handles that, but altering it might be complicated because it seems like it currently assumes that civil partnerships are implicitly same-sex.

ChrisBAshton:
Following on from Ben's comment earlier, I thought I'd review and see if the outcomes are in the place the content designer would expect.

Looking at the [app/flows/marriage_abroad_flow/outcomes/countries](https://github.com/alphagov/smart-answers/tree/main/app/flows/marriage_abroad_flow/outcomes/countries) directory and picking any country at random, we tend to have three subfolders: `ceremony_country`, `partner_local` and `partner_other`. It looks as though the flow is chosen based on the answer to "Where do you live?". Let's take [Vietnam](https://www.gov.uk/marriage-abroad/y/vietnam) as an example:

- Choosing "Vietnam" would trigger the `ceremony_country` flow
- Choosing "UK" would trigger the `uk` flow
- Choosing "Elsewhere" would trigger the `third_country` flow

The logic changes in the card seem only to apply to outcomes in the ceremony country (i.e. "THEN go to new outcome about opposite sex civil partnership **in Vietnam** (see later in doc)"), so I'd imagine our new outcomes should go in the `ceremony_country` folder.

In that folder, there are more subdirectories: `partner_british`, `partner_local` and `partner_other`. This seems to correlate to the "[What is your partner’s nationality?](https://www.gov.uk/marriage-abroad/y/vietnam/ceremony_country)" question.

For Japan and Panama, the answer to that question doesn't affect the outcome for opposite sex civil partnerships, but for Vietnam, the logic should only apply `IF ‘British’ OR ‘National of another country’`.

So my conclusion would be that we should probably aim for this sort of file structure (my, this calculator is complicated!):

```
japan/
  ceremony_country/
    partner_british/
      _opposite_sex_civil_partnership.erb
    partner_local/
      _opposite_sex_civil_partnership.erb
    partner_other/
      _opposite_sex_civil_partnership.erb
panama/
  ceremony_country/
    partner_british/
      _opposite_sex_civil_partnership.erb
    partner_local/
      _opposite_sex_civil_partnership.erb
    partner_other/
      _opposite_sex_civil_partnership.erb
vietnam/
  ceremony_country/
    partner_british/
      _opposite_sex_civil_partnership.erb
    partner_local/
      # no new file here, as opposite sex civil partnerships do not apply when the partner is Vietnamese
    partner_other/
      _opposite_sex_civil_partnership.erb
```

For each country, the `_opposite_sex_civil_partnership.erb` file should be duplicated across the subdirectories, like we already do for the other outcomes (e.g. [1](https://github.com/alphagov/smart-answers/blob/0b785e830268ecd9d14eb9fd1184e94a919f3135/app/flows/marriage_abroad_flow/outcomes/countries/vietnam/ceremony_country/partner_british/_opposite_sex.erb) [2](https://github.com/alphagov/smart-answers/blob/0b785e830268ecd9d14eb9fd1184e94a919f3135/app/flows/marriage_abroad_flow/outcomes/countries/vietnam/ceremony_country/partner_local/_opposite_sex.erb) [3](https://github.com/alphagov/smart-answers/blob/0b785e830268ecd9d14eb9fd1184e94a919f3135/app/flows/marriage_abroad_flow/outcomes/countries/vietnam/ceremony_country/partner_other/_opposite_sex.erb)). This gives the content designer the freedom to make granular changes for very specific outcomes (even if it does look like quite a maintenance nightmare!).

Does that make sense? Is there a reason I've missed as to why we should avoid the outcome duplication?

beccapearce:
This one of the options I considered when moving the outcomes. I wasn't sure what the best approach would be so I asked the team and the general consensus was that adding around 18 outcomes for both Vietnam and Panama (Japan has a different structure) would turn into a bit of a nightmare for updating the content. I also suggested using a partial but we thought that might get confusing. So this was seen as a way to avoid those issues.

config/smart_answers/marriage_abroad_data.yml
```diff
@@ -185,10 +184,12 @@ countries_with_18_outcomes:
   - uzbekistan
   - vanuatu
   - venezuela
-  - vietnam
   - western-sahara
   - yemen
   - zambia
+countries_with_18_outcomes_consular_civil_partnerships:
```
ChrisBAshton, line 190: This threw me a bit - I understand the 18 outcomes refer to the `_opposite_sex`/`_same_sex` files in each of the `partner_british`/`partner_local`/`partner_other` directories in each of the `ceremony_country`/`third_country`/`uk` directories (2*3*3=18).

We're now adding 3 more outcomes for Japan and Panama, and 2 more outcomes for Vietnam, [following the logic](https://github.com/alphagov/smart-answers/pull/5507#issuecomment-897416005). Should this particular subset be renamed `countries_with_21_outcomes`?

ChrisBAshton (commented):
Thanks @beccapearce - I vaguely remember that Slack conversation now 👍 probably worth explaining in the commit message why we've gone a different route. I do think the new outcomes should be moved down one level, though, into `ceremony_country`, as the outcomes only apply in that flow.

config/smart_answers/marriage_abroad_data.yml
beccapearce: > Thanks @beccapearce - I vaguely remember that Slack conversation now 👍 probably worth explaining in the commit message why we've gone a different route. I do think the new outcomes should be moved down one level, though, into `ceremony_country`, as the outcomes only apply in that flow.

Hm, I think if we move the outcome down a level then we need to replicate it across the other levels as well. I believe this is an outcome for `ceremony_country` as well as `third country` and `partner_local` that would be 9 places to add the outcome for panama, 6 for vietnam and then the one place for japan.

beccapearce:
Thanks @beccapearce - I vaguely remember that Slack conversation now 👍 probably worth explaining in the commit message why we've gone a different route. I do think the new outcomes should be moved down one level, though, into ceremony_country, as the outcomes only apply in that flow.

Hm, I think if we move the outcome down a level then we need to replicate it across the other levels as well. I believe this is an outcome for ceremony_country as well as third country and partner_local that would be 9 places to add the outcome for panama, 6 for vietnam and then the one place for japan.

config/smart_answers/marriage_abroad_data.yml
beccapearce: Would it be ok to name them countries_with_21_outcomes? I think Panama would have 27 and Vietnam would have 24 (if I counted correctly)

ChrisBAshton:
> Hm, I think if we move the outcome down a level then we need to replicate it across the other levels as well. I believe this is an outcome for ceremony_country as well as third country and partner_local that would be 9 places to add the outcome for panama, 6 for vietnam and then the one place for japan.

Apologies, @beccapearce. As discussed on Slack, I'd misunderstood this and thought the outcome would only apply to people answering "Vietnam"/"Japan"/"Panama" to the "Where do you live?" question.

config/smart_answers/marriage_abroad_data.yml
ChrisBAshton: Perhaps `countries_with_multiple_outcomes_including_consular_civil_partnerships`?

beccapearce: Hmmm if we keep the file structure as it is currently then we can do 19 outcomes. It's only a problem if we move the outcome files within the subfolders

ChrisBAshton: That's true. Let's do that 👍  `countries_with_19_outcomes`?

ChrisBAshton (approved):
Great job 👏  this is not a simple smart answer!

I do think it would be worth editing the commit message to explain why you've gone a different route to the "duplicate all the outcomes" approach used in the other flows, but there's also enough context in this PR that people can dig out that history if they need to.

PR #5508, opened 2021-07-21
Rename the flow files to regular Ruby conventions

kevindew:

This PR may look scary but it doesn't actually do a lot. It essentially only does 3 things:

- it renames all the flow classes from kebab-case names that don't match Ruby conventions, to snake_case that does. For example: AdditionalCommodityCodeFlow has moved from `lib/smart_answer_flows/additional-commodity-code.rb` to `lib/smart_answer_flows/additional_commodity_code_flow.rb`
- it renames the directory within `lib/smart_answer_flows` to match the changed class names. These directories are also used for class namespacing (as per [MaternityPaternityCalculatorFlow](https://github.com/alphagov/smart-answers/tree/main/lib/smart_answer_flows/maternity-paternity-calculator))
- it fixes any configuration and problems caused by this renaming.

It isn't anywhere near as scary as it looks, I promise. It has been done as part of implementing [ADR-3](https://github.com/alphagov/smart-answers/blob/main/docs/arch/003-testing-flows.md) and is the last change to lib/smart_answer_flows before they are moved to app/flows.

To review this: 

- Go through commit by commit - only the second one is scary, with the third a bit of a pain, most are tiny.
- For the second commit you don't actually need to review the changes - you can verify this is done correctly by the small change in https://github.com/alphagov/smart-answers/commit/8dc94ae144afdee625eb32ef82b0b42986a2519e and the fact that all Smart Answers can render a start page.


gclssvglx (approved):
Thanks @kevindew 👍  nice shell script skills!  I've not reviewed all ~4k files of course, but the 'real' changes all look great.

PR #5514, opened 2021-07-22
Rewrite Benefits Cap Calculator flow test

kevindew:

This converts the Benefits Cap Calculator flow tests to use the node based
testing approach introduced in [ADR-3][1].

This flow is somewhat unique in that it has a dynamic question system - which
needed considering in how to test it.

In putting this together there were some decisions to make about what
outcome logic to test as there is a setting up of values in the
calculator that is then output on the outcome. I went quite modest here
by just testing the benefit calculation logic. I could be persuaded that
it would be good to test the figures are outcome, but these are a little
annoying to test as the same thing is output on four different outcomes.


test/flows/benefit_cap_calculator_flow_test.rb
```diff
@@ -0,0 +1,311 @@
...
+      assert_rendered_outcome text: "Your Housing Benefit will be: £20"
+    end
+
+    should "reduce housing benefit amount to a minimum of 50p" do
+      cap = benefit_cap(region: :london)
+      add_responses jsa_amount?: cap.to_s,
+                    housing_benefit_amount?: cap.to_s
+      assert_rendered_outcome
+      assert_match "Your Housing Benefit will be: 50p", @test_flow.outcome_body_text
+      assert_match "Your Housing Benefit won’t be reduced to less than 50p", @test_flow.outcome_body_text
```
leenagupte, line 252: It's great that these values are being tested, especially as I can't see that they're being tested anywhere else.

Is that the case for the other values. e.g. `total_over_cap`. Is that value being tested anywhere else? If not, perhaps it and the others could be tested for this outcome? I wouldn't be expect them to be duplicated in the other test. 

From a quick look, I think there are 4 values in total that are calculated in the flow for use in the outcome:

`benefit_cap`
`housing_benefit_amount`
`total_benefits_amount`
`total_over_cap`
 
It would be great if one of the lines that each appears on could be tested.

test/flows/benefit_cap_calculator_flow_test.rb
```diff
@@ -0,0 +1,311 @@
...
+      testing_node :outcome_not_affected_exemptions
+      add_responses receive_housing_benefit?: "yes",
+                    working_tax_credit?: "no",
+                    receiving_exemption_benefits?: %w[carers_allowance war_pensions]
+    end
+
+    should "render the exemption benefits" do
+      assert_rendered_outcome
+      assert_match "Carer's Allowance", @test_flow.outcome_body_text
+      assert_match "War pensions", @test_flow.outcome_body_text
```
leenagupte, line 296: Nice!

leenagupte (commented):
The Benefit Cap Calculator is one of the really gnarly calculators, so well done on figuring out a readable way to test it. 🎉 

I've left a couple of inline suggestions.

test/flows/benefit_cap_calculator_flow_test.rb
```diff
@@ -0,0 +1,311 @@
...
+                      receiving_exemption_benefits?: [],
+                      receiving_non_exemption_benefits?: [benefit]
+      end
+
+      should "render question" do
+        assert_rendered_question
+      end
+
+      context "next_node" do
+        should "have a next node of housing_benefit_amount? if there aren't other benfits remaining" do
```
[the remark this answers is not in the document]
kevindew: Yes, your understanding is correct. I can see how that wasn't easy to work out!

kevindew:
In light of https://github.com/alphagov/smart-answers/pull/5527 I'm going to close this PR to pause work on this. We can resume work if that PR drags on for a while.

Thanks very much for the review @leenagupte, the comments sound good points.

PR #5516, opened 2021-07-23
Convert Business Coronavirus Support Finder flow tests to use node-based testing

leenagupte:
Trello: https://trello.com/c/TXtphN02

# What's changed and why?

Converts the Business Coronavirus Support Finder flow tests to use the
node based testing approach introduced in [ADR-3][1].

Also removes the existing RSpec feature tests as they are no longer
needed.


test/flows/business_coronavirus_support_finder_flow_test.rb
```diff
@@ -0,0 +1,260 @@
...
+    end
+
+    should "render the question" do
+      assert_rendered_question
+    end
+
+    context "next_node" do
+      %w[nurseries retail_hospitality_or_leisure nightclubs_or_adult_entertainment personal_care none].each do |response|
+        should "have a next node of closed_by_restrictions? for a #{response} response" do
+          assert_next_node :closed_by_restrictions?, for_response: [response]
```
Rosa-Fox, line 128:
```suggestion
          assert_next_node :closed_by_restrictions?, for_response: response
```

test/flows/business_coronavirus_support_finder_flow_test.rb
```diff
@@ -0,0 +1,260 @@
...
+        end
+      end
+
+      should "have a next node of closed_by_restrictions? for a multiple responses" do
+        assert_next_node :closed_by_restrictions?, for_response: %w[nightclubs_or_adult_entertainment personal_care]
+      end
+    end
+  end
+
+  context "question: sectors?" do
```
Rosa-Fox, line 138:
```suggestion
  context "question: closed_by_restrictions?" do
```

test/flows/business_coronavirus_support_finder_flow_test.rb
```diff
@@ -0,0 +1,260 @@
...
+    end
+
+    should "render the question" do
+      assert_rendered_question
+    end
+
+    context "next_node" do
+      %w[local_1 local_2 national none].each do |response|
+        should "have a next node of results for a #{response} response" do
+          assert_next_node :results, for_response: [response]
```
Rosa-Fox, line 156:
```suggestion
          assert_next_node :results, for_response: response
```

test/flows/business_coronavirus_support_finder_flow_test.rb
```diff
@@ -0,0 +1,260 @@
...
+    end
+
+    should "render vat_reduction if sectors include retail_hospitality_or_leisure" do
+      add_responses sectors?: %w[retail_hospitality_or_leisure]
+
+      assert_rendered_outcome text: "VAT reduction for hospitality, accommodation and attractions"
+    end
+
+    should "render traineeships if business_based is england" do
+      add_responses sectors?: %w[retail_hospitality_or_leisure]
```
Rosa-Fox, line 201: I don't think we need `add_responses sectors?: %w[retail_hospitality_or_leisure]`?

test/flows/business_coronavirus_support_finder_flow_test.rb
```diff
@@ -0,0 +1,260 @@
...
+
+      assert_rendered_outcome text: "Traineeships"
+    end
+
+    should "render apprenticeships if business_based is england" do
+      add_responses sectors?: %w[retail_hospitality_or_leisure]
```
Rosa-Fox, line 207: I don't think we need `add_responses sectors?: %w[retail_hospitality_or_leisure]`?

test/flows/business_coronavirus_support_finder_flow_test.rb
```diff
@@ -0,0 +1,260 @@
...
+
+      assert_rendered_outcome text: "Apprenticeships"
+    end
+
+    should "render tlevels if business_based is england" do
+      add_responses sectors?: %w[retail_hospitality_or_leisure]
```
Rosa-Fox, line 213: I don't think we need `add_responses sectors?: %w[retail_hospitality_or_leisure]`?

test/flows/business_coronavirus_support_finder_flow_test.rb
```diff
@@ -0,0 +1,260 @@
...
+    end
+
+    should "render council_grants if sector retail_hospitality_or_leisure" do
+      add_responses sectors?: %w[retail_hospitality_or_leisure]
+
+      assert_rendered_outcome text: "Additional schemes and grants from your local council"
+    end
+
+    should "render additional_restrictions_grant if business_based is england" do
+      add_responses sectors?: %w[retail_hospitality_or_leisure]
```
Rosa-Fox, line 225: I don't think we need `add_responses sectors?: %w[retail_hospitality_or_leisure]`?

Rosa-Fox (changes requested):
Nice! I made a few small comments.

test/flows/business_coronavirus_support_finder_flow_test.rb
leenagupte: `sectors?` is a checkbox question. If more than one answer is selected, they are sent as an array, so whilst a string response for a single answer does work, I thought it would be more confusing for one test for `sectors?` to pass in a string and the other one below to pass an array.

test/flows/business_coronavirus_support_finder_flow_test.rb
leenagupte: 🤦 Good spot!

test/flows/business_coronavirus_support_finder_flow_test.rb
leenagupte: It's the same here as for `sectors?`. `closed_by_restrictions?` is a checkbox question, so I choose to pass the responses as an array to be consistent with the other test in this context.

leenagupte: Would it be better to combine the tests? i.e. rather than have two tests, one for a single response, and another for multiple responses, would it be better, to just have one for all responses? e.g.

```
should "have a next node of results for all responses" do
  assert_next_node :results, for_response: %w[local_1 local_2 national none]
end
```

I could do the same for `sectors?`.

test/flows/business_coronavirus_support_finder_flow_test.rb
leenagupte: Good spot! I'll remove `retail_hospitality_or_leisure` from the tests below as well.

leenagupte:
@Rosa-Fox Thank you for your review. I've responded to your comments and fixed up any changes.

test/flows/business_coronavirus_support_finder_flow_test.rb
Rosa-Fox: @leenagupte yeah that makes sense!

test/flows/business_coronavirus_support_finder_flow_test.rb
Rosa-Fox: @leenagupte yeah I think that would be better! It would assume that if multiple selection works then single selection should too... which I think is fine.

test/flows/business_coronavirus_support_finder_flow_test.rb
Rosa-Fox: 👍

leenagupte:
@Rosa-Fox I've changed the tests for the checkbox questions to have one test for all responses, and one test for a "none" response. I've checked the UI, and user cannot select "none" and any of the other options at the the same time, so I've matched that behaviour in the tests.

test/flows/business_coronavirus_support_finder_flow_test.rb
```diff
@@ -0,0 +1,255 @@
...
+
+  context "question: business_based?" do
+    setup { testing_node :business_based? }
+
+    should "render the question" do
+      assert_rendered_question
+    end
+
+    context "next_node" do
+      %w[england scotland wales northern_ireland].each do |response|
```
kevindew, line 21: As per the review on FindCoronavirusSupport just testing one of these should suffice since there is no logic in the Flow for this.

test/flows/business_coronavirus_support_finder_flow_test.rb
```diff
@@ -0,0 +1,255 @@
...
+    should "render the question" do
+      assert_rendered_question
+    end
+
+    context "next_node" do
+      should "have a next node of results for any response" do
+        assert_next_node :results, for_response: %w[local_1 local_2 national]
+      end
+
+      should "have a next node of results for not closed_by_restrictions" do
```
kevindew, line 163: I don't think this test is needed as the condition of "have a next node of results for any response" in the previous test covers this - and there is no additional code for this.

kevindew (commented):
This looks great too 👍  thanks

leenagupte:
@kevindew Thanks for your review. I've updated the tests with your suggestions and fixed up.

test/flows/business_coronavirus_support_finder_flow_test.rb
kevindew: Thanks, I should have been clearer but this comment applies to all of the places you've got a loop of responses - as they're all in the same situation where they don't have any logic for the next node.

leenagupte:
@kevindew I've removed the remaining loops in the test and fixed up.

kevindew (approved):
Super, thanks @leenagupte

PR #5531, opened 2021-07-29
Convert Calculate Agricultural Holiday Entitlement flow tests

leenagupte:
Trello: https://trello.com/c/Z33RyIY5

# What's changed and why?

Converts the Calculate Agricultural Holiday Entitlement flow tests to use the node based testing approach introduced in [ADR-3][1].

Also removes the existing flow integration and unit tests as they are no longer needed.


test/flows/calculate_agricultural_holiday_entitlement_flow_test.rb
```diff
@@ -0,0 +1,125 @@
...
+      add_responses work_the_same_number_of_days_each_week?: "different-number-of-days"
+    end
+
+    should "render the question" do
+      assert_rendered_question
+    end
+
+    context "next_node" do
+      should "have a next node of how_many_total_days? for any response in current year" do
+        travel_to Time.zone.parse("2021-01-01")
```
kevindew, line 60:
```suggestion
      should "have a next node of how_many_total_days? for any response" do
```

I don't think we need the "current year" qualifier because it doesn't do anything if the year is incorrect (these lines seem to do nothing https://github.com/alphagov/smart-answers/blob/f27f74227f27f47dfce636d18b859f453da16577/app/flows/calculate_agricultural_holiday_entitlement_flow.rb#L46-L47 - I guess at one point they set boundaries on select inputs but I can't work out any affect they have now).


Oh and moot point now, but for future, this repo seems to have used timecop for all time travel so it'd be better to stay conventional with that until such a time that we swap that over (_wonders whether it could be done entirely with a find & replace_)

test/flows/calculate_agricultural_holiday_entitlement_flow_test.rb
```diff
@@ -0,0 +1,125 @@
...
+      travel_to Time.zone.parse("2021-01-01")
+      testing_node :how_many_total_days?
+      add_responses work_the_same_number_of_days_each_week?: "different-number-of-days",
+                    what_date_does_holiday_start?: "2021-12-01"
+    end
+
+    should "render the question" do
+      assert_rendered_question
+    end
+
```
kevindew, line 99: We should have a: `context "validation"` here to assert that it can be invalid.

A simple way to trigger the validation is to put in a big number like 365.

test/flows/calculate_agricultural_holiday_entitlement_flow_test.rb
```diff
@@ -0,0 +1,125 @@
...
+    setup do
+      testing_node :how_many_weeks_at_current_employer?
+      add_responses work_the_same_number_of_days_each_week?: "same-number-of-days",
+                    how_many_days_per_week?: "7-days",
+                    worked_for_same_employer?: "multiple-employers"
+    end
+
+    should "render the question" do
+      assert_rendered_question
+    end
```
kevindew, line 117: Same point about validation here - I imagine 52 will do the trick.

test/flows/calculate_agricultural_holiday_entitlement_flow_test.rb
```diff
@@ -0,0 +1,125 @@
...
+
+      should "have a next node of how_many_weeks_at_current_employer? for a 'multiple-employers' response" do
+        assert_next_node :how_many_weeks_at_current_employer?, for_response: "multiple-employers"
+      end
+    end
+  end
+
+  context "question: how_many_total_days?" do
+    setup do
+      travel_to Time.zone.parse("2021-01-01")
```
kevindew, line 90: As before I don't think you need this.

kevindew (commented):
This is looking good, just a few small things.

test/flows/calculate_agricultural_holiday_entitlement_flow_test.rb
kevindew: Yup, find and replaced away: https://github.com/alphagov/smart-answers/pull/5535

leenagupte:
@kevindew Thanks for your review. I've addressed your comments and fixed up.

kevindew (approved):
Nice one 👍

PR #5533, opened 2021-07-30
Check outcomes against title and body text

leenagupte:
# What's changed and why?

Not all outcomes have body text, some of them only have titles,
for example: https://www.gov.uk/additional-commodity-code/y/5/30/70

If an outcome didn't have body text, the `assert_next_node` check failed
with `Expected "" to not be empty.` Checking against the title as well fixes
that.

### DIFF
--- a/test/support/flow_test_helper.rb
+++ b/test/support/flow_test_helper.rb
@@ -23,7 +23,7 @@ module FlowTestHelper
     when :question
       assert_not_empty test_flow.question_title
     when :outcome
-      assert_not_empty test_flow.outcome_body
+      assert_not_empty test_flow.outcome_text
     else
       raise "Unknown node type for #{next_node}"
     end
@@ -71,8 +71,8 @@ module FlowTestHelper
   def assert_rendered_outcome(text: nil)
     ensure_valid_and_correct_node
 
-    assert_not_empty test_flow.outcome_body
-    assert_match text, test_flow.outcome_body_text if text
+    assert_not_empty test_flow.outcome_text
+    assert_match text, test_flow.outcome_text if text
   end
 
   class TestFlow
@@ -146,13 +146,25 @@ module FlowTestHelper
     end
 
     def outcome_body
-      raise "#{state.current_node_name} is not an outcome" unless current_node_type == :outcome
+      outcome_presenter.body
+    end
 
-      OutcomePresenter.new(flow.node(state.current_node_name), nil, state).body
+    def outcome_title
+      outcome_presenter.title
     end
 
     def outcome_body_text
       Nokogiri::HTML::DocumentFragment.parse(outcome_body).text
+    end
+
+    def outcome_text
+      [outcome_title, outcome_body_text].reject(&:blank?).join("\n")
+    end
+
+    def outcome_presenter
+      raise "#{state.current_node_name} is not an outcome" unless current_node_type == :outcome
+
+      OutcomePresenter.new(flow.node(state.current_node_name), nil, state)
     end
   end
 end


test/support/flow_test_helper.rb
```diff
@@ -151,8 +151,26 @@ def outcome_body
...
     def outcome_body_text
       Nokogiri::HTML::DocumentFragment.parse(outcome_body).text
     end
+
+    def outcome_title_text
+      Nokogiri::HTML::DocumentFragment.parse(outcome_title).text
+    end
+
+    def outcome_text
+      "#{outcome_title}#{outcome_body}"
```
kevindew, line 169:
```suggestion
      [outcome_title, outcome_body].reject(&:blank?).join("\n")
```

As this method is suffixed with text I think we should use the outcome_body_text in it so that it only returns text, the [outcome_title can only return text](https://github.com/alphagov/smart-answers/blob/06bda75726f104de6ceb25927415cbff16a7046b/lib/smart_answer/erb_renderer/format_capture_helper.rb#L3-L9).

I think a new line might be a reasonable separator for these so we can assert on it, but we need to only apply that if one of them isn't empty otherwise it'll blow the empty check.

test/support/flow_test_helper.rb
```diff
@@ -151,8 +151,26 @@ def outcome_body
...
+    end
+
+    def stripped_outcome_text
+      "#{outcome_title_text} #{outcome_body_text}"
+    end
```
kevindew, line 174: I don't think we'll actually need this method.

test/support/flow_test_helper.rb
```diff
@@ -151,8 +151,26 @@ def outcome_body
...
+      OutcomePresenter.new(flow.node(state.current_node_name), nil, state).title
+    end
+
     def outcome_body_text
       Nokogiri::HTML::DocumentFragment.parse(outcome_body).text
     end
+
+    def outcome_title_text
+      Nokogiri::HTML::DocumentFragment.parse(outcome_title).text
+    end
```
kevindew, line 166: This shouldn't be needed to as we can't have a HTML title (sorry I only learnt that during this review)

test/support/flow_test_helper.rb
```diff
@@ -151,8 +151,26 @@ def outcome_body
       OutcomePresenter.new(flow.node(state.current_node_name), nil, state).body
     end
 
+    def outcome_title
+      raise "#{state.current_node_name} is not an outcome" unless current_node_type == :outcome
+
+      OutcomePresenter.new(flow.node(state.current_node_name), nil, state).title
+    end
```
kevindew, line 158: Rather than having this be very similar to `outcome_body` I think it'd make sense to have a

```
    def outcome_presenter
      raise "#{state.current_node_name} is not an outcome" unless current_node_type == :outcome

      OutcomePresenter.new(flow.node(state.current_node_name), nil, state)
    end
```

and then we can delegate outcome_body and outcome_title to use that

test/support/flow_test_helper.rb
```diff
@@ -71,8 +71,8 @@ def assert_rendered_question
   def assert_rendered_outcome(text: nil)
     ensure_valid_and_correct_node
 
-    assert_not_empty test_flow.outcome_body
-    assert_match text, test_flow.outcome_body_text if text
+    assert_not_empty test_flow.outcome_text
+    assert_match text, test_flow.stripped_outcome_text if text
```
kevindew, line 75: This looks good 👍 

I imagine if needed we could also have optional arguments like `assert_rendered_outcome(title: /bah/)` or `def assert_rendered_outcome(body_html: /<a href/)` if we wanted to do more granular assertions in a shorthand syntax.

kevindew (commented):
Nice one finding out about this problem ⭐ , it's my bad not to realise that an outcome needn't have a body.

Approach looks good. I've made some suggestions to try reduce some ambiguities and duplications.

test/support/flow_test_helper.rb
leenagupte: Yeah, I think this is best left for when there is a known need for it.

kevindew (approved):
Nice one

PR #5535, opened 2021-07-30
Remove timecop gem 

kevindew:
This isn't needed any more as ActiveSupport provides time helpers.

We also don't need an equivalent of `Timecop.return` as
ActiveSupport::Testing::TestHelpers hooks into the teardown method and
resets the time (it also transpires that most of the Timecop.return
weren't needed).


leenagupte (approved):
👍 Looks good.

PR #5537, opened 2021-08-02
Replace minimum wage flow tests

kevindew:
Trello: 

- https://trello.com/c/fU4RQYvF/494-foundational-work-to-implement-smart-answer-flow-test-adr
- https://trello.com/c/ngaMSUL7

This applies the flow test approach introduced in ADR-3 [1] to the two
minimum wage flows. It removes the previous unit and integration flow
tests.

The minimum wage flow is rather special in that it is essentially the
same flow for two different Smart Answers with different templates for
each of them at the outcome stage. This makes it a bit of a pain to test
as it is superfluous to write the same tests out twice but we should
still execute the flow through all of the distinct files (mostly
templates).

The approach I've taken is to include the test through helper methods.
This has the advantage of very low repetition between the files, but has
the disadvantage that the tests aren't re-runnable based on the line
they fail on. I can't think of a good way to solve this and think it
might just be a pain point that needs to be lived with for shared
minitests.

Both of these flows had very similar, but ultimately different outcome
pages so I've put together individual tests for these in each flow.


test/flows/am_i_getting_minimum_wage_flow_test.rb
```diff
@@ -0,0 +1,153 @@
...
+                    current_additional_work_outside_shift?: "yes",
+                    current_paid_for_work_outside_shift?: "yes"
+    end
+
+    should "render living wage copy if a user is eligible for the national living wage" do
+      add_responses how_old_are_you?: @living_wage_age
+      assert_rendered_outcome text: "you appear to be getting the National Living Wage"
+    end
+
+    should "render minimum wage copy if a user is not eligible for the national living wage" do
```
leenagupte, line 44:
```suggestion
    should "render minimum wage copy if a user is not eligible for the national minimum wage" do
```

test/flows/am_i_getting_minimum_wage_flow_test.rb
```diff
@@ -0,0 +1,153 @@
...
+                    current_additional_work_outside_shift?: "yes",
+                    current_paid_for_work_outside_shift?: "yes"
+    end
+
+    should "render living wage copy if a user is eligible for the national living wage" do
+      add_responses how_old_are_you?: @living_wage_age
+      assert_rendered_outcome text: "you appear to be not getting the National Living Wage"
+    end
+
+    should "render minimum wage copy if a user is not eligible for the national living wage" do
```
leenagupte, line 87:
```suggestion
    should "render minimum wage copy if a user is not eligible for the national minimum wage" do
```

test/flows/minimum_wage_calculators_employers_flow_test.rb
```diff
@@ -0,0 +1,133 @@
...
+                    current_additional_work_outside_shift?: "yes",
+                    current_paid_for_work_outside_shift?: "yes"
+    end
+
+    should "render living wage copy if an employee is eligible for the national living wage" do
+      add_responses how_old_are_you?: @living_wage_age
+      assert_rendered_outcome text: "you appear to be paying the National Living Wage"
+    end
+
+    should "render minimum wage copy if an employee is not eligible for the national living wage" do
```
leenagupte, line 44:
```suggestion
    should "render minimum wage copy if an employee is not eligible for the national minimum wage" do
```

test/flows/minimum_wage_calculators_employers_flow_test.rb
```diff
@@ -0,0 +1,133 @@
...
+                    current_additional_work_outside_shift?: "yes",
+                    current_paid_for_work_outside_shift?: "yes"
+    end
+
+    should "render living wage copy if an employee is eligible for the national living wage" do
+      add_responses how_old_are_you?: @living_wage_age
+      assert_rendered_outcome text: "you appear to be not paying the National Living Wage"
+    end
+
+    should "render minimum wage copy if an employee is not eligible for the national living wage" do
```
leenagupte, line 77:
```suggestion
    should "render minimum wage copy if an employee is not eligible for the national minimum wage" do
```

leenagupte (commented):
This is looking good. There are a few copy and paste errors to fix, but nothing major there.

I like the idea of having the shared tests to reduce duplication, however as mentioned inline, I'm not convinced that they each need to be in separate methods, and believe that the tests will be easier to follow and update if they were all in one.

test/support/shared_flows/minimum_wage_flow_test_helper.rb
```diff
@@ -0,0 +1,753 @@
...
+    test_question_does_employer_charge_for_job_requirements
+    test_question_did_employer_charge_for_job_requirements
+    test_question_current_additional_work_outside_shift
+    test_question_past_additional_work_outside_shift
+    test_question_current_paid_for_work_outside_shift
+    test_question_past_paid_for_work_outside_shift
+  end
+
+  def test_question_what_would_you_like_to_check
+    context "question: receive_housing_benefit?" do
```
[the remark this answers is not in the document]
kevindew: Thanks, good spot 🕵️‍♀️

test/support/shared_flows/minimum_wage_flow_test_helper.rb
```diff
@@ -0,0 +1,753 @@
...
+    test_question_current_accommodation_charge
+    test_question_past_accommodation_charge
+    test_question_current_accommodation_usage
+    test_question_past_accommodation_usage
+    test_question_does_employer_charge_for_job_requirements
+    test_question_did_employer_charge_for_job_requirements
+    test_question_current_additional_work_outside_shift
+    test_question_past_additional_work_outside_shift
+    test_question_current_paid_for_work_outside_shift
+    test_question_past_paid_for_work_outside_shift
```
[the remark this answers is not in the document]
kevindew: Yeah that seems a reasonable shout. 

I think I put them this way from initially thinking we'd call more of the methods in the flow, before I realised these were exactly the same.

Putting them altogether does have the downside of one monster-method rather than lots of followable ones, but does seem to cut out the annoyance of writing the question name about 4 times per file and, since there aren't other methods in the file, doesn't need much navigation.

test/flows/am_i_getting_minimum_wage_flow_test.rb
kevindew: The logic is actually based on the else clause of a living wage conditional. 

https://github.com/alphagov/smart-answers/blob/aa5425c0939f94f290463c7bcb65bc71242923ca/app/flows/am_i_getting_minimum_wage_flow/outcomes/current_payment_above.erb#L3-L7

Similar situation in the other cases you've flagged.

kevindew:
Thanks @leenagupte, I've pushed up some amends.

PR #5539, opened 2021-08-03
Use node based testing for Next steps for you business

theseanything:
This converts the Next steps for your business tests to use the node based testing approach introduced in [ADR-3][1]. This replaces the RSpec feature tests.

1]: https://github.com/alphagov/smart-answers/blob/main/docs/arch/003-testing-flows.md


leenagupte (commented):
I think the tests for the question flow look great! 🎉 

I am concerned about the outcome tests though. The outcome relies heavily on the `NextStepsForYourBusinessCalculator` calculator, however I haven't been able to find any tests for it. Are/were there any plans to add the missing tests?

Are the group headings e.g. "Good to know" always rendered on the outcome regardless of the answers provided? 

As there is conditional logic for displaying some of the results [1] I think some outcome tests should are needed to make sure that the content is displaying correctly.

For example, if `annual_turnover_over_85k` = "yes", we should check for:

```
assert_rendered_outcome text: "Register for VAT"
```

[1]: https://github.com/alphagov/smart-answers/blob/main/lib/smart_answer/calculators/next_steps_for_your_business_calculator.rb#L47-L61

theseanything:
@leenagupte thanks for the review 😄 !!

> I am concerned about the outcome tests though. The outcome relies heavily on the `NextStepsForYourBusinessCalculator` calculator, however I haven't been able to find any tests for it. Are/were there any plans to add the missing tests?

Yes, you are correct we don't have any yet - and I'm doing a follow up PR to add them in,

> Are the group headings e.g. "Good to know" always rendered on the outcome regardless of the answers provided?

Yup. We don't have a state where those groups would be empty - so they should always be present.

> As there is conditional logic for displaying some of the results [1](https://github.com/alphagov/smart-answers/blob/main/lib/smart_answer/calculators/next_steps_for_your_business_calculator.rb#L47-L61) I think some outcome tests should are needed to make sure that the content is displaying correctly.

Yeah I see the concern. I wonder if it's enough to cover them in the calculator tests (in the separate PR)? I'm just worried, that I'll be essentially writing the same tests here. Also not sure how scalable this would be, as we are potentially adding a lot more results.

leenagupte:
> Yeah I see the concern. I wonder if it's enough to cover them in the calculator tests (in the separate PR)? I'm just worried, that I'll be essentially writing the same tests here. Also not sure how scalable this would be, as we are potentially adding a lot more results.

We do have precedence for checking conditionals in the outcome views [1]. 

I'm not sure what the calculator tests will cover, but it would be good to have something to check that the logic in the views is working correctly. I don't expect a test to cover all of the potential outcomes but a test that checks the correct group name and section title are being rendered for a calculator value would be good. 


[1]: https://github.com/alphagov/smart-answers/blob/aa5425c0939f94f290463c7bcb65bc71242923ca/test/flows/find_coronavirus_support_flow_test.rb#L292

theseanything:
Thanks Leena! 

> I don't expect a test to cover all of the potential outcomes but a test that checks the correct group name and section title are being rendered for a calculator value would be good.

For this, did you want a test for every result item or just one to check it was in the correct group or section? I feel that is we verify it works for one the fundamental logic works, and should work for _n_ results items.

leenagupte:
> For this, did you want a test for every result item or just one to check it was in the correct group or section? I feel that if we verify it works for one, then fundamental logic should work for _n_ results items. (A little worried about scalability and fragility of testing all - for example if we have 100 results, that going to be difficult)

I'm not sure I understand the question, but I think it would be good to check that the content displays as expected when responses have been selected. I think you could get away with 17 assertions over a handful of tests.

One that checks each of:

```ruby
      r1: ->(calculator) { calculator.registered_for_corp_tax != true },
      r16: ->(calculator) { calculator.annual_turnover_over_85k == "yes" },
      r17: ->(calculator) { calculator.annual_turnover_over_85k == "not_sure" },
      r18: ->(calculator) { calculator.employer != "no" },
      r19: ->(calculator) { calculator.needs_financial_support == "yes" },
      r20: ->(calculator) { calculator.needs_financial_support == "yes" },
      r21: ->(calculator) { calculator.employer != "no" },
      r22: ->(calculator) { calculator.business_premises.include?("home") },
      r23: ->(calculator) { calculator.business_premises.include?("rented") },
      r24: ->(calculator) { (calculator.business_premises & %w[rented owned none]).present? },
      r25: ->(calculator) { (calculator.business_premises & %w[rented owned none]).present? },
      r26: ->(calculator) { calculator.activities.include?("import_goods") },
      r27: ->(calculator) { calculator.activities.include?("export_goods_or_services") },
      r28: ->(calculator) { calculator.annual_turnover_over_85k == "yes" },
      r29: ->(calculator) { calculator.annual_turnover_over_85k == "no" },
      r30: ->(calculator) { calculator.activities.include?("export_goods_or_services") },
```
and one from r2 - r15 as they all do the same thing so if one works, the others should.

A lot of these can be combined, so the test where `calculator.annual_turnover_over_85k == "yes"` can check that titles for r16 and r28 are displayed. Also the test for r23 can be included in the test for r24 and r25 content if business_premises is "rented" etc.

theseanything:
@leenagupte 

Ah I guess I'm not sure what the assertions should be. Do you mean something like this:

```ruby
  context "outcome: results" do
    setup do
      testing_node :results
      add_responses annual_turnover_over_85k: "no",
                    employ_someone: "no",
                    activities: "none",
                    financial_support: "no",
                    business_premises: "none"
    end

    should "render the default results" do
      assert_rendered_outcome text: "Check if you need any licences for your business"
      assert_rendered_outcome text: "Check if you need insurance"
      assert_rendered_outcome text: "Check if you need to send a Self Assessment"
      assert_rendered_outcome text: "Selling goods or services"
      assert_rendered_outcome text: "Avoid anti-competitive activity"
      assert_rendered_outcome text: "Learn how to deal with personal information"
      assert_rendered_outcome text: "Health and safety at work"
      assert_rendered_outcome text: "How to run a limited company"
      assert_rendered_outcome text: "Starting a business overview"
      assert_rendered_outcome text: "Include your company details on stationery"
      assert_rendered_outcome text: "How to dispose of your company's waste"
      assert_rendered_outcome text: "Protect your brand or designs"
      assert_rendered_outcome text: "Get help from the Business Support Helpline"
      assert_rendered_outcome text: "Sign up for Companies House email reminders"
    end

    should "render the results if not registered for corporation tax" do
      add_responses ct: true
      assert_rendered_outcome text: "Register for Corporation Tax"
    end

    should "render the results if business needs financial support" do
      add_responses financial_support: "yes"
      assert_rendered_outcome text: "See what finance and support is available"
      assert_rendered_outcome text: "Get financial support during coronavirus"
    end

    should "render the results if business has annual turnover more than 85k" do
      add_responses annual_turnover_over_85k: "yes"
      assert_rendered_outcome text: "Register for VAT"
      assert_rendered_outcome text: "How and when to charge VAT"
    end
```

Wasn't sure whether it should be checking the right result should be under the correct group or section? (Also unsure how to right an assertion for that)

leenagupte:
> @leenagupte
> 
> Ah I guess I'm not sure what the assertions should be. Do you mean something like this:
> 
> ```ruby
>   context "outcome: results" do
>     setup do
>       testing_node :results
>       add_responses annual_turnover_over_85k: "no",
>                     employ_someone: "no",
>                     activities: "none",
>                     financial_support: "no",
>                     business_premises: "none"
>     end
> 
>     should "render the default results" do
>       assert_rendered_outcome text: "Check if you need any licences for your business"
>       assert_rendered_outcome text: "Check if you need insurance"
>       assert_rendered_outcome text: "Check if you need to send a Self Assessment"
>       assert_rendered_outcome text: "Selling goods or services"
>       assert_rendered_outcome text: "Avoid anti-competitive activity"
>       assert_rendered_outcome text: "Learn how to deal with personal information"
>       assert_rendered_outcome text: "Health and safety at work"
>       assert_rendered_outcome text: "How to run a limited company"
>       assert_rendered_outcome text: "Starting a business overview"
>       assert_rendered_outcome text: "Include your company details on stationery"
>       assert_rendered_outcome text: "How to dispose of your company's waste"
>       assert_rendered_outcome text: "Protect your brand or designs"
>       assert_rendered_outcome text: "Get help from the Business Support Helpline"
>       assert_rendered_outcome text: "Sign up for Companies House email reminders"
>     end
> 
>     should "render the results if not registered for corporation tax" do
>       add_responses ct: true
>       assert_rendered_outcome text: "Register for Corporation Tax"
>     end
> 
>     should "render the results if business needs financial support" do
>       add_responses financial_support: "yes"
>       assert_rendered_outcome text: "See what finance and support is available"
>       assert_rendered_outcome text: "Get financial support during coronavirus"
>     end
> 
>     should "render the results if business has annual turnover more than 85k" do
>       add_responses annual_turnover_over_85k: "yes"
>       assert_rendered_outcome text: "Register for VAT"
>       assert_rendered_outcome text: "How and when to charge VAT"
>     end
> ```
> 

Yes! Exactly that. We don't need to test all of the content. We only need enough to be sure that the expected content has rendered.

theseanything:
Thanks @leenagupte! Updated!

test/flows/next_steps_for_your_business_flow_test.rb
```diff
@@ -0,0 +1,184 @@
...
+      assert_rendered_outcome text: "Include your company details on stationery"
+      assert_rendered_outcome text: "How to dispose of your company's waste"
+      assert_rendered_outcome text: "Protect your brand or designs"
+      assert_rendered_outcome text: "Get help from the Business Support Helpline"
+      assert_rendered_outcome text: "Sign up for Companies House email reminders"
+    end
+
+    should "render the results if not registered for corporation tax" do
+      add_responses ct: "true"
+      assert_rendered_outcome text: "Register for Corporation Tax"
```
leenagupte, line 130: I don't understand how this test is passing.

It looks like `registered_for_corp_tax ` is being set to "true" if `ct` is "true" and "false" for any other value:

```ruby
calculator.registered_for_corp_tax = (ct == "true")
```

However, in the calculator, the text for `r1` should be displayed if `registered_for_corp_tax` is not true:

```ruby
r1: ->(calculator) { calculator.registered_for_corp_tax != true },
```

So I don't understand how the "Register for Corporation Tax" text is being rendered. Nor do I understand how the value for `ct` is being set in the flow.

leenagupte: I've tried this manually, and it looks like "Register for Corporation Tax" does NOT display if `ct=true` is [appended to the querystring](https://www.gov.uk/next-steps-for-your-business/results?activities%5B%5D=export_go[…]B%5D=home&employ_someone=not_sure&financial_support=yes&ct=true ) but DOES display if it's [not appended](https://www.gov.uk/next-steps-for-your-business/results?activities%5B%5D=export_go[…]B%5D=home&employ_someone=not_sure&financial_support=yes) so I'm still confused.

leenagupte: I changed the response in the test to `add_responses ct: "false"` and it still passes, so I think we need to re-think how the value if `ct` is set in the test. 

Perhaps @kevindew has some ideas?

test/support/flow_test_helper.rb
```diff
@@ -98,9 +98,10 @@ def testing_node
 
     def add_responses(responses)
       question_names = flow.questions.map(&:name)
+      valid_parameters = question_names + flow.additional_parameters
 
       responses.each do |key, value|
-        raise "Invalid question #{key}" unless question_names.include?(key.to_sym)
+        raise "Invalid response key #{key}" unless valid_parameters.include?(key.to_sym)
```
leenagupte, line 104: I like the explicitness of the original error, so how about:
```suggestion
        raise "Invalid question or query parameter: #{key}" unless valid_parameters.include?(key.to_sym)
```

leenagupte (commented):
Thanks for adding the extra outcome tests 🏅 🎉 All but one of them looks really good. There are more comments/questions about that in the inline comments.

I think more thought is needed about how to handle extra non-question querystring parameters in the tests as I don't think this is working as expected.

test/flows/next_steps_for_your_business_flow_test.rb
theseanything: Ah sooo this is super confusing, because I made a mistake - the test should have `add_response ct: "false"`.

But it's weird that it's still passing - I'll take a look.

---
Just an explainer for the behaviour:
```
calculator.registered_for_corp_tax = (ct == "true")
```

is setting the the attribute `registered_for_corp_tax` to a boolean if the ct query parameters is a String "true". (i.e. type casting)

Then the rule to show the result:
```
r1: ->(calculator) { calculator.registered_for_corp_tax != true }
```

returns false, if the user has already registered for corp tax. (e.g. ct="true"). This is because the result is "Register for corporation tax", which we only want to show user's that haven't already. Overall the behaviour is hide the results for users with `ct=true` appended to the query parameters.

Btw, the context is Companies House sends links to our tool to users. They generate a link with the `ct` query parameter already set.

kevindew: I was just looking at this after @leenagupte's tag so I can save you a little time @theseanything.

The lack of this doing anything is caused by string/symbol key indexing of the underlying hash in response store. 
Basically: 
https://github.com/alphagov/smart-answers/blob/48c9b5d5a507dbc5fcb66cc9133a14333ce25b28/lib/smart_answer/state_resolver.rb#L9-L11
looks up the response store by a symbol

whereas:
https://github.com/alphagov/smart-answers/blob/48c9b5d5a507dbc5fcb66cc9133a14333ce25b28/lib/smart_answer/state_resolver.rb#L38-L40

uses a string.

This only affects the tests because the data that goes into the [response store in tests is a regular hash](https://github.com/alphagov/smart-answers/blob/48c9b5d5a507dbc5fcb66cc9133a14333ce25b28/test/support/flow_test_helper.rb#L85), whereas in [FlowHelper](https://github.com/alphagov/smart-answers/blob/main/app/helpers/flow_helper.rb#L16) it's a HashWithIndifferentAccess.

So we can sort this out if we ensure that a hash with indifferent access is used consistently in the response store class. 

There'd be a few ways to resolve that, I wonder if one of the best ones is to call `.with_indifferent_access` on the initial argument in ResponseStore as I believe that will also dup the hash (which seems like desirable behaviour as you'd not want anything to be actually modifying request.query_parameters which is what is probably already happening)

leenagupte: Thanks for the explanation @theseanything. So it looks like the default behaviour is to display the "Register for corporation tax" text, so this check could probably be added to the first test. It looks like what we need for this test is to check that the text _isn't_ being displayed.

theseanything: Thanks both! I agree I think adding `.with_indifferent_access` might be the thing to do - I'll add a commit to fix that here.

Happy to move this into the first test - however, I'm worried about adding tests to show that results are hidden. If I do this for this result, then why not all the others? Which essential takes closer to trying to test every possible permutation of results - which is going to be super cumbersome for these node tests (and dups of the calculator unit tests).

leenagupte: I think the test for "hidden" results is needed here, because that's the purpose of the `ct` parameter. How will we know if the `. with_indifferent_access` changes are working if don't test it?

theseanything: But you could argue that for any result - e.g. the financial support results are hidden if the user answers "no" etc?

If we want to verify the behaviour of the test helpers, they'd need to their own unit tests. These tests are for testing the flow not the test helpers.

leenagupte: I agree, these tests are for the for flow, and this behaviour, which I struggled to understand is, set in the flow:

```ruby
calculator.registered_for_corp_tax = (ct == "true")
```

I don't believe there are any plans to test it elsewhere?

theseanything: Ah right, sorry I thought you meant about testing the response store for a HashWithIndifferentAccess behaviour rather than a Hash.

As for that line, I think what I'm struggling with is _what are we trying to test_ with for these node based tests? Are we unit testing the on_response, next_node etc code blocks? Are we unit testing the page renders?

That particular line gets executed in the default test where the results is displayed. However, we don't have a test specifically for the permutation where `ct="true"`. But if I were to add that in, why would I not add in a test for all the other permutation of the other results? E.g.employ_someone="in_future" shows the "Check what to do before employing someone" result.

I'm just unclear around the reasoning around what assertions we need to make.

(Also not sure how to assert that text is not present - guessing I'd have to add another test helper?)

kevindew: I can help on this. I'll drop a reply once I'm out of a meeting.

kevindew: **TL;DR - no tests for outcome would meet the standard for testing, a few outcome tests would make this a sufficiently thorough integration test**

---

Great and interesting discussion here, it often seems that deciding the boundaries of what to test varies on personal perspective. I find it's often good to think in terms of [necessary and sufficient tests] (https://blog.testdouble.com/posts/2020-02-25-necessary-and-sufficient/) to try to establish what the boundaries should be.

[As referenced in the ADR](https://github.com/alphagov/smart-answers/blob/main/docs/arch/003-testing-flows.md#decision) the purpose of outcome tests are to explicitly test specific logic in an outcome. Motivation for this is mostly coverage related, and it's really on the basic side of testing that a lot of Smart Answer's lack (for an example of why just this level of basic is enough, yet not currently met, is https://github.com/alphagov/smart-answers/blob/0f1daad3d148b384ce65e06ae738a759a8315a61/app/flows/maternity_paternity_calculator_flow/outcomes/adoption_not_entitled_to_leave_or_pay.erb#L5 where `a_leave_employment_threshold` will raise a variable not defined error).

There is not specific logic in the outcome, so I'd argue it's not necessary that we actually have any outcome tests. We achieve the goal of the code being executed by at least one of `calculator.grouped_results` returning true. So I'd be satisfied that having a unit test for `grouped_results` is enough that we're confident we have the necessary tests to meet the outcome testing goal.

The more ambiguous area seems to be deciding if the tests we have are sufficient. In that case we'll want to consider the domain of this individual test and the impact of individual lines of code in the flow. So just the basic testing thing of "if this line is code is removed will something break" (TBH I doubt all Smart Answers will 100% comply with this level of testing due to the tedium involved in some cases).

It could also be deciding that we want these tests to assert that the YAML file integrates with flow. My view would be that testing YAML file integration can be done in the calculator unit tests so doesn't belong here.

I'd consider having a few assertions to do with the 6 attributes that are stored on the calculator to be enough for these tests to be sufficient (example below). I think they do enough to confirm that the flow integrates with the calculator and then anything further needed should be testable at calculator level.

```
  context "outcome: results" do
    setup do
      testing_node :results
      add_responses annual_turnover_over_85k: "no",
                    employ_someone: "no",
                    activities: "none",
                    financial_support: "no",
                    business_premises: "none"
    end

    should "render guidance specific to annual turnover" do
      add_responses annual_turnover_over_85k: "yes"
      assert_rendered_outcome text: "Register for VAT"    
    end

    should "hide corporation tax guidance based on ct parameter" do
      assert_match "Register for Corporation Tax", @test_flow.outcome_text
      add_responses ct: "true"
      assert_no_match "Register for Corporation Tax", @test_flow.outcome_text
    end

    should "render guidance specific to employing someone" do
      add_responses employ_someone: "yes"
      assert_rendered_outcome text: "Check what to do before employing someone"    
    end

    should "render guidance specific to a businesses activities" do
      add_responses activities: %w[import_goods]
      assert_rendered_outcome text: "How to buy items from abroad"   
    end

    should "render guidance specific to financial support" do
      add_responses financial_support: "yes"
      assert_rendered_outcome text: "See what finance and support is available"
    end

    should "render guidance specific to business premise" do
      add_responses business_premises: %w[home]
      assert_rendered_outcome text: "Check the rules on running a business from home"
    end
  end
```

does this help?

theseanything: I've updated the tests, which hopefully covers at least the some permutation of the results being shown or hidden. Let me know that make sense.

kevindew: Great thank you! Yes they look very thorough.

kevindew (approved):
Thanks ⭐ 

As this is the last RSpec flow test you can have the honours of checking an item off the list of: https://github.com/alphagov/smart-answers/issues/5350

PR #4823, opened 2020-09-03
Add option to use title as outcome page heading

gclssvglx:
- Add support to use the page title instead of flow name as the
  outcome page heading (`h1`)
- Add tests
- Add documentation
- Update AllSmartAnswers flow outcome page to use title as heading

:warning: This application is Continuously Deployed: :warning:

- Merged changes are automatically deployed to staging and production.

- Make sure you follow [the guidance for deployments](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) **before** you merge.

- Check your branch is being deployed in the [Release app](https://release.publishing.service.gov.uk/applications/smartanswers), after merging.

[Trello](https://trello.com/c/DXCVrn1z/446-results-page-layout-and-styling)

PR #5543, opened 2021-08-05
Update headings to use context inside option for outcome page

jon-kirwan:
## What?

- Smart answers have the same title for both Landing page and Outcome page - which is a duplicate page title and fails the accessibility guidelines.
- The outcome pages currently use the title component (where the caption / context is not nested in the H1) https://components.publishing.service.gov.uk/component-guide/title/with_context.
- We need to update the Outcome page so that the page title and H1 has the agreed heading which is `[Service name]: What you need to know based on the answers you gave`.
- This will use the design system pattern where the caption (service name) is part of the heading (what you need to know based on the answers you gave). for example https://design-system.service.gov.uk/styles/typography/captions-inside/index.html. This has been recently added as an option to the title component - https://components.publishing.service.gov.uk/component-guide/title/with_context_inside

## Why?

Fail of WCAG SC 2.4.6

## Visuals Before & After

| Before | After |
| --- | --- |
|  |  |

## Anything else

Example of where the original issue exists:
https://www.gov.uk/calculate-your-redundancy-pay/y/2020-01-01/20/1.0

### DIFF
--- a/app/flows/all_smart_answer_questions_flow/outcomes/outcome.erb
+++ b/app/flows/all_smart_answer_questions_flow/outcomes/outcome.erb
@@ -1,5 +1,3 @@
-<% use_title_as_h1 true %>
-
 <% text_for :title do %>
   Congratulations
 <% end %>
--- a/app/flows/am_i_getting_minimum_wage_flow/outcomes/current_payment_above.erb
+++ b/app/flows/am_i_getting_minimum_wage_flow/outcomes/current_payment_above.erb
@@ -1,9 +1,9 @@
 <% govspeak_for :body do %>
   $C
   <% if calculator.eligible_for_living_wage? %>
-    Based on your answers, you appear to be getting the National Living Wage.
+    You appear to be getting the National Living Wage.
   <% else %>
-    Based on your answers, you appear to be getting the National Minimum Wage.
+    You appear to be getting the National Minimum Wage.
   <% end %>
   $C
 
--- a/app/flows/am_i_getting_minimum_wage_flow/outcomes/current_payment_below.erb
+++ b/app/flows/am_i_getting_minimum_wage_flow/outcomes/current_payment_below.erb
@@ -1,9 +1,9 @@
 <% govspeak_for :body do %>
   $C
   <% if calculator.eligible_for_living_wage? %>
-    Based on your answers, you appear to be not getting the National Living Wage.
+    You appear to be not getting the National Living Wage.
   <% else %>
-    Based on your answers, you appear to be not getting the National Minimum Wage.
+    You appear to be not getting the National Minimum Wage.
   <% end %>
   $C
 
--- a/app/flows/am_i_getting_minimum_wage_flow/outcomes/past_payment_above.erb
+++ b/app/flows/am_i_getting_minimum_wage_flow/outcomes/past_payment_above.erb
@@ -1,6 +1,6 @@
 <% govspeak_for :body do %>
   $C
-    Based on your answers, you appear to have been getting the National Minimum Wage.
+    You appear to have been getting the National Minimum Wage.
   $C
 
   <% if calculator.potential_underpayment? %>
--- a/app/flows/am_i_getting_minimum_wage_flow/outcomes/past_payment_below.erb
+++ b/app/flows/am_i_getting_minimum_wage_flow/outcomes/past_payment_below.erb
@@ -1,6 +1,6 @@
 <% govspeak_for :body do %>
   $C
-   Based on your answers, you appear to have not been getting the National Minimum Wage.
+    You appear to have not been getting the National Minimum Wage.
   $C
 
   This is because you were not being paid the [minimum rate](/national-minimum-wage-rates) you are entitled to.
--- a/app/flows/business_coronavirus_support_finder_flow/outcomes/results.erb
+++ b/app/flows/business_coronavirus_support_finder_flow/outcomes/results.erb
@@ -1,5 +1,3 @@
-<% use_title_as_h1 true %>
-
 <% text_for :title do %>
   Financial support your business might be entitled to
 <% end %>
--- a/app/flows/calculate_married_couples_allowance_flow/outcomes/_done_body.erb
+++ b/app/flows/calculate_married_couples_allowance_flow/outcomes/_done_body.erb
@@ -9,4 +9,4 @@ $C
 $C
 
 
-%This result is an estimate based on your answers.%
+%This result is an estimate.%
--- a/app/flows/calculate_your_redundancy_pay_flow/outcomes/done.erb
+++ b/app/flows/calculate_your_redundancy_pay_flow/outcomes/done.erb
@@ -1,5 +1,5 @@
 <% text_for :title do %>
-  Based on your answers, your statutory redundancy payment is <%= format_money(statutory_redundancy_pay) %> (<%= format_money(statutory_redundancy_pay_ni) %> in Northern Ireland).
+  Your statutory redundancy payment is <%= format_money(statutory_redundancy_pay) %> (<%= format_money(statutory_redundancy_pay_ni) %> in Northern Ireland).
 
 <% end %>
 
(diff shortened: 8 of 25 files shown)


theseanything:
Is the extra title text is going to make sense for all flows? In the example, its seems like there's repetition - "What you need to know based on the answers you gave" which is followed immediately by "Based on your answers,..."

Also for the [business coronavirus support finder]( https://www.gov.uk/business-coronavirus-support-finder/y/england/0_to_249/no/yes/yes/personal_care/national), is already quite a long page and worried this is going to make the heading very verbose.

maxgds:
> Is the extra title text is going to make sense for all flows? In the example, its seems like there's repetition - "What you need to know based on the answers you gave" which is followed immediately by "Based on your answers,...". They ["find-coronavirus-support"](https://www.gov.uk/find-coronavirus-support/results) also already has the title "Information based on your answers..."
> 
> Also for the [business coronavirus support finder](https://www.gov.uk/business-coronavirus-support-finder/y/england/0_to_249/no/yes/yes/personal_care/national), is already quite a long page and worried this is going to make the heading very verbose.

Hey Sean, thanks for your input. Unfortunately we have been struggling with this issue for far too long and trying to figure out the right way forward. We know that this may not be ideal for every single outcome but it's the compromise our team has come up with to fix the accessibility issue on this. Where we spot cases of it not working I think we're open to updating the content - I'll ask Tobi to take a look at this conversation.

Tobiogunsina:
Max is right that this pattern works for most, especially where the service name is the same as outcome page title, which is most. Content changes can be made where we spot that it doesn't work. We can probably proactively remove 'Based on your answers' in most places. @maxgds is this possible?

The new pattern (service name and what you need to know) would replace 'information based on your answers' in "find-coronavirus-support", which already has that repetition too. 

Although, I now wonder if 'information based on your answer' was considered at all as it will be less verbose and seems to communicate the same thing. I'll check with content on that

maxgds:
> We can probably proactively remove 'Based on your answers' in most places. @maxgds is this possible?

Technically, yes, they're hard coded strings, we can find them here: https://github.com/alphagov/smart-answers/search?q=%22Based+on+your+answers%22&type=

I assume there'd be no issue on the content/owner side with doing this. Jon could take it on as part of this ticket and add it to this PR, I expect.

theseanything:
So just also an FYI - Smart Answers has the ability to specify custom page templates for results pages. So not all flows will pick up this pattern in the default template. At the moment, its just "Next steps for you business", but could be others in future.

Regarding the ["Next steps for your business" flow ](https://www.gov.uk/next-steps-for-your-business/results?activities%5B%5D=export_goods_or_services&activities%5B%5D=import_goods&annual_turnover_over_85k=no&business_premises%5B%5D=owned&business_premises%5B%5D=rented&employ_someone=yes&financial_support=yes), I think the start page and the result page have slightly different titles and may offer enough context(?), but it's good to get your thoughts on that too.

More broadly is there a plan to review the content for all outcome pages? I think that might need to be done before this can be rolled out.

Tobiogunsina:
I think that's okay, I would expect that where the default is not used, it means the content has been written in a way that's accessible. 

We've already checked that this pattern will work for most but will check again on integration before it's published. We initially reviewed the content of all the pages and attempted to update all the titles individually but that was problematic because some outcome pages have dynamic content and some pages have multiple outcome answers

jon-kirwan:
Thanks for the feedback @theseanything @maxgds @Tobiogunsina 

I'll make further changes to update the templates to remove the 'Based on your answers' text.

kevindew:
There's a feature, that I think isn't used any more, that would cause problems for this: https://github.com/alphagov/smart-answers/pull/4823 you probably want to remove that if you're applying this. 

Are you working with the content support on this? It feels like something they'd potentially have concerns about. To me it feels like a lot of things aren't going to read correctly with this title, for example: "What you need to know based on the answers you gave: Financial support your business might be entitled to" - did I _need_ to know that? It feels like the heading phrases it like it's always going to tell you something legal.

Lots of Smart Answers define a title in their outcome, it could be that using that for the heading is preferable to the generic phrase?

Tobiogunsina:
Hmmm, I'm not sure what you mean by 'Smart Answers define a title in their outcome' If they've deviated from the default and defined a title already that accurately describes the page content, then that's fine from an accessibility perspective but I'm not sure if I've understood you correctly?

For Context, we initially did a content audit and attempted to identify what the heading of each page should be based on the content of the page. Sometimes the H2 was sufficient but we’ve not been able to do this effectively because
- Some answers are dynamic and there’s no way to preview what the title will be. We’ve been guessing based on the code and that has already caused errors
- Some answers are complex i.e have multiple outcomes on one page
@chris-gds may have more details

A generic title that sufficiently explains the page purpose, combined with the service name seems to fix the problem.

We've decided to change it to 'information based on your answers' instead of 'What you need to know based on the answers you gave' We planned to let Content people know before it's live but I think I'll just let them know today.   

From my understanding of the feature Kevin proposed we remove, It looks like it will be redundant if we don't remove it cc @jon-kirwan @maxgds

kevindew:
> Hmmm, I'm not sure what you mean by 'Smart Answers define a title in their outcome' If they've deviated from the default and defined a title already that accurately describes the page content, then that's fine from an accessibility perspective but I'm not sure if I've understood you correctly?

Sure. I'll show a screenshot to illustrate. Basically Smart Answer outcomes can have a title but it doesn't replace the existing heading merely appears under it.

For example, Plan Adoption Leave has a title of "Your Adoption Leave dates" which could replace the heading but currently sits below the flow name:

Screenshot 2021-08-09 at 14 34 37

> 
> For Context, we initially did a content audit and attempted to identify what the heading of each page should be based on the content of the page. Sometimes the H2 was sufficient but we’ve not been able to do this effectively because
> 
> * Some answers are dynamic and there’s no way to preview what the title will be. We’ve been guessing based on the code and that has already caused errors
> * Some answers are complex i.e have multiple outcomes on one page
>   @chris-gds may have more details

Sure, a title isn't required for an outcome so I could imagine hitting problems with the h2. I think the problems you've experienced are basically the same concerns raised that a generic phrase might not be applicable for the various outcome scenarios.

> A generic title that sufficiently explains the page purpose, combined with the service name seems to fix the problem.
> 
> We've decided to change it to 'information based on your answers' instead of 'What you need to know based on the answers you gave' We planned to let Content people know before it's live but I think I'll just let them know today.

Yup, I think bring content in early is a good idea. They're likely the most invested group in the consequences of this.

> From my understanding of the feature Kevin proposed we remove, It looks like it will be redundant if we don't remove it cc @jon-kirwan @maxgds

If you don't remove that feature then I expect individual Smart Answers will have the ability to override the generic text. You may find this is actually preferred in some circumstances, but life is probably just simpler to remove the feature since nothing is currently using it.

Tobiogunsina:
@kevindew Ha, I see what you mean with the titles now. Yes, that's what we attempted to do initially - Move the H2 into H1. The problem that caused is that on a lot of pages, it means the entire result page needs to be edited/reviewed because other headings on the pages need to move up the hierarchy to make sure the page still has a logical heading structure. It's a bit of a minefield. 

I've not seen any examples where the generic approach doesn't work though. Apart from the ones with repetition that Sean mentioned earlier?

kevindew:
> @kevindew Ha, I see what you mean with the titles now. Yes, that's what we attempted to do initially - Move the H2 into H1. The problem that caused is that on a lot of pages, it means the entire result page needs to be edited/reviewed because other headings on the pages need to move up the hierarchy to make sure the page still has a logical heading structure. It's a bit of a minefield.

Yup sure, I could understand that.

> I've not seen any examples where the generic approach doesn't work though. Apart from the ones with repetition that Sean mentioned earlier?

Yup, it seems less problematic now with the text "Information based on your answers". That seems less likely to clash with the next sentence, than the previous iteration. I do feel a little uneasy that we're not really conducting an audit and thus may just not know if there's a clash - maybe that's unlikely now :-/

jon-kirwan:
Hi all - i've now completed the following changes:
- title text amended to 'Information based on your answers' from 'What you need to know based on the answers you gave'
- updated multiple instances of the outcome page to remove the text 'Based on your answers, '. (for outcome pages affected please see https://github.com/alphagov/smart-answers/pull/5543/commits/51da6e9ff9bd7f2104eb959ec66ae50ca1918869) 
- updated tests based on the previous change
- reverted commit 'Add option to use title as outcome page heading #4823' and resolved conflicts

maxgds (approved):
Looks good to me - albeit more complicated than we initially expected!

PR #5544, opened 2021-08-05
Add ability to dynamically define question options

theseanything:
This PR introduces the ability dynamic define options for radio and checkbox question at request time. This is useful for flows that have differing options depending on answers to previous questions.

Dynamic options are specified as a block within the flow definition. The block needs to return an array of string or symbols that are option keys.
```ruby
radio :node_name do
  options { return <array of option keys> }
end
```
For example:
```ruby
class ExampleFlow < SmartAnswer::Flow
  def define
    name "example"
    ...
    radio :question1 do
      option :yes
      option :no

     on_response { |response| self.x = response }

      next_node { question :question2 }
    end

    radio :question2 do
      options do
        x == "yes" ? ["a", "b"] : ["a", "b", "c"]
      end

      next_node { question :question3 }
    end
    ...
  end
end
```

The option block runs when before the node is rendered, but after the state has resolved for previous nodes. This allows you to modify options based on previous responses.

The existing method and syntax  to specify static options (`option <key_name>`) is still support and unaffected. If both static and dynamic options are defined, the dynamic options will take precedence and replace the static options.

Without dynamic options, flow's currently must define duplicate version of the same question and have addition logic within the next node block to handle which options to show the user. The duplicate question means that redundant content must specified as each question will need its own erb content template, which will be largely that same. An example of flow that suffers from this is the [additional-commodity-code flow](https://github.com/alphagov/smart-answers/blob/main/app/flows/additional_commodity_code_flow.rb). This flow currently has 11 question nodes, however there's only 4 different questions asked to the user. An example of how this flow could be refactor is [seen here.](https://github.com/alphagov/smart-answers/pull/5545/files#diff-3a0bcabf152b44fe0c875947e0c911e9b5c767349058872da04a53d54ee05d76L2)

In addition to dynamic options, the PR introduces a setup block that can be specified in the flow definition before node blocks. For example:

```ruby
class ExampleFlow < SmartAnswer::Flow
  def define
    name "example"
    ...
    setup do                                <-
      self.calculator = SomeCalculator.new  <- NEW SETUP BLOCK SYNTAX
    end                                     <-

    radio :question1 do
    end
    ...
  end
end
```

This is needed because previously the `on_response` block in the first question was relied upon to initial state. For example [set up calculator classes](https://github.com/alphagov/smart-answers/blob/6303598b1af8dae71592e85746e8a681ad11deab/app/flows/additional_commodity_code_flow.rb#L17). It meant the initial state wasn't configure until after the first question was rendered and a response was given. This prevents the dynamic options block from using initialisation logic, as it executes before page render. This aside, setting up state was also a misuse of the purpose of the `on_response` blocks.

### DIFF
--- a/app/presenters/graph_presenter.rb
+++ b/app/presenters/graph_presenter.rb
@@ -49,7 +49,7 @@ private
     when SmartAnswer::Question::Radio
       text << word_wrap(node_title(node))
       text << "\n\n"
-      text << node.permitted_options.map { |o| "( ) #{o}" }.join("\n")
+      text << node.option_keys.map { |o| "( ) #{o}" }.join("\n")
     when SmartAnswer::Question::Checkbox
       text << word_wrap(node_title(node))
       text << "\n\n"
--- a/app/presenters/question_with_options_presenter.rb
+++ b/app/presenters/question_with_options_presenter.rb
@@ -4,7 +4,7 @@ class QuestionWithOptionsPresenter < QuestionPresenter
   end
 
   def option_attributes(key)
-    option = @renderer.option(key.to_sym)
+    option = @renderer.option(key)
 
     if option.is_a?(String)
       { label: option, value: key }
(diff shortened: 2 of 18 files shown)


alex-ju (approved):
Excellent PR description! I made my way through the code and played a bit with the app, nothing I can come up with. Don't hold from asking a 2nd reviewer if you want, but to me it looks great!

PR #5545, opened 2021-08-05
Refactor commodity code flow

theseanything:
This PR refactors the commodity code flow to use dynamic options and store responses using query parameters

Using dynamic options simplifies the flow as previously it had multiple question nodes with the same question, but different options. This created redundant and duplicative content to maintain and also made the flow definition more difficult to read. This was a precursor step required to enable the flow to use query parameters, as the node names would be visible in the url and this makes them more user friendly.

This switches the commodity code flow to use query parameters as the response store instead of the URL path. This change is a part of a migration for all flows that previously used URL path to use query parameters.


theseanything:

PR #5547, opened 2021-08-06
Convert Redundancy Pay flow tests

leenagupte:
Trello: 

- https://trello.com/c/gVmHUVKq
- https://trello.com/c/vLDsFQOj

# What's changed and why?
    
Converts the Calculate Employee Redundancy and Calculate Your Redundancy flow tests to use the node based testing approach introduced in [ADR-3][1].

Calculate Employee Redundancy Flow and Calculate Your Redundancy Flow have the same questions and outcomes. As there is no conditional logic in the outcomes, we can DRY-up the tests for both flows, as seen previously with the minimum wage flow tests.

Also removes the existing flow integration tests as they are no longer needed.


test/support/shared_flows/redundancy_pay_flow_test_helper.rb
```diff
@@ -0,0 +1,108 @@
...
+          add_responses age_of_employee?: "17"
+          assert_next_node :weekly_pay_before_tax?, for_response: "2"
+        end
+      end
+    end
+
+    context "question: weekly_pay_before_tax?" do
+      setup do
+        testing_node :weekly_pay_before_tax?
+        add_responses date_of_redundancy?: 3.years.ago.strftime("%Y-%m-%d"),
```
Rosa-Fox, line 92: This threw me a bit as it feels like it should be invalid if the date of redundancy is longer ago than the number of years worked... however it is valid in the flow as I suppose it doesn't effect the calculation. Could change it to a time within the years_employed? though I also appreciate using this value keeps consistency.

Rosa-Fox (commented):
Looks good! I like how you DRYed it up. A few small comments/questions.

test/support/shared_flows/redundancy_pay_flow_test_helper.rb
leenagupte: The redundancy flows doesn't assume that the redundancy occurred recently, it could have happened a few years in the past.

In reality the time frames for `years_employed?` and `date_of_redundancy?` can never overlap, because once you've been made redundant, you are no longer employed. I think the confusion comes from my use of "X years ago" for the `date_of_redundancy?` rather than using a specific date. I used "X years ago" years for the redundancy date because it passed all of the conditionals for `first_selectable_date` and `last_selectable_date` (which changes every year) without needing to resort to time travelling (with `travel_to`) to make sure the redundancy date is always valid.

`years_employed?` refers to how many years they were employed, not _when_ they were employed. So I guess for this question, they could have been employed somewhere 5 years ago, worked there for 2 years and were made redundant 3 years ago. 

`years_employed?` is also affected by the `age_of_employee?` They must have been over 17 when they were made redundant to have been able to have even worked enough years (2 or more) to claim redundancy pay.

test/support/shared_flows/redundancy_pay_flow_test_helper.rb
```diff
@@ -0,0 +1,108 @@
...
+                      age_of_employee?: "16"
+      end
+
+      should "render question" do
+        assert_rendered_question
+      end
+
+      context "validation" do
+        should "be valid for a valid number of years employe" do
```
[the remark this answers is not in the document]
leenagupte: 😱

leenagupte:
@Rosa-Fox Thanks for your review. I've fixed up the typos and test name changes, and left a response to the query about the years of employment and date of redundancy time frames.

test/support/shared_flows/redundancy_pay_flow_test_helper.rb
Rosa-Fox: Thanks for explaining :) yeah sounds good to keep.

Rosa-Fox (approved):

PR #5554, opened 2021-08-11
Use query parameters for report lost or stolen passport flow

theseanything:
This switches the flow to use query parameters as the response store instead of the URL path. This change is a part of a migration for all flows that previously used URL path to use query parameters. This also renames the nodes to be more human readable, as they will now show in the url.


theseanything:

PR #5558, opened 2021-08-16
Rewrite flow tests for MaternityPaternityCalculatorFlow

kevindew:
https://trello.com/c/tyDMt1tJ/537-smart-answer-flow-tests-adoption-calculator

This rewrites the tests for [MaternityPaternityCalculatorFlow](https://www.gov.uk/maternity-paternity-calculator) to apply the [ADR-3](https://github.com/alphagov/smart-answers/blob/main/docs/arch/003-testing-flows.md) style of node testing.

This MaternityPaternityCalculatorFlow is very complex, perhaps the most complex Smart Answer flow, and is split across 5 interconnected flow files. I've mapped the tests to match these files so they are easier to digest (at first it was a single 2000+ line file) and used a helper file to reduce repetition. As a mark of this flows' complexity I also found out a bug that was previously undiscovered and a number of unreachable nodes.

There's lots more details in the commits.


test/support/flows/maternity_paternity_calculator_flow_test_helper.rb
```diff
@@ -0,0 +1,157 @@
...
+    payday_eight_weeks = payday_eight_weeks ? Date.parse(payday_eight_weeks) : due_date - 6.months
+
+    responses = { what_type_of_leave?: "maternity",
+                  baby_due_date_maternity?: due_date.to_s,
+                  date_leave_starts?: due_date.to_s,
+                  did_the_employee_work_for_you_between?: "yes",
+                  last_normal_payday?: last_normal_payday.to_s,
+                  payday_eight_weeks?: payday_eight_weeks.to_s }
+
+    responses.merge!(pay_frequency_responses(pay_frequency, pay_per_frequency))
```
GDSNewt, line 19: Hi Kevin, I'm not sure I fully understand the purpose of the suffix arguments on line 48, 87, 118 but I was wondering why we don't seem to have an equivalent for line 19 i.e `suffix: " maternity" `

GDSNewt:
Hi Kevin, so I spent the past few days tracking the flows and seeing how they match up in the tests, very impressed you've managed to capture the convoluted web of flows in the tests - even visualising it in a flow chart it's hard to keep track. 

I looked into the flows as they exist now was unable to reach the nodes you pointed out were impossible to get to, so nice spot! 

I like the use of helper methods to make the tests less verbose, at first it was a little confusing but I quickly got used to the approach and felt it was easy enough to see what responses were being tested. 

I am somewhat curious from a code quality standpoint as to how we make that decision to split up the files. That is, is it an ad hoc approach based on the circumstances or does it stem from a set of 'guidelines' that we adhere to. 'Thou shall not exceed the line limit' for example.

test/flows/maternity_paternity_calculator_flow/paternity_calculator_flow_test.rb
```diff
@@ -0,0 +1,791 @@
...
+      add_responses what_type_of_leave?: "paternity",
+                    leave_or_pay_for_adoption?: "yes"
+    end
+
+    should "render the question" do
+      assert_rendered_question
+    end
+
+    context "next_node" do
+      should "have a next node of padoption_date_of_adoption_placement?" do
```
GDSNewt, line 62: Nitpickiest of nitpicks, I understood what was meant by padoption, given the context, but might be slightly ambiguous at a glance

test/support/flows/maternity_paternity_calculator_flow_test_helper.rb
kevindew: Sure, good question. So the same questions can exist with different names for maternity, paternity and adoption. Those for maternity only exist in an un-suffixed form. For reference:

https://github.com/alphagov/smart-answers/blob/e96abd0d1784866e9ae4f76e5cbdc66e2e311435/app/flows/maternity_paternity_calculator_flow/maternity_calculator_flow.rb#L77

https://github.com/alphagov/smart-answers/blob/e96abd0d1784866e9ae4f76e5cbdc66e2e311435/app/flows/maternity_paternity_calculator_flow/adoption_calculator_flow.rb#L156

https://github.com/alphagov/smart-answers/blob/e96abd0d1784866e9ae4f76e5cbdc66e2e311435/app/flows/maternity_paternity_calculator_flow/paternity_calculator_flow.rb#L250

I assume this is a product of the way the flow was originally written, I'm guessing this was first wrote for maternity and then expanded for paternity and adoption - this probably left some scars along the way, including this inconsistency.

test/flows/maternity_paternity_calculator_flow/paternity_calculator_flow_test.rb
kevindew: Yup good point, I don't see that as something that is in the scope for me to change as part of writing these tests as it is the node name in the Smart Answer (and if I was to start trying to improve this where to stop 😬 )

https://github.com/alphagov/smart-answers/blob/e96abd0d1784866e9ae4f76e5cbdc66e2e311435/app/flows/maternity_paternity_calculator_flow/paternity_calculator_flow.rb#L60

kevindew:
> Hi Kevin, so I spent the past few days tracking the flows and seeing how they match up in the tests, very impressed you've managed to capture the convoluted web of flows in the tests - even visualising it in a flow chart it's hard to keep track.

Great thanks for that, yes this is an exceptionally hard Smart Answer to follow.

> I looked into the flows as they exist now was unable to reach the nodes you pointed out were impossible to get to, so nice spot!
> 
> I like the use of helper methods to make the tests less verbose, at first it was a little confusing but I quickly got used to the approach and felt it was easy enough to see what responses were being tested.

Yeah, adding a helper method is always a trade off. If you add one in it's a context switch for anyone reading the code to go find the method and understand what it does, if you don't add one the code reading and code maintenance can be tedious and repetitive.

For these Smart Answers flow tests I veer along the lines of trying to not use helper methods where possible for test clarity and isolation, but once you've got 10 lines of answers repeating many times you need another tool.

> I am somewhat curious from a code quality standpoint as to how we make that decision to split up the files. That is, is it an ad hoc approach based on the circumstances or does it stem from a set of 'guidelines' that we adhere to. 'Thou shall not exceed the line limit' for example. Incidentally I like that you've split it up, just curious as to how we justify/rationalise it in a universal 'this is the approach we take' sense.

I based the split I did here on the way the files were split for the actual Smart Answer so the 4 different answer flows: https://github.com/alphagov/smart-answers/tree/e96abd0d1784866e9ae4f76e5cbdc66e2e311435/app/flows/maternity_paternity_calculator_flow. So I've followed the unit test approach here of mapping code files to individual test files.

Ideally I'd have probably preferred this to be one single file, to reflect more that this an integration test of all the things but that was just too overwhelming. I don't think the way the flow source classes are spread out is ideal really as they're all interdependent so to understand what's going on you need to basically have all 4 files open and jump between them.

Thanks again for the review 👍

GDSNewt (approved):

GDSNewt:
Thanks for the responses, very informative 😃. Happy for you to merge :shipit:

PR #5561, opened 2021-08-16
Remove no-op `from` and `to` date definitions

kevindew:
This PR changes the behaviour of setting `from` and `to` dates on a date question to validate that the input matches that range. It then removes all the cases where `from` and `to` were used without validation. These have been removed because setting them had no effect on the behaviour of the application, prior to https://github.com/alphagov/smart-answers/pull/4598 these used to set the range of `<select>` elements but these are now boundless `<input type="text">` elements.

I think a further improvement could be to remove these `from` and `to` attributes from date questions and define this validation in a `validate` block. However I couldn't tell how big a can of worms that would open and felt this was a sufficient net improvement alone.

More info in the commits...

### DIFF
--- a/lib/smart_answer/question/date.rb
+++ b/lib/smart_answer/question/date.rb
@@ -5,10 +5,6 @@ module SmartAnswer
     class Date < Base
       PRESENTER_CLASS = DateQuestionPresenter
 
-      def validate_in_range
-        @validate_in_range = true
-      end
-
       def from(&block)
         if block_given?
           @from_func = block
@@ -49,18 +45,6 @@ module SmartAnswer
         end
       end
 
-      def range
-        if from && to
-          raise "to date must be after the from date" if from >= to
-
-          from..to
-        elsif !from && !to
-          false
-        else
-          raise "Both from and to must be defined to validate a date question"
-        end
-      end
-
       def parse_input(input)
         date = case input
                when Hash, ActiveSupport::HashWithIndifferentAccess
@@ -81,7 +65,7 @@ module SmartAnswer
                  raise InvalidResponse, "Bad date", caller
                end
 
-        validate_input(date) if @validate_in_range
+        validate_input(date)
         date
       rescue ArgumentError => e
         if e.message =~ /invalid date/
@@ -94,16 +78,14 @@ module SmartAnswer
       def date_of_birth_defaults
         from { 122.years.ago.beginning_of_year.to_date }
         to { ::Time.zone.today.end_of_year }
-        validate_in_range
       end
 
     private
 
       def validate_input(date)
-        return unless range
+        raise "from date must not be after the to date" if from && to && from > to
 
-        min, max = [range.begin, range.end].sort
-        if date < min || date > max
+        if (from && date < from) || (to && date > to)
           raise InvalidResponse, "Provided date is out of range: #{date}", caller
         end
       end
--- a/lib/smart_answer/question/date.rb
+++ b/lib/smart_answer/question/date.rb
@@ -50,7 +50,15 @@ module SmartAnswer
       end
 
       def range
-        @range ||= from && to ? from..to : false
+        if from && to
+          raise "to date must be after the from date" if from >= to
+
+          from..to
+        elsif !from && !to
+          false
+        else
+          raise "Both from and to must be defined to validate a date question"
+        end
       end
 
       def parse_input(input)
--- a/app/flows/all_smart_answer_questions_flow.rb
+++ b/app/flows/all_smart_answer_questions_flow.rb
@@ -43,8 +43,6 @@ class AllSmartAnswerQuestionsFlow < SmartAnswer::Flow
       from { Time.zone.today }
       to { 4.years.since(Time.zone.today) }
 
-      validate_in_range
-
       next_node do
         question :which_date_this_year?
       end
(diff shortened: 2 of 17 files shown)


Rosa-Fox (commented):
Overall it looks good. I think you are right about fixing the validations being a can of worms… I am a bit confused about how you figured out which `to` and `from` values are needed for validation and which are not.

I guess the `to` and `from` in [RedundancyPayFlow](https://github.com/alphagov/smart-answers/blob/main/app/flows/shared/redundancy_pay_flow.rb#L4) are so that the calculations still work? The validation to the user just reads ‘Please answer this question’ so it is hard to understand the error. That error seemed to be like that before you made changes though…  

However, I am confused how RedundancyPayFlow is validating, whereas [this]( https://github.com/alphagov/smart-answers/blob/main/app/flows/calculate_married_couples_allowance_flow.rb#L39) would raise an exception if it wasn’t removed. Maybe we can have a quick call tomorrow?

Shall we add something to the planning board about improving validations?

kevindew:
> Overall it looks good. I think you are right about fixing the validations being a can of worms… I am a bit confused about how you figured out which `to` and `from` values are needed for validation and which are not.

Sure, I worked these out by the presence of the `validate_in_range` method with the `from` and `to`. If this didn't exist then the date range did nothing.

You can see the removal of that method (as it becomes meaningless once validation is implicit) in https://github.com/alphagov/smart-answers/pull/5561/commits/4dcf080992022cc4121ae0ac4aa19e74094baeef

> I guess the `to` and `from` in [RedundancyPayFlow](https://github.com/alphagov/smart-answers/blob/main/app/flows/shared/redundancy_pay_flow.rb#L4) are so that the calculations still work? The validation to the user just reads ‘Please answer this question’ so it is hard to understand the error. That error seemed to be like that before you made changes though…

Yes, that's one that has [`validate_in_range`](https://github.com/alphagov/smart-answers/blob/2571835f8a85498475b3fc9d62cd997ae42cfd7d/app/flows/shared/redundancy_pay_flow.rb#L6) so that those dates actually have an affect.

And yeah, the error is poor. I understand it's a consequence of https://github.com/alphagov/smart-answers/pull/4598 as when a select input existed with only valid dates then any invalid input would have nothing selected - thus the error would make sense. As it's now text based an invalid input will show and not make sense.

> However, I am confused how RedundancyPayFlow is validating, whereas [this](https://github.com/alphagov/smart-answers/blob/main/app/flows/calculate_married_couples_allowance_flow.rb#L39) would raise an exception if it wasn’t removed. Maybe we can have a quick call tomorrow?

Happy to have a call :)

> Shall we add something to the planning board about improving validations?

I'm not sure we want to take on this as it's only tangentially relevant to our work. I think given the Platform Health change that introduced this is relatively recent we should start by taking it to them to see if they'll take it on as a bug fix.

lib/smart_answer/question/date.rb
```diff
@@ -50,7 +50,15 @@ def default_year(&block)
...
-        @range ||= from && to ? from..to : false
+        if from && to
+          raise "to date must be after the from date" if from >= to
+
+          from..to
+        elsif !from && !to
+          false
+        else
+          raise "Both from and to must be defined to validate a date question"
+        end
```
theseanything, line 57: Nit: I think this will only be run at request time (when an input is being validated). It would be great if we could check the node definition at load time, so developers immediately know that definition is wrong. Could we put the checks in the initialiser?

theseanything (approved):
Thanks @kevindew, this is a good improvement! I've also been thinking about how we do validations, do you mind if I join discussions about potential work?

kevindew:
Thanks @theseanything, yup absolutely join in - I'm mostly trying to defer it away 😆

lib/smart_answer/question/date.rb
kevindew: Is there an applicable place to do this prior to runtime without ending up in a 🐰 🕳️ ? I was under the impression that due to the DSL nature that you can essentially adjust the values at any point until runtime and by virtue of them being defined separately you can only handle all cases in this place (you could handle a little in the `to` method)

theseanything: 🤦🏻 You're right, forgot that the `to`/`from` dates are defined as code blocks, which are resolved at request time, so won't be possible the validation at definition load time.

Just thinking how this error surfaces, I think it would be a 5xx page in production and the exception shown in development (?). Not much we can do about that I guess, except making sure we have good testing for the `to`/`from` blocks to help reduce chances of bugs.

theseanything: I wonder if we even need the range method? Could we simplify the validate_input method to:
```ruby
def validate_input(date)
  raise "to date must be after the from date" if from && to && from >= to

  if from && date < from || to && date > to
    raise InvalidResponse, "Provided date is out of range: #{date}", caller
  end
end
```

This would remove the constraint of requiring both `to` and `from` to be defined (not sure why we'd need it).

kevindew: Thanks, that's a great point that we no longer need a public range method.  You're right that it allows it to be simpler where we can either from, to or both and for the values to always have an effect. The concept of a range seems rooted in the past system of select options that needed a start and an end.

I wasn't too concerned about 5xx's in production as, given the values are rather static, I thought it was unlikely to pass through testing and more likely to just linger in the background as code that does nothing and remains. However with your suggested switch I think we can remove the exceptions altogether - if a developer sets up dates where `from > to` they will just create a node that cannot be progressed from - which should hopefully be caught in testing.

I'll just amend the commit now. Only lingering thought I have is that the code will need to be a bit more complex to show more relevant error messages - but that feels a challenge for another time (and also think it would warrant moving this code outside of the question - this seems to be the only question with built in validation - and into validate blocks)

theseanything: Agree with all those points! Thanks Kev, this looks great!

PR #5564, opened 2021-08-17
Rewrite flow tests for ChildBenefitTaxCalculatorFlow

kevindew:
Trello: https://trello.com/c/I5UDYONN/541-smart-answer-flow-tests-child-benefit-tax-calculator

This replaces the previous unit/integration flow tests with the node
based approach proposed in ADR 3 [1].

This flow is somewhat unusual in that it has a dynamic quantity of nodes
that are based on the answers provided to how_many_children? and
how_many_children_part_year? - where that could be up to 90 dynamic
nodes added. Due to this the flow tests for this apply a loop to test
these dynamic nodes (using the same approach that is used to define
them).


test/flows/child_benefit_tax_calculator_flow_test.rb
```diff
@@ -0,0 +1,399 @@
...
+    should "render the question" do
+      assert_rendered_question
+    end
+
+    context "validation" do
+      should "be invalid for a number less than 1" do
+        assert_invalid_response "0"
+      end
+
+      should "be invalid for a number greater than 30" do
```
Rosa-Fox, line 25: Unrelated to your changes but... https://www.guinnessworldrecords.com/world-records/most-prolific-mother-ever

test/flows/child_benefit_tax_calculator_flow_test.rb
```diff
@@ -0,0 +1,399 @@
...
+                      which_tax_year?: "2021",
+                      is_part_year_claim?: "no",
+                      income_details?: "100000",
+                      add_allowable_deductions?: "no"
+
+        assert_rendered_outcome text: "This is an estimate based on your adjusted net income of £100,000.00"
+      end
+    end
+
+    should "render specific guidance when the tax year is 2012-2013" do
```
Rosa-Fox, line 380: There doesn't seem to be an outcome test for [this other 2012 if statement](https://github.com/alphagov/smart-answers/blob/main/app/flows/child_benefit_tax_calculator_flow/outcomes/results.erb#L59)

Rosa-Fox (commented):
Looks good, I have been through it thoroughly and mostly seemed great :) I had a couple of comments about the outcomes. I wondered if we should lower the max children in the tests to improve performance... but 3 seconds doesn't seem too bad considering and it gives a more accurate representation of the flow so I think its ok.

test/flows/child_benefit_tax_calculator_flow_test.rb
kevindew: Haha wow, that is incredible

test/flows/child_benefit_tax_calculator_flow_test.rb
kevindew: Our precedence is to just test conditionals once in outcomes - we get the same coverage either way. Otherwise it'd be pretty arduous in some cases where the same conditional is multiple times in an outcome.

kevindew:
Thanks @Rosa-Fox for the review. You raised a good point about the max children, I don't think I gained anything by testing all of them, so I've just switched it to one which has the same level of coverage and takes 2s off my local running time.

test/flows/child_benefit_tax_calculator_flow_test.rb
Rosa-Fox: 👍

Rosa-Fox (approved):
Definitely prefer this. I wasn't sure how strict we needed to be about covering all the potential children but as you say, it uses the same code so should be no surprises. 1.6 seconds now 🎉 .

kevindew:
> Definitely prefer this. I wasn't sure how strict we needed to be about covering all the potential children but as you say, it uses the same code so should be no surprises. 1.6 seconds now 🎉 .

Yup, I think of it as just aiming for coverage as anything beyond that is so tedious to implement.

PR #5568, opened 2021-08-19
Rewrite CheckUkVisaFlow tests

kevindew:
Trello: https://trello.com/c/bMMeYqKl/542-smart-answer-flow-tests-check-uk-visa

This rewrites the CheckUkVisaFlow test in the node style introduced in
ADR-3 [1].

CheckUkVisa is one of the most complicated Smart Answers (it ranks at
number 4 most complicated based on my crude filesize metric). There is a
lot of different routes through this flow and a daunting 38 different
outcome files.

The most complicated bit of logic, the purpose of visit, is repeated in
two questions - I therefore moved that into a helper file so it could be
re-used without engulfing the top of the file (as test helper methods
need to be static, they need to be defined before their usage).

The other challenging area is setting up the stubs for the world
locations so that individual scenarios will work. There seems to be a
quirk with the world locations stub where it can't be called multiple
times in a flow (presumably it stubs an index endpoint and subsequent
stubbings don't clear the previous one), so you can't adjust this
stubbing in particular test scenarios. Because of this I decided to have
one stub at the top of the file for all scenarios and set ivars for
country types. This carries with it an expense that all tests carry a
stubs they don't need (the stubs are surprisingly time expensive).

As a mark of the complexity of this I did find a number of unreachable
scenarios. I'll resolve those in a separate commit.


test/flows/check_uk_visa_flow_test.rb
```diff
@@ -0,0 +1,741 @@
...
+      %w[republic_of_ireland somewhere_else].each do |response|
+        should "have a next node of outcome_no_visa_needed for a non-visa national passport and a " \
+               "'#{response}' response" do
+          add_responses what_passport_do_you_have?: @non_visa_national_country
+          assert_next_node :outcome_no_visa_needed, for_response: response
+        end
+
+        should "have a next node of outcome_no_visa_needed for an EEA passport and a '#{response}' response" do
+          add_responses what_passport_do_you_have?: @eea_country
+          assert_next_node :outcome_no_visa_needed, for_response: "republic_of_ireland"
```
Rosa-Fox, line 181: Should be `for_response: response`, same for the other assertions in the block.

Rosa-Fox (commented):
Need a lie down after that 🤣. Have followed through it all and all seems to be covered. Nice work! I just had a comment about using a variable rather than string.

PR #5590, opened 2021-09-02
Rewrite tests for help if you are arrested abroad

JonathanHallam:
This replaces the previous unit/integration flow tests with the node
based approach proposed in [ADR 3].

[ADR 3]: https://github.com/alphagov/smart-answers/blob/main/docs/arch/003-testing-flows.md


test/flows/help_if_you_are_arrested_abroad_flow_test.rb
```diff
@@ -0,0 +1,55 @@
...
+
+    should "render the Iran specific text for Iran" do
+      assert_rendered_outcome text: "Information pack for British nationals imprisoned in Iran"
+    end
+
+    should "render the standard text for other countries" do
+      add_responses which_country?: "france"
+
+      assert_no_match "Information pack for British nationals imprisoned in Iran", @test_flow.outcome_text
+    end
```
leenagupte, line 53: There's a missing test for `calculator.has_extra_downloads` being true in this outcome.

test/flows/help_if_you_are_arrested_abroad_flow_test.rb
```diff
@@ -0,0 +1,55 @@
...
+  end
```
leenagupte, line 54: There are also missing outcome tests for `transfer_back`. For example, `answer_one_generic` uses a calculator value so we should probably test that if the use picks a country that that doesn't have a transfer back that the content doesn't show.

Also, it's very strange that `iran_downloads` partial doesn't just show the content by default as Iran is not in `COUNTRIES_WITHOUT_TRANFSERS_BACK` so will always show the content.

test/flows/help_if_you_are_arrested_abroad_flow_test.rb
leenagupte: It looks as though not every download is displayed for every country:

```
    <%= calculator.pdf %>
    <%= calculator.doc %>
    <%= calculator.benefits %>
    <%= calculator.judicial %>
    <%= calculator.prison %>
    <%= calculator.police %>
    <%= calculator.consul %>
    <%= calculator.lawyer %>
```
It seems to be dependent on the [prisoner packs config](config/smart_answers/prisoner_packs.yml). So we should probably a pick a country that does have a download and check that this section appears.

leenagupte (commented):
This is looking good. Though this smart-answer is not as simple as it looks. I've added more detail in the inline comments, but I feel like the tests for outcome `answer_one_generic` could be expanded a bit.

leenagupte (approved):
Looks good! 👍

PR #5595, opened 2021-09-08
Resolve InvalidNode exception for visualise

kevindew:
https://github.com/alphagov/smart-answers/pull/5544 changed the purpose
of the options method on a checkbox question. This had the side effect
of breaking some (presumably untested) code in visualise which tried to
use that method.

I've resolved the problem by changing the options method to option_keys
and adapted the test so that this code has coverage.

### DIFF
--- a/test/fixtures/flows/graph_flow.rb
+++ b/test/fixtures/flows/graph_flow.rb
@@ -12,12 +12,12 @@ class GraphFlow < SmartAnswer::Flow
       end
     end
 
-    radio :q2? do
+    checkbox_question :q2? do
       option :a
       option :b
 
       next_node do |response|
-        if response == "a"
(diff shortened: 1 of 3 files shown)


test/fixtures/flows/graph_flow.rb
```diff
@@ -12,12 +12,12 @@ def define
       end
     end
 
-    radio :q2? do
+    checkbox_question :q2? do
```
leenagupte, line 15: This is great! ⭐

leenagupte (approved):
I've run the visualisation locally and it looks great:

Screenshot 2021-09-08 at 16 50 56

I've checked that the checkbox flow is working as expected. The flow tests are all passing so it LGTM 👍

PR #5604, opened 2021-09-16
Remove ability to define multiple on_response blocks

theseanything:
This removes the ability to define more than one on_response block for a question node. No flow currently has this requirement, removing the code to simplify the codebase. If multiple definitions of on_response are defined an error is thrown at app start app.

### DIFF
--- a/test/unit/question_base_test.rb
+++ b/test/unit/question_base_test.rb
@@ -5,6 +5,15 @@ class QuestionBaseTest < ActiveSupport::TestCase
     @question = SmartAnswer::Question::Base.new(nil, :example_question)
   end
 
+  context "#on_response" do
+    should "not allow multiple definitions of on_response" do
+      assert_raises(RuntimeError, "Multiple on_response blocks not allowed") do
+        @question.on_response { |_| nil }
+        @question.on_response { |_| nil }
+      end
+    end
+  end
+
   context "#transition" do
     should "pass input to next_node block" do
       input_was = nil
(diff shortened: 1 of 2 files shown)


test/unit/question_base_test.rb
```diff
@@ -5,6 +5,15 @@ class QuestionBaseTest < ActiveSupport::TestCase
     @question = SmartAnswer::Question::Base.new(nil, :example_question)
   end
 
+  context "#on_response" do
+    should "not allow multiple definitions of on_response" do
+      assert_raises do
```
leenagupte, line 10: I think this should assert on the message as well so we have confidence that failure is occurring where expected.

leenagupte (commented):
This is an interesting problem. What prompted it? Could you add a some context to the commit message? For example, did you get a strange error? What was it?

I've added a few inline comments. I'm not sure I'd feel comfortable approving this until all of the new flow tests are completed and we have confidence that every node is being tested. I think there are only 2 or 3 left to pick up.

Bye the way, thanks for adding a test 💯

leenagupte:
> We simply don't have a need to define multiple on_response block for a node; none of the current flows do, and there's no reason logic couldn't be combined into a single on_response block. I'm pretty sure this was just left behind through past iterations of Smart Answers.

I'd forgotten about this PR 😆. I was just curious about how you stumbled across the problem. There's also a couple of inline comments. We're not far off having the flow tests completed. I think the last few are being completed / reviewed.

sihugh (approved):
This seems reasonable, and I think we'd see any problems pretty quickly 👍

PR #5623, opened 2021-09-28
Convert Student finance calculator flow tests

GDSNewt:
[Trello](https://trello.com/c/JAZ46tEM/599-smart-answer-flow-tests-student-finance-calculator)

### **What's changed and why?**

Converts the student finance calculator tests to use the node based testing approach introduced in [ADR-3][1].


test/flows/student_finance_calculator_flow_test.rb
```diff
@@ -0,0 +1,362 @@
...
+    end
+
+    context "next_node" do
+      should "have a next node of how_much_are_your_tuition_fees_per_year? for any response" do
+        assert_next_node :how_much_are_your_tuition_fees_per_year?, for_response: "uk-full-time"
+      end
+    end
+  end
+
+  context "UK full time students" do
```
kevindew, line 44: We shouldn't use contexts to organise nodes. The reason for this is that there's many many permutations of contexts that could occur and there's no way to apply it logically and you'll end up testing the same node logic multiple times. Life ends up simpler if you fully test a node within a single context for that node.

Some of this is explained in the ADR. This example should be useful: https://github.com/alphagov/smart-answers/blob/main/docs/arch/003-testing-flows.md#asserting-a-question-goes-to-a-next-node

test/flows/student_finance_calculator_flow_test.rb
```diff
@@ -0,0 +1,362 @@
...
+        end
+      end
+
+      context "next_node" do
+        should "have a next node of outcome_eu_students" do
+          assert_next_node :outcome_eu_students, for_response: "6935"
+        end
+      end
+    end
+  end
```
kevindew, line 361:
```suggestion
```

As per previous comment, that these contexts should be covered in the individual nodes

test/flows/student_finance_calculator_flow_test.rb
```diff
@@ -0,0 +1,362 @@
...
+        add_responses when_does_your_course_start?: "2020-2021",
+                      what_type_of_student_are_you?: "uk-full-time"
+      end
+
+      should "render the question" do
+        assert_rendered_question
+      end
+
+      context "validation" do
+        should "be invalid if a fee over 9250 is entered for full time students" do
```
kevindew, line 57:
```suggestion
        should "be invalid for a fee greater than the maximum" do
```

It's nicer to capture the intent of the validation rather than the specifics -  particularly since we don't intend to test each permutation of it at this level of testing.

test/flows/student_finance_calculator_flow_test.rb
```diff
@@ -0,0 +1,362 @@
...
+          assert_invalid_response "9251"
```
kevindew, line 58:
```suggestion
          max_for_uk_full_time = SmartAnswer::Calculators::StudentFinanceCalculator::TUITION_FEE_MAXIMUM["full-time"]
          assert_invalid_response (max_for_uk_full_time + 1).to_s
```

We can utilise a (admittedly rather long) constant to make this less of a magic number

test/flows/student_finance_calculator_flow_test.rb
```diff
@@ -0,0 +1,362 @@
...
+        end
+      end
+
+      context "next_node" do
+        should "have a next node of where_will_you_live_while_studying?" do
```
kevindew, line 63: We should have tests here for the next steps you can reach from this node and use `add_responses` to replace the previous response. There's an example of that in [inherit_someone_dies_without_will_flow_test.rb](https://github.com/alphagov/smart-answers/blob/65c061b5fb62ba0d81267bd269b492d4789cdae0/test/flows/inherits_someone_dies_without_will_flow_test.rb#L46-L59).

So I think you'd have something like:

```
%w[uk-full-time uk-part-time].each do |uk_student_type|
  should "have a next node of where_will_you_live_while_studying? for a #{uk_student_type} student" do
    add_responses what_type_of_student_are_you?: uk_student_type
    assert_next_node :where_will_you_live_while_studying?, for_response: "100"
  end
end

%w[eu-full-time eu-part-time].each do |eu_student_type|
  should "have a next node of outcome_eu_students for a #{eu_student_type} student" do
    add_responses what_type_of_student_are_you?: eu_student_type
    assert_next_node :outcome_eu_students, for_response: "100"
  end
end
```

test/flows/student_finance_calculator_flow_test.rb
```diff
@@ -0,0 +1,362 @@
...
+
+    context "question: what_course_are_you_studying?," do
+      setup do
+        testing_node :what_course_are_you_studying?
+        add_responses when_does_your_course_start?: "2020-2021",
+                      what_type_of_student_are_you?: "uk-full-time",
+                      how_much_are_your_tuition_fees_per_year?: "9250",
+                      where_will_you_live_while_studying?: "at-home",
+                      whats_your_household_income?: "50,000",
+                      do_any_of_the_following_apply_uk_full_time_students_only?: "children-under-17,dependant-adult,has-disability,low-income"
```
kevindew, line 137: it'd be an easier read if we went for a simpler response here (I imagine we just need the first one?)

test/flows/student_finance_calculator_flow_test.rb
```diff
@@ -0,0 +1,362 @@
...
+      end
+
+      context "next_node" do
+        should "have a next node of outcome_eu_students" do
+          assert_next_node :outcome_eu_students, for_response: "6935"
+        end
+      end
+    end
+  end
+end
```
kevindew, line 362: We should add tests for the logic in the outcomes. Some examples of this are:

https://github.com/alphagov/smart-answers/blob/65c061b5fb62ba0d81267bd269b492d4789cdae0/app/flows/student_finance_calculator_flow/outcomes/outcome_uk_full_time_students.erb#L10-L48
https://github.com/alphagov/smart-answers/blob/65c061b5fb62ba0d81267bd269b492d4789cdae0/app/flows/student_finance_calculator_flow/outcomes/_nhs_full_time_funding.erb#L3-L29
https://github.com/alphagov/smart-answers/blob/65c061b5fb62ba0d81267bd269b492d4789cdae0/app/flows/student_finance_calculator_flow/outcomes/_tuition_maintenance_summary.erb#L5-L19

There's a brief guide on how to test these in the ADR: https://github.com/alphagov/smart-answers/blob/main/docs/arch/003-testing-flows.md#testing-logic-in-an-outcome. The motivation for doing this is to explore all the executable code in the test suite, which basically offers a guarantee against any basic errors (such as syntax).

But for a good example of it applied to an actual flow see https://github.com/alphagov/smart-answers/blob/65c061b5fb62ba0d81267bd269b492d4789cdae0/test/flows/uk_benefits_abroad_flow_test.rb#L1170-L1196

kevindew (commented):
Great stuff Alex, this is progressing well. It looks like you tripped over one of the core conventions which might mean you need to do a bit of surgery - there's also some logic exploring of outcomes to go

test/flows/student_finance_calculator_flow_test.rb
```diff
@@ -0,0 +1,519 @@
...
+    end
+
+    context "next_node" do
+      should "have a next node do_any_of_the_following_apply_uk_full_time_students_only? for any response" do
+        assert_next_node :do_any_of_the_following_apply_uk_full_time_students_only?, for_response: "50,000"
+      end
+
+      should "have a next node of how_many_credits_will_you_study?" do
+        add_responses what_type_of_student_are_you?: "uk-part-time",
+                      how_much_are_your_tuition_fees_per_year?: "6935",
```
JonathanHallam, line 129: You can delete lines 129 and 130, it only cares about the uk-part-time bit

test/flows/student_finance_calculator_flow_test.rb
```diff
@@ -0,0 +1,519 @@
...
+    should "render the question" do
+      assert_rendered_question
+    end
+
+    context "next_node" do
+      should "have a next node do_any_of_the_following_apply_uk_full_time_students_only? for any response" do
+        assert_next_node :do_any_of_the_following_apply_uk_full_time_students_only?, for_response: "50,000"
+      end
+
+      should "have a next node of how_many_credits_will_you_study?" do
```
JonathanHallam, line 127: I'd add "for part time students" to this line and "for full time students" on 124 just to make it a bit clearer what's being tested

test/flows/student_finance_calculator_flow_test.rb
```diff
@@ -0,0 +1,519 @@
...
+      should "have a next node of outcome_uk_full_time_students for any of the non-medical responses" do
+        assert_next_node :outcome_uk_full_time_students, for_response: "teacher-training"
+      end
+
+      should "have a next node of are_you_a_doctor_or_dentist? for the medical response" do
+        assert_next_node :are_you_a_doctor_or_dentist?, for_response: "dental-medical-healthcare"
+      end
+
+      should "have a next node of outcome_uk_part_time_students for any response" do
+        add_responses what_type_of_student_are_you?: "uk-part-time",
```
JonathanHallam, line 260: Don't need all the responses, just "uk-part time"

test/flows/student_finance_calculator_flow_test.rb
```diff
@@ -0,0 +1,519 @@
...
+                    whats_your_household_income?: "50,000",
+                    do_any_of_the_following_apply_uk_full_time_students_only?: "children-under-17"
+    end
+
+    should "render the question" do
+      assert_rendered_question
+    end
+
+    context "next_node" do
+      should "have a next node of outcome_uk_full_time_students for any of the non-medical responses" do
```
JonathanHallam, line 251: I think we're missing a test here, To make things a bit simpler I'd add the following `should` statements

```
should "have a next node of are_you_a_doctor_or_dentist? for full time uk students given a dental-medical-healthcare response" do

should "have a next node of outcome_uk_full_time_students for full time, non medical, uk students" do

should "have a next node of outcome_uk_part_time_students for part time, non medical, uk students" do

should "have a next node of outcome eu students for eu students" do
```

test/flows/student_finance_calculator_flow_test.rb
```diff
@@ -0,0 +1,519 @@
...
+                    whats_your_household_income?: "50,000",
+                    do_any_of_the_following_apply_uk_full_time_students_only?: "children-under-17",
+                    what_course_are_you_studying?: "teacher-training"
+    end
+
+    should "render text for when the student is eligible for extra grants and allowances" do
+      assert_rendered_outcome text: "Depending on your income and circumstances"
+    end
+
+    should "render text for when the student is not eligible for extra grants and allowances" do
```
JonathanHallam, line 315: Super minor, can we make this the first test because it's the first thing in that outcome

test/flows/student_finance_calculator_flow_test.rb
```diff
@@ -0,0 +1,519 @@
...
+      assert_rendered_outcome text: "Funding for teacher training"
+    end
+
+    should "render text if student is eligible for social work funding" do
+      add_responses what_course_are_you_studying?: "social-work"
+
+      assert_rendered_outcome text: "Social Work Bursary"
+    end
+
+    should "render the full time student text within the extra help partial" do
```
JonathanHallam, line 373: As long as we're testing the `if` statement in that partial in one of the other tests we can delete this test. If we're not testing the `if` somewhere else we can change this to look for the full time text

test/flows/student_finance_calculator_flow_test.rb
```diff
@@ -0,0 +1,519 @@
...
+      add_responses what_course_are_you_studying?: "none-of-the-above"
+
+      assert_rendered_outcome text: "To qualify for a Maintenance Loan, you must be studying one of the following courses"
+    end
+
+    should "not render text if the student is not eligible for a maintenance loan" do
+      add_responses how_many_credits_will_you_study?: "2",
+                    how_many_credits_does_a_full_time_student_study?: "10"
+
+      assert_no_match "How your Maintenance Loan is calculated", @test_flow.outcome_text
```
JonathanHallam, line 412: I'd add the counter test to this as well, so add a test that has `assert_rendered_outcome text: "how your maintenance....

test/flows/student_finance_calculator_flow_test.rb
```diff
@@ -0,0 +1,519 @@
...
+                    what_type_of_student_are_you?: "uk-full-time",
+                    how_much_are_your_tuition_fees_per_year?: "9250",
+                    where_will_you_live_while_studying?: "at-home",
+                    whats_your_household_income?: "50,000",
+                    do_any_of_the_following_apply_uk_full_time_students_only?: "children-under-17",
+                    what_course_are_you_studying?: "dental-medical-healthcare",
+                    are_you_a_doctor_or_dentist?: "yes"
+    end
+
+    should "render text for when the student is undertaking a medical course" do
```
JonathanHallam, line 460: can delete this test because the text isn't hidden behind any logic

JonathanHallam (commented):
This is a properly horrible smart answer given all the view logic but it's really close, just a few minor tweaks and we're there, well done

test/flows/student_finance_calculator_flow_test.rb
GDSNewt: I think they might be necessary here

test/flows/student_finance_calculator_flow_test.rb
GDSNewt: I don't think the EU path ever gets this far so there are only 3 tests I believe

test/flows/student_finance_calculator_flow_test.rb
GDSNewt: Think they are needed here 😬

JonathanHallam: I was wrong, definitely need more than just "uk-part-time" but you can get rid of the ones that are the same in the setup so where-will-you-live and whats-your-household-income

JonathanHallam: in fact, if you change the setup to have `how_much_are_your_tuition_fees_per_year?: "6935",` you can remove that key from the 2nd setup too

test/flows/student_finance_calculator_flow_test.rb
JonathanHallam: you're right. That means that the code itself needs updating because there's logic to deal with eu students that can't be reached. I'd be inclined to just add a commit to this

test/flows/student_finance_calculator_flow_test.rb
JonathanHallam: Sorry, missed this comment when I looked at it earlier. If you change the setup to have tuition fees of 6935 you can remove these two lines

JonathanHallam (approved):
Nice one, thank you :)

PR #5628, opened 2021-10-01
Update business support smart answer outcomes

johnpauldickie:
This PR updates or removes sections in the outcome for the ['Find coronavirus financial support for your business' smart answer](https://www.gov.uk/business-coronavirus-support-finder).

The following sections have been updated in the outcome:

- Coronavirus Job Retention Scheme (furlough scheme)
- Statutory Sick Pay rebate
- Support for businesses paying tax: Time To Pay Arrangement (was 'Support for businesses paying tax: Time To Pay Service')
- VAT reduction for hospitality, accommodation and attractions
- Recovery Loan Scheme
- Apprenticeships

The 'Self-Employment Income Support Scheme (SEISS)' section was removed as the fifth grant as now closed for claims.

⚠️ This repo is Continuously Deployed: make sure you [follow the guidance](https://docs.publishing.service.gov.uk/manual/development-pipeline.html#merge-your-own-pull-request) ⚠️

PR #5630, opened 2021-10-01
Fix time sensitive test

1pretz1:
This test started to fail when the date hit "01/10/2021" due to the
dynamic question `how_many_total_days?, which works out how many days
are available from 1st of October (start of leave year) to the current
date. As the response was 50 days, this test will fail for the next 50
days. Freezing time at "2021-12-01" made available enough days to comply
with the 50 day response.

### DIFF
--- a/test/flows/calculate_agricultural_holiday_entitlement_flow_test.rb
+++ b/test/flows/calculate_agricultural_holiday_entitlement_flow_test.rb
@@ -86,6 +86,7 @@ class CalculateAgriculturalHolidayEntitlementFlowTest < ActiveSupport::TestCase
 
   context "question: how_many_total_days?" do
     setup do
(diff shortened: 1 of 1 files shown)


deborahchua (approved):

kevindew:
@1pretz1 thanks for this, however I think we've managed to double fix this now after @leenagupte fixed this as part of https://github.com/alphagov/smart-answers/pull/5628 with: https://github.com/alphagov/smart-answers/commit/071d11cf3d89023d8700823c8d122db752565458

We should remove one of the fixes, could you revert or edit as appropriate?

1pretz1:
The duplicate fix was resolved in: https://github.com/alphagov/smart-answers/pull/5630

PR #5648, opened 2021-10-19
Add redirect from path to query parameters based URLs

theseanything:
This adds support for legacy path based URLs for flows that have been converted to use query parameters. When navigating to a query parameter based flow with a path based URL, the user will automatically be redirected to equivalent query parameter based URL to continue.

This helps the migration of existing path based flows by preventing user interruption by supporting existing user sessions whilst the updated flow is deployed.


theseanything:

PR #5658, opened 2021-10-29
Remove BaseStateTransitionError and LoggedError classes

theseanything:
This removes the unused LoggedError and BaseStateTransitionError classes.

### BaseStateTransitionError
BaseStateTransitionError is the parent class for the InvalidResponse class, however BaseStateTransitionError is an empty class. Therefore InvalidResponse can inherit directly from StandardError. 

e.g.
before: StandardError -> BaseStateTransitionError (empty class) -> InvalidResponse
after: StandardError -> InvalidResponse

### LoggedError

LoggedError is never instantiated in the codebase and the class is never used. We have a check for LoggedError objects, however this will never be true.

### DIFF
--- a/lib/smart_answer/base_state_transition_error.rb
+++ b/lib/smart_answer/base_state_transition_error.rb
@@ -1,3 +1,0 @@
-module SmartAnswer
-  class BaseStateTransitionError < StandardError; end
-end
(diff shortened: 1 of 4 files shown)


sihugh (approved):
🔪

PR #5667, opened 2021-11-04
Sally Creasey visit visa update Nov 2021

sallycreasey:

======

PH TRELLO: https://trello.com/c/RbOVXmAT/2767-check-if-you-need-a-uk-visa-standard-visitor-updates
RED TEAM TRELLO: Part of https://trello.com/c/IRiPGFex/2942-visitor-visa-guides-part-2-check-if-you-need-a-uk-visa-and-ppe

IMPORTANT NOTE: Needs a fact check.

SUMMARY
Updating information for people who do not need a visa (non visa nationals) to tell them they need to meet the Standard Visitor eligibility requirements even though they don't need a visa.

We've also:
- changed all /standard-visitor-visa links to /standard-visitor for better link hygiene
- correcting negative contractions, style errors and a few accessibility fails across all outcomes
- corrected some factual inaccuracies where we were telling users to apply for the wrong visa or sending them to visa mainstream browse when there was only one visit visa option

### DIFF
--- a/app/flows/check_uk_visa_flow/outcomes/outcome_work_n.erb
+++ b/app/flows/check_uk_visa_flow/outcomes/outcome_work_n.erb
@@ -7,9 +7,9 @@
 
   * invited as an expert in your profession for a ‘Permitted Paid Engagement’
   * visiting for certain business or academic activities, but not working in the UK
-  
-  %You may want to apply for a visa if you have a criminal record or you’ve previously been refused entry into the UK.%
-  
+
+  You must meet the eligibility requirements and only be doing permitted activities.
+
   ##If you’re invited as an expert in your profession
   
   You can stay in the UK for up to 1 month without a visa, but you can only be paid by a UK-based organisation to do certain things, for example:
@@ -18,9 +18,15 @@
   * provide advocacy in legal proceedings
   * take part in arts, entertainment or sporting activities
   
-  Check the list of what you can and cannot be paid to do - it’s the same as [what you can do with a Permitted Paid Engagement visa](/permitted-paid-engagement-visa).
+  Find out what you can and cannot be paid to do if you're [visiting the UK for a Permitted Paid Engagement](/permitted-paid-engagement-visa).
+
+  %You may want to [apply for a Permitted Paid Engagement Visitor visa](/permitted-paid-engagement-visa/apply) if you have a criminal record or you’ve previously been refused entry into the UK.%
 
   ###What you need at the UK border
+
+  <% if calculator.passport_country_in_epassport_gate_list? %>
+    You must go to a border control officer to get a stamp in your passport. Do not use the automatic ePassport gates. You need a stamp to do the activities you came to the UK to do.
+  <% end %>
 
   You must provide a valid passport or travel document. Your passport should be valid for the whole of your stay in the UK.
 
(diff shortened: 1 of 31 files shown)


app/flows/check_uk_visa_flow/outcomes/outcome_work_n.erb
```diff
@@ -7,9 +7,9 @@
 
   * invited as an expert in your profession for a ‘Permitted Paid Engagement’
   * visiting for certain business or academic activities, but not working in the UK
-  
-  %You may want to apply for a visa if you have a criminal record or you’ve previously been refused entry into the UK.%
-  
+
+  You must meet the eligibility requirements and only be doing permitted activities.
+
```
ChrisBAshton, line 12:
```suggestion

```

app/flows/check_uk_visa_flow/outcomes/outcome_work_n.erb
```diff
@@ -62,29 +68,35 @@
...
 
   You must bring a [certificate of sponsorship and evidence of savings](/creative-worker-visa/eligibility) to show officers at the UK border.
 
-  ^ You must see a Border Force officer - do not use the automatic ePassport gates.
-<% end %>
+  <% if calculator.passport_country_in_epassport_gate_list? %>
+    ^ You must see a Border Force officer - do not use the automatic ePassport gates.
+  <% end %>
+
+  <% end %>
```
ChrisBAshton, line 101: Looks like there are two `<% end %>`'s here - should one be removed?

Bit concerning if so, as the test passed regardless! 😬

app/flows/check_uk_visa_flow/questions/channel_islands_or_isle_of_man.erb
```diff
@@ -16,4 +16,4 @@
   <% end %>
 <% end %>
 
-<%= render partial: 'purpose_of_visit_options', locals: { tourism_friends_family: true } %>
```
ChrisBAshton, line 19: Nice 👍

lib/smart_answer/calculators/uk_visa_calculator.rb
```diff
@@ -514,11 +514,42 @@ def study_or_work
...
+      poland
+      portugal
+      romania
       singapore
+      slovakia
+      slovenia
       south-korea
+      spain
+      sweden
+      switzerland
```
ChrisBAshton, line 552: This _looks_ like the intention is to add all EEA countries to this list, which could be done with:

```ruby
COUNTRY_GROUP_EPASSPORT_GATES = %w[
  COUNTRY_GROUP_EEA,
  australia,
  canada,
  japan,
  new-zealand,
  singapore,
  south-korea,
].flatten.freeze
```

...however, [saint-barthelemy](https://github.com/alphagov/smart-answers/blob/10afbc65154fa804c1a93e81a4783e1c7e1828e9/lib/smart_answer/calculators/uk_visa_calculator.rb#L499) and [st-martin](https://github.com/alphagov/smart-answers/blob/10afbc65154fa804c1a93e81a4783e1c7e1828e9/lib/smart_answer/calculators/uk_visa_calculator.rb#L503) are missing from the list you've added, so maybe I've oversimplified? Is it worth checking with content whether these countries should be on the list?

(I wasn't too sure myself, but they do seem to be classed as EEA, according to 40e3a92a68a37501a503edde08e0bc883fd7993a).

ChrisBAshton (commented):
Hi @callumknights - the bulk of the changes look good, but I've spotted a couple of things below, if you could take a look 👀

lib/smart_answer/calculators/uk_visa_calculator.rb
callumknights: This has already been checked with HO, cause I thought the same thing about saint-barthelemy and st-martin, but according to HO they shouldn't be on the list so I had to create a new group with most of the EEA countries included.

ChrisBAshton: Thanks for checking. An alternative would be to do something like `COUNTRY_GROUP_EEA - %w[st-barthelemy st-martin]`, but I don't think it's any better than what you have already 👍 duplicating the country names seems the best approach here, in case the situation gets more complex later.

app/flows/check_uk_visa_flow/outcomes/outcome_work_n.erb
callumknights: Turns out this needs to be there but wasn't indented correctly, it's the `<% end %>` for the govspeak block that starts  on line 5.

ChrisBAshton: Ah, that makes sense 🤦 I figured it was probably needed, sorry if that sent you on a wild goose chase!

ChrisBAshton (approved):
LGTM ✅

PR #5690, opened 2021-12-09
Simplified expenses checker

GDSNewt:

Converts the simplified expenses checker tests to use the node based testing approach introduced in [ADR-3][1].

I've also made some changes to the 'you can use result' outcome view to remove some of the logic that seemed to be no longer relevant. Many of the conditionals in that view were impossible to achieve as the questions would route the user to another view should those conditions be met.  

I also removed some conditions which would appear to be a given, for example if a user has given an input to determine a vehicles cost, something that would only be possible after an initial 'do you have a vehicle question', it seems odd to then have a conditional asking whether they had a vehicle or not. 

[Trello](https://trello.com/c/l7nEU2MJ/34-smart-answer-flow-tests-simplified-expenses-checker)


test/flows/simplified_expenses_checker_flow_test.rb
```diff
@@ -0,0 +1,584 @@
...
+      testing_node :you_can_use_result
+      add_responses vehicle_expense?: "car",
+                    home_or_business_premises_expense?: "using_home_for_business",
+                    buying_new_vehicle?: "no",
+                    capital_allowances?: "simplified_expenses",
+                    hours_work_home?: "100",
+                    current_claim_amount_home?: "10"
+    end
+
+    should "render default outcome text for you_can_use_result" do
```
JonathanHallam, line 463: wouldn't bother with the default stuff. This test is big enough already! :D

test/flows/simplified_expenses_checker_flow_test.rb
```diff
@@ -0,0 +1,584 @@
...
+                    buying_new_vehicle?: "new",
+                    how_much_expect_to_claim?: "50000",
+                    is_vehicle_green?: "low",
+                    price_of_vehicle?: "50",
+                    vehicle_business_use_time?: "50",
+                    drive_business_miles_car_van?: "1000"
+
+      assert_rendered_outcome text: "for the running costs of your car, van or motorcycle"
+    end
+
```
JonathanHallam, line 543: I think we're missing tests for lines 82 - 91 of the outcome. Testing for `of capital allowances for your car, van or motorcycle` and `for working from home`

test/flows/simplified_expenses_checker_flow_test.rb
```diff
@@ -0,0 +1,584 @@
...
+                    home_or_business_premises_expense?: "no_expense",
+                    buying_new_vehicle?: "no",
+                    capital_allowances?: "simplified_expenses"
+    end
+
+    should "render text if respondent picked the simplified_expenses option and did not have any previous expenses" do
+      assert_rendered_outcome text: "You can’t claim capital allowances for your vehicle because you’ve already claimed simplified expenses for it"
+    end
+  end
+end
```
JonathanHallam, line 584: I'd delete these because of the lack of logic. We're already testing that they show up in other bits of this test

JonathanHallam (changes requested):
Most of the comments here are super minor. couple more things to add in another commit but nothing too gnarly, happy to chat about it offline if you like. Cracking work, this stuff is all a bit gross!

test/flows/simplified_expenses_checker_flow_test.rb
GDSNewt: Ahhh I think I omitted tests here because essentially that logic had already been tested for lines 35 - 49 in the outcome. I can flesh out the commit message to make that clearer if you agree with my assessment.

test/flows/simplified_expenses_checker_flow_test.rb
```diff
@@ -0,0 +1,584 @@
...
+                    price_of_vehicle?: "10000",
+                    vehicle_business_use_time?: "50",
+                    drive_business_miles_car_van?: "20000"
+    end
+
+    should "render the question" do
+      assert_rendered_question
+    end
+
+    context "next_node" do
```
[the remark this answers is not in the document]
JonathanHallam: ah sorry, can you also add the validation block to the flow? Should have said, my fault!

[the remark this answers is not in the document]
JonathanHallam: Good work on the test though :D

JonathanHallam (approved):
Brilliant work mate, thanks for persevering

PR #5694, opened 2021-12-15
Add autocomplete component

andysellick:

## What
Add an autocomplete component. This uses the [accessible autocomplete](https://github.com/alphagov/accessible-autocomplete) to enhance an `input` element with an associated `datalist`.

This component is based on [this closed PR in the components gem](https://github.com/alphagov/govuk_publishing_components/pull/2481) but has been significantly stripped down - specifically by enhancing an `input` (with an associated [datalist](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/datalist)) instead of a `select`. The code is also partly based on the autocomplete in `content publisher`.

This PR also adds the code needed to test components in this application, which wasn't present.

This component works by in essence duplicating an existing input inside an autocomplete, then removing the original from the DOM. This means that the user is typing their selection into the input that will be submitted to the backend. 

Other versions of the autocomplete relied upon a hidden select element that was dynamically updated, but there was a potential disconnect between those two elements i.e. it was possible for the user to enter something into the autocomplete that wasn't an option in the select. Worse, it was possible to type in something valid (successfully updating the hidden select) but then change it to something invalid (with no resulting change to the hidden select), making it possible for the page to submit something different to what the user believes they have entered.

The other advantage of this approach is that when JS is unavailable, modern browsers can provide an autocomplete-like functionality to the `input` using the `datalist` values.

## Why
We want to use an autocomplete to let users select countries.

## Visual changes

Screenshot 2021-12-15 at 09 23 29

Without JavaScript:

Screenshot 2022-01-10 at 14 12 56

Trello card: https://trello.com/c/i3brOfzz/431-add-multiple-country-select-question-type-to-smart-answers

### DIFF
--- a/app/assets/javascripts/components/autocomplete.js
+++ b/app/assets/javascripts/components/autocomplete.js
@@ -1,0 +1,47 @@
+/* global accessibleAutocomplete */
+//= require accessible-autocomplete/dist/accessible-autocomplete.min.js
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
+(function (Modules) {
+  function Autocomplete ($module) {
+    this.$module = $module
+    this.$input = this.$module.querySelector('input')
+    this.$datalist = this.$module.querySelector('datalist')
+  }
+
+  Autocomplete.prototype.init = function () {
+    if (!this.$datalist) {
+      return
+    }
+
+    var configOptions = {
+      element: this.$module,
+      id: this.$input.id,
+      defaultValue: this.$input.value,
+      source: this.getDataListContents()
+    }
+
+    accessibleAutocomplete(configOptions) // eslint-disable-line new-cap
+    this.copyInputAttributes()
+  }
+
+  Autocomplete.prototype.getDataListContents = function () {
+    var options = this.$datalist.querySelectorAll('option')
+    var sources = []
+    for (var i = 0; i < options.length; i++) {
+      sources.push(options[i].getAttribute('value'))
+    }
+    return sources
+  }
+
+  // the name attribute isn't copied by the accessible autocomplete code
+  Autocomplete.prototype.copyInputAttributes = function () {
+    var name = this.$input.getAttribute('name')
+    var autoCompleteInput = this.$module.querySelector('.autocomplete__input')
+    autoCompleteInput.setAttribute('name', name)
+    this.$input.parentNode.removeChild(this.$input)
+  }
+
+  Modules.Autocomplete = Autocomplete
+})(window.GOVUK.Modules)
--- a/app/assets/stylesheets/components/_autocomplete.scss
+++ b/app/assets/stylesheets/components/_autocomplete.scss
@@ -1,0 +1,7 @@
+@import "accessible-autocomplete/dist/accessible-autocomplete.min"; // stylelint-disable scss/at-import-partial-extension
+
+.app-c-autocomplete {
+  .autocomplete__option {
+    @include govuk-font(19);
+  }
+}
--- a/app/views/components/_autocomplete.html.erb
+++ b/app/views/components/_autocomplete.html.erb
@@ -1,0 +1,43 @@
+<%
+  id ||= "autocomplete-#{SecureRandom.hex(4)}"
+  name ||= "autocomplete"
+  label ||= nil
+  value ||= nil
+  options ||= []
+
+  error_id = "error-#{SecureRandom.hex(4)}"
+  error_items ||= []
+
+  classes = %w(app-c-autocomplete govuk-form-group)
+  classes << "govuk-form-group--error" if error_items.any?
+
+  if !options.any? || !label
+    raise ArgumentError, "The autocomplete component needs at least one option and a label in order to render."
+  end
+%>
+<%= tag.div class: classes, data: { module: "autocomplete" } do %>
+  <%= render "govuk_publishing_components/components/label", {
+    text: label,
+    html_for: id,
+  } %>
+
+  <% if error_items.any? %>
+    <%= render "govuk_publishing_components/components/error_message", {
+      id: error_id,
+      items: error_items,
+    } %>
+  <% end %>
+
+  <%= tag.input name: name,
+                value: value,
+                id: id,
+                class: "govuk-input",
+                list: "#{id}-list",
+                type: "text" %>
+
+  <%= tag.datalist id: "#{id}-list" do %>
+    <% options.each do |option| %>
+      <%= tag.option(value: option) %>
+    <% end %>
+  <% end %>
+<% end %>
--- a/config/initializers/assets.rb
+++ b/config/initializers/assets.rb
@@ -11,4 +11,10 @@ Rails.application.config.assets.version = "1.0"
 # Precompile additional assets.
 # application.js, application.css, and all non-JS/CSS in the app/assets
 # folder are already added.
-Rails.application.config.assets.precompile += %w[test-dependencies.js]
+Rails.application.config.assets.precompile += %w[
+  test-dependencies.js
+  accessible-autocomplete/dist/accessible-autocomplete.min.js
+  accessible-autocomplete/dist/accessible-autocomplete.min.css
+]
+
+Rails.application.config.assets.paths << Rails.root.join("node_modules")
--- a/yarn.lock
+++ b/yarn.lock
@@ -4,14 +4,14 @@
 
 "@babel/code-frame@^7.0.0", "@babel/code-frame@^7.10.4", "@babel/code-frame@^7.12.11":
   version "7.12.11"
-  resolved "https://registry.yarnpkg.com/@babel/code-frame/-/code-frame-7.12.11.tgz#f4ad435aa263db935b8f10f2c552d23fb716a63f"
+  resolved "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.12.11.tgz"
   integrity sha512-Zt1yodBx1UcyiePMSkWnU4hPqhwq7hGi2nFL1LeA3EUl+q2LQx16MISgJ0+z7dnmgvP9QtIleuETGOiOH1RcIw==
   dependencies:
     "@babel/highlight" "^7.10.4"
 
 "@babel/core@>=7.9.0":
   version "7.12.10"
-  resolved "https://registry.yarnpkg.com/@babel/core/-/core-7.12.10.tgz#b79a2e1b9f70ed3d84bbfb6d8c4ef825f606bccd"
+  resolved "https://registry.npmjs.org/@babel/core/-/core-7.12.10.tgz"
   integrity sha512-eTAlQKq65zHfkHZV0sIVODCPGVgoo1HdBlbSLi9CqOzuZanMv2ihzY+4paiKr1mH+XmYESMAmJ/dpZ68eN6d8w==
   dependencies:
     "@babel/code-frame" "^7.10.4"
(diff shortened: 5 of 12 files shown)


config/initializers/assets.rb
```diff
@@ -11,4 +11,12 @@
...
-Rails.application.config.assets.precompile += %w[test-dependencies.js]
+Rails.application.config.assets.precompile += %w[
+  test-dependencies.js
+  accessible-autocomplete/dist/accessible-autocomplete.min.js
+  accessible-autocomplete/dist/accessible-autocomplete.min.css
+]
+
+Rails.application.config.assets.paths += %W[
+  #{__dir__}/../../node_modules/
+]
```
kevindew, line 22:
```suggestion
Rails.application.config.assets.paths << Rails.root.join("node_modules")
```

is a little cleaner

kevindew (commented):
Code looks good to me 👍  nice job. It's definitely a world simpler without enhancing a select.

The thing that comes to mind is that we might be setting users up for a tough time by not using a select - e.g it'll accept things like a country name in a different case, or missing a character but then user will get caught by validation. Are we planning to be as strict with validation as a select?

Also a minor annoyance I get trying this out in Chrome is the browser autofill appearing over the top. I guess this is because of the label (since the name isn't specific) and looks like there's already attributes to disable autocomplete so I guess browsers are just ignoring it:

Screenshot 2021-12-15 at 11 22 10

app/assets/javascripts/components/autocomplete.js
```diff
@@ -0,0 +1,48 @@
...
+    accessibleAutocomplete(configOptions) // eslint-disable-line new-cap
+    this.copyInputAttributes()
+  }
+
+  Autocomplete.prototype.getDataListContents = function () {
+    var options = this.$datalist.querySelectorAll('option')
+    for (var i = 0; i < options.length; i++) {
+      this.sources.push(options[i].getAttribute('value'))
+    }
+  }
```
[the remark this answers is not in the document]
andysellick: Good shout, changed.

config/initializers/assets.rb
andysellick: Thanks for the suggestion, have changed.

app/assets/javascripts/components/autocomplete.js
```diff
@@ -0,0 +1,48 @@
...
+
+  // the name attribute isn't copied by the accessible autocomplete code
+  Autocomplete.prototype.copyInputAttributes = function () {
+    var name = this.$input.getAttribute('name')
+    var autoCompleteInput = this.$module.querySelector('.autocomplete__input')
+    autoCompleteInput.setAttribute('name', name)
+    this.$input.parentNode.removeChild(this.$input)
+  }
```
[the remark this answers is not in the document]
andysellick: No, it doesn't. It's a bit annoying, because that would be the best solution - the same input would be used for both the user to type in and for the backend to accept the value. But instead we have to manually update the cloned input that the autocomplete creates and delete the original one from the DOM to ensure its still submitted correctly.

This is the behaviour of the existing component in `content-publisher` (although I'm not sure it updates the `name` attribute).

andysellick:
@kevindew thanks for the review, I've made a few changes based on your comments.

> The thing that comes to mind is that we might be setting users up for a tough time by not using a select - e.g it'll accept things like a country name in a different case, or missing a character but then user will get caught by validation. Are we planning to be as strict with validation as a select?

I agonised over this a bit. The thinking behind it was that if we enhance a `select` there's still a disconnect between what the user types and what is actually submitted. These are:

- the user could type anything in the autocomplete, so we need the backend to check that a valid country was chosen anyway
- the user could choose something valid in the autocomplete (like `United Kingdom`) but then retype it to something that they think is valid but doesn't match the value in the hidden select (like `United States of America`, when the select contains `United States`). They then submit the form thinking they've chosen one option, but in fact have chosen a different one, because the hidden select wouldn't be updated as no matching option would be found by the autocomplete code.

> Also a minor annoyance I get trying this out in Chrome is the browser autofill appearing over the top.

That's concerning, I'll catch up with direct about that.

andysellick:
@kevindew had a chat with the DS team and apparently the autofill problem you found is a known issue: https://github.com/alphagov/reported-bugs/issues/47

It's a bug/feature of Chrome, which is that Chrome doesn't respect `autocomplete="off"`. This [chromium bug report](https://bugs.chromium.org/p/chromium/issues/detail?id=914451#c53) contains way more detail and a number of suggested hacks, but I'm not comfortable implementing a "hack" on GOV.UK.

Not sure where that leaves us - a known problem without a solution. Can ask - when you got this, did it override the accessible-autocomplete's dropdown when you attempt to use it with a keyboard? In the screenshot I can see the autofill overlaps it but I wasn't sure which gets precedence when the cursor keys are invoked.

app/assets/javascripts/components/autocomplete.js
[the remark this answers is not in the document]
kevindew: Sure thing - just an oddity from the library.

app/views/components/_autocomplete.html.erb
```diff
@@ -0,0 +1,41 @@
...
+  value ||= nil
+  options ||= []
+
+  error_id = "error-#{SecureRandom.hex(4)}"
+  error_items ||= []
+
+  classes = %w(app-c-autocomplete govuk-form-group)
+  classes << "govuk-form-group--error" if error_items.any?
+%>
+<% if label && options.any? %>
```
kevindew, line 14: Shouldn't we error if these aren't provided. It feels like unexpected behaviour for the component to quietly do nothing when the wrong or missing inputs are given (I'm getting flash backs to the incident behind https://github.com/alphagov/govuk_publishing_components/pull/2335)

kevindew (commented):
> I agonised over this a bit. The thinking behind it was that if we enhance a `select` there's still a disconnect between what the user types and what is actually submitted. These are:
> 
> * the user could type anything in the autocomplete, so we need the backend to check that a valid country was chosen anyway
> * the user could choose something valid in the autocomplete (like `United Kingdom`) but then retype it to something that they think is valid but doesn't match the value in the hidden select (like `United States of America`, when the select contains `United States`). They then submit the form thinking they've chosen one option, but in fact have chosen a different one, because the hidden select wouldn't be updated as no matching option would be found by the autocomplete code.

Yeah, I have recollections of us battling this for Content Publisher and (having just played with CP now) I don't think we got it to somewhere ideal.

I imagine what you want is for it to always change back to your last selection or blank on blur whenever the input contents don't match a select option - like this one behaves: https://react-select.com/home

Well this seems like something to try out 👍 

> @kevindew had a chat with the DS team and apparently the autofill problem you found is a known issue: [alphagov/reported-bugs#47](https://github.com/alphagov/reported-bugs/issues/47)
> 
> It's a bug/feature of Chrome, which is that Chrome doesn't respect `autocomplete="off"`. This [chromium bug report](https://bugs.chromium.org/p/chromium/issues/detail?id=914451#c53) contains way more detail and a number of suggested hacks, but I'm not comfortable implementing a "hack" on GOV.UK.
> 
> Not sure where that leaves us - a known problem without a solution. Can ask - when you got this, did it override the accessible-autocomplete's dropdown when you attempt to use it with a keyboard? In the screenshot I can see the autofill overlaps it but I wasn't sure which gets precedence when the cursor keys are invoked.

Thanks for digging into it, that does sound thorny - that Chrome bug thread is a bit depressing!

Looks like Chrome's receives precedence I'm afraid. I made a GIF to demo:

Dec-17-2021 17-40-20

app/views/components/_autocomplete.html.erb
andysellick: I think most of our components fail quietly if the right data isn't passed to them? Do you have an example of component that doesn't?

andysellick: @kevindew I've found an example and updated this PR so that the autocomplete raises an error if the right things aren't passed to it - see the extra commit (will squash once approved).

app/assets/stylesheets/components/_autocomplete.scss
```diff
@@ -0,0 +1,7 @@
+@import url(asset-path("accessible-autocomplete/dist/accessible-autocomplete.min.css")); // stylelint-disable function-url-quotes
```
kevindew, line 1: If i understand right this is creating a css import? Whereas we just want the CSS file imported and save a request as per: https://github.com/alphagov/content-publisher/blob/8240b79bbfbab91e24530b7f9f178f0ef541cc71/app/assets/stylesheets/components/_autocomplete.scss#L2

We made the change in Content Publisher in: https://github.com/alphagov/content-publisher/commit/40bc73a1540f51d59fe688de7be3bbf8befe7d60

app/views/components/_autocomplete.html.erb
```diff
@@ -0,0 +1,43 @@
...
+
+  error_id = "error-#{SecureRandom.hex(4)}"
+  error_items ||= []
+
+  classes = %w(app-c-autocomplete govuk-form-group)
+  classes << "govuk-form-group--error" if error_items.any?
+
+  if !options.any? || !label
+    raise ArgumentError, "The autocomplete component needs at least one option and a label in order to render."
+  end
```
kevindew, line 16: This looks good 👍  much safer than failing silently

yarn.lock
```diff
@@ -4,14 +4,14 @@
 
 "@babel/code-frame@^7.0.0", "@babel/code-frame@^7.10.4", "@babel/code-frame@^7.12.11":
   version "7.12.11"
-  resolved "https://registry.yarnpkg.com/@babel/code-frame/-/code-frame-7.12.11.tgz#f4ad435aa263db935b8f10f2c552d23fb716a63f"
+  resolved "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.12.11.tgz"
```
kevindew: How odd that these all changed from yarnpkg.com to npmjs.org 😕

kevindew (commented):
Looks good - sorry to not just approve I think there's the one CSS thing to fix.

app/views/components/_autocomplete.html.erb
kevindew: Super. It's a shame that this is a pattern that has crept into the components, we should try reverse that.

app/assets/stylesheets/components/_autocomplete.scss
andysellick: Ah that's much cleaner, thanks, have updated.

app/assets/stylesheets/components/_autocomplete.scss
```diff
@@ -0,0 +1,7 @@
+@import "accessible-autocomplete/dist/accessible-autocomplete.min.css";
```
kevindew, line 1: Do you know if this definitely imports correctly? in Content Publisher we don't use the extension of ".css" and I'm not sure whether that triggers a CSS import

app/views/components/_autocomplete.html.erb
andysellick: One to consider definitely. It looks as though that change to the image card has been reverted, although I don't remember the exact reasons behind it.

app/assets/stylesheets/components/_autocomplete.scss
andysellick: I wrote it first like in the example you gave and it worked in the browser but raised a linting error, so I tried it with the `.css` extension and both worked.

kevindew: Looking on the heroku deploy I can see it comes out as:

```
@import url(/assets/smartanswers/accessible-autocomplete/dist/accessible-autocomplete.min-53a3e4a863bb49fa038824ac9bb9f71099b3ca20b68ab846f6942c700da65770.css);.govuk-button,.govspeak .gem-c-button,.gem-c-govspeak .gem-c-button{font-family:"GDS Transport",arial,sans-serif;-webkit-font-smoothing:antialiased;-moz-osx-font-smoothing:grayscale;font-weight:400;font-size:16px;font-size:1rem;line-height:1.1875;box-sizing:border-box;display:inline-block;position:relative;width:100%;margin-top:0;margin-right:0;margin-left:0;margin-bottom:22px;padding:8px 10px 7px;border:2px solid transparent;border-radius:0;color:#fff;background-color:#00703c;box-shadow:0 2px 0 #002d18;text-align:center;vertical-align:top;cursor:pointer;-webkit-appearance:none}/*! 
```

andysellick: That doesn't look great... I've changed it back to just `.min` and put in a linter disable rule.

kevindew: Super - yeah my understanding is that adding ".css" converts it to a CSS import - which is rarely what you want.

kevindew (approved):

PR #5725, opened 2022-01-19
Remove vaccination-status from Google Analytics

gclssvglx:
### What

Remove the response to vaccination-status in the covid-travel-abroad flow from all data that we send to Google Analytics.

### Why

A users vaccination status is considered health data so we should remove it from being exposed outside of the application.

### Before

Screenshot 2022-01-19 at 16 23 19

### After

Screenshot 2022-01-19 at 16 21 57

[Trello](https://trello.com/c/DInfO64C/540-innout-decide-what-to-do-from-a-data-privacy-angle)

### DIFF
--- a/app/assets/javascripts/modules/track-responses.js
+++ b/app/assets/javascripts/modules/track-responses.js
@@ -10,14 +10,25 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     this.$module.addEventListener('submit', this.handleFormSubmit.bind(this))
   }
 
+  TrackResponses.prototype.filterLabel = function () {
+    return 'vaccination-status'
+  }
+
+  TrackResponses.prototype.filterResponse = function (questionKey, label) {
+    var filteredLabel = label
+    if (this.filterLabel() === questionKey) {
+      filteredLabel = '[FILTERED]'
+    }
+    return { transport: 'beacon', label: filteredLabel }
+  }
+
   TrackResponses.prototype.handleFormSubmit = function (event) {
     var submittedForm = event.target
     var questionKey = this.getQuestionKey(submittedForm)
     var responseLabels = this.getResponseLabels(submittedForm)
 
     for (var i = 0; i < responseLabels.length; i++) {
-      var label = responseLabels[i]
-      var options = { transport: 'beacon', label: label }
+      var options = this.filterResponse(questionKey, responseLabels[i])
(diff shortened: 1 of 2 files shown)


spec/javascripts/modules/track-responses.spec.js
```diff
@@ -82,6 +82,17 @@ describe('Track responses', function () {
       )
     })
 
+    it('filters out sensitive responses', function () {
```
andysellick, line 85: The tests you've added appear to be the same but in different contexts, is that correct? Just wanted to check are they all necessary?

app/assets/javascripts/modules/track-responses.js
```diff
@@ -10,14 +10,25 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     this.$module.addEventListener('submit', this.handleFormSubmit.bind(this))
   }
 
+  TrackResponses.prototype.filterLabels = function () {
+    return ['vaccination-status']
```
andysellick, line 14: This returns an array of only one value. Are we expecting to have to expand this at some point, or could it return a single string? If we're expecting to expand it, can we have a test that checks the code works if an array of more than one value is returned from this function?

andysellick (commented):
Looks good, just a couple of questions. I assume the alternative way of stripping stuff from GA using meta tags didn't work here?

spec/javascripts/modules/track-responses.spec.js
gclssvglx: Yes you're right, two are the same - the third one (select) is subtlety different. I wanted to test this only once - ideally - however the tests are structured around input type (radio, checkbox and select). Both of these caused me to duplicate it. Would testing this just once - in a separate `describe` block - be ok and work for all input types? That would be my preference.

app/assets/javascripts/modules/track-responses.js
gclssvglx: Ah! You caught me - I was thinking this wouldn't be the only one over time - but currently it is.  I'll update to a single string for now.

gclssvglx: Updated for single value

spec/javascripts/modules/track-responses.spec.js
andysellick: Thanks for the explanation. No, it sounds sensible the way you've done it, thanks.

andysellick (approved):

PR #5736, opened 2022-01-25
[DO NOT MERGE] Add config to remove vaccination status from url

gclssvglx:
This adds the flag that's checked by the publishing components gem so that vaccination status is stripped from the URL saved in GA. It also adds a test to check that the validation status exists and hasn't been renamed. This has been done to make sure that this can't be changed without something breaking. If it were renamed there's a chance it would silently fail and expose data. 

Linked - https://github.com/alphagov/govuk_publishing_components/pull/2582/

[Trello](https://trello.com/c/DInfO64C/540-mvp-innout-decide-what-to-do-from-a-data-privacy-angle)

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -25,7 +25,8 @@
       <%= render "govuk_publishing_components/components/meta_tags",
         content_item: content_item,
         strip_dates_pii: true,
-        strip_postcode_pii: true %>
+        strip_postcode_pii: true,
+        strip_vaccination_status_pii: true %>
     <% end %>
     <% if @meta_section %>
       <meta name="govuk:section" content="<%= @meta_section %>">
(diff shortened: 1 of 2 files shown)


kevindew:
If I understand correctly you don't need to make a change to govuk_publishing_components to do this. You can use https://github.com/alphagov/govuk_publishing_components/blob/dcaa0800ca66c920ccb195b796a67496c6d28e23/app/assets/javascripts/govuk_publishing_components/analytics/pii.js#L34 as a way to tell govuk_publishing_components which query string parameters to remove.

So if we have a meta tag of: `<meta name="govuk:static-analytics:strip-query-string-parameters" content="vaccination_status">` it'll do it without any changes to components gem.

gclssvglx:
> So if we have a meta tag of: <meta name="govuk:static-analytics:strip-query-string-parameters" content="vaccination_status"> it'll do it without any changes to components gem.

Great suggestion, will take a look at this - however it does look like we'll be going down the `session` flow route.

PR #5747, opened 2022-01-31
Honour Content Store Cache-Control Header

bilbof:
Partially implements [RFC 144](https://github.com/alphagov/govuk-rfcs/blob/main/rfc-144-consistent-cache-expiry.md)

https://trello.com/c/OaUhq6uI/722

### DIFF
--- a/app/controllers/smart_answers_controller.rb
+++ b/app/controllers/smart_answers_controller.rb
@@ -3,9 +3,7 @@ class SmartAnswersController < ApplicationController
 
   before_action :find_smart_answer, except: %w[index]
   before_action :redirect_response_to_canonical_path, only: %w[show]
-  before_action :setup_content_item, except: %w[index]
-
-  attr_accessor :content_item
+  before_action :content_item, except: %w[index]
 
   rescue_from SmartAnswer::FlowRegistry::NotFound, with: :error_404
   rescue_from SmartAnswer::InvalidNode, with: :error_404
@@ -25,9 +23,8 @@ class SmartAnswersController < ApplicationController
   def show
     @title = @presenter.title
 
-    render page_type, formats: [:html]
-
     set_expiry
+    render page_type, formats: [:html]
   end
 
   def visualise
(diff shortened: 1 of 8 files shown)


app/controllers/smart_answers_controller.rb
```diff
@@ -73,13 +70,18 @@ def redirect_response_to_canonical_path
...
-  def setup_content_item
-    @content_item = ContentItemRetriever.fetch(params[:id])
+  def set_expiry
+    return unless Rails.configuration.set_http_cache_control_expiry_time
+
+    public_cache = content_item.dig("cache_control", "public")
+
+    expires_in(
+      content_item.dig("cache_control", "max-age") || 5.minutes.to_i,
+      public: public_cache.nil? ? true : public_cache,
```
sengi, line 84: I found this quite hard to understand. Also it looks like `public_cache.nil? ? true : public_cache` inverts the sense of the boolean, if I'm reading it right. I'm assuming it's intended to be `true` if `{cache_control => "public"}` and `false` otherwise?

Consider something like this, which would make it clearer that it's supposed to produce a bool, as well as simplifying the logic:

```suggestion
    expires_in(
      content_item.dig("cache_control", "max-age") || 5.minutes.to_i,
      public: !!content_item.dig("cache_control", "public"),
```

I had a look through the style guide and I think this would be an acceptable use of `!!`, since we actually want to be explicit about producing a bool here, as opposed to using it unnecessarily in a control flow conditional. If Ruby had an explicit `boolean()` cast or something then that'd be better still, but sadly I couldn't find any such thing.

bilbof: You're right, this is convoluted. This is handling the case where `content_item.dig("cache_control", "public")` is `nil`. We want the default to be `true` but unfortunately `!!nil == false`.

Is this clearer? (second arg to `fetch` is a default)

```
!!content_item.fetch("cache_control", {}).fetch("public", true)
```

sengi: chitty chitty bang bang

sengi: Oh, sorry I hadn't twigged that we want it to default to public. Your suggestion LGTM though I would probably drop the `!!` now since we no longer need it (since the "true" makes it fairly clear that we're dealing with a bool).

sengi (approved):

PR #5756, opened 2022-02-03
Add ability to override the Before you start text

gclssvglx:
### What

The content designers would like to override the 'Before you start' text on the COVID travel Start page.

Unfortunately it is part of the Smart Answer landing template.

### Why

The desired content doesn't fit in with the current approach of a list of prerequisites, or pre-start information.

### Before

Screenshot 2022-02-03 at 14 22 19

### After

Screenshot 2022-02-03 at 14 20 57

Co-authored-by: Alex Newton <alex.newton@digital.cabinet-office.gov.uk>

[Trello](https://trello.com/c/TLpYQ1s4/588-mvp-if-possible-remove-the-before-you-start-text-from-the-start-page)

### DIFF
--- a/app/flows/check_travel_during_coronavirus_flow/start.erb
+++ b/app/flows/check_travel_during_coronavirus_flow/start.erb
@@ -24,13 +24,11 @@
   $CTA
 <% end %>
 
-<% govspeak_for :post_body do %>
-  <%= render "govuk_publishing_components/components/heading", {
-    text: "Travelling to England from within the UK, Channel Islands and the Isle of Man",
-    font_size: "m",
-    margin_bottom: 4,
-  } %>
+<% text_for :post_body_header do %>
+  Travelling to England from within the UK, Channel Islands and the Isle of Man
+<% end %>
 
+<% govspeak_for :post_body do %>
   <p>If you’re travelling to England from within the Common Travel Area (England, Scotland, Wales and Northern Ireland, Ireland, the Channel Islands and the Isle of Man) and haven’t been anywhere else within the 10 days before you arrive, there are no entry requirements for coming into England.</p>
   <p>If you’ve been to a country or territory outside the Common Travel Area within the 10 days before you arrive in England, you must follow the rules for entering England from that country. You can use this tool to find out those rules.</p>
 <% end %>
--- a/app/presenters/start_node_presenter.rb
+++ b/app/presenters/start_node_presenter.rb
@@ -19,6 +19,11 @@ class StartNodePresenter < NodePresenter
     @renderer.content_for(:body)
   end
 
+  def post_body_header
+    custom_header_text = @renderer.content_for(:post_body_header)
+    custom_header_text.presence || "Before you start"
+  end
+
   def post_body
     @renderer.content_for(:post_body)
   end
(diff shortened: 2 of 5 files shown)


test/unit/start_node_presenter_test.rb
```diff
@@ -56,5 +56,17 @@ class StartNodePresenterTest < ActiveSupport::TestCase
...
     end
+
+    test "#post_body_header returns custom text when supplied" do
+      @renderer.stubs(:content_for).with(:post_body_header).returns("post-body-header-text")
+
+      assert_equal "post-body-header-text", @presenter.post_body_header
+    end
+
+    test '#post_body_header returns "Before you start" when there is no custom header text' do
+      @renderer.stubs(:content_for).with(:post_body_header).returns("Before you start")
```
JonathanHallam, line 67: Should "Before you start" in this test not be `nil`?

JonathanHallam: Just to replicate what would happen when `:post_body_heaader` isn't defined

GDSNewt: Yes you are absolutely right, thanks Jon 👍

app/flows/check_travel_during_coronavirus_flow/start.erb
```diff
@@ -24,13 +24,11 @@
   $CTA
 <% end %>
 
-<% govspeak_for :post_body do %>
-  <%= render "govuk_publishing_components/components/heading", {
-    text: "Travelling to England from within the UK, Channel Islands and the Isle of Man",
-    font_size: "m",
-    margin_bottom: 4,
-  } %>
+<% text_for :post_body_header do %>
```
leenagupte, line 27: Very clever! 👏

app/presenters/start_node_presenter.rb
```diff
@@ -19,6 +19,11 @@ def body
     @renderer.content_for(:body)
   end
 
+  def post_body_header
+    custom_header_text = @renderer.content_for(:post_body_header)
+    custom_header_text.presence || "Before you start"
```
leenagupte, line 24: This is nice! I think that the 23 other smart-answers remain unaffected.

leenagupte (commented):
This is a really neat solution. The only thing that's missing is a bit of documentation about how it works. See: https://github.com/alphagov/smart-answers/blob/main/docs/design/erb-templates/landing-page-template.md

app/presenters/start_node_presenter.rb
gclssvglx: Yeah, we tried not to mess with any of the other flows

gclssvglx:
> This is a really neat solution. The only thing that's missing is a bit of documentation about how it works. See: https://github.com/alphagov/smart-answers/blob/main/docs/design/erb-templates/landing-page-template.md

Ah yes, quite right - thanks for pointing this out @leenagupte 👍  Doc updated.

docs/design/erb-templates/landing-page-template.md
```diff
@@ -51,4 +51,12 @@ Used to generate supplementary content (appearing below the start button). Expec
 <% end %>
 ```
 
+Optionally, it is possible to supply a custom header for this section. If supplied, this will override the default "Before you start" heading and can only be text. Example:
```
leenagupte, line 54: Thanks!

leenagupte (approved):
This is a really nice solution 👍

PR #5764, opened 2022-02-08
Codify vaccination_status in check_travel_during_coronavirus flow

gclssvglx:
## What

Currently the `vaccination_status` question in the `check_travel_during_coronavirus` flow has the following possible responses:

* "vaccinated": "I will have had a second dose of an approved COVID-19 vaccine 14 or more days before I travel"
* "in_trial": "I'm taking part in an approved COVID-19 vaccine trial in the UK"
* "exempt": "I cannot have a COVID-19 vaccination for a medical reason approved by a clinician"
* "none": "None of the above"

As this flow is URL-based, we would see something like this `vaccination_status=vaccinated` in the URL when the question is answered.

We need to obfuscate (codify) the response in the URL. For example:

* "123456": "I will have had a second dose of an approved COVID-19 vaccine 14 or more days before I travel"
* "234567": "I'm taking part in an approved COVID-19 vaccine trial in the UK"
* "345678": "I cannot have a COVID-19 vaccination for a medical reason approved by a clinician"
* "456789": "None of the above"

`SecureRandom.hex(8)` has been used to generate the obfuscated codes as this is unlikely to generate duplicates if more response values are added, are long enough to unique and short enough to be human usable.

## Why

Vaccination status is considered PII Health data and so we need to replace this question/response as plain text from the URL with a codified version.

## Before

Screenshot 2022-02-08 at 08 10 08

Screenshot 2022-02-08 at 08 22 50

Screenshot 2022-02-08 at 09 50 57

## After

Screenshot 2022-02-08 at 08 08 38

Screenshot 2022-02-08 at 08 22 28

Screenshot 2022-02-08 at 09 50 11

[Trello](https://trello.com/c/wwMtYcRM/621-codify-vaccinationstatus)

### DIFF
--- a/test/flows/check_travel_during_coronavirus_flow_test.rb
+++ b/test/flows/check_travel_during_coronavirus_flow_test.rb
@@ -352,11 +352,11 @@ class CheckTravelDuringCoronavirusFlowTest < ActiveSupport::TestCase
 
       should "render guidance for people who aren't fully vaccinated" do
         add_responses vaccination_status: "9ddc7655bfd0d477"
-        assert_rendered_outcome text: "people who aren't fully vaccinated"
+        assert_rendered_outcome text: "Returning to England if you're not fully vaccinated"
       end
 
       should "render guidance for people who are fully vaccinated" do
-        assert_rendered_outcome text: "fully vaccinated people"
+        assert_rendered_outcome text: "Returning to England if you're fully vaccinated"
       end
 
       should "render guidance for people travelling with children" do
--- a/test/flows/check_travel_during_coronavirus_flow_test.rb
+++ b/test/flows/check_travel_during_coronavirus_flow_test.rb
@@ -172,7 +172,7 @@ class CheckTravelDuringCoronavirusFlowTest < ActiveSupport::TestCase
     context "next_node" do
       should "have a next node of travelling_with_young_people " \
                 "if the user is not travelling to a red list country " do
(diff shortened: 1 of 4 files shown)


test/flows/check_travel_during_coronavirus_flow_test.rb
```diff
@@ -352,11 +352,11 @@ class CheckTravelDuringCoronavirusFlowTest < ActiveSupport::TestCase
 
       should "render guidance for people who aren't fully vaccinated" do
         add_responses vaccination_status: "9ddc7655bfd0d477"
-        assert_rendered_outcome text: "people who aren't fully vaccinated"
+        assert_rendered_outcome text: "Returning to England if you're not fully vaccinated"
```
JonathanHallam, line 355: We should probably think about internationalising this at some point to avoid having to do things like this

JonathanHallam (approved):
Looks good to me. I've checked in browser and the URL bit works which means it won't end up in the logs 🎉

test/flows/check_travel_during_coronavirus_flow_test.rb
gclssvglx: Agree. However, and perhaps surprisingly - at the moment the app is [not I18N enabled](https://github.com/alphagov/smart-answers/blob/main/config/locales/en.yml). I suspect to do so would be quite a large task.

PR #5771, opened 2022-02-09
Change content around summary text and vaccine exemption

JonathanHallam:
Some changes were agreed to the summary text if a user is coming from Ireland and only Ireland. This makes those changes, I had to include a new method to check if a user was only coming from Ireland to keep the frontend logic free.

This also adds a section to the "fully vaccinated" partial to include people who are exempt from vaccination.

## Ireland
Screenshot 2022-02-09 at 09 40 49

## Vax exempt
Screenshot 2022-02-09 at 14 20 57

## Trello 
https://trello.com/c/3YfXGaqp/628-adding-inbound-results-for-people-exempt-from-vax-mvp
https://trello.com/c/fPtwW9zX/635-content-changes-to-summary-text-because-you-said-mvp

### DIFF
--- a/app/flows/check_travel_during_coronavirus_flow/outcomes/_fully_vaxed.erb
+++ b/app/flows/check_travel_during_coronavirus_flow/outcomes/_fully_vaxed.erb
@@ -12,17 +12,35 @@
 
   <p class="govuk-body">You do not need to take any COVID-19 tests before you travel to England or after you arrive.</p>
 
-  <p class="govuk-body">You must be able to prove that you’ve been fully vaccinated. If you live in England, you can prove your vaccination status using the <a href="https://www.gov.uk/guidance/nhs-covid-pass" class="govuk-link">NHS COVID Pass</a>.</p>
+  <p class="govuk-body">You must <a href="/provide-journey-contact-details-before-travel-uk" class="govuk-link">complete a passenger locator form</a> in the 48 hours before you arrive in England.</p>
 
-  <p class="govuk-body">If you were vaccinated in another country or territory, see <a href="https://www.gov.uk/guidance/countries-with-approved-covid-19-vaccination-programmes-and-proof-of-vaccination#countries-with-an-approved-proof-of-vaccination-and-examples-of-proof-required" class="govuk-link">examples of what you can use as proof of vaccination</a>.</p>
+  <%= render "govuk_publishing_components/components/heading", {
+    text: "Proof of vaccination",
+    heading_level: 3,
+    font_size: "m",
+    margin_bottom: 4,
+  } %>
 
-  <p class="govuk-body">You must <a href="https://www.gov.uk/provide-journey-contact-details-before-travel-uk" class="govuk-link">complete a passenger locator form</a> in the 48 hours before you arrive in England.</p>
+  <p class="govuk-body">You must be able to prove that you’ve been fully vaccinated against COVID-19. If you live in England, you can prove your vaccination status using the <a href="/guidance/nhs-covid-pass" class="govuk-link">NHS COVID Pass</a>.</p>
+
+  <p class="govuk-body">If you were vaccinated in another country or territory, see <a href="/guidance/countries-with-approved-covid-19-vaccination-programmes-and-proof-of-vaccination#countries-with-an-approved-proof-of-vaccination-and-examples-of-proof-required" class="govuk-link">examples of what you can use as proof of vaccination</a>.</p>
+
+  <%= render "govuk_publishing_components/components/heading", {
+    text: "Proof that you cannot have a COVID-19 vaccination",
+    heading_level: 3,
+    font_size: "m",
+    margin_bottom: 4,
+  } %>
+
+  <p class="govuk-body">If you cannot have a COVID-19 vaccination for a medical reason approved by a clinician, you will need to show proof using your <a href ="/guidance/nhs-covid-pass" class="govuk-link">NHS COVID Pass</a>.</p>
+
+  <p class="govuk-body">If you do not have proof in your NHS COVID Pass, you will need to <a href="/guidance/covid-19-medical-exemptions-proving-you-are-unable-to-get-vaccinated" class="govuk-link">apply using the NHS COVID Pass service on 119</a> and ask for a medical exemptions application form.</p>
 <% end %>
 
 <%
   data = {
-    title: "Returning to England if you’re fully vaccinated",
-    reasons: ['You are fully vaccinated'],
+    title: "Returning to England if you’re fully vaccinated or cannot have a COVID-19 vaccination",
+    reasons: ['You are fully vaccinated or cannot have a COVID-19 vaccination'],
     content: content
   }
 %>
--- a/app/flows/check_travel_during_coronavirus_flow/outcomes/_fully_vaxed.erb
+++ b/app/flows/check_travel_during_coronavirus_flow/outcomes/_fully_vaxed.erb
@@ -40,7 +40,6 @@
 <%
   data = {
     title: "Returning to England if you’re fully vaccinated or cannot have a COVID-19 vaccination",
-    reasons: ['You are fully vaccinated or cannot have a COVID-19 vaccination'],
     content: content
   }
 %>
--- a/app/flows/check_travel_during_coronavirus_flow/outcomes/_travel_result_layout.erb
+++ b/app/flows/check_travel_during_coronavirus_flow/outcomes/_travel_result_layout.erb
@@ -7,21 +7,14 @@
 
 <div class="govuk-grid-row travel-abroad__section">
   <div class="govuk-grid-column-one-third travel-abroad__reasons">
-    <p class="govuk-body govuk-body-s">Because you said</p>
-    <ul class="govuk-list govuk-list--bullet govuk-body-s">
-      <% data[:reasons].each do |reason| %>
-        <li><%= reason %></li>
-      <% end %>
-      <% if calculator.travelling_with_young_people? %>
-        <li>You are travelling with someone aged 17 and under</li>
-      <% end %>
-      <% if calculator.travelling_with_children.include?("zero_to_four") %>
-        <li>You are travelling with someone aged 4 and under</li>
-      <% end %>
-      <% if calculator.travelling_with_children.include?("five_to_seventeen") %>
-        <li>You are travelling with someone aged 5 to 17</li>
-      <% end %>
-    </ul>
+    <% if calculator.summary_text_fields.any? %>
+      <p class="govuk-body govuk-body-s">Because you said:</p>
+      <ul class="govuk-list govuk-list--bullet govuk-body-s">
+        <% calculator.summary_text_fields.each do |field| %>
+          <li><%= t("check_travel_during_coronavirus.summary_text.#{field}") %></li>
+        <% end %>
+      </ul>
+    <% end %>
   </div>
 
   <div class="govuk-grid-column-two-thirds">
--- a/app/flows/check_travel_during_coronavirus_flow/outcomes/results.erb
+++ b/app/flows/check_travel_during_coronavirus_flow/outcomes/results.erb
@@ -24,14 +24,16 @@
   <% calculator.country_locations.each do |country| %>
     <div class="govuk-grid-row travel-abroad__country">
       <div class="govuk-grid-column-one-third travel-abroad__reasons">
-        <p class="govuk-body govuk-body-s">Because you said</p>
-        <ul class="govuk-list govuk-list--bullet govuk-body-s">
-          <% if calculator.transit_countries.include?(country.slug) %>
-            <li>You are travelling through <%= country.title %></li>
-          <% else %>
-            <li>You are travelling to <%= country.title %></li>
-          <% end %>
-        </ul>
+        <% unless calculator.travelling_to_ireland? && calculator.single_journey? %>
+          <p class="govuk-body govuk-body-s">Because you said:</p>
+          <ul class="govuk-list govuk-list--bullet govuk-body-s">
+            <% if calculator.transit_countries.include?(country.slug) %>
+              <li>You are travelling through <%= country.title %></li>
+            <% else %>
+              <li>You are travelling to <%= country.title %></li>
+            <% end %>
+          </ul>
+        <% end %>
       </div>
 
       <div class="govuk-grid-column-two-thirds">
--- a/lib/smart_answer/calculators/check_travel_during_coronavirus_calculator.rb
+++ b/lib/smart_answer/calculators/check_travel_during_coronavirus_calculator.rb
@@ -22,6 +22,22 @@ module SmartAnswer::Calculators
       end
     end
 
+    def summary_text_fields
+      fields = []
+
+      return fields if travelling_to_ireland? && single_journey?
+
+      fields << (vaccination_status == "9ddc7655bfd0d477" ? "not_vaxed" : "fully_vaxed")
+      fields << "red_list" if travelling_to_red_list_country?
+      if travelling_with_children?
+        fields << travelling_with_children
+      elsif travelling_with_young_people?
+        fields << "young_person"
+      end
+
+      fields.flatten
+    end
+
     def travelling_with_children=(travelling_with_children)
       travelling_with_children.split(",").each do |response|
         @travelling_with_children << response
--- a/test/flows/check_travel_during_coronavirus_flow_test.rb
+++ b/test/flows/check_travel_during_coronavirus_flow_test.rb
@@ -563,7 +563,7 @@ class CheckTravelDuringCoronavirusFlowTest < ActiveSupport::TestCase
         assert_no_match "Returning to England if you’re fully vaccinated", @test_flow.outcome_text
       end
 
-      should "not render the summary text if only travelling to irealnd" do
+      should "not render the summary text if only travelling to Ireland" do
         assert_no_match "Because you said:", @test_flow.outcome_text
       end
 
--- a/test/flows/check_travel_during_coronavirus_flow_test.rb
+++ b/test/flows/check_travel_during_coronavirus_flow_test.rb
@@ -563,6 +563,10 @@ class CheckTravelDuringCoronavirusFlowTest < ActiveSupport::TestCase
         assert_no_match "Returning to England if you’re fully vaccinated", @test_flow.outcome_text
       end
 
+      should "not render the summary text if only travelling to irealnd" do
+        assert_no_match "Because you said:", @test_flow.outcome_text
+      end
+
       should "render Ireland country guidance and red list guidance if user also travelled to red list country" do
         add_responses any_other_countries_1: "yes",
                       which_1_country: "spain",
--- a/test/unit/calculators/check_travel_during_coronavirus_calculator_test.rb
+++ b/test/unit/calculators/check_travel_during_coronavirus_calculator_test.rb
@@ -120,5 +120,52 @@ module SmartAnswer::Calculators
       @calculator.countries = %w[spain]
       assert @calculator.single_journey?
     end
+
+    context "summary_text_fields" do
+      should "return an empty array if only travelling to ireland, even with other fields" do
+        @calculator.countries = %w[ireland]
+        @calculator.travelling_with_young_people = "yes"
+
+        assert_equal [], @calculator.summary_text_fields
+      end
+
+      should "not include young person if not travelling_with_young_people?" do
+        @calculator.travelling_with_young_people = "no"
+
+        assert_not @calculator.summary_text_fields.include?("young_person")
+      end
+
+      should "include young person if travelling_with_young_people?" do
+        @calculator.travelling_with_young_people = "yes"
+
+        assert @calculator.summary_text_fields.include?("young_person")
+      end
+
+      should "include responses to the travelling_with_children question" do
+        @calculator.travelling_with_children = "zero_to_four,five_to_seventeen"
+
+        %w[zero_to_four five_to_seventeen].each do |age|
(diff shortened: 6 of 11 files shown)


app/flows/check_travel_during_coronavirus_flow/outcomes/results.erb
```diff
@@ -20,14 +20,16 @@
   <% calculator.country_locations.each do |country| %>
     <div class="govuk-grid-row travel-abroad__country">
       <div class="govuk-grid-column-one-third travel-abroad__reasons">
-        <p class="govuk-body govuk-body-s">Because you said</p>
+        <% unless calculator.only_travelling_to_ireland? %>
+        <p class="govuk-body govuk-body-s">Because you said:</p>
```
leenagupte, line 24: I think this needs to be indented.

app/flows/check_travel_during_coronavirus_flow/outcomes/_travel_result_layout.erb
```diff
@@ -7,21 +7,23 @@
...
+          <li>You are travelling with someone aged 17 and under</li>
+        <% end %>
+        <% if calculator.travelling_with_children.include?("zero_to_four") %>
+          <li>You are travelling with someone aged 4 and under</li>
+        <% end %>
+        <% if calculator.travelling_with_children.include?("five_to_seventeen") %>
+          <li>You are travelling with someone aged 5 to 17</li>
+        <% end %>
+      </ul>
+    <% end %>
```
leenagupte, line 26: This "because you said" logic has gotten quite unwieldy, and wrapping the existing `if` statements in and `unless` only compounds that.

Perhaps instead, "Because you said" could be built in the calculator?  A new methods that uses the existing methods, i.e. `travelling_with_young_people?` to return a list of content instead of populating `data[:reasons]`.
Another benefit to this is that the first line of the method could return an empty array if the user is only travelling to Ireland. i.e.:

`return if travelling_to_ireland? && single_journey?` 

This is a bit more work but would make the views easier to manage.

leenagupte (commented):
@JonathanHallam  I haven't finished reviewing the content, but thought perhaps you'd like to discuss the new method thoughts in person?

lib/smart_answer/calculators/check_travel_during_coronavirus_calculator.rb
```diff
@@ -56,6 +56,10 @@ def travelling_to_ireland?
       countries.include?("ireland")
     end
 
+    def only_travelling_to_ireland?
```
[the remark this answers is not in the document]
JonathanHallam: have fixed this and just removed the new method :)

app/flows/check_travel_during_coronavirus_flow/outcomes/results.erb
JonathanHallam: done!

app/flows/check_travel_during_coronavirus_flow/outcomes/_travel_result_layout.erb
JonathanHallam: this was a bigger bit of work. Done now although tests are failing so I'll fix whatever that is before a rereview

lib/smart_answer/calculators/check_travel_during_coronavirus_calculator.rb
```diff
@@ -22,6 +22,15 @@ def country_locations
...
     end
 
+    def summary_text_fields
+      return [] if travelling_to_ireland? && single_journey?
+
+      @summary_text_fields = @travelling_with_children.dup
+      @summary_text_fields << "young_person" if travelling_with_young_people?
+      @summary_text_fields << "red_list" if travelling_to_red_list_country?
+      @summary_text_fields << (vaccination_status == "9ddc7655bfd0d477" ? "not_vaxed" : "fully_vaxed")
+    end
```
leenagupte, line 32: So let's see if I understand correctly. You are creating an array called `@summary_text_fields` (why is it an instance variable?)

And in `@summary_text_fields` you're adding codes to look up the text value of what should be displayed in the "Because you said" column?

I think this would be much easier if you started with populating the text here directly in this commit and then try moving to a config file, as at that point you'd have working tests to make sure you're plucking out the correct thing.

Regardless I still think a few `if/else` statements would make things easier. For example:

```
if travelling_with_children?
  @summary_text_field << travelling_with_children
elsif travelling_with_young_people?
  @summary_text_field << "young_person"
end

if vaccination_status == "9ddc7655bfd0d477"
  @summary_text_field << "not_vaxed"
elsif vaccination_status == "code for fully vaxed" || vaccination_status == "code for vaccination trial"
   @summary_text_field << "fully_vaxed"
end

@summary_text_fields << "red_list" if travelling_to_red_list_country?
```
etc

I think `travelling_with_children` returns an array, so you may need to return `@summary_text_fields.flatten` on the last line to make it work.

As mentioned before I'd also recommend adding a method here that takes the result of `summary_text_fields`  and uses it to look up the long text from the YAML file, and return a list that can be looped through in the view.

test/unit/calculators/check_travel_during_coronavirus_calculator_test.rb
```diff
@@ -120,5 +120,52 @@ class CheckTravelDuringCoronavirusCalculatorTest < ActiveSupport::TestCase
...
+
+      should "include responses to the travelling_with_children question" do
+        @calculator.travelling_with_children = "zero_to_four,five_to_seventeen"
+
+        %w[zero_to_four five_to_seventeen].each do |age|
+          assert @calculator.summary_text_fields.include?(age)
+        end
+      end
+
+      should "include the correct vaccination status for unvaccinted users" do
```
leenagupte, line 146: Typo
```suggestion
      should "include the correct vaccination status for unvaccinated users" do
```

test/unit/calculators/check_travel_during_coronavirus_calculator_test.rb
```diff
@@ -120,5 +120,52 @@ class CheckTravelDuringCoronavirusCalculatorTest < ActiveSupport::TestCase
...
+        @calculator.vaccination_status = "529202127233d442"
+
+        assert @calculator.summary_text_fields.include?("fully_vaxed")
+      end
+
+      should "include young person if travelling_with_young_people?" do
+        @calculator.travelling_with_young_people = "yes"
+
+        assert @calculator.summary_text_fields.include?("young_person")
+      end
```
leenagupte, line 162: I think this "young person" test should either be moved next to other test for young people, or replace it. For a moment I thought I was looking at a duplicate test!

test/unit/calculators/check_travel_during_coronavirus_calculator_test.rb
```diff
@@ -120,5 +120,52 @@ class CheckTravelDuringCoronavirusCalculatorTest < ActiveSupport::TestCase
       @calculator.countries = %w[spain]
       assert @calculator.single_journey?
     end
+
+    context "summary_text_fields" do
```
leenagupte, line 124: These tests are great! ⭐

app/flows/check_travel_during_coronavirus_flow/outcomes/_fully_vaxed.erb
```diff
@@ -12,17 +12,35 @@
 
   <p class="govuk-body">You do not need to take any COVID-19 tests before you travel to England or after you arrive.</p>
 
-  <p class="govuk-body">You must be able to prove that you’ve been fully vaccinated. If you live in England, you can prove your vaccination status using the <a href="https://www.gov.uk/guidance/nhs-covid-pass" class="govuk-link">NHS COVID Pass</a>.</p>
+  <p class="govuk-body">You must <a href="https://www.gov.uk/provide-journey-contact-details-before-travel-uk" class="govuk-link">complete a passenger locator form</a> in the 48 hours before you arrive in England.</p>
```
leenagupte, line 15: Not a blocker: This should be a relative link. i.e. `<a href="/provide-journey-contact-details-before-travel-uk" class="govuk-link">`

app/flows/check_travel_during_coronavirus_flow/outcomes/_fully_vaxed.erb
```diff
@@ -12,17 +12,35 @@
...
+
+  <%= render "govuk_publishing_components/components/heading", {
+    text: "Proof of vaccination",
+    heading_level: 3,
+    font_size: "m",
+    margin_bottom: 4,
+  } %>
+
+  <p class="govuk-body">You must be able to prove that you've been fully vaccinated. If you live in England, you can prove your vaccination status using the <a href="https://www.gov.uk/guidance/nhs-covid-pass" class="govuk-link">NHS COVID Pass</a>.</p>
```
leenagupte, line 24: I think there's some missing text here, missing curly quotes and another absolute like that should be relative:
```suggestion
  <p class="govuk-body">You must be able to prove that you’ve been fully vaccinated against COVID-19. If you live in England, you can prove your vaccination status using the <a href="/guidance/nhs-covid-pass" class="govuk-link">NHS COVID Pass</a>.</p>
```

leenagupte: The content changes in this view partial don't seem to be related to the rest of the "Because you said" changes. I think they have a separate Trello card? You could separate them out into a different PR to get them approved more quickly.

app/flows/check_travel_during_coronavirus_flow/outcomes/_fully_vaxed.erb
```diff
@@ -12,17 +12,35 @@
...
 
-  <p class="govuk-body">You must <a href="https://www.gov.uk/provide-journey-contact-details-before-travel-uk" class="govuk-link">complete a passenger locator form</a> in the 48 hours before you arrive in England.</p>
+  <%= render "govuk_publishing_components/components/heading", {
+    text: "Proof that you cannot have a COVID-19 vaccination",
+    heading_level: 3,
+    font_size: "m",
+    margin_bottom: 4,
+  } %>
+
+  <p class="govuk-body">If you cannot have a COVID-19 vaccination for a medical reason approved by a clinician, you will need to show proof using your <a href ="https://www.gov.uk/guidance/nhs-covid-pass" class="govuk-link">NHS COVID Pass</a>.</p>
```
leenagupte, line 35: Not a blocker: Same here about the link.

config/smart_answers/check_travel_during_coronavirus_summary_text.yml
```diff
@@ -0,0 +1,8 @@
+---
+summary_text:
+  young_person: You are travelling with someone aged 17 and under
+  zero_to_four: You are travelling with someone aged 4 and under
+  five_to_seventeen: You are travelling with someone aged 5 to 17
+  fully_vaxed: You are fully vaccinated or cannot have a COVID-19 vaccination
+  not_vaxed: You are not fully vaccinated
+  red_list: You will be in a red list country within 10 days of returning to England
```
leenagupte, line 8: I think the ordering is not quite right. In the previous world the content in `data[:reasons]` was displayed, first, so I think the order should be:

- fully_vaxed
- not_vaxed
- red_list
- young_person
- zero_to_four
- five_to_seventeen

This will ensure the order of the reasons displayed to the user remains unchanged.

leenagupte (changes requested):
@JonathanHallam I've put a red ❌  against these changes as some of the content doesn't seem quite right, and the order of the "Because you said" reasons has changed unexpectedly.

I think you've made a good start (the tests especially, are great!) with creating a lookup for the "Because you said" reasons, but as stated in the inline comments, some of the logic should be moved from the view and into the calculator.

I've added more detail inline about how the code can be simplified and I'm happy to talk through them in person. I don't think there's that much that needs to be done.

The commit structure could do with a bit of tidying to keep the related changes together, but that's just a case of squashing everything together and then picking out the changes.

app/flows/check_travel_during_coronavirus_flow/outcomes/results.erb
```diff
@@ -24,14 +24,16 @@
...
       <div class="govuk-grid-column-one-third travel-abroad__reasons">
-        <p class="govuk-body govuk-body-s">Because you said</p>
-        <ul class="govuk-list govuk-list--bullet govuk-body-s">
-          <% if calculator.transit_countries.include?(country.slug) %>
-            <li>You are travelling through <%= country.title %></li>
-          <% else %>
-            <li>You are travelling to <%= country.title %></li>
-          <% end %>
-        </ul>
+        <% unless calculator.travelling_to_ireland? && calculator.single_journey? %>
```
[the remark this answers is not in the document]
JonathanHallam: Yeah, this was one of the things in the card https://trello.com/c/fPtwW9zX/635-content-changes-to-summary-text-because-you-said-mvp

app/flows/check_travel_during_coronavirus_flow/outcomes/_travel_result_layout.erb
```diff
@@ -7,21 +7,23 @@
...
-        <li>You are travelling with someone aged 17 and under</li>
-      <% end %>
-      <% if calculator.travelling_with_children.include?("zero_to_four") %>
-        <li>You are travelling with someone aged 4 and under</li>
-      <% end %>
-      <% if calculator.travelling_with_children.include?("five_to_seventeen") %>
-        <li>You are travelling with someone aged 5 to 17</li>
-      <% end %>
-    </ul>
+    <% unless calculator.travelling_to_ireland? && calculator.single_journey? %>
```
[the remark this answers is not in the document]
JonathanHallam: I've moved these changes, I'll amend further because there's more comments on the next commit. Excellent shout with the `summary_text_fields.any?` thanks!

app/flows/check_travel_during_coronavirus_flow/outcomes/_fully_vaxed.erb
JonathanHallam: oops, copy paste error, thanks!

config/smart_answers/check_travel_during_coronavirus_summary_text.yml
```diff
@@ -0,0 +1,8 @@
+---
+summary_text:
```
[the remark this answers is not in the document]
JonathanHallam: good shout, done

config/smart_answers/check_travel_during_coronavirus_summary_text.yml
JonathanHallam: yeah that was silly, have fixed this

lib/smart_answer/calculators/check_travel_during_coronavirus_calculator.rb
```diff
@@ -22,6 +22,15 @@ def country_locations
       end
     end
 
+    def summary_text_fields
+      return [] if travelling_to_ireland? && single_journey?
+
+      @summary_text_fields = @travelling_with_children.dup
```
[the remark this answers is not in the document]
JonathanHallam: So I figured the `@travelling_with_children` object was a good place to start because it's an array that has some of the values I want and the only place where there can currently be multiple values that I'd need to loop through. The `.dup` is needed because without it you end up with that horrible problem I had on Friday where `fully_vaxed` was added everywhere

app/flows/check_travel_during_coronavirus_flow/outcomes/_fully_vaxed.erb
```diff
@@ -40,7 +40,6 @@
 <%
   data = {
     title: "Returning to England if you’re fully vaccinated or cannot have a COVID-19 vaccination",
-    reasons: ['You are fully vaccinated or cannot have a COVID-19 vaccination'],
```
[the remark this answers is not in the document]
JonathanHallam: sorted!

lib/smart_answer/calculators/check_travel_during_coronavirus_calculator.rb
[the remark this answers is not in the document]
leenagupte: I think that was because you were assigning to the instance variable directly, rather than calling the method?

i.e. What happens if you try `@summary_text_fields = travelling_with_children`. (I'm still not sure why `@summary_text_fields` is an instance method.

config/smart_answers/check_travel_during_coronavirus_summary_text.yml
leenagupte: Thinking about this morning, I think the order changing is due to the order of the conditionals in the `summary_text_fields` method rather than in this file. Oops!

lib/smart_answer/calculators/check_travel_during_coronavirus_calculator.rb
leenagupte: Actually, my brain seems to working better this morning. And I have a better suggestion. Rather than creating a list of "codes" i.e. ["fully_vaxed", "young_person"], why not create a method that loads the YAML and lets you look up the text string by code. For example, if you have a method called `def find_summary_text_from_code(code)` then if the `summary_text_fields` you could call:

```
if vaccination_status == "9ddc7655bfd0d477"
  @summary_text_field << find_summary_text_from_code("not_vaxed")
elsif vaccination_status == "code for fully vaxed" || vaccination_status == "code for vaccination trial"
   @summary_text_field << find_summary_text_from_code("fully_vaxed")
end
```

config/smart_answers/check_travel_during_coronavirus_summary_text.yml
leenagupte: Also, I think this is a locale file isn't it? Perhaps it should be in `config/locales`? If you did that you wouldn't need a method to load the YAML file because Rails would do it for you.

JonathanHallam: Have sorted this. Added a locale file with proper file structure and fixed the order that things are added to the `summary_text_fields` array that I've now renamed and stopped being an instance variable

app/flows/check_travel_during_coronavirus_flow/outcomes/_travel_result_layout.erb
```diff
@@ -10,18 +10,10 @@
     <% unless calculator.travelling_to_ireland? && calculator.single_journey? %>
       <p class="govuk-body govuk-body-s">Because you said:</p>
       <ul class="govuk-list govuk-list--bullet govuk-body-s">
-        <% data[:reasons].each do |reason| %>
+        <% summary_text = YAML.load_file(Rails.root.join("config/smart_answers/check_travel_during_coronavirus_summary_text.yml"))['summary_text'] %>
+        <% summary_text.slice(*calculator.summary_text_fields).values.each do |reason| %>
```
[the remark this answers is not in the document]
JonathanHallam: Sorted this by using the locales stuff and a method rather than an instance variable

lib/smart_answer/calculators/check_travel_during_coronavirus_calculator.rb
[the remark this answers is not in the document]
JonathanHallam: yeah, have fixed all of this, silly error, sorry!

lib/smart_answer/calculators/check_travel_during_coronavirus_calculator.rb
JonathanHallam: Again, sorted by internationalising :)

test/unit/calculators/check_travel_during_coronavirus_calculator_test.rb
JonathanHallam: oops!

test/unit/calculators/check_travel_during_coronavirus_calculator_test.rb
JonathanHallam: sorted! Sorry!

app/flows/check_travel_during_coronavirus_flow/outcomes/_fully_vaxed.erb
JonathanHallam: Nice one, sorted. I'll leave it in this PR because I don't think we're a million miles away

leenagupte (commented):
This is looking really good! Well done 🎉 

I really like the use of locales / Internationalisation to lookup at the text for the "Because you said content". 🏅 

I've checked the content changes and they look good. So does the rendering of the "Because you said" reasons.

I've added some inline comments, but the blocking changes have been summarised here:

1.  Add a flow test to make sure that "Because you said" isn't being populated when Ireland is selected on its own. This is something that came out in manual testing and we should make sure it's covered in our automated tests.
2. Ideally all of the locale file config would be in their own commit, but regardless, as wilth all config changes it will really useful for future-us to know why the config change was necessary.

It's almost there!

config/locales/en.yml
```diff
@@ -1,33 +0,0 @@
-# Files in the config/locales directory are used for internationalization
```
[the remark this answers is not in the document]
JonathanHallam: I think now that I've added the line changing the load path this won't work as expected. I'll add a note to the commit message explaining this

app/flows/check_travel_during_coronavirus_flow/outcomes/_fully_vaxed.erb
JonathanHallam: Sorted!

test/flows/check_travel_during_coronavirus_flow_test.rb
```diff
@@ -563,6 +563,10 @@ class CheckTravelDuringCoronavirusFlowTest < ActiveSupport::TestCase
         assert_no_match "Returning to England if you’re fully vaccinated", @test_flow.outcome_text
       end
 
+      should "not render the summary text if only travelling to irealnd" do
```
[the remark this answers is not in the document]
JonathanHallam: Have fixed this :)

PR #5778, opened 2022-02-14
Add validation to transiting question

leenagupte:
Trello: https://trello.com/c/vp9op3Dd

# What's changed?

Validate that users cannot select all of their previously selected countries as transiting countries. At least one country must be unchecked to be considered a destination country.

# Why?

In the results, the guidance for transiting countries is much less than the guidance for destination countries. This check guards against users misreading the question, selecting all of the countries thinking that they are selecting destinations and only being provided with limited guidance.

# Expected changes

## New error
Screenshot 2022-02-15 at 13 44 13

## Existing error (remains unchanged)
Screenshot 2022-02-15 at 13 43 59

### DIFF
--- a/app/flows/check_travel_during_coronavirus_flow.rb
+++ b/app/flows/check_travel_during_coronavirus_flow.rb
@@ -64,6 +64,10 @@ class CheckTravelDuringCoronavirusFlow < SmartAnswer::Flow
         calculator.transit_countries = response unless response == "none"
       end
 
+      validate(:error_no_destination) do
+        calculator.transit_countries.size < calculator.countries.size
+      end
+
       next_node do
         if calculator.red_list_countries.any? && calculator.countries.length > 1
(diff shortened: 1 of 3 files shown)


Rosa-Fox (commented):
I noticed a problem with this which is that this overwrites the default error message when no questions are checked 'Please answer this question' with 'You must leave at least one country unselected' which looks a bit weird:

Screenshot 2022-02-15 at 12 35 14. 

I think it will also need to check that the countries are not empty.

leenagupte:
> I noticed a problem with this which is that this overwrites the default error message when no questions are checked 'Please answer this question' with 'You must leave at least one country unselected' which looks a bit weird:
> 
> I think it will also need to check that the countries are not empty.

Good spot! I've updated the error block so that both errors are now being displayed.

Rosa-Fox (approved):

PR #5803, opened 2022-02-25
Dockerise Smartanswers

fredericfran-gds:
Note that we install npm so that we can install yarn
in the recommended [way](https://classic.yarnpkg.com/lang/en/docs/install).

Yarn is then used to install the node dependencies, yarn is the recommended
GOV.UK Node package manager, see [here](https://docs.publishing.service.gov.uk/manual/conventions-for-rails-applications.html#javascript-package-management)

Ref:
1. [trello card](https://trello.com/c/vR41bV93/840-package-smart-answers-for-migration)


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,19 @@
...
+on:
+  workflow_dispatch:
+    branches:
+      - main
+  push:
+    branches:
+      - main
+    paths-ignore:
+      - "Jenkinsfile"
+      - ".git**"
```
bilbof, line 12: Would be nice if paths-ignore could read from a file like `.dockerignore` :)

bilbof (approved):
Looks good, thanks 👍

.github/workflows/ci.yaml
fredericfran-gds: don't think this is possible, maybe ask @issyl0 lol

PR #5807, opened 2022-02-28
Filter vaccination_status response from change link event

Rosa-Fox:
[Trello](https://trello.com/c/r4lZ5FYx/712-link-tracking-hide-vaccination-status-value-from-being-sent-to-ga-when-change-answer-link-is-clicked)

The `vaccination_status` response is health data. Despite having no personally identifing questions in this Smart Answer, we are required to prevent health data from being sent to Google Analytics.

### Redacting vaccination_status from travel flow / country question response remains on the label

change-link-vax-status-and-country-change

### Non travel flow 

tracking-non-travel-flow

### Non travel flow checkbox question 
change-link-tracking-checkbox-question

Co-authored-by: Leena Gupte <leena.gupte@digital.cabinet-office.gov.uk> as a result of approach suggested on [this branch](https://github.com/alphagov/smart-answers/commit/991964cf3691c8da5afb707ccf0bceee87acbf81)

### DIFF
--- a/app/presenters/question_presenter.rb
+++ b/app/presenters/question_presenter.rb
@@ -71,6 +71,14 @@ class QuestionPresenter < NodePresenter
     value
   end
 
+  def change_link_track_label(value)
+    if redacted?
+      "#{title} / [FILTERED]"
+    else
+      "#{title} / #{response_label(value)}"
+    end
+  end
+
   def partial_template_name
     "#{@node.class.name.demodulize.underscore}_question"
   end
--- a/app/flows/check_travel_during_coronavirus_flow.rb
+++ b/app/flows/check_travel_during_coronavirus_flow.rb
@@ -97,6 +97,8 @@ class CheckTravelDuringCoronavirusFlow < SmartAnswer::Flow
     radio :vaccination_status do
       options { calculator.vaccination_option_keys }
 
+      redact true
+
       on_response do |response|
         calculator.vaccination_status = response
       end
--- a/app/presenters/node_presenter.rb
+++ b/app/presenters/node_presenter.rb
@@ -18,4 +18,8 @@ class NodePresenter
   def view_template_path
     @node.view_template_path
   end
+
+  def redacted?
+    @node.redact
+  end
 end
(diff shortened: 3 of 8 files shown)


app/presenters/flow_presenter.rb
```diff
@@ -54,6 +54,14 @@ def change_answer_link(question, forwarding_responses)
     end
   end
 
+  def change_link_track_label(question, accepted_response)
+    if question.node_name == :vaccination_status
```
leenagupte, line 58: It feels a bit odd to have a reference to a flow specific question `vaccination_status` in the generic Flow Presenter.

Is it possible to delegate this method the "Flow" class? Or does "Flow" not have access to the `question` object? Or maybe QuestionPresenter?

I need a little time to think about this.

Rosa-Fox: I can give it a try :) The reason I went with flow_presenter was that it already had `change_answer_link` and the change answer section is on all of the flows. Agree that having the `vaccine_status` specific stuff in there is a bit weird but I thought it would be ok as it also displayed the label for the other questions... however if I can delegate it and then make use of inheritance then than would be better so i'll see if I can... thanks!

Rosa-Fox: That approach did work but testing it was really difficult and it didn't feel right having an overriding method in the `check_travel_during_coronavirus_flow` because it wasn't in keeping with our usual flow test style. I came up with a new solution which I think works better, where the method lives on the `question_presenter`. Let me know what you think :) @leenagupte

app/presenters/question_presenter.rb
```diff
@@ -71,6 +71,14 @@ def response_label(value)
     value
   end
 
+  def change_link_track_label(value)
+    if redacted?
```
JonathanHallam, line 75: technically would be slightly more performant if these were swapped but it wouldn't read as nicely so I don't really mind

JonathanHallam: although actually no it wouldn't because it would have to check `redacted?` regardless so ignore me entirely!

JonathanHallam (approved):
Happy with where all of the changes are. Tests look good, GA redaction works (took me an aaaagge to prove that, couldn't get my debugger to work!) All good, nice one :)

PR #5814, opened 2022-03-02
Add margin option to autocomplete component

andysellick:

## What
Add an option to the autocomplete component to allow a custom margin bottom to be set. Uses the standard shared_helper approach from the gem to do margin bottom.

## Why
We're going to need to reduce the margin where we're about to use it but the design isn't finalised to having a flexible option for margin bottom seems the best way forward.

## Visual changes

Screenshot 2022-03-02 at 16 53 38

Trello card: https://trello.com/c/i3brOfzz/431-add-multiple-country-select-question-type-to-smart-answers

### DIFF
--- a/app/views/components/_autocomplete.html.erb
+++ b/app/views/components/_autocomplete.html.erb
@@ -4,12 +4,15 @@
   label ||= nil
   value ||= nil
   options ||= []
+  local_assigns[:margin_bottom] ||= 6
+  shared_helper = GovukPublishingComponents::Presenters::SharedHelper.new(local_assigns)
 
   error_id = "error-#{SecureRandom.hex(4)}"
   error_items ||= []
 
   classes = %w(app-c-autocomplete govuk-form-group)
   classes << "govuk-form-group--error" if error_items.any?
+  classes << shared_helper.get_margin_bottom
 
   if !options.any? || !label
     raise ArgumentError, "The autocomplete component needs at least one option and a label in order to render."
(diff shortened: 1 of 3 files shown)


app/views/components/_autocomplete.html.erb
```diff
@@ -4,12 +4,15 @@
   label ||= nil
   value ||= nil
   options ||= []
+  local_assigns[:margin_bottom] ||= 6
+  shared_helper = GovukPublishingComponents::Presenters::SharedHelper.new(local_assigns)
```
leenagupte, line 8: Could you link to this helper in the commit message? Also, what is the benefit of using the helper?

leenagupte: Could you also add why "6" is being used as the default to the commit message as well?

leenagupte (commented):
I'm not sure how to test this locally, as I don't think the autocomplete component is being used yet?

I've added a couple of inline comments about updating the commit message. It would be great if the "why" from the PR description could also be added to the commit message so future developers understand the full context of this change.

Otherwise the tests are passing 🎉 , so I think it's ok.

app/views/components/_autocomplete.html.erb
andysellick: Commit message updated. The [shared helper](https://github.com/alphagov/govuk_publishing_components/blob/master/docs/component_conventions.md#shared-helper) is basically a way of not having to write the same bit of code for more than one component - it handles things like margins, heading levels, etc. in a consistent way, and already has tests written for it so we can keep testing per component for these things to a minimum.

leenagupte: This ☝️  would be a great paragraph to add to the commit message!

leenagupte (approved):

PR #5817, opened 2022-03-03
Revert "Change gds::worldwide_api to retrieve information inside VPC"

bilbof:
After discussing with @kevindew we have decided to continue consuming these APIs via the public endpoint rather than using an internal endpoint.

The reasoning is that we prefer all requests to frontend applications to go through Fastly and router. This is particularly useful because Fastly caches requests to the APIs, and we have seen an incident in the past where these APIs could not handle high volumes of requests (and so require caching). We believe these APIs should probably not be served by frontend apps, and instead should be served by a different backend service, but this is out of scope of the replatforming project.

Reverts alphagov/smart-answers#5812

### DIFF
--- a/app/models/world_location.rb
+++ b/app/models/world_location.rb
@@ -15,12 +15,11 @@ class WorldLocation
 
   def self.all
     cache_fetch("all") do
-      world_locations = WorldwideApi
-                        .endpoint
-                        .world_locations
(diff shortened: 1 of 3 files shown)


fredericfran-gds (approved):
Thanks!

sengi (approved):

PR #5832, opened 2022-03-10
Fix and update self assessment calculator

1pretz1:
Contains two changes, firstly it fixes a bug where covid easement was being applied for paper submissions, which is incorrect. Secondly, interest rates have been updated backdated to the 21st February 2022.

Trello:
https://trello.com/c/B79tAlpm/1134-self-assessment-penalty-calculator-interest-rate-change-and-covid-easement-error  

### DIFF
--- a/lib/smart_answer/calculators/self_assessment_penalties.rb
+++ b/lib/smart_answer/calculators/self_assessment_penalties.rb
@@ -202,7 +202,7 @@ module SmartAnswer::Calculators
       covid_easement_first_year = (tax_year == "2019-20" && filing_date < Date.parse("2021-03-01"))
       covid_easement_second_year = (tax_year == "2020-21" && filing_date < Date.parse("2022-03-01"))
 
-      covid_easement_first_year || covid_easement_second_year
+      (covid_easement_first_year || covid_easement_second_year) && submission_method == "online"
     end
 
     def payment_deadline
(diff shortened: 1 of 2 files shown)


lib/smart_answer/calculators/self_assessment_penalties.rb
```diff
@@ -214,18 +220,11 @@ def calculate_interest_for_date(date)
...
-        0.0275 / 365.0
-      elsif date >= rate_change_date
-        0.026 / 365.0
-      else
-        0.0325 / 365.0
+      rate_for_date = INTEREST_RATES.find do |rate|
+        date >= Time.zone.parse(rate[:start_date]) && date <= Time.zone.parse(rate[:end_date])
       end
+
+      rate_for_date ? rate_for_date[:value] / 365.0 : 0.0325 / 365.0
```
edwardkerry, line 227: This _might_ read easier as an if/else rather than a ternary...but usually I would also go for a ternary so I'm not too concerned.
Nice refactor of the rates 👍

edwardkerry (approved):
One non-blocking comment, otherwise looks good!

lib/smart_answer/calculators/self_assessment_penalties.rb
1pretz1: Good point - the ternary was a bit of a mouthful with the calculations, have used `if/else` and `present?` on `rate_for_date` to be even more explicit.

PR #5833, opened 2022-03-14
(Uprating) Minimum wage calculator 2022

AgaDufrat:
National Living Wage for those aged 23 and over will increase
from £8.91 per hour to £9.50 per hour.

National Minimum Wage will also increase:
- for 21 to 22 year olds – from £8.36 per hour to £9.18
- for 18 to 20 year olds – from £6.56 per hour to £6.83
- for those under 18 – from £4.62 per hour to £4.81
- for apprentices – from £4.30 per hour to £4.81

Accommodation offset rates will also change:
- daily – from £8.36 to £8.70

https://trello.com/c/eoazcTDo/868-deploy-on-1-april-national-minimum-wage-and-living-wage-calculators

---

### DIFF
--- a/config/smart_answers/rates/minimum_wage.yml
+++ b/config/smart_answers/rates/minimum_wage.yml
@@ -242,7 +242,7 @@
   :apprentice_rate: 4.15
   :accommodation_rate: 8.20
 - :start_date: 2021-04-01
-  :end_date: 3000-01-01
+  :end_date: 2022-03-31
   :minimum_rates:
   - :min_age: 0
     :max_age: 18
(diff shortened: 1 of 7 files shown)


AgaDufrat:
This has been content reviewed and fact checked. Changes were addressed in separate PRs that can be deployed before the 1 April.

config/smart_answers/rates/minimum_wage.yml
```diff
@@ -258,3 +258,20 @@
     :rate: 8.91
   :apprentice_rate: 4.30
   :accommodation_rate: 8.36
+- :start_date: 2022-04-01
```
hannako, line 261: Do you know why we need to keep data dating back to 2016? I can't see how it's used for the two services referenced in your PR description. Is it used for other services that support checking pay entitlement dating back that far? If it is, perhaps we could add a comment at the top of this file saying so?

hannako: OK it doesn't look like it's used anywhere else. I think we should remove the old data and cut down the tests, as it makes these uprating tasks more complicated than they need to be. As it's a bit time consuming lets make a card for that and do it as a separate task.

hannako (approved):
Code change looks good if content folks are happy

config/smart_answers/rates/minimum_wage.yml
AgaDufrat: I'm not sure why we keep data from that long ago but this it the case with a lot of those calculators. Removing it seems like something worth doing but since it will require a lot of changes to the tests, it's probably out of scope for this PR.
The services only use current and previous year's data, I will add a comment to reflect that.

hannako: https://trello.com/c/MslkvTUc/319-remove-legacy-data-from-smart-answers-to-make-uprating-easier

PR #5874, opened 2022-04-13
Rails 7 Update

KludgeKML:


Gemfile
```diff
@@ -2,7 +2,7 @@ source "https://rubygems.org"
 
 ruby File.read(".ruby-version").chomp
 
-gem "rails", "6.1.4"
+gem "rails", "7.0.2"
```
kevindew, line 5: As an aside you'll want to set this to 7.0.2.3 which is the current release.

spec/support/jasmine-browser.json
```diff
@@ -2,7 +2,8 @@
   "srcDir": "public/assets",
   "srcFiles": [
     "**/*.js",
-    "!express/**/*.js"
+    "!express/**/*.js",
+    "!**/visualise-*.js"
```
edwardkerry, line 6: Do we have a note or card somewhere to remind us to add this back when possible?

lib/support_ticket.rb
```diff
@@ -41,12 +41,6 @@ def payload
...
 
   def send
-    self.class.client.ticket.create!({
-      subject: subject,
-      priority: PRIORITY,
-      requester: REQUESTER.merge(email: requester_email),
-      tags: TAGS,
-      comment: { body: body },
-    })
+    self.class.client.zendesk_client.tickets.create!(payload)
```
edwardkerry, line 44: this seems to be undoing a change from the `Fix deprecation notices` commit. Is it possible to remove the change from that commit?

edwardkerry (approved):
This looks OK to me - a couple of questions but not blockers.
I notice the heroku app has failed to deploy but think that is a Heroku issue. Has it been deployed/tested on Integration?

lib/support_ticket.rb
KludgeKML: It's... not quite undoing it (you'll see that the end result is slightly different), but yeah, I might be able to tinker with the commits to go straight from the original version to this, let me see if I can do that.

app/assets/config/manifest.js
```diff
@@ -0,0 +1,13 @@
...
+//= link joint.css
+//= link visualise.css
+//= link accessible-autocomplete/dist/accessible-autocomplete.min.js
+//= link application.js
+//= link dagre.js
+//= link joint.layout.DirectedGraph.js
+//= link joint.js
+//= link joint.patch.js
+//= link test-dependencies.js
+//= link visualise.js
```
kevindew, line 13: I imagine you can simplify this a little:

```
//= link_directory ./stylesheets
//= link joint.css
//= link application.js
//= link dagre.js
//= link joint.layout.DirectedGraph.js
//= link joint.js
//= link joint.patch.js
//= link test-dependencies.js
//= link visualise.js
```

You should also remove the assets_precompile stuff from https://github.com/alphagov/smart-answers/blob/d3fc1b1ff37a9163cdbb020e2e2fd7c52d4f77ea/config/initializers/assets.rb#L14-L18 and https://github.com/alphagov/smart-answers/blob/d3fc1b1ff37a9163cdbb020e2e2fd7c52d4f77ea/config/application.rb#L63-L72

A couple of suggestions that could make this part of the update a little easier are:

- Do the sprockets update as part of the dependabot PR: https://github.com/alphagov/smart-answers/pull/5879 (which updates Sprockets) so you can separate it out from the Rails 7 stuff
- - Perhaps use asset-pipeline to require the various joint.* js files from visualise.js, I don't think they're used anywhere else.

package.json
```diff
@@ -7,7 +7,10 @@
   "scripts": {
     "lint": "yarn run lint:js && yarn run lint:scss",
     "lint:js": "standardx 'app/assets/javascripts/**/*.js' 'spec/javascripts/**/*.js'",
-    "lint:scss": "stylelint app/assets/stylesheets/"
+    "lint:scss": "stylelint app/assets/stylesheets/",
+    "jasmine:prepare": "bundle exec rake assets:precompile",
+    "jasmine:ci": "yarn run jasmine:prepare && PATH=~/.webdrivers:$PATH yarn run jasmine-browser-runner runSpecs",
```
kevindew, line 12:
```suggestion
    "jasmine:ci": "yarn run jasmine:prepare && yarn run jasmine-browser-runner runSpecs",
```

This suggestion is for consistency with our other apps: https://github.com/alphagov/content-publisher/blob/6511019bc2313b1cea3f1fcd7f5e77aef6913368/package.json#L12

This app doesn't have the webdrivers gem (it was removed from GOV.UK test) https://github.com/alphagov/govuk_test/pull/43

kevindew (commented):
Excuse my drive by review - I get email notifications as I think I already commented. I thought I'd offer some suggestions having worked on a few JS testing and sprockets things.

KludgeKML:
Superceded by: https://github.com/alphagov/smart-answers/pull/5892

PR #5890, opened 2022-05-05
Remove user intent banner from smart answers

MartinJJones:
## What

Remove the user intent banner from the application view template, the shared component will remain in place should anyone need to use it in the future, for example, this was previously used on the /check-travel-during-coronavirus smart answer.

## Why

The only page using the user-intent banner, is the '/next-steps-for-your-business' smart answer, which it needs to be removed from.

The team working on the '/next-steps-for-your-business' smart answer are no longer doing user research, so they'd like this to be removed. This will ensure the team working on it don't get requests for user research help that they don't need anymore. Users don't also waste their time.

## Visual Changes

| **Before** | **After** |
| - | - |
| Screenshot 2022-05-05 at 10 39 26 | Screenshot 2022-05-05 at 10 38 38 |


maxgds (changes requested):
I think we should remove the shared file at /shared/_banner.html.erb too. It's currently tied to a particular path/answer so if anyone wanted to use it they'd need to edit that anyway. I'm not keen on leaving unused templates lying around just in case - it leads to confusion when someone tries to figure out what they're there for and where they're used. Because of version control anyone remembering the previous incarnation and wanting to reimplement it can dig it out anyway, and there's a good chance people won't and would wind up implementing a parallel version with a different design.

MartinJJones:
Thanks @maxgds, I have removed the shared file /shared/_banner.html.erb

maxgds (approved):
👍

PR #5905, opened 2022-05-19
Make Docker image run as non-root.

sengi:
Basically the same as alphagov/feedback#1432.

- Add some files to .dockerignore.
- Don't run as root by default.
- Symlink /app/tmp and /home/app to /tmp so that gems don't try to write to the (read-only) image filesystem. (Ruby/Rails doesn't seem to have a consistent convention for telling gems where to write temp files.)
- Run `rails assets:precompile` in the build stage.
- Avoid installing "suggested" and "recommended" dependencies of Debian packages.
- Avoid installing gem documentation.
- Avoid including the APT cache in the output image.
- Increase build parallelism (saves a couple of minutes on a 2019 MBP).

Tested: `docker run -it --rm -e GOVUK_APP_DOMAIN=x -e GOVUK_WEBSITE_ROOT=x feedback`

Rollout: this app is push-on-green, so watch out.


lwilts (approved):
As with 'feedback' - LGTM, giving benefit of the doubt on Ruby aspects.

PR #5928, opened 2022-06-06
Update check-benefits-support results to only show results user is eligible for

edwardkerry:
This PR adds a yml file holding information on each possible result .

One key is `condition`, which corresponds to a method on the calculator with the same name as the key's value.

If a benefit does not have a condition, or if the user input from the flow passes the logic check in the corresponding method, the benefit is selected.

A single view is used to iterate over and display all selected benefits. This view is currently using the `result_item` partial which will be replaced by a new custom partial when developed.

### DIFF
--- a/config/smart_answers/check_benefits_support_data.yml
+++ b/config/smart_answers/check_benefits_support_data.yml
@@ -66,6 +66,7 @@
     - name: 15hrs_free_childcare_3_4yr_olds
       title: 15hrs free childcare for 3 and 4yrs
       description:
+      condition: eligible_for_15hrs_free_childcare_3_4yr_olds?
       url: /help-with-childcare-costs/free-childcare-and-education-for-2-to-4-year-olds
       url_text: Anchor link text
       group: Help with childcare costs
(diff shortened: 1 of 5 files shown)


config/smart_answers/check_benefits_support_data.yml
```diff
@@ -0,0 +1,146 @@
...
+      url: https://www.mygov.scot/child-disability-payment
+      url_text: Anchor link text
+      group: Help with childcare costs
+    - name: carers_allowance
+      title: Carer's Allowance
+      description:
+      url: /carers-allowance
+      url_text: Anchor link text
+      group: Supporting your income
+    - name: personal_indepdence_payment
```
Rosa-Fox, line 122: Typo indepdence -> independence

edwardkerry: good eyes! 👍

Rosa-Fox (approved):
Approving - flow test for outcomes will be added in a future PR. There will also be changes to the flow/logic in a future PR.

PR #5930, opened 2022-06-08
Add result card app component to smart answers

MartinJJones:
## What
Add a new result card component intended for use on the results page.

## Why
The custom design uses visual grouping to bring clarity to the results, to help improve legibility and readability of the results page.

## Visual
result-card-component

### Trello
https://trello.com/c/6gjI5Ami/22-finalise-results-page
https://trello.com/c/6oJJ00i8/78-create-new-component-for-benefits-results-item

### DIFF
--- a/app/views/components/_result-card.html.erb
+++ b/app/views/components/_result-card.html.erb
@@ -1,0 +1,27 @@
+<%
+  title ||= nil
+  description ||= nil
+  url ||= nil
+  url_text ||= nil
+%>
+<div class="app-c-result-card">
+  <%= render "govuk_publishing_components/components/heading", {
+    text: title,
+    heading_level: 2,
+    margin_bottom: 2,
+    font_size: "m"
+  } %>
+
+  <p class="govuk-body app-c-result-card__description">
+    <%= description %>
+  </p>
+
+  <%= link_to url_text, url,
+  class: "govuk-link app-c-result-card__link",
+  data: {
+    module: "gem-track-click",
+    "track-action": "Cost of living results",
+    "track-category": "SmartAnswerClicked",
+    "track-label": url
+  } %>
+</div>
--- a/app/views/components/docs/result-card.yml
+++ b/app/views/components/docs/result-card.yml
@@ -1,0 +1,11 @@
+name: Result card
+description: Presents a card style component with a title, description and link
+shared_accessibility_criteria:
+  - link
+examples:
+  default:
+    data:
+      title: Tax-Free childcare
+      description: You may be able to get up to £2,000 a year for each of your children to help with the costs of childcare.
+      url: /tax-free-childcare
+      url_text: Check if you’re eligible for Tax-Free childcare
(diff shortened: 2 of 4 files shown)


MartinJJones:
@injms good spot, I will fix this!, thanks

app/views/components/_result-card.html.erb
```diff
@@ -0,0 +1,24 @@
...
+
+  <div class="app-c-result-card__description">
+    <p class="govuk-body">
+      <%= description %>
+    </p>
+  </div>
+
+  <div class="app-c-result-card__link">
+    <p><%= link_to url_text, url, class: "govuk-link" %></p>
+  </div>
```
injms, line 23: Could we remove the need for a wrapping `div` and `p` elements by having the two classes on the same element?
```suggestion
    <%= link_to url_text, url, class: "govuk-link app-c-result-card__link" %>
```

app/views/components/_result-card.html.erb
```diff
@@ -0,0 +1,22 @@
+<%
+  title ||= nil
+  description ||= nil
+  url ||= nil
+  url_text ||= nil
+%>
+<div class="app-c-result-card">
+  <%= render "govuk_publishing_components/components/heading", {
+    text: title,
+    heading_level: 2,
```
injms, line 10: Are we sure that this is only going to be needed as a level 2 heading? If so, fine - otherwise it'd be worth looking into whether the [shared helper from the components gem][1] could be used:

```erb
shared_helper = GovukPublishingComponents::Presenters::SharedHelper.new(local_assigns)
```
```suggestion
    heading_level: shared_helper.get_heading_level,
```

The shared helper defaults to a level two heading, so this wouldn't change the output unless the `heading_level` parameter was set.

[1]: https://github.com/alphagov/govuk_publishing_components/blob/99938773aef7a14e8f67c878f5ed0d9deddd09e0/lib/govuk_publishing_components/presenters/shared_helper.rb#L28-L32

app/views/components/_result-card.html.erb
MartinJJones: Thanks, I have made this change and removed the `div` and `p` elements. I tested it locally and still renders as expected.

MartinJJones:
Removed unused CSS rule

app/views/components/docs/result-card.yml
```diff
@@ -0,0 +1,11 @@
+name: Result card
+description: Presents a card style component with a title, description and link
+shared_accessibility_criteria:
+  - [Heading](https://github.com/alphagov/govuk_publishing_components/blob/main/app/views/govuk_publishing_components/components/docs/heading.yml)
```
injms, line 4: This unfortunately throws an error when viewing the page on the component guide.

I think that `link` is needed as this component has a link - but I can't find any use of `heading` anywhere, so we could leave it off?

```suggestion
  - link
```

injms: There are only currently two shared accessibility criteria in the component gem at the moment - link and button[^1]. We should add heading - though for now I think that's outside of the scope of this task.


[^1]: https://github.com/alphagov/govuk_publishing_components/blob/99938773aef7a14e8f67c878f5ed0d9deddd09e0/app/models/govuk_publishing_components/shared_accessibility_criteria.rb

MartinJJones:
Added missing custom tracking for GA

injms (approved):
Looks good - thanks for your work on this!

PR #5934, opened 2022-06-09
Update check_benefit_support flow and question copy

edwardkerry:
This PR removes a branching question, `unpaid_care_hours`, and incorporates it into the previous questions.

it also makes some copy updates to other questions.


KludgeKML (approved):
LGTM

PR #5939, opened 2022-06-09
Update check-financial-benefit-support results page

edwardkerry:
This PR updates the results page for the check-financial-benefit-support to use the new results-card component introduced in https://github.com/alphagov/smart-answers/pull/5930 and updates the copy should a user see no results.

Screenshot 2022-06-09 at 17 14 38


MartinJJones (approved):
nice work 👍

PR #5940, opened 2022-06-09
Update jobseekers allowance calculator test for check-benefit-support

edwardkerry:
Adds more robust and thorough tests for eligible_for_jobseekers_allowance in check-benefits-support calculator.


Rosa-Fox (approved):
Discussed offline - tests with multiple assertions will be split out into separate tests in a separate task.

PR #5950, opened 2022-06-10
Update hint text open in a new tab link

Rosa-Fox:
Screenshot 2022-06-10 at 16 30 22


edwardkerry (approved):

edwardkerry (approved):
Approving on advice of content, although there are accessibility issues
https://www.w3.org/WAI/WCAG21/Techniques/html/H83
and goes against design system advice
https://design-system.service.gov.uk/styles/typography/#links

edwardkerry (approved):

PR #5952, opened 2022-06-13
WIP: Fix duplicate GA events firing

MartinJJones:


app/assets/javascripts/modules/track-results.js
```diff
@@ -24,11 +24,14 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
 
   TrackResults.prototype.trackInternalLinks = function () {
     var currentHost = document.location.protocol + '//' + document.location.hostname
-    var internalLinkSelector = 'a[href^="' + currentHost + '"], a[href^="/"]'
+    var internalLinkSelector = '#result-info a[href^="' + currentHost + '"], #result-info a[href^="/"]'
```
andysellick, line 27: I think an alternative approach might work better here. The problem is that currently the code finds all links on the page, regardless of where they are, and tries to apply the tracking to them. Your change scopes that to the `#result-info` element (the main content of the page) but that still includes links in the 'Your answers' section that have the gem-track-click script.

The alternative is to embrace the scoped nature of this module. It's being initialised on an element in the main content of the page that doesn't include the 'Your answers' section, so instead of finding all links and applying the event listener to them, you could find all links within the element that the module is acting on, something like this:

```
var internalLinks = this.$module.querySelector('a[href^="' + currentHost + '"], a[href^="/"]')
for (var i = 0; i < internalResultLinks.length; i++) {
  internalResultLinks[i].addEventListener('click', this.trackClickEvent.bind(this))
}
```

(I haven't tested this, you might need to do something a bit cleverer with the initial var declaration because it's a multi part selector, but it might just work)

Now the only links that this code will apply to are those in the element shown.

Screenshot 2022-06-16 at 16 53 16

andysellick: I've checked the markup layout on [this page](https://www.gov.uk/check-uk-visa/y/afghanistan/work/longer_than_six_months/academic) but the results pages can vary so worth checking a few to make sure the layout is consistent.

MartinJJones: Thanks Andy. I have updated the code as suggested and works as expected.

After further testing and checking different results page, I found a couple of issues which have been fixed and updated in the PR description.

andysellick (approved):
Looks good, happy to approve, although could you add a note explaining the change to the result card component? We discussed it and it makes sense, but I think it's worth documenting - I think maybe if you put a bit more detail into the commit message generally, and include that detail in that as well, thanks.

MartinJJones:
Thanks Andy, I added a note explaining the change to the result card component and updated the commit message.

PR #5969, opened 2022-06-15
Check benefits financial support flow tests

Rosa-Fox:
This PR adds outcome tests to the check-benefits-support flow test file. It only tests "happy" paths to ensure desired outcomes are present in the result given the necessary criteria.

### DIFF
--- a/test/flows/check_benefits_support_flow_test.rb
+++ b/test/flows/check_benefits_support_flow_test.rb
@@ -374,6 +374,15 @@ class CheckBenefitsSupportFlowTest < ActiveSupport::TestCase
       end
     end
 
+    should "render Universal Credit (Northern Ireland) when eligible" do
+      add_responses where_do_you_live: "northern-ireland",
+                    over_state_pension_age: "no",
+                    assets_and_savings: "under_16000"
+
+      assert_rendered_outcome text: "Universal Credit"
+      assert_rendered_outcome text: "Check if you’re eligible for Universal Credit on the nidirect website"
+    end
+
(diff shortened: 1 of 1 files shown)


test/flows/check_benefits_support_flow_test.rb
```diff
@@ -455,5 +455,12 @@ class CheckBenefitsSupportFlowTest < ActiveSupport::TestCase
         assert_rendered_outcome text: "Check if you’re eligible for 30 hours free childcare"
       end
     end
+
+    should "render Carer's Allowance when eligible" do
+      add_responses carer_disability_or_health_condition: "yes"
+
+      assert_rendered_outcome text: "Carer's Allowance"
```
leenagupte, line 462: This is unrelated to the changes you are making, but the apostrophe in Carer's Allowance should be a typographic quote, i.e. `’`.  I think this one was missed by accident as the other quotes seem to be correct.

leenagupte (commented):
Ed this is looking great.  There are couple of copy and paste errors that need to be fixed but nothing major.

The structure of the YAML data file and the calculator made this really easy to review, so thank you (and Rosa)! 🎉

test/flows/check_benefits_support_flow_test.rb
edwardkerry: good spot!

edwardkerry:
@leenagupte thanks for catching those errors, I've fixed them up now so ready for a re-review. Thank you!

leenagupte (approved):
Looks good! 💯

PR #5979, opened 2022-06-20
Pass GOVUK_CI_GITHUB_TOKEN to deploy workflow

theseanything:
The deploy workflow now requires the GOVUK_CI_GITHUB_TOKEN to make the API call to check if the actor has the correct permissions to trigger a deployment.


kevindew (approved):
Thanks

PR #5988, opened 2022-06-27
Removing unneeded files from public

andysellick:

## What
Remove files from [public](https://github.com/alphagov/smart-answers/tree/main/public).

## Why
These are the error pages, favicon and robots.txt from a default rails app. They were added 11 years ago and haven't been modified since. Our application `static` handles error pages, favicon and (I believe) robots.txt, so it seems like we don't need these files.

## Visual changes
None.


andysellick:
@kevindew can you take a quick look and confirm I'm not going to horribly break anything with this PR? Thanks.

kevindew (approved):
Yes, these aren't used in public rendering and I agree with their removal.

There are a couple of scenarios I was concerned about and have done brief investigations on:

- whether a public directory is needed, i.e. an error will occur without it and we need to put a .gitkeep in it. I notice Content Publisher doesn't have one, https://github.com/alphagov/content-publisher, so conclude that is ok
- whether Rails will error when trying to render a 404 without these, in a quick test it seems to handle this fine - although I found a Smart Answers bug that should be fixed (separate to this PR)

andysellick:
Thanks @kevindew. Yes I wasn't sure if removing the files would automatically remove the directory, so I removed only the files in case the dir was needed - but it does and it isn't so that's fine. Thanks!

PR #5999, opened 2022-07-04
Update property fire safety payment copy

edwardkerry:
This PR introduces multiple changes in copy to questions and views.
It also makes a small number of technical changes
- changes a date validation from 1945 to 1900
- updates a percentage validation to check between 0.1 and 1 rather than 0 and 100
- moves a percentage validation from the flow to the calculator, following smart-answer ADR guidance
- some refactoring to move a magic number to a constant


KludgeKML (approved):
LGTM, and I learned what a carat does in govspeak.

PR #6003, opened 2022-07-06
Marriage abroad Russia update July 22

stwcontent:
https://govuk.zendesk.com/agent/tickets/4862149
https://trello.com/c/OKW1xllv/3899-getting-married-abroad-russia-uk-cni-exchange-process-fee

3 commits to add information about certificates of no impediment and fee payment to 3 outcomes.

Also:

* reworked list of register offices
* fixed missing link
* reworked headers to deal with new information

### DIFF
--- a/app/flows/marriage_abroad_flow/outcomes/countries/russia/uk/partner_british/_opposite_sex.erb
+++ b/app/flows/marriage_abroad_flow/outcomes/countries/russia/uk/partner_british/_opposite_sex.erb
@@ -2,22 +2,71 @@ Contact the [Embassy of Russia](/government/publications/foreign-embassies-in-th
 
 ^You should [get legal advice](/government/collections/list-of-lawyers) and check the [travel advice for Russia](/foreign-travel-advice/russia) before making any plans.
 
-##What you need to do
+## What you need to do
 
 You’ll be asked to provide a certificate of no impediment (CNI) to prove you’re allowed to marry.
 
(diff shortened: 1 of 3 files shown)


stwcontent:
Preview shared on Zendesk. Fact check amends received. Uopdating with fresh commits.

Comes down to 2 things: 

* Whole section about CNI validity seems off to them (the comments aren’t especially clear, but we’ve got to the bottom of them I think)
* Payment info is inaccurate. 

1.

CHANGE
### How long a CNI is valid for

Your CNI must be valid both: 

* in the issuing country 

* under local law in Russia

A CNI issued in England, Wales or Northern Ireland is valid forever. A CNI issued in Scotland is valid for 3 months.

If you live in the Isle of Man, Jersey or Guernsey, check with your local registrar to find out how long a CNI is valid for.

You must check with the local authorities in Russia to find out how long a CNI is valid under local law.

TO

^ A Scottish CNI is only valid for 3 months from the date it’s issued



REASON
It looks like we went over the top here - no need to mention that other CNIs do not expire - Scotland is really the exception, and we can get away with warning about that particular case. 
Also no need to mention checking local validity/expiration - as they say, your CNI will become valid in Russia when you legalise or exchange it. Checking that it is valid is redundant.
So the whole section seems redundant. 

2.

CHANGE

### If you want to exchange your CNI

You must [make an appointment at the British embassy to exchange your CNI](https://www.book-consular-appointment.service.gov.uk/TimeSelection?location=132&service=4).

It will cost £50.

You normally have to pay fees for consular services in the local currency – these are shown in the list of [consular fees for Russia](/government/publications/russia-consular-fees).

If you’re paying by credit card, you can use either MasterCard or Visa.

TO


### If you want to exchange your CNI

You must [make an appointment at the British embassy to exchange your CNI](https://www.book-consular-appointment.service.gov.uk/TimeSelection?location=132&service=4).

You’ll have to pay a fee. Check how much it will cost on the list of [consular fees for Russia](/government/publications/russia-consular-fees).

You can pay using MasterCard or Visa. 

REASON
After pushing for the figure previously, they’ve now suggested (on previous GDS advice) that we just give the link. OK!
No messing about with local currency - it’s just mastercard and visa at this point.

PR #6006, opened 2022-07-06
Fix raising InvalidNode errors based on user input

kevindew:

This resolves a regression introduced by https://github.com/alphagov/smart-answers/commit/152f74a3e6a3739fc45f45734b5e145e83bfb6f1 which led to 500 errors in Smart Answers when a user provided incorrect input. The situation was when a user was requesting a Smart Answer and had a path that had responses beyond what produces an outcome the application raises an error. (the example situation is: given a request of `/state-pension-age/y/age/2022-01-11/male` we only need `age/2022-01-11` to produce an outcome and `/male` is superfluous - when the application tries to process the `male` argument the application errors).

To resolve this I've created a new type of exception to cover this event that can be converted into a 404, so that this can be distinct from the configuration issues that InvalidNode exposes.

More info in the commits.


maxgds:
I'll merge to quieten sentry, thanks both.

PR #6008, opened 2022-07-07
Move PropertyFireSafetyPayment currency methods from view to calculator

edwardkerry:
This PR extracts some rather verbose and repeated currency helper methods from the view to the calculator.

It also allows us to move some methods from public to private, and update tests accordingly


1pretz1 (approved):
Awesome!

PR #6012, opened 2022-07-07
Change property-fire-safety-payment to check-fire-safety-costs

edwardkerry:
This PR changes the slug of the smart-answer from property-fire-safety-payment to check-fire-safety-costs.

It updates all the associated files and class names to match this new name.

It also makes some small changes to copy in the start page and outcomes.


KludgeKML (approved):
LGTM

PR #6020, opened 2022-07-12
Remove track results module

MartinJJones:
## What

Remove the "track-results" JavaScript module used for firing Google Analytics events.

## Why

- The "track-results" module in Smart Answers does not appear to be required. The same functionality can be achieved using the "track-click" and "auto-track-event" modules from "govuk_publishing_components"
- Help avoid duplicate GA events, It is still possible to use both modules to track Google Analytics events in Smart Answers
- Less code and tests to maintain

## Visual Changes
There are some key differences if we remove the "track-results" module and instead use "govuk_publishing_components"

### First Difference: The data sent for click events

There are some differences between the modules in how they get the Event Data to see to GA, because of this, there will be some differences, but this should be consistent with the approach used across GOV.UK

| | Before | After |
| - | - | - |
| Event Action | /student-finance/who-qualifies | student-finance-calculator results |
| Event Category | Internal Link Clicked | Internal Link Clicked |
| Event Label | check if you qualify | /student-finance/new-fulltime-students |

### Second Difference: Removing the track-click module from Smart Answers allows External Links to be tracked
Currently the "track-results" module only tracks internal links on the Smart Answers results page, the "track-click" module will include clicks on all links, both internal and external.

Therefor, we would likely see an increase for click events on the Smart Answers results pages, as external links would also be tracked.

An example of this is shown in the videos below for the cost of living smart answer results page.

#### Before
https://user-images.githubusercontent.com/28779939/179217887-3e698895-53ea-4297-b335-323e3ecee497.mov

#### After
https://user-images.githubusercontent.com/28779939/179217999-583f0cf6-3aad-4d7c-aaaa-1a6ebc30fa74.mov

Trello: https://trello.com/c/Vl2qEqFw/1353-review-the-track-result-module-and-requirements

---


andysellick (approved):
Code changes look good. Well done for removing this module 🚀

PR #6024, opened 2022-07-13
Rename check-fire-safety-costs to  check-building-safety-costs and update copy

edwardkerry:
This PR renames the smart-answer from check-fire-safety-costs to check-building-safety-costs. It renames all associated files and classes.

It also updates copy across questions and answers, and adds an additional method to the controller in order to present the minimum valuation limit to an outcome.


KludgeKML (approved):
LGTM

PR #6030, opened 2022-07-14
Update logic for under_valuation_limit to exclude lower limit

edwardkerry:
check-building-safety-costs has two calculator methods to compare a property valuation to a lower valuation limit for both inside and outside London.

Currently the comparison is `valuation less than or equal to lower limit`
This updates the comparison to `valuation less than lower limit`


KludgeKML (approved):
Looks okay!

KludgeKML (approved):
LGTM again!

PR #6045, opened 2022-07-22
Update Dockerfile to use new GOV.UK Ruby base images

samsimpson1:
Updates the Dockerfile to use the new [GOV.UK Ruby base images](https://github.com/alphagov/govuk-ruby-images). Also updates the .ruby-version file to not specify a patch version, to allow for easier Ruby patching.

Context: https://trello.com/c/Zy0fd25w/970-use-base-builder-images-in-all-the-app-dockerfiles


fredericfran-gds (approved):
ok

PR #6055, opened 2022-07-26
Turn off font

jon-kirwan:
## What

Prevent the font from being included in `smart-answers`'s stylesheet.

## Why

We should be serving the font files from `static` so they're cached across the entirety of GOVUK - so the font shouldn't be included in `smart-answers`'s stylesheet.

With the font being served by each application, users need to re-download it every time they went from a page rendered in one application to a page rendered in another application.

Serving it from `static` means that the same font URI is used in each application, allowing the fonts to be cached and reused.

## Visual changes

None - the same font is being served from `static`, instead of `smart-answers`.

Before (the font paths contains `smart-answers`):

After (the font path contains `static`):

---

## Search page examples to sanity check:

- https://smart-answers-pr-6055.herokuapp.com/additional-commodity-code

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -2,6 +2,11 @@ $govuk-compatibility-govuktemplate: false;
 $govuk-use-legacy-palette: false;
 $govuk-new-link-styles: true;
 
+// This flag stops the font from being included in this application's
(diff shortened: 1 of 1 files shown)


KludgeKML (approved):
LGTM

injms (approved):
Super - thanks @jon-kirwan!

PR #6064, opened 2022-08-01
Refactor check benefits support Smart Answer

1pretz1:
PR defines the where the user lives in YAML file, rather than as a conditional.

Has a few benefits:

1) DRYs the code out, as previously there were individual methods with
   the same logic for one or more regions. We can now use one generic
   method if the logic is the same. We can also remove methods entirely,
   if the benefit only depends on the country.

   The above also has means that a large amount of duplicate tests can
   be removed.

2) We can remove the `where_do_you_live` conditional from all the
   methods, meaning much lighter weight tests as they don't have to
   iterate over 4 countries per test.

3) Removing tests from point 1 and 2 won't reduce test coverage, as the
   flow tests (check_benefits_financial_support_flow_test.rb) cover the
   all the outcomes.

3) Modifying benefit countries is slightly easier in a YAML file, rather
   than through conditionals.

Trello:
https://trello.com/c/96VGgATX/1398-smart-answer-new-benefits-to-the-benefits-checker-cost-of-living-checker-v2


KludgeKML (approved):
LGTM

PR #6093, opened 2022-08-17
Add tests for result section and item components

MartinJJones:
## What

Add the missing tests for the `result-sections` and `result-item` components.

## Why

No tests exist for these components.


app/views/components/_result-sections.html.erb
```diff
@@ -2,6 +2,10 @@
   highlighted ||= true
   group_index ||= 1
   result_index ||= 1
+
+  if !topics.any?
+    raise ArgumentError, "The result section component requires topics data"
+  end
 %>
```
jon-kirwan, line 9: I think I've noticed something with `!topics.any?` which prevents the `ArgumentError` from being raised from the tests.

I've run locally with some simple logging, `bin/rails test test/components/result_sections_test.rb:39`:

```
test "fails to render when no data is given" do
  error = assert_raises do
    render_component({})
  end

  puts error
end
```

And see:

`undefined local variable or method topics for #<ActionView::Base:0x0000000001e3c0>`

However, when I add `topics ||= []` and rerun the tests (and logging), I see:

`The result section component requires topics data`.

So, whilst the tests do pass, I think the test can't execute the `.any?` method on topics (because it's `undefined` here) and the test passes but for a different reason to the expected `ArgumentError`.

jon-kirwan (commented):
Hi Martin, looks really good thanks 👍.

I think I may have noticed something with the result sections tests and have added some comments for `app/views/components/_result-sections.html.erb` (although I'm not 100% sure so it might be fine)

app/views/components/_result-sections.html.erb
MartinJJones: @jon-kirwan Thanks very much for the info, really helpful.

I tested this out myself again and decided to go with the approach of ensuring that topic is given a default value. This helps to avoid the undefined error mentioned, as well as make it clear that it is an available attribute, this approach is also used in the autocomplete component - https://github.com/alphagov/smart-answers/blob/main/app/views/components/_autocomplete.html.erb#L6

jon-kirwan (approved):
Thanks for the making the change. All looks good to me! 🚀 .

PR #6141, opened 2022-09-20
Update check-uk-visa results

MartinJJones:
This PR moves the outcome content for "work" related visas in the check-uk-visa smart-answer from ERB partials into a YAML file.

It also introduces a new "visa card" frontend component to display each available visa on the outcome page.

[trello 1](https://trello.com/c/NWxiv3CU/1455-add-work-visa-content-to-a-yml-config-file)
[trello 2](https://trello.com/c/TW774RVY/1453-create-new-card-component-in-smart-answers)


app/views/components/check_uk_visa/_result-card.html.erb
```diff
@@ -0,0 +1,60 @@
+<%
+  type ||= nil
+  title ||= nil
+  description ||= nil
+  attributes ||= []
+  url ||= nil
```
1pretz1, line 6: Minor: we should remove the default values if we are requiring values to be present.

app/views/components/check_uk_visa/_result-card.html.erb
```diff
@@ -0,0 +1,60 @@
...
+
+  if !type || !title || !url || !attributes.any? || !calculator
```
1pretz1, line 8: Is the [results card validation](https://github.com/alphagov/smart-answers/pull/6141/commits/e90f88ee135ba8aed644f542c58a8fa94412ce95#diff-bb4685394a5e3a9f03f288b3a808404314e69033c7e806394f9de77aae8eb6c6R8) to ensure the YAML data is valid? If so, I think we should probably convert this validation into a test, rather than at runtime validation as the goal is to test our own config, not users input. Errors could also slip through to production if our checks / test coverage isn't comprehensive.

I'd suggest adding a test that parses the YAML file and checks the data integrity within. The [calculator spec](https://github.com/alphagov/smart-answers/blob/main/test/unit/calculators/uk_visa_calculator_test.rb) probably makes most sense, or potentially the [result card spec](https://github.com/alphagov/smart-answers/blob/91eeede53e1cc80f2bfc35e1e521304ae425b0ad/test/components/check_uk_visa/result_card_test.rb.

Thoughts?

app/views/components/check_uk_visa/_result-card.html.erb
```diff
@@ -0,0 +1,60 @@
+<%
```
1pretz1, line 1: Is this component intended to be reused? If not, then maybe the best place is: https://github.com/alphagov/smart-answers/tree/main/app/flows/check_uk_visa_flow, as app/components feels like a shared directory.

1pretz1 (commented):
Nice work, looking great! Just have a comment about how we're handling the results card validation.

app/views/components/check_uk_visa/_result-card.html.erb
edwardkerry: I believe this component is intended to be reused in future work, although all within the check-uk-visa flow. I think there are plans for a mix of generic and flow-specific components, so `app/views/components` is being used to hold all these templates and the `app/flows/*` folders will hold "content" templates.

app/views/components/check_uk_visa/_result-card.html.erb
edwardkerry: This sounds good to me, although it seems we have a pattern here based on the other components in
https://github.com/alphagov/smart-answers/tree/91eeede53e1cc80f2bfc35e1e521304ae425b0ad/app/views/components

app/views/components/check_uk_visa/_result-card.html.erb
1pretz1: Ah okay, thanks for explaining!

app/views/components/check_uk_visa/_result-card.html.erb
1pretz1: Me and Ed discussed the above offline and agreed that tests would be more suited, rather than view validation for this use case. As this is an emerging pattern and affects other components, it's deemed as non-blocking. However, we have created a tech-debt card regarding this issue: https://trello.com/c/47h58IB7/1609-convert-view-level-validation-pattern-to-equivalent-tests-for-components-in-smart-answers.

1pretz1 (approved):

app/views/components/check_uk_visa/_result-card.html.erb
jon-kirwan: Thanks Pete, I'll leave them in for now but sounds good to me to update the view level validation to tests (and remove the default values whilst doing so).

jon-kirwan (approved):

PR #6156, opened 2022-09-30
Remove all the extra binstubs

kevindew:
I'm not sure quite why we have all these, I guess at one point in time someone generated them for every gem.

Anyway this deletes all but the ones Rails installs.

This was prompted by finding that this app referenced the Slimmer render_slimmer_error bin that will soon be deleted.


1pretz1:
Nice spot! From looking at other apps the common binstubs are `bundle rails  rake  setup  update`. I can see we've removed `bundle` and missing `update`. Do we want to add these?

kevindew:
@1pretz1 So I just added all the ones `rails app:update` adds and deleted every other one. I'm not sure anyone has a use for the ./bundle or ./update ones. It'd seem more prudent to add them in if/when people were surprised that they were missing, as I imagine we just build up ones that aren't needed.

1pretz1 (approved):
Sounds sensible - thanks for explaining! 💯

PR #6163, opened 2022-10-05
Use `Plek.new` as `current` is deprecated

1pretz1:
This is causing logging noise.

Trello:
https://trello.com/c/LVVUbd4m/1551-replace-plekcurrent-with-pleknew-in-smartanswer

### DIFF
--- a/lib/link_checker.rb
+++ b/lib/link_checker.rb
@@ -49,7 +49,7 @@ private
 
   def link_checker_api
     @link_checker_api ||= GdsApi::LinkCheckerApi.new(
-      Plek.current.find("link-checker-api"),
+      Plek.new.find("link-checker-api"),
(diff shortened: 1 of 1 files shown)


KludgeKML (approved):
LGTM

lib/link_checker.rb
```diff
@@ -49,7 +49,7 @@ def generate_report
 
   def link_checker_api
     @link_checker_api ||= GdsApi::LinkCheckerApi.new(
-      Plek.current.find("link-checker-api"),
+      Plek.new.find("link-checker-api"),
```
leenagupte, line 52: I think you can just use `Plek.find` here. See: https://github.com/alphagov/plek/blob/0f9c8e8bccc53f01d180a2fe80aabfc853dcbdc6/lib/plek.rb#L73

1pretz1: That is cleaner!

PR #6212, opened 2022-11-30
Update Check Benefits Financial Support copy and results

edwardkerry:
This PR updates the copy for the start and results page. It reorders Warm Home Discounts results so they appear more prominently. It also adds new results for Scottish Child Payment and Older Persons Bus Pass (Scotland and Northern Ireland)


KludgeKML (approved):
LGTM

PR #6256, opened 2023-01-09
Remove autocomplete component code

jon-kirwan:
## What

Remove the Autocomplete component code since it's not being used anywhere and there are currently no plans to use the component.

## Why

The component is not used anywhere.

## Anything else

The commit reverts most of the changes added here: https://github.com/alphagov/smart-answers/pull/5694.

The one exception not removed is `test/component_test_helper.rb` - now used with app components like `_result-card`.


config/initializers/assets.rb
```diff
@@ -9,9 +9,4 @@
 # Precompile additional assets.
 # application.js, application.css, and all non-JS/CSS in the app/assets
 # folder are already added.
-# Rails.application.config.assets.precompile += %w( admin.js admin.css )
-
-# Add Yarn node_modules folder to the asset load path.
-# Rails.application.config.assets.paths << Rails.root.join('node_modules')
-
-Rails.application.config.assets.paths << Rails.root.join("node_modules")
+Rails.application.config.assets.precompile += %w[test-dependencies.js]
```
MartinJJones, line 12: We should be able to remove this line of code as well, it was later removed as part of this commit - [Move to manifest.js for new sprockets](https://github.com/alphagov/smart-answers/pull/5881/commits/1961b28a5e30eea30c929b5d754cba08f3d2e1bd)

MartinJJones (commented):
All of the other changes look good to me, think we just need to remove the 1 line of code in `assets.rb`

PR #6326, opened 2023-02-27
Update same sex outcomes for Switzerland

richard-gds:
https://govuk.zendesk.com/agent/tickets/5222150
https://trello.com/c/rGsCqakL/4393-getting-married-abroad-switzerland-same-sex-marriage-content-change-request

Same sex marriage is now legal in Switzerland as of 1 July 2022. As a result, it is no longer possible to enter into same sex civil partnership in Switzerland.

Same sex couples can now follow the same process as opposite sex couples, but can also convert their registered partnership into a marriage, or enter into a new marriage.


KludgeKML (approved):
Looks good to me

PR #6380, opened 2023-04-19
Update State Pension age review link

richardskipper:
https://govuk.zendesk.com/agent/tickets/5296413
https://trello.com/c/tFchB6Lm/4252-state-pension-age-review

The review into changing the State Pension age has concluded, so we need to update information on the start page of this tool to alert users to the current situation.


gclssvglx (approved):
Content only change 👍

PR #6469, opened 2023-07-21
Revert "Update world location code for breaking changes in `gds-api-adapters` version 89.0.0"

ChrisBAshton:
Reverts alphagov/smart-answers#6463, as it causes issues with the alphabetical ordering of countries on the "Check if you need a UK visa" smart answer.

Zendesk: https://govuk.zendesk.com/agent/tickets/5416650

PR #6473, opened 2023-07-24
Revert "Revert "Update world location code for breaking changes in `gds-api-adapters` version 89.0.0""

brucebolt:
Reverts alphagov/smart-answers#6469

We can revert the change, as the issue was fixed in https://github.com/alphagov/whitehall/pull/8002.

[Trello card](https://trello.com/c/eKjf7gPC)


ChrisBAshton (approved):
Thanks Bruce ⭐

PR #6486, opened 2023-07-28
Remove uk-mission-to-the-wto-un-and-other-international-organisations…

Rosa-Fox:
…-geneva from country select

- This is causing the [Marriage Abroad flow to break](https://www.gov.uk/marriage-abroad/y/uk-mission-to-the-wto-un-and-other-international-organisations-geneva/third_country/partner_british) as this location does not appear in the marriage_abroad_data.yml for `countries_with_x_outcomes`.
- This requires further investigation but it is likely that an update to the World Location API has caused this to appear in the country select unexpectedly.
- In the short term it has been excluded from the country select.

### DIFF
--- a/app/flows/marriage_abroad_flow.rb
+++ b/app/flows/marriage_abroad_flow.rb
@@ -12,7 +12,7 @@ class MarriageAbroadFlow < SmartAnswer::Flow
     name "marriage-abroad"
     status :published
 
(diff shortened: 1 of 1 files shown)


brucebolt:
This location is set incorrectly as a "World Location" (rather than "International Delegation") in Whitehall.  Changing that will fix the issue.

The filtering code is [here](https://github.com/alphagov/smart-answers/blob/43cec8531b3518853b5cbd92756215c0eac0830a/app/models/world_location.rb#L39).

PR #6536, opened 2023-09-06
Revert "CW/SS - Remove reference to a repo during checkout in minitest.yml"

jon-kirwan:
## What

Add back in reference to `alphagov/smart-answers` repo.

## Why

Removing the repo seems to be generating an error when running **Test Content Schemas** workflows from Publishing API. See https://github.com/alphagov/publishing-api/actions/runs/6084918130/job/16536652209.


FelixMarcusMillne (approved):
Reverting for now as suspected of breaking other systems

PR #6555, opened 2023-09-18
Implement routes redirect for register-a-death abroad

anatron:
This PR will implement a basic redirect solution for orphaned register a death abroad URLs.

The implemented code specifically handles register a death abroad only. Arguably to improve quality this could be made more generic, however we believe this is beyond the scope of the register a death abroad ticket and we will raise another tech debt ticket to cover the generic case.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -10,6 +10,9 @@ Rails.application.routes.draw do
   get "/:id", to: "flow#landing", as: :flow_landing
 
   get "/:id/y/visualise(.:format)", to: "smart_answers#visualise", as: :visualise
+
+  get "/register-a-death/y/overseas/*other", to: redirect("/register-a-death/y/overseas")
+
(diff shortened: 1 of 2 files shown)


config/routes.rb
```diff
@@ -10,6 +10,9 @@
   get "/:id", to: "flow#landing", as: :flow_landing
 
   get "/:id/y/visualise(.:format)", to: "smart_answers#visualise", as: :visualise
+
+  get "/register-a-death/y/overseas/*other", to: redirect("/register-a-death/y/overseas")
```
FelixMarcusMillne, line 14: Looks good - one small request, could we please add a comment in here so that it's easy for any future devs (including us) to remember why this is in here? 🙏

anatron: @FelixMarcusMillne Hey Felix, thanks for the review. We have now added a comment! 👍

PR #6614, opened 2023-11-24
Update timelines for swearing an affirmation

richardskipper:
https://govuk.zendesk.com/agent/tickets/5535950
https://trello.com/c/lFPh9BEz/4655-inconsistency-in-thailand-content-in-getting-married-abroad-smart-answer

Removed the '45 days' timeline for swearing an affirmation, as this is not accurate. Users can swear their affirmation anytime after they've applied for it, which should be within 3 months of the wedding day.


cynthia-anya (approved):
LGTM

PR #6616, opened 2023-11-28
Use govuk frontend v5

patrickpatrickpatrick:
[Trello](https://trello.com/c/srVLB7Bk/2275-update-frontend-rendering-applications-using-govukpublishingcomponents-xl)

## What

- Replace Uglifier with Terser, add patch which adds Sprockets support for Terser
- Update how JS modules are initialised in-line with changes in govuk-frontend and govuk_publishing_components
- Move components that rely on govuk-frontend modules to separate `es6-components.js` file
- Remove deprecated scss variables

## Why

govuk_publishing_components uses v5 of govuk-frontend. There are several breaking changes in this release and so we need to update the applications that use govuk_publishing_components. Please read what has changed in [govuk_publishing_components](https://github.com/alphagov/govuk_publishing_components/pull/3700) and [govuk-frontend](https://github.com/alphagov/govuk-frontend/releases/tag/v5.0.0-beta.0). This specific changes this PR addresses are:

- Replace Uglifier with Terser, add patch which adds Sprockets support for Terser
  *  govuk-frontend v5 UMD component JS uses ES6 features which aren't supported by Uglifier, so we have to make the change to use Terser instead
  * Sprockets doesn't support Terser by default, need to add in a patch which adds Terser to list of supported JS compressors
- Update initialisation method of application JS modules
  *   Initialisation is now called upon module creation, so init methods have been moved to the constructors to reflect this
- Move components that rely on govuk-frontend modules to seperate `es6-components.js` file
  * In the event that a browser below the target for `govuk-frontend` loads a page with JS on it, attempting to parse the JS from `govuk-frontend` will cause an error. To avoid this from happening, JS that contains `govuk-frontend` JS has been moved to seperate file which will be loaded in a script tag with `type="module"`. This will prevent the JS from being parsed and so prevent the error
- Remove deprecated scss variables
  * In govuk-frontend v5 including $govuk-compatibility-govuktemplate (removed), $govuk-use-legacy-palette (deprecated) and $govuk-new-link-styles (now always default) in a file will have no effect


patrickpatrickpatrick:
This is blocked by https://github.com/alphagov/govuk_publishing_components/pull/3700

PR #6634, opened 2023-12-20
Migrate to Dart Sass from LibSass [DO NOT MERGE] 

matthillco:
## What

Migrate Smart Answers to Dart Sass from LibSass

## Why

- LibSass is now [deprecated](https://sass-lang.com/blog/libsass-is-deprecated/)
- Making changes ahead of Design System team dropping support for LibSass https://github.com/alphagov/govuk-frontend/discussions/2632.

## Visual Changes

None.

## Anything else

Once the PR gets the green light, any temporary commits and references to local gems e.g. `gem "govuk_publishing_components", path: "../govuk_publishing_components"` will be removed.

See:
- https://github.com/alphagov/govuk_publishing_components/pull/3726
- https://github.com/alphagov/email-alert-frontend/pull/1655
- https://github.com/alphagov/static/pull/3190
- https://github.com/alphagov/govuk-developer-docs/pull/4216.


bin/dev
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env sh
+
+if ! gem list foreman -i --silent; then
+	echo "Installing foreman..."
+	gem install foreman
+fi
```
jon-kirwan, line 6: Minor comment regards indentation. Can you please modify to 2 spaces:
```
if ! gem list foreman -i --silent; then
  echo "Installing foreman..."
  gem install foreman
fi
```

config/initializers/dartsass.rb
```diff
@@ -0,0 +1,11 @@
+# config/initializers/dartsass.rb
+APP_STYLESHEETS = {
```
jon-kirwan, line 2: Another small comment. I think you can remove the comment on line 1 (apologies, I know that it might be included in the docs in which case I'll also remove it from there)

config/environments/production.rb
```diff
@@ -26,8 +26,6 @@
 
   # Compress JavaScripts.
   config.assets.js_compressor = :uglifier
-  # Compress CSS using a preprocessor.
-  # config.assets.css_compressor = :sass
 
   # Do not add source map line comments into css files
   config.sass.line_comments = false
```
jon-kirwan, line 31: It looks like `config.sass.line_comments` was added twice (probably because of a Rails update). Could you please remove it and the comment on line 30.

config/initializers/dartsass.rb
```diff
@@ -0,0 +1,11 @@
...
+  "application.scss" => "application.css",
+  "components/_result-card.scss" => "components/_result-card.css",
+  "components/_result-item.scss" => "components/_result-item.css",
+  "components/_result-sections.scss" => "components/_result-sections.css",
+  "visualise.scss" => "visualise.css",
+}.freeze
+
+all_stylesheets = APP_STYLESHEETS.merge(GovukPublishingComponents::Config.all_stylesheets)
+Rails.application.config.dartsass.builds = all_stylesheets
```
jon-kirwan, line 11: This hasn't been documented yet (I'll make a note to include it in the documentation), but could you please add an additional line of configuration to silence dependency deprecation warnings like so:

```
Rails.application.config.dartsass.builds = all_stylesheets

Rails.application.config.dartsass.build_options << " --quiet-deps"
```
I initially thought that adding this line in the GOV.UK Publishing Components would suffice, but it appears that including it here is also necessary.

.gitignore
```diff
@@ -23,4 +23,8 @@ node_modules
 yarn-error.log
 
 # IntelliJ
-.idea/
\ No newline at end of file
+.idea/
+
+# DartSass
```
jon-kirwan, line 28: Very minor suggestion but could you please change `# DartSass` to `# Sass` or `# Dart Sass`.

jon-kirwan (commented):
Thanks Matt, great work. All looks good and works fine. I've just added a couple of small comments and a general note around vendor CSS

jon-kirwan:
I'm adding a note regarding the vendor CSS which is currently uncompressed. It seems that `joint.css` is designed for internal use, meant to visualize Smart Answer journeys. To compress CSS files, it looks like the typical approach involves adding Sass configuration, like this:

```
config.sass.style = :compressed
config.sass.line_comments = false
```

However, this process seems closely linked to the sassc-rails gem, which has been removed. The only alternative I've come across for CSS compression is to add `config.assets.css_compressor = :yui` which requires the yui-compressor gem https://guides.rubyonrails.org/asset_pipeline.html#css-compression.

At present, the vendor CSS folder doesn't seem to be used in other UX applications that are being migrated to Dart Sass.

Otherwise, Sass files are compressed before being handled by the asset pipeline.

matthillco:
@jon-kirwan I've made the requested changes, please could you have a final check? Thanks.

jon-kirwan (approved):
All working fine for me and the changes look great! Thanks Matt 👍

PR #6639, opened 2024-01-05
Move Qatar from 'visa national' to 'non-visa national'

cynthia-anya:
As a result of moving Qatar from 'visa national' to 'non-visa national', changes needed to be made to the expected outcomes and questions. 

As an electronic travel authorisation country, Qatar does not have the same behaviour as other non-visa national countries. 

For more detail view [Trello ticket](https://trello.com/c/C9PPOd6d/224-eta-related-move-qatar-from-visa-national-to-non-visa-national)


anatron (approved):
Looks good!

PR #6650, opened 2024-01-29
Fix formatting in .govuk_dependabot_merger.yml

ChrisBAshton:
This invalid YAML was causing build failures in govuk-dependabot-merger (now fixed in https://github.com/alphagov/govuk-dependabot-merger/pull/41).

This PR fixes the configuration so that Whitehall dependency PRs  can be auto-merged again.

It also rearranges the dependencies alphabetically.


ChrisBAshton:
(Raised in wrong repo - apologies!)

PR #6654, opened 2024-01-31
Remove obsolete Jenkinsfile.

sengi:
This is a bulk change and will be bulk-approved.

There is no need to review individually.


dj-maisy (approved):
🔥

PR #6702, opened 2024-03-05
Update Belgium civil partnerships information

richardskipper:
Removing any mention of civil partnerships, as the embassy in Belgium cannot help users with getting a civil partnership.

Also removing old links about documents you'll need, as the information in the link does not exist anymore.


cynthia-anya (approved):
LGTM

PR #6729, opened 2024-03-22
Eta September 2024 nvn/eea warning callout

isabelA211:
Content storytime: https://trello.com/c/SwjCds1e/263-3-april-eta-callout-for-2-user-groups-on-check-if-you-need-a-uk-visa

Adding in a callout for non-visa national and EEA user groups that they may need an ETA if coming to the UK after September 2024.


mtaylorgds:
Closing PR as the preview environment has expired–see #6788 instead now.

PR #6736, opened 2024-04-03
Simplify path extraction for ga4_url.

sengi:
Should maybe use [URI] here, but that might change some edge cases that I'm not entirely sure are covered by the tests. So let's just simplify what's there, preserving the existing behavior.

Might resolve https://github.com/alphagov/smart-answers/security/code-scanning/1. (Not an actual security issue, just a warning from a heuristic.)

[URI]: https://docs.ruby-lang.org/en/3.3/URI.html


AshGDS (approved):
Thanks 👍

PR #6739, opened 2024-04-05
Fix redundancy calculator tests date issue

mtaylorgds:
A couple of tests were comparing the current *Time* to the end date specified by the retrieved rates. This comparison was wrong in a couple of ways:

1. It did a less-than comparison, rather than a less-than-or-equal-to, which meant that running the test on the end date itself would cause the test to fail, even though that is a valid date (start and end dates are inclusive).
2. It used the current *Time*, not the current *Date*, which meant that even fixing the above comparison issue, would still cause the tests to fail, as it was comparing, for example, 10am (now) with the end date (00h).

[Trello ticket
](https://trello.com/c/3fQj2IAW)


anatron (approved):
LGTM

PR #6753, opened 2024-04-16
Updating employee parental pay calculator for users in Northern Ireland

anatron:
[Trello card](https://trello.com/c/6jejLsQQ/269-updating-employee-parental-pay-calculator-for-users-in-northern-ireland)

For this PR we have:
- Added a new question to the calculator
- Updated the calculations for NI users on the outcome page
- Removed the notice for NI users from the start page


test/flows/maternity_paternity_pay_leave_flow_test.rb
```diff
@@ -38,4015 +50,4038 @@ class MaternityPaternityPayLeaveFlowTest < ActiveSupport::TestCase
...
-                    employment_status_of_mother: "employee"
-    end
 
-    should "render the question" do
-      assert_rendered_question
-    end
+    context "question: employment_status_of_mother" do
+      setup do
+        testing_node :employment_status_of_mother
+        add_responses two_carers: "yes",
```
mtaylorgds, line 85: Since the tests inside this context block don't all use the same response for `two_carers`, (in fact 2/3 require a "no" response), I'd push this `add_response` down into the individual tests (i.e. add it to the one test that doesn't already do it).

mtaylorgds (changes requested):
Happy to discuss any of these comments if they aren't clear, or you want to push back on any.

mtaylorgds (changes requested):
A few minor comments.

test/flows/maternity_paternity_pay_leave_flow_test.rb
```diff
@@ -177,6 +202,7 @@ class MaternityPaternityPayLeaveFlowTest < ActiveSupport::TestCase
       should "have a next node of partner_started_working_before_continuity_start_date " \
                "for a 'yes' response " \
                "when there are two carers " \
+               "and the partner is from England"  \
```
mtaylorgds, line 205: It seems that two spaces have been added between the double-quotes and the backslash, rather than one before the closing double-quote, and one after (applies to subsequent lines as well).

mtaylorgds (changes requested):
One change left!

test/flows/maternity_paternity_pay_leave_flow_test.rb
```diff
@@ -177,6 +202,7 @@ class MaternityPaternityPayLeaveFlowTest < ActiveSupport::TestCase
       should "have a next node of partner_started_working_before_continuity_start_date " \
                "for a 'yes' response " \
                "when there are two carers " \
+               "and the partner is from England" \
```
mtaylorgds, line 205: This is still not quite right–it should be:
```
"and the partner is from England " \
```

mtaylorgds (changes requested):

PR #6834, opened 2024-07-03
Remove Universal Analytics

AshGDS:
## What / Why
- Removes Universal Analytics, now that GA4 is Google's current analytics offering
- https://trello.com/c/fKGle0gD/187-remove-ua-code-from-applications


andysellick (approved):
👏

PR #6846, opened 2024-07-19
Use type=module for application.js

AshGDS:
## What / Why
- Use type=module for this application's JS file
- To coincide with https://github.com/alphagov/govuk_publishing_components/pull/4111
- Has [DO NOT MERGE] as we want the above PR live on `static` first before this is merged.
- We will combine `es6-components.js` again as a separate task unless you disagree with this
- To accurately test this you will need to use local static that is using a local version the `govuk_publishing_components` branch linked above. If you test without this, this app's `application.js` seems to crash.
- Trello card: https://trello.com/c/Hoa45SMD/141-turn-off-javascript-in-legacy-browsers

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -28,7 +28,7 @@
     <%= stylesheet_link_tag "application", :media => "all" %>
     <%= javascript_include_tag "test-dependencies" if Rails.env.test? %>
     <%= javascript_include_tag "es6-components", type: "module" %>
-    <%= javascript_include_tag "application", defer: true %>
+    <%= javascript_include_tag "application", defer: true, type: "module" %>
     <%= yield :head %>
     <% if content_item %>
       <%= render "govuk_publishing_components/components/meta_tags",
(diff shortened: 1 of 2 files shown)


MartinJJones (commented):
Looks good to me, the JS error appears to be caused by the closest polyfill imported from the publishing components gem and using type="module".

If this PR could also include the upgrade to the gem that removes support for older browsers and polyfills when released, this will resolve the issue.

Further info on the error:
```
Uncaught TypeError:
Cannot read properties of undefined (reading 'Element')
    at application.js:3:6
```
this-undefined

The reason for this appears to be that a top level `this` is `undefined` when used in a module, before this change it would be the `window` object - https://exploringjs.com/es6/ch_modules.html#sec_modules-vs-scripts

app/views/layouts/application.html.erb
```diff
@@ -28,7 +28,7 @@
     <%= stylesheet_link_tag "application", :media => "all" %>
     <%= javascript_include_tag "test-dependencies" if Rails.env.test? %>
```
MartinJJones, line 29: I wonder if we should also add type="module" here as well. It appears that test-dependencies.js is used to pull in [JS used by static to replicate the approach used on dev/production](https://github.com/alphagov/frontend/blob/main/app/assets/javascripts/test-dependencies.js#L1-L2), not 100% sure if this would actually make any difference to the tests, but would just align the approach used (to be used) more closely.

MartinJJones (approved):

Issue #6848, opened 2024-07-22
`gds_zendesk` gem is deprecated


AgaDufrat:
## What

Remove Zendesk-ticket creation and usage of `gds_zendesk` gem from the Smart Answers app.
Use Support API to raise the tickets instead. The [/support-tickets endpoint](https://github.com/alphagov/support-api/blob/c0b6ca3587f6673c9512573deeecd66a2aaa0d98/app/controllers/support_tickets_controller.rb) is exposed via [GDA API Adapters's `.raise_support_ticket`](https://github.com/alphagov/gds-api-adapters/blob/9aabf9d/lib/gds_api/support_api.rb#L25-L39).

It’ll require adding Bearer token for Support API. It’s created in Signon and added in helm-charts, [env var example](https://github.com/alphagov/govuk-helm-charts/blob/e16d6ae3dcfb86d8667cdb3a35d3df9809c1b252/charts/app-config/values-production.yaml#L986-L990). It can the be used in the app, [example](https://github.com/alphagov/support/blob/main/app/lib/services.rb).

For local development you may need to update govuk-docker to add support-api as a dependency.

Note, `GovukStatsd` is deprecated and should be removed as it doesn't do anything.

You will also need to remove `ZENDESK_CLIENT_USERNAME` and `ZENDESK_CLIENT_PASSWORD` from app's environment variables in helm-charts once the porting is complete.

## Why

We want to consolidate onto one centralised way of raising support tickets. This will minimise the maintenance overhead if we ever switch away from Zendesk (rumoured to be happening in 2024).

## Related

* https://github.com/alphagov/support/pull/1435
* https://github.com/alphagov/feedback/issues/1926

PR #6877, opened 2024-08-29
Remove links:send_report rake task

mtaylorgds:
The `gds_zendesk` gem is being retired and so we need to remove Smart Answer's dependency on it.

There is a single rake task that depends on the gem, and it does not
 appear to be used anymore. The last support ticket raised by the gem
 was on April 15th 2023, and we cannot find any code anywhere that would
 be triggering the rake task. Therefore, instead of updating the rake
 task to send the report to Zendesk without the `gds_zendesk` gem, we
 are deleting the rake task.

[Trello card](https://trello.com/c/CD1vEsdR)


Tetrino (approved):
LGTM

AgaDufrat:
Closes https://github.com/alphagov/smart-answers/issues/6848

PR #6909, opened 2024-10-15
Fix inline styling of bold text on a results page

MartinJJones:
## What

Fixed a styling issue with bold text on the check benefits financial support results page.

## Why

Noticed a console error when completing a review of app components, it seems inline styles are no longer applied.

### Visual Changes

| Before | After |
| --- | --- |
| bold-before | bold-after |

Page example: https://www.gov.uk/check-benefits-financial-support/y/england/no/no/yes/yes/yes_unable_to_work/no/no/dont_know/none_16000


davidtrussler (approved):
LGTM 👍

PR #6990, opened 2024-12-06
Force autoload of lib for Zeitwerk

Tetrino:
Zeitwerk requries lib to be autoloaded early so it can be properly initialised. Previously this had been set to eager-load erroneously.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -64,7 +64,7 @@ module SmartAnswers
     config.action_dispatch.ignore_accept_header = true
 
     # Force lib autoload, which was removed by Rails 3.0 and enforced by Zeitwerk
-    config.eager_load_paths << Rails.root.join("lib")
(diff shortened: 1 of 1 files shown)


ChrisBAshton (approved):
I'll approve if this is urgent and you've tested it and it works as you expect, but a few questions:

1) Why define this here rather than in the array [on line 33](https://github.com/alphagov/smart-answers/blob/d8e721d328007477b5d314323809c15c16f6d45a/config/application.rb#L33)?
2) I've looked in [Frontend](https://github.com/alphagov/frontend/blob/main/config/application.rb#L30-L36) and [Whitehall](https://github.com/alphagov/whitehall/blob/main/config/application.rb#L64-L73) for comparison and both of them use `eager_load_paths`. Not entirely clear why this change is needed here when it doesn't seem to be needed for other paths.

syed-ali-tw (approved):

PR #7002, opened 2024-12-13
Update wfp_going_abroad_eea_maybe_outcome.erb

Hannahmackay:
Updating this wording as a temporary fix so users have an onwards journey and out-of-date wording is removed.


ellohez (approved):
Changes make sense to me,

PR #7050, opened 2025-01-27
Replace title component with heading component

JamesCGDS:

## What
This PR replaces instances of the [title component](https://components.publishing.service.gov.uk/component-guide/title) with the [heading component](https://components.publishing.service.gov.uk/component-guide/heading).

In order to maintain spacing, I've added a `govuk-main-wrapper` class to the `main` element which adds `40px` worth of `padding` to the top and bottom of the page. This gives the heading some spacing above it, which is something the title component included but isn't something the heading component provides (further details in the [commit message](https://github.com/alphagov/smart-answers/pull/7050/commits/1a4695438ab6132ef60c4ccc2e2c0415e9f5859c)). It has introduced some visual differences (i.e. a bit less spacing above the heading and more spacing below the contextual footer - please see the screenshots below) but I think they may not be significant enough to be a blocker. Happy to discuss.

## Why
[Trello card](https://trello.com/c/QjeqZdIf/454-replace-title-with-heading-in-smart-answers)

## Visual changes

There are spacing differences at the top (slight) and bottom of the page (pronounced)

| ### Top of the page (before) | ### Top of the page (after) |
|------------------------------|-----------------------------|
| image | image |

| ### Bottom of the page (before) | ### Bottom of the page (after) |
|------------------------------|-----------------------------|
| image | image  |

### DIFF
--- a/app/views/smart_answers/custom_result.html.erb
+++ b/app/views/smart_answers/custom_result.html.erb
@@ -26,9 +26,11 @@
       "action": "complete",
       "tool_name": @presenter.title,
     }.to_json %>">
-    <%= render "govuk_publishing_components/components/title", {
-      title: outcome.title,
+    <%= render "govuk_publishing_components/components/heading", {
+      text: outcome.title,
       margin_bottom: 6,
+      heading_level: 1,
+      font_size: "xl",
     } %>
 
     <%= outcome.description %>
--- a/app/views/smart_answers/index.html.erb
+++ b/app/views/smart_answers/index.html.erb
@@ -1,7 +1,10 @@
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-full">
-    <%= render "govuk_publishing_components/components/title", {
-      title: "Smart Answers",
+    <%= render "govuk_publishing_components/components/heading", {
+      text: "Smart Answers",
+      heading_level: 1,
+      font_size: "xl",
+      margin_bottom: 8,
     } %>
 
     <div class="smart-answers-list">
--- a/app/views/smart_answers/landing.html.erb
+++ b/app/views/smart_answers/landing.html.erb
@@ -23,8 +23,11 @@
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <%= render "govuk_publishing_components/components/title", {
-      title: start_node.title,
+    <%= render "govuk_publishing_components/components/heading", {
+      text: start_node.title,
+      heading_level: 1,
+      font_size: "xl",
+      margin_bottom: 8,
     } %>
   </div>
 </div>
--- a/app/views/smart_answers/result.html.erb
+++ b/app/views/smart_answers/result.html.erb
@@ -30,10 +30,13 @@
     data-ga4-ecommerce-start-index="1"
     data-ga4-list-title="<%= @presenter.title %>">
     <%= render "smart_answers/shared/debug" %>
-    <%= render "govuk_publishing_components/components/title", {
-      title: title,
+    <%= render "govuk_publishing_components/components/heading", {
+      text: title,
       context: @presenter.title + ":",
       context_inside: true,
+      heading_level: 1,
+      font_size: "xl",
+      margin_bottom: 8,
     } %>
 
     <div
--- a/app/views/smart_answers/visualise.html.erb
+++ b/app/views/smart_answers/visualise.html.erb
@@ -14,9 +14,12 @@
   var adjacencyList = <%= JSON.pretty_generate(@graph_data).html_safe %>;
 <%- end %>
 
-<%= render "govuk_publishing_components/components/title", {
+<%= render "govuk_publishing_components/components/heading", {
   context: "Flow visualisation for",
-  title: @title,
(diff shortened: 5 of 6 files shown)


app/views/smart_answers/custom_result.html.erb
```diff
@@ -26,10 +26,14 @@
       "action": "complete",
       "tool_name": @presenter.title,
     }.to_json %>">
-    <%= render "govuk_publishing_components/components/title", {
-      title: outcome.title,
-      margin_bottom: 6,
-    } %>
+    <div class="govuk-!-margin-top-8">
```
matthillco, line 29: I've avoided doing this extra top margin on the ones I've done so far as there's a different way of handling this using classes on the `main` element now. If you have a look at this section in the design system, it explains which classes to use: https://design-system.service.gov.uk/styles/layout/#add-vertical-space

app/views/smart_answers/custom_result.html.erb
```diff
@@ -26,10 +26,14 @@
...
+      <%= render "govuk_publishing_components/components/heading", {
+        text: outcome.title,
+        margin_bottom: 6,
```
matthillco, line 32: Interesting that this is value 6 here for bottom margin. I've used 8 for the ones I've done and there's been no visual difference. Does it look wrong if you set it to 8?

app/views/smart_answers/index.html.erb
```diff
@@ -1,8 +1,13 @@
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-full">
-    <%= render "govuk_publishing_components/components/title", {
-      title: "Smart Answers",
-    } %>
+    <div class="govuk-!-margin-top-8">
```
matthillco, line 3: Could the `margin-top` go on the column itself, rather than adding an extra element?

app/views/smart_answers/landing.html.erb
```diff
@@ -23,9 +23,14 @@
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <%= render "govuk_publishing_components/components/title", {
-      title: start_node.title,
-    } %>
+    <div class="govuk-!-margin-top-8">
```
matthillco, line 26: Could the `margin-top` go on the column itself, rather than adding an extra element?

app/views/smart_answers/result.html.erb
```diff
@@ -30,11 +30,16 @@
     data-ga4-ecommerce-start-index="1"
     data-ga4-list-title="<%= @presenter.title %>">
     <%= render "smart_answers/shared/debug" %>
-    <%= render "govuk_publishing_components/components/title", {
-      title: title,
-      context: @presenter.title + ":",
-      context_inside: true,
-    } %>
+    <div class="govuk-!-margin-top-8">
```
matthillco, line 33: Could the `margin-top` go on the column itself, rather than adding an extra element?

app/views/smart_answers/visualise.html.erb
```diff
@@ -14,10 +14,15 @@
   var adjacencyList = <%= JSON.pretty_generate(@graph_data).html_safe %>;
 <%- end %>
 
-<%= render "govuk_publishing_components/components/title", {
-  context: "Flow visualisation for",
-  title: @title,
-} %>
+<div class="govuk-!-margin-top-8">
```
matthillco, line 17: I can see why this is necessary, but I'm not a fan of adding an extra wrapper to adjust the vertical spacing. Is there another way this could be done more cleanly? Looking at the surrounding markup it seems not. What does this page look like if there's no top margin added? Will it be inconsistent with the others?

matthillco (changes requested):
I've added a few comments around the addition of top margins. Hopefully it makes sense, but happy to chat if you need more info or want to discuss anything.

andysellick:
Haven't looked at this in detail but just dropping a general comment before I finish for today, generally the govuk `!` classes are intended for use where an override is necessary, such as where existing styles exist but can't be changed. For cases like this the guidance from the DS team is to write custom CSS using the `govuk-spacing` mixin.

matthillco:
@JamesCGDS My apologies, I should have looked at this more deeply before adding those comments about top margins. I think in your case, if these are all H1 page titles, you can solve the top margins by updating the `main` element in `views/layouts/application.html.erb` which currently has this for `main`:

```
<% content_for :body do %>
  <div class="smart_answer">
    <%= yield :breadcrumbs %>
    <main id="content" role="main">
      <%= yield %>
    </main>
  </div>
<% end %>
```

If you add the class `govuk-main-wrapper` to the `main` element, this should sort out all the vertical spacing above the heading. See the associated [design system doc](https://design-system.service.gov.uk/styles/layout/#add-vertical-space) for clarification.

JamesCGDS:
Thanks for your comments both - I've made the change Matt suggested as that seemed like a nice way to do things. I've updated the PR description to reflect this.

matthillco (approved):

matthillco:
Looks good to me. The small visual differences are acceptable, we've approved those on other similar PRs. Nice job, thanks.

PR #7125, opened 2025-03-20
Improve integration test reliability

mtaylorgds:
Following a recent release of Chrome/Chromium, integration tests have become less reliable. It seems like when a page change occurs (for example, when visiting a URL, or after submitting a form) the new page is less likely to have loaded by the time the next line in the test occurs than was previously the case.

This commit updates the `ChangingAnswerTest` integration tests to "find" the title on the page that is being expected to be loaded—find will wait until the element is present, rather than instantly fail (up until the configured timeout period).


baisa (approved):
It makes sense to me

PR #7137, opened 2025-03-25
Update docs for the `RATES_QUERY_DATE` env var

pezholio:
The fact that `app.json` changes don’t take effect after a review app has been deployed tripped me up a bit here, so updating the docs to take this into account.

### DIFF
--- a/docs/tasks/fact-check.md
+++ b/docs/tasks/fact-check.md
@@ -24,4 +24,23 @@ environment variable in the [`app.json`](../../app.json), for example:
 `"RATES_QUERY_DATE": "2020-04-01"` - don't forget to remove this before
 merging.
 
+**Note** This will only work if your review app has not yet been created. Pushing
+a change to an app.json after a review app has been created will not update the
+environment variables. 
+
+If you want to do change the variable after your review app has been deployed, 
+you'll have to manually set the  environment variable like so (where `$DATE` is 
+the date you want to set and `$APP` is the name of the review app (e.g. 
+`smart-answers-pr-7132`)):
+
+```bash
+heroku config:set RATES_QUERY_DATE=$DATE -a $APP
+```
+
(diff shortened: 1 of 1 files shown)


docs/tasks/fact-check.md
```diff
@@ -24,4 +24,20 @@ environment variable in the [`app.json`](../../app.json), for example:
...
+you'll have to manually set the  environment variable like so (where `$DATE` is 
+the date you want to set and `$APP` is the name of the review app (e.g. 
+`smart-answers-pr-7132`)):
+
+```bash
+heroku config:set RATES_QUERY_DATE=$DATE -a $APP
+```
+
+This assumes you have [access to the shared Heroku account][].
+
```
baisa, line 41: Maybe we can add that as a result, we should be redirected from heroku page back into console with success message like
`Logging in... done
RATES_QUERY_DATE: $DATE`
just to clarify how the change is confirmed

pezholio: That will only happen if you haven't already logged in via `heroku login`. I'd be a bit wary of adding any details of what to expect in case that behaviour changes in the future. I've updated it with a link to the Heroku CLI docs though, just so it's a bit clearer.

pezholio: Actually, I've made a further tweak that gives a clue without going into too many specifics 👍

docs/tasks/fact-check.md
```diff
@@ -24,4 +24,23 @@ environment variable in the [`app.json`](../../app.json), for example:
 `"RATES_QUERY_DATE": "2020-04-01"` - don't forget to remove this before
```
Harriethw, line 24: while you're here, could you perhaps add how to get the login creds for the review apps frontend as well?

baisa (approved):

baisa (commented):
Great to have this documented!

PR #7303, opened 2025-08-21
Upgrade Smart Answers to Ruby 3.4.5

dj-maisy:
## What?
This updates Smart Answers from Ruby 3.2.4 up to Ruby 3.4.5.

Also:

* Updates bundler to 2.7
* Fixes a broken test.


Gemfile
```diff
@@ -1,11 +1,12 @@
 source "https://rubygems.org"
 
-ruby "~> 3.2.0"
+ruby "~> 3.4.0"
 
 gem "rails", "8.0.2.1"
 
 gem "ast"
 gem "bootsnap", require: false
+gem "csv"
```
AgaDufrat, line 9: Does this app use this gem?

AgaDufrat (commented):
We don't own this app. We should live it to the team to upgrade.

Gemfile
dj-maisy: So it was in the lock file before, but actually doing a deeper dig, doesn't look like it is actually used anywhere!

I'll pull it out and see if it still works.

test/unit/money_test.rb
```diff
@@ -43,7 +43,7 @@ class MoneyTest < ActiveSupport::TestCase
...
+        SmartAnswer::Money.new 1_111_111_111_111_111_122_222_222_222_222_222_222_222_222_222_222_222_222_222_222_222_222_222_222_222_222_222_222_222_222_222_222_225_555_555_555_555_555_555_555_555_555_555_555_555_555_555_555_555_555_555_555_222_222_222_222_222_222_222_211_111_111_111_111_111_133_333_333_333_333_333_333_555_555_555_555_555_555_555_555_555_555_555_555_555_555_555_555_555_555_555_555_555_555_555_555_555_555_555_555_555_555_555_444_444_444_444_444_444_444_444_444_448_888_888_888_888_888_888_888_888_888_888_888_888_888_888_888_888_888_888_888_887_777_777_777_777_778_888_888_888_888_889_999_999_999.0 # rubocop:disable Lint/FloatOutOfRange
```
AgaDufrat, line 46: I'm not a fan of disabling linting rules in line instead of addressing them. 
Alternatively we could use BigDecimal instead of Float here or update the value in the test (but we don't have enough product knowledge).

PR #7509, opened 2026-02-19
Update check UK visa smart answer

mtaylorgds:
Add a new question to ask about dual British or Irish citizenship, along with a new outcome for those that do (or don't know if they do) hold such dual citizenship.

[MAIN-7272](https://gov-uk.atlassian.net/browse/MAIN-7272)

[MAIN-7272]: https://gov-uk.atlassian.net/browse/MAIN-7272?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/flows/check_uk_visa_flow.rb
+++ b/app/flows/check_uk_visa_flow.rb
@@ -16,7 +16,24 @@ class CheckUkVisaFlow < SmartAnswer::Flow
       end
 
       next_node do
-        if calculator.passport_country_is_israel?
+        if calculator.passport_country_is_ireland?
+          outcome :outcome_no_visa_needed_ireland
+        else
+          question :dual_british_or_irish_citizenship?
+        end
+      end
+    end
+
+    # Q2
+    radio :dual_british_or_irish_citizenship? do
+      option :yes
+      option :no
+      option :dont_know
+
+      next_node do |response|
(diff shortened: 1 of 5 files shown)


app/flows/check_uk_visa_flow.rb
AlexRyanShep, line 47: I'm not super familiar with this flow, but reading through it, it seems like this conditional might not be needed anymore? The catch for "If passport is irish, go straight to the no visa outcome" is now happening in Q1 right?

I don't think it's doing any harm by being here, just don't currently see a way for a user to actually reach this code.

AlexRyanShep (approved):
Looks good to me! Approving with one minor comment

app/flows/check_uk_visa_flow.rb
PeterHattyar: I _think_ you might be right here yeah! I know the new question block here relates to the dual citizenship parts, and yeah not sure if it needs to be included, if they're being sent to no visa needed outcome by declaring an Irish passport at Q1. I'll deploy this is as as it's been long outstanding, but will whip up a quick correction PR, if I tested it locally that it doesn't break anything. (It really shouldn't, logically.) Great spot!

PR #7588, opened 2026-04-13
Upgrade Rails from 8.0.3 to 8.1.3

young-rob:
### Changes

- Upgrade from Rails 8.0.3 to Rails 8.1.3
- Config and bin script changes from running `rails app:update`


config/environments/production.rb
```diff
@@ -4,12 +4,9 @@
   # Settings specified here will take precedence over those in config/application.rb.
 
   # Code is not reloaded between requests.
-  config.cache_classes = true
```
AlexRyanShep, line 7: Was this removed on purpose? It looks like this is something non-standard that we added in that got wiped by the update?

young-rob: I did wonder why it replaced this but I checked and `config.cache_classes = true` [is equivalent to ](https://guides.rubyonrails.org/configuring.html#config-cache-classes)`config.enable_reloading = false`

baisa (approved):

PR #7660, opened 2026-06-03
Student finance calculator adding new routes

stacy-hayes:
[MAIN-7387](https://gov-uk.atlassian.net/browse/MAIN-7387)
(populate the link above to make sure your PR is linked to the Jira ticket)

## Changes

Adds new flows related to the lifelong learning entitlement coming in in 2027/2028, which can be broadly split into:

- Under 60 learners who are learning in person, or remotely as a result of a disability
- Under 60 learners who are learning remotely
- Over 60 learners who are learning in person, or remotely as a result of a disability
- Over 60 learners who are learning remotely

These routes are displayed on the mural board linked from the ticket. 

The testing PRs are separate and can be merged into this one once they are approved - this was an intentional decision to allow the flow to enter fact check as soon as possible while still writing the flow tests.

[MAIN-7387]: https://gov-uk.atlassian.net/browse/MAIN-7387?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/flows/student_finance_calculator_flow/questions/whats_your_household_income.erb
```diff
@@ -3,7 +3,7 @@
 <% end %>
 
 <% text_for :hint do %>
-  This is your parents’ or partner’s taxable income. It affects how much help you'll get with living costs.
+  This is your <%= calculator.age != "60-or-more" ? "parents’ or ": ""%>partner’s taxable income plus your own. It affects how much help you'll get with living costs.
```
AlexRyanShep: `<%= calculator.age != "60-or-more" ? "parents’ or ": ""%>`

No problems with this functionally, but I think in terms of readability this would be pretty confusing if a non-dev came across it because of all the mixed `"` and `'` symbols (it took me a few read-overs to properly process it).

I'm usually a big fan of the ternary operator, but it might be worth simplifying this into some kind of if statement for readability by non-devs?

app/flows/student_finance_calculator_flow.rb
```diff
@@ -113,7 +123,12 @@ def define
       end
 
       next_node do
-        question :how_many_credits_does_a_full_time_student_study?
+          case calculator.course_type
+          when "full-time"
+            question :how_much_are_your_tuition_fees_per_year?
```
AlexRyanShep: Does this need a `full-time` option? As far as I can see, you can only access this question through `:whats_your_household_income?` when `course_type` is `part-time`?

AlexRyanShep: Also, if I'm reading it right and you did somehow manage to make it to this question with a `course_type` of `full-time`, I think it would dump you in a loop (old pre-lle flow):

- You can only access this question from `:whats_your_household_income?`
- This question full-time leads to`:how_much_are_your_tuition_fees_per_year?` 
- `:how_much_are_your_tuition_fees_per_year?` with tuition-and-maintenance leads to `:where_will_you_live_while_studying?`
- :`where_will_you_live_while_studying?` in old flow leads back to `:whats_your_household_income?`

app/flows/student_finance_calculator_flow.rb
```diff
@@ -209,6 +233,173 @@ def define
       end
     end
 
+    # Q2
```
AlexRyanShep: I think this comment might need updating or differentiating somehow from the other `Q2`? Right now we've got the old flow where all the questions have a `Q#` comment and none of the new ones do? Just a minor thing, but might matter for non-dev readability

app/flows/student_finance_calculator_flow.rb
```diff
@@ -76,7 +76,7 @@ def define
       option :'at-home'
       option :'away-outside-london'
       option :'away-in-london'
-      option :'overseas'
+      option :overseas
```
AlexRyanShep: Cursing Rubocop here 😢  Is it worth thinking about disabling this specific rule for this file? Pretty sure you could avoid having an inline comment by putting it into the Rubocop.yml:

```
Style/SymbolLiteral:
    Exclude:
        - app/flows/student_finance_calculator_flow.rb
```

app/flows/student_finance_calculator_flow/outcomes/_extra_student_funding_lle.erb
```diff
@@ -0,0 +1,54 @@
...
+<% if calculator.eligible_for_nhs_bursary %>
+  You can apply for [NHS funding towards your fees and living costs](/nhs-bursaries).
+  If you’re on a 4 year accelerated graduate course, you’ll have to apply for an NHS bursary from your second year.
+<% elsif calculator.course_studied == "teacher-training" %>
+  You can apply for [funding for teacher training](/teacher-training-funding).
+<% elsif calculator.course_studied == "social-work" %>
+  You can apply for a [Social Work Bursary](/social-work-bursaries).
+<% end %>
+
+<% extra_student_funding = [] %>
```
AlexRyanShep, line 14: Having a thought. Could we maybe move the logic around building this set of available extra student funding into the calculator,  and pass it in as one List local variable rather than building this list here?

That could get us away from having so much logic inside the view, and from having to pass in the calculator itself as a local variable which I remember you weren't thrilled with on the team day.

AlexRyanShep: The compromise side of that might end up with the messages themselves being in the calculator I guess. Though we could maybe simplify to something like

```
EXTRA_STUDENT_FUNDING_MESSAGES = {
    condition: "message",
    condition2: "message 2",
     ...
}

<%  extra_student_funding.each do |funding|  %>
  - <%=  EXTRA_STUDENT_FUNDING_MESSAGES[funding]  %>
  <%  end  %>
```

And then in the calculator we build extra_student_funding as a list of symbols? Since we're doing it in the calculator it'll have access to all its own state so we could copy the logic used here for the most part.
 
It's still some logic in the view here but less, and I think arguably easier for a content designer to understand and update with well named symbols. Hard to tell without getting it in front of one of them, but I suspect the if statements below (especially the elsif) might be pushing a bit far in what they are used to seeing?

AlexRyanShep: It could also probably include the nhs/teacher/social work messages above!

app/flows/student_finance_calculator_flow/outcomes/_maintenance_summary_lle.erb
```diff
@@ -0,0 +1,32 @@
+<% if maintenance_loan_amount > 0 %>
```
AlexRyanShep, line 1: I think having this logic in the partial here and leaving the views clean is the right way to go here 👍  It does mean the views always have to render this whether or not maintenance loan is 0, but the performance overhead of that for something simple loading once like this is minuscule

app/flows/student_finance_calculator_flow/outcomes/_maintenance_summary_lle.erb
```diff
@@ -0,0 +1,32 @@
...
+  ## Maintenance Loan
+
+  <% if uk_ft_circumstances.include?('care-leaver') || uk_all_circumstances.include?('care-leaver') %>
+    You can choose to borrow the maximum amount: <%= format_money(max_loan_amount, pounds_only: true) %>.
```
AlexRyanShep: Am I understanding right that the `format_money` method is purely around string formatting for the views and then the `SmartAnswer::Money` class (using BigDecimal 🥳 ) is for keeping calculations accurate?

lib/smart_answer/calculators/student_finance_calculator.rb
```diff
@@ -12,19 +12,30 @@ class StudentFinanceCalculator
...
+                    :age,
+                    :studied_before,
+                    :attend_in_person,
+                    :disability_status,
+                    :eligible_for_nhs_bursary
 
       LOAN_MAXIMUMS = {
         "2025-2026" => {
-          "at-home" => 8877,
+          "at-home" => 8_877,
```
AlexRyanShep, line 24: It's going to be a pain so I'm kinda hesitant to mention it sorry, but should we be defining these values (just pinning this comment to the first one for readability but we've got a lot in this file) using `SmartAnswer::Money` (like in the married couples allowance calculator)?

app/flows/student_finance_calculator_flow.rb
```diff
@@ -209,6 +264,245 @@ def define
...
+            when "no"
+              outcome :outcome_under_60_distance_learner
+            end
+          elsif response == "no"
+            case calculator.attend_in_person
+            when "yes"
+              outcome :outcome_under_60_students
+            when "no"
+              outcome :outcome_under_60_distance_learner
+            end
```
AlexRyanShep: I think this if statement can be stripped out? Both sides of it go to the same outcomes

app/flows/student_finance_calculator_flow/outcomes/outcome_over_60_distance_learner.erb
```diff
@@ -0,0 +1,13 @@
+<% govspeak_for :body do %>
+  <%= render partial: 'disclaimer_lle' %>
+
+  <% if calculator.uk_ft_circumstances.include?('has-disability') || calculator.uk_all_circumstances.include?('has-disability') %>
+    Because you are a distance learner, you are not eligible for a Special Support loan. Special Support loans are available if you are not able to attend your course in person due to having a disability.
```
AlexRyanShep: (I know this is in a different file in the test version.) Just checking here as a bit confused, this sentence is telling them they're not eligible for the SSL, but then immediately telling them they are? Was the first half of this a placeholder that didn't get deleted?

AlexRyanShep: Just had half a memory pop up sorry. Is this where we're giving them a nudge that we think they might have not selected an option they should have around this?

app/flows/student_finance_calculator_flow/outcomes/_disclaimer_lle.erb
```diff
@@ -0,0 +1,5 @@
+$!
+Your result is an estimate. You'll be told your exact amount, including any extra student funding, after your application is approved.
+
+You may get a smaller amount of funding to start with while Student Finance England (SFE) reviews your application. They'll send you a top-up payment later if you're eligible for it – [check your online account for updates](/student-finance-register-login).
```
AlexRyanShep: Very much one to pass onto the content designer and not for you, but I've noticed that some of the language in this flow is inconsistent (i.e. we switch between You are and you're quite a few times over the flow).

app/flows/student_finance_calculator_flow/questions/how_much_are_your_tuition_fees_course_or_module.erb
```diff
@@ -0,0 +1,11 @@
+<% text_for :title do %>
+  How much are your tuition fees for this year of your course or module?
+<% end %>
+
+<% text_for :hint do %>
+    The maximum Tuition Fee Loan is <%= format_money(calculator.max_tuition_fee_amount) %> for a <%= calculator.part_time_credits.to_i %> credit course.
+<% end %>
+
+<% text_for :error_message do %>
+    Please enter a number up to the maximum amount of <%= format_money(calculator.max_tuition_fee_amount) %>, don’t enter ‘£’
```
AlexRyanShep:
```suggestion
    Please enter a number up to the maximum amount of <%= format_money(calculator.max_tuition_fee_amount) %>, do not enter ‘£’
```

https://guidance.publishing.service.gov.uk/writing-to-gov-uk-standards/style-guides/a-to-z-style-guide/#contractions recommends avoiding negative contractions like don't

lib/smart_answer/calculators/student_finance_calculator.rb
```diff
@@ -264,12 +275,26 @@ def max_loan_amount
...
     private
 
       def min_loan_amount
         LOAN_MINIMUMS[@course_start][@residence]
       end
 
+      def ssl_reduction_based_on_income
+        return 0 if @household_income <= 25_000
+
+        ((@household_income - 25_000) / INCOME_PENALTY_RATIO_OVER_60).floor
```
AlexRyanShep, line 313: Just double-checking, have we had a steer at all from SLC on rounding up or down? (`.floor` here is rounding us down to a whole pound right?)

app/flows/student_finance_calculator_flow/outcomes/_extra_student_funding_lle.erb
```diff
@@ -0,0 +1,54 @@
...
+
+<% if calculator.uk_ft_circumstances.include?('low-income') || calculator.uk_all_circumstances.include?('low-income') %>
+  You could get a [bursary or scholarship](/extra-money-pay-university) from your university or college.
+<% else %>
+  You could get a [bursary, scholarship or financial hardship funding](/extra-money-pay-university) from your university or college.
+<% end %>
+
+You might be able to get [help with the costs of travel for study or work placements](/travel-grants-students-england) as part of your course.
+
+Use the [Turn2us grant search](http://grants-search.turn2us.org.uk/) to check whether you qualify for funding from a charitable trust.
```
AlexRyanShep: This is missing a closing newline

AlexRyanShep (commented):
Nice work Stacy! This was a mammoth of a ticket! 🎉 

There are a fair few comment sorry, but a lot of them are fairly minor style/readability and some are more for the content designer than you.

The key ones to me are the first one about the ternary operator readability, and the one about using `SmartAnswer::Money` in the calculator definitions.

As discussed on Slack, this review is purely looking at this from a code quality/logic point of view. Aside from one or two bits where something jumped out at me I've not looked at the actual flow of the journeys. We're going to get together and look at those aspects together.

app/flows/student_finance_calculator_flow.rb
stacy-hayes: I considered disabling the rule, but then we end up with nothing holding a consistency and we use `:yes` and `:no` in here too so I leant towards keeping the rule and accepting the slight difference in term. alternatively we could refactor to `living-overseas` to be able to use quotes?

app/flows/student_finance_calculator_flow/outcomes/_extra_student_funding_lle.erb
stacy-hayes: I think the difficulty there is that those are messages/links a content designer is more likely to need to change, and I think thats easier to follow doing in the view since thats the standard work process (and also how the old pathways do it)?

app/flows/student_finance_calculator_flow/outcomes/_maintenance_summary_lle.erb
stacy-hayes: exactly that!

app/flows/student_finance_calculator_flow/outcomes/outcome_over_60_distance_learner.erb
stacy-hayes: yes exactly, this is when they've said theyre disabled, but havent said the reason they cant attend in person is due to disability, to give the hint that if those things are related they might get extra support

lib/smart_answer/calculators/student_finance_calculator.rb
stacy-hayes: yeah its rounding down to the whole pound, maintaining behaviour with all the current calculations, although I will triple check theyre happy with it

app/flows/student_finance_calculator_flow/outcomes/_extra_student_funding_lle.erb
AlexRyanShep: Yeah I did have that thought after I wrote it. So in what I suggested the controller might have:

```
if eligible_for_childcare_grant_one_child?
    extra_support << :childcare_grant_one_child
    childcare_grant_amount_one_child = format_money(childcare_grant_one_child)
    childcare_grant_amount_more_than_one_child = format_money(childcare_grant_more_than_one_child)
elsif eligible_for_childcare_grant_more_than_one_child?   
    extra_support << childcare_grant_more_than_one_child
    childcare_grant_amount_more_than_one_child = format_money(childcare_grant_more_than_one_child)
    ....
```

and then the view could have

```
EXTRA_STUDENT_FUNDING_MESSAGES = {
    childcare_grant_one_child: "up to 85% of your childcare costs (maximum #{childcare_grant_amount_one_child} a week for a single child or #{childcare_grant_amount_more_than_one_child} a week if you have more than one) in Childcare Grant",
    childcare_grant_more_than_one_child: ""up to 85% of your childcare costs to a maximum #{childcare_grant_amount_more_than_one_child} a week in Childcare Grant if you have 2 or more children"",
     ...
}

<%  extra_student_funding.each do |funding|  %>
  - <%=  EXTRA_STUDENT_FUNDING_MESSAGES[funding]  %>
  <%  end  %>
```

I'm not super attached to the idea if you're not keen, my mind just saw what felt like too much logic in the view and was trying to think of ways to reduce it 🙂

lib/smart_answer/calculators/student_finance_calculator.rb
stacy-hayes: we've reviewed this and we are useing Money when we do calculations in the file

app/flows/student_finance_calculator_flow/questions/whats_your_household_income.erb
stacy-hayes: Yeah I think youre right, updated now :)

app/flows/student_finance_calculator_flow.rb
stacy-hayes: oh you're right, i think this is left over from before i split out some other questions

lib/smart_answer/calculators/student_finance_calculator.rb
```diff
@@ -280,6 +280,10 @@ def adjusted_ssl_loan_amount
         reduced_amount = ssl_loan_amount - ssl_reduction_based_on_income
         SmartAnswer::Money.new([reduced_amount, SPECIAL_SUPPORT_ELEMENT_OF_ML_OVER_60_MINIMUM].max * loan_proportion)
       end
+      
+      def lle_scheme?
+        %w[2027-2028].include?(@course_start)
```
stacy-hayes, line 301: oooh this is really nice, especially with it letting us add extra years as time progresses!

lib/smart_answer/calculators/student_finance_calculator.rb
```diff
@@ -318,13 +318,14 @@ def loan_proportion
...
 
-        # 120 is taken as the standard for full-time/full amount, so >= here accounts for students studying more credits than that
+        # 120 is taken as the standard for full-time/full amount, so >= here accounts for students studying 120+ credits
         return 1 if course_intensity >= 100
         return 0.75 if course_intensity >= 75
         return 0.666 if course_intensity >= 66.6
         return 0.5 if course_intensity >= 50
         return 0.333 if course_intensity >= 33.3
         return 0.25 if course_intensity >= 25
+        return (course_intensity / 100).round(3) if lle_scheme?
```
stacy-hayes: it feels weird that we chunk this for everything above 25% and then are specific under, I'm assuming this is an SLC choice?

app/flows/student_finance_calculator_flow/questions/how_many_credits_will_you_study_course_module.erb
```diff
@@ -7,5 +7,5 @@
 <% end %>
 
 <% text_for :error_message do %>
-  Enter between 30 and 180 credits
+  Enter between 30 and 180 credits <%= calculator.credits_studied %>
```
stacy-hayes: I think this might be a mis-paste? It renders things like `Enter between 30 and 180 credits 25` now if you input 25 credits - is it meant to be a fullstop?

lib/smart_answer/calculators/student_finance_calculator.rb
AlexRyanShep: That's an SLC choice yep. As I said, this ask felt kinda wooly to me which is why I left it as its own commit, easy to identify/rollback if we need to.

app/flows/student_finance_calculator_flow/questions/how_many_credits_will_you_study_course_module.erb
AlexRyanShep: Agh sorry! This is a debug statement (like I mentioned on Slack, byebug is somehow broken for this test file) that slipped past me. I'll remove and re-push

lib/smart_answer/calculators/student_finance_calculator.rb
AlexRyanShep: I also asked Heather (though we might need to prod her replacement/fill-in as she's now on leave) in general to check if that approach of using stepped chunks for the intensity is still correct for LLE

AlexRyanShep: Closing this as irrelevant now: Anna has asked us to change this so that we use the old behaviour for old flows, and just calculate the exact percentage in all cases in LLE

app/flows/student_finance_calculator_flow/outcomes/outcome_tuition_fee_only.erb
```diff
@@ -11,7 +11,7 @@
 
   ##Extra help
 
-  You could get a [bursary or scholarship](/extra-money-pay-university) from your university or college. Usually, this doesn't have to be paid back.
+  You could get a [bursary or scholarship](/extra-money-pay-university) from your university, college or training provider. Usually, this doesn't have to be paid back.
```
AlexRyanShep: Does this need changing? It's the old pre-LLE flow

app/flows/student_finance_calculator_flow/outcomes/_uk_extra_help.erb
```diff
@@ -1,6 +1,6 @@
 ##Extra help
 
-You could get a [bursary or scholarship](/extra-money-pay-university) from your university or college.
+You could get a [bursary or scholarship](/extra-money-pay-university) from your university, college or training provider.
```
AlexRyanShep: I think this is also only used in the pre-LLE flows?

AlexRyanShep (commented):
Looks good, two minor comments 🙂

app/flows/student_finance_calculator_flow/outcomes/_uk_extra_help.erb
stacy-hayes: good catch, fixed

app/flows/student_finance_calculator_flow/outcomes/outcome_tuition_fee_only.erb
stacy-hayes: good catch, fixed

app/flows/student_finance_calculator_flow/questions/whats_your_household_income.erb
```diff
@@ -3,7 +3,7 @@
 <% end %>
 
 <% text_for :hint do %>
-  [Check what counts as household income if you're not sure](/apply-for-student-finance/household-income)
+  <a href="/apply-for-student-finance/household-income">Check what counts as household income if you're not sure.</a>
```
young-rob: Do we need to apply govuk styling to this link? e.g. with `class="govuk-link"`

stacy-hayes: Oh good catch, adding now!

app/flows/student_finance_calculator_flow/outcomes/outcome_under_60_distance_learner.erb
```diff
@@ -5,7 +5,7 @@
     Because you are not attending the course in person, you are not eligible for a Maintenance Loan and some types of extra funding.
   <% end %>
 
-  <%= render partial: 'tuition_summary_lle', locals: { tuition_fee_amount: calculator.tuition_fee_amount, tuition_fee_url: "/student-finance/new-fulltime-students", maintenance_loan_url: "/student-finance/new-fulltime-students", uk_ft_circumstances: calculator.uk_ft_circumstances, uk_all_circumstances: calculator.uk_all_circumstances, studied_before: calculator.studied_before == "yes", course_studied: calculator.course_studied } %>
+  <%= render partial: 'tuition_summary_lle', locals: { tuition_fee_amount: calculator.tuition_fee_amount, max_tuition_fee_amount_lle: calculator.max_tuition_fee_amount_lle, tuition_fee_url: "/student-finance/new-fulltime-students", uk_ft_circumstances: calculator.uk_ft_circumstances, uk_all_circumstances: calculator.uk_all_circumstances, studied_before: calculator.studied_before == "yes", course_studied: calculator.course_studied } %>
```
young-rob: If `tuition_fee_amount` is no longer used in the `tuition_summary_lle.erb` partial (being replaced by `max_tuition_fee_amount_lle`), I think we can remove it from the locals in the render calls in the files in this commit?

young-rob:
Looking good to me up to this point, just 2 minor comments! 👍

young-rob (dismissed):
Looks good to me, great job on a huge piece of work! 🚀

app/flows/student_finance_calculator_flow.rb
```diff
@@ -165,11 +165,21 @@ def define
...
+            else
+              question :whats_your_household_income?
+            end
+          when "60-or-more"
+            if response.include?("care-leaver") && !response.include?("children-under-17") && !response.include?("dependant-adult")
+              calculator.household_income = 0
+              outcome :outcome_over_60_students
+            else
+              question :whats_your_household_income?
+            end
```
young-rob: Looking good! My only suggestion would be to simplify this logic slightly if that works?

```suggestion
          if response.include?("care-leaver") && !response.include?("children-under-17") && !response.include?("dependant-adult")
            calculator.household_income = 0
            if calculator.age == "under-60"
              question :are_you_studying_one_of_these_courses?
            else
              outcome :outcome_over_60_students
            end
          else
            question :whats_your_household_income?
          end
```

young-rob (approved):
LGTM ✅

PR #7662, opened 2026-06-04
Update Dart Sass to v1.100.0 to prepare for Design System v6 upgrade

luke-booker:
## What

- Ran bundle update sass-embedded, which successfully resolved and updated the gem to `sass-embedded (1.100.0)` in `Gemfile.lock`.

## Why

GOV.UK Frontend v6.0 will be dropping support for Ruby Sass, LibSass, and older versions of Dart Sass. To prepare the application for this upgrade, we need to ensure our Sass compiler is running Dart Sass v1.79.0 or later.

Jira card: https://gov-uk.atlassian.net/browse/NAV-18914

## Visual changes

None. This is a dependency requirement upgrade only.

## Testing carried out locally

✅  Verified stylesheets compile successfully. 
✅  All tests passing successfully.
✅  Application is building without issues locally.


MartinJJones (approved):
Nice work, the changes look good to me 👍

non-blocking: it could be worth adding a commit message linking to the changelog entry on the design system https://github.com/alphagov/govuk-frontend/blob/main/CHANGELOG.md#use-dart-sass-v1790-or-later-to-compile-your-sass-stylesheets, but not essential

PR #7671, opened 2026-06-09
Remove data-ga4-track-links-only references

AshGDS:
## What / Why
- This attribute was removed from govuk_publishing_components as it wasn't actually needed, so we can remove it from this app.
- JIRA Card: https://gov-uk.atlassian.net/browse/PNP-10043


sairamya93 (approved):
LGTM!

PR #7683, opened 2026-06-16
provide a fallback for no tax year

stacy-hayes:
Currently if you go down a path without a tax year it crashes, this provides a fallback for those calculations


PeterHattyar (approved):
Nice fix, well done!

AlexRyanShep (approved):
LGTM

young-rob (approved):
LGTM 👍

PR #7700, opened 2026-07-01
Add GA4 tracking to 'next steps for your business'

AshGDS:
## What / Why
- Adds GA4 tracking to the 'next steps for your business' route: (example [here](https://smart-answers-pr-7700.herokuapp.com/next-steps-for-your-business/results?activities%5B%5D=export_goods_or_services&annual_turnover_over_90k=not_sure&business_premises%5B%5D=owned&employ_someone=yes&financial_support=yes))
- I also removed the `noreferrer` property from internal links that were opening in a new tab. This is so GA4 will track the referrer when they click a result that's still on gov.uk. I had to duplicate the link to do this since the linter expects `noopener` as an explicit string without any Ruby magic in it, so dynamically setting the `rel` value didn't work.

### DIFF
--- a/app/flows/next_steps_for_your_business_flow/outcomes/results.erb
+++ b/app/flows/next_steps_for_your_business_flow/outcomes/results.erb
@@ -5,6 +5,10 @@
   ======================================================================
 %>
 
+<%
+  @website_root ||= Plek.new.website_root
+%>
+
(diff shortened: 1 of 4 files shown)


KludgeKML (approved):
LGTM!

PR #7735, opened 2026-07-30
CR-38 Remove Content Block support from Smart Answers

edavey:
## Remove Content Blocks from Smart Answers

see Jira [CR-37](https://gov-uk.atlassian.net/browse/CR-37)

We are removing Content Block support from Smart Answers for now as complying
with the stricter format of `base_path` has brought it to our attention
that we don't currently have a real use-case for publishing blocks to the
Content Store -- given that Smart Answers do not currently make any use of
Content Blocks.

This work we are reverting was a valuable proof of concept but for now it's
an architectural and maintenance overhead we can't support.

BACKGROUND AND FURTHER DETAIL

[RFC-192: Restrict base_paths to a smaller character set][rfc_192] is restricting
GOV.UK base paths to `a-z 0-9 . / -` (no underscores, no uppercase).
Content blocks are affected because their base_path includes the `document_type`
(e.g. `content_block_time_period`) which contains underscores, as in this
`base_path`:

```
/content-blocks/content_block_time_period/tax-year
```

Support for Content Blocks was originally added in https://github.com/alphagov/smart-answers/pull/7378

[rfc_192]:
https://github.com/alphagov/govuk-rfcs/pull/192/

[CR-37]: https://govukverify.atlassian.net/browse/CR-37?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


baisa (approved):
LGTM

PR #7752, opened 2026-08-10
Rename method to prevent false missing assertions warnings

stacy-hayes:
The method starting with the word 'test' was flagging for missing assertions, renaming it correctly identifies that it is a helper method, not a test in itself, therefore reducing false alerts in the test output and making it easier to identify missing assertions


test/support/flow_test_helper.rb
```diff
@@ -66,37 +66,37 @@ def assert_rendered_start_page(text = nil)
...
   def assert_no_rendered_outcome(text: nil)
     ensure_valid_and_correct_node
 
-    assert_not_empty test_flow.outcome_text
-    assert_no_match text, test_flow.outcome_text if text
+    assert_not_empty flow_being_tested.outcome_text
+    assert_no_match text, flow_being_tested.outcome_text if text
   end
 
   class TestFlow
```
hezzimo, line 102: I am guessing that this doesn't cause the same issues?

hezzimo (approved):
Just a question but looks good. It'll be lovely to see fewer warnings in the test output!

test/support/flow_test_helper.rb
stacy-hayes: I'd not noticed it causing them, but I'll double check

stacy-hayes: It does not cause the same issues

PR #7768, opened 2026-08-19
Validate user input

hezzimo:
Check user input float value before calling float methods, to prevent triggering a FloatDomainError.


Tetrino (approved):
Nice and clean!

